Issue #1, opened 2026-03-22
Research: tmux control mode integration for future consideration


cheapsteak:
## Context

We initially implemented tmux integration using control mode (`tmux -CC attach`) but replaced it with grouped sessions + direct PTY attachment due to rendering issues. This issue documents what we learned and what a future control mode implementation would require.

## Current Approach: Grouped Sessions

Each terminal panel creates a tmux grouped session (`tmux new-session -t main -s view-<uuid>`) and attaches via a native PTY. SwiftTerm connects directly — input, output, resize all work through the standard terminal driver.

**Pros:** Simple, reliable, SwiftTerm works as designed
**Cons:** One tmux client process per visible panel, no background output detection for non-visible panes

## Why Control Mode Failed (v1 attempt)

1. **Size synchronization** — `refresh-client -C` sets control client size which constrains ALL windows. Multiple panels of different sizes cause constant resize fights.
2. **Staircase rendering** — `capture-pane -p` output uses `\n` line endings, but SwiftTerm expects `\r\n`. Raw `%output` streaming worked but `capture-pane` for state replay didn't.
3. **Input encoding** — `send-keys -l` can't handle control characters (Enter, Escape). `send-keys -H` (hex) works but adds complexity.
4. **Actor/threading conflicts** — blocking `availableData` read loop on a Swift actor starved all other method calls (registerPane, sendKeys). Fixed by using a real Thread, but highlighted architectural complexity.
5. **View lifecycle conflicts** — SwiftUI's polling + `@Published` updates caused terminal views to be destroyed/recreated, losing `%output` handler registrations. `%output` only streams changes, not full state.

## What Control Mode Would Give Us

- **Single connection per repo** — 1 tmux client vs N (one per visible panel)
- **Background output detection** — `%output` for ALL panes, even non-visible ones. Could power notification badges without Claude Code hooks.
- **Programmatic window lifecycle events** — `%window-add`, `%window-close`, `%layout-change`

## iTerm2's Implementation (Reference)

iTerm2 has ~8,000 lines across 12 files for tmux integration:

| Component | Lines | Role |
|---|---|---|
| `TmuxController.m/.h` | ~4,300 | Central orchestrator — window/pane registry, affinity system, state persistence, resize coordination |
| `TmuxGateway.m/.h` | ~1,350 | Protocol parser — command queue with callbacks, `%output`/`%layout-change` dispatch |
[...]

PR #32, opened 2026-03-27
Feat: Pin worktrees and terminal panes for persistent split view across sessions

cheapsteak:
## Summary
- **Worktree pinning** — pin worktrees so they auto-select on launch and stay visible in split view, with panes rendered in pin/click order instead of arbitrary Set order
- **Terminal pane pinning** — pin individual terminal panes so they persist in a right-side dock when navigating between worktrees; dock hides when viewing the pinned terminal's home worktree (no duplication)
- **Resizable dock** — draggable divider between main content and pinned terminal dock, ratio persisted across sessions
- Pin icon on hover (worktree rows + pane headers) with Pin/Unpin in context menus

## Test plan
- [ ] Pin a worktree via hover icon or context menu → verify it auto-selects on app restart
- [ ] Pin multiple worktrees → verify split view shows them in pin order
- [ ] Cmd+click a pinned worktree → verify it can't be deselected (must unpin first)
- [ ] Pin a terminal pane via the pane header icon → verify dock appears on the right when viewing other worktrees
- [ ] View the pinned terminal's home worktree → verify dock hides (terminal visible in-place)
- [ ] Pin multiple terminals from different worktrees → verify dock stacks them vertically
- [ ] Drag the dock divider → verify ratio persists after app restart
- [ ] Unpin terminal → verify it disappears from the dock
- [ ] `swift test` passes (155 tests including 4 new pin tests)

### DIFF
--- a/Sources/TBDApp/AppState+Terminals.swift
+++ b/Sources/TBDApp/AppState+Terminals.swift
@@ -55,4 +55,21 @@ extension AppState {
             handleConnectionError(error)
         }
     }
+
+    /// Toggle pin state for a terminal.
+    func setTerminalPin(id: UUID, pinned: Bool) async {
+        // Optimistic local update
+        for worktreeID in terminals.keys {
+            if let idx = terminals[worktreeID]?.firstIndex(where: { $0.id == id }) {
+                terminals[worktreeID]?[idx].pinnedAt = pinned ? Date() : nil
+            }
+        }
+
+        do {
+            try await daemonClient.setTerminalPin(id: id, pinned: pinned)
+        } catch {
+            logger.error("Failed to set terminal pin: \(error)")
+            handleConnectionError(error)
+        }
+    }
 }
--- a/Sources/TBDApp/AppState+Worktrees.swift
+++ b/Sources/TBDApp/AppState+Worktrees.swift
@@ -99,6 +99,61 @@ extension AppState {
         }
     }
 
+    /// Toggle pin state for a worktree.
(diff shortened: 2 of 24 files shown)


cheapsteak:
Thanks for the review!

**Addressed in 65a7eda:**
- **#1** — Removed extra DB round-trips in both `handleWorktreeSetPin` and `handleTerminalSetPin`. Now derives `pinnedAt` from `params.pinned` directly (`Date()` or `nil`).
- **#3** — Added `max(0.1, min(0.6, ...))` clamp when restoring `dockRatio` from UserDefaults.

**Not addressing:**
- **#2 (ScrollView in dock)** — The dock cells contain live terminal views that need fixed frames. Each cell gets equal height within available space. A ScrollView would conflict with terminal rendering — terminals aren't scrollable cards, they need stable geometry for PTY resize signals.
- **#4 (UserDefaults writes during drag)** — Agree with the reviewer's own "take or leave" assessment. Crash-safety is worth the cheap writes.

cheapsteak:
Addressed in b552335:

- **#1** — Added `at: Date` parameter to `WorktreeStore.setPin` and `TerminalStore.setPin`. Handlers now create one `Date()` and pass it to both the DB write and the broadcast delta — no more T1/T2 mismatch.
- **#2** — `archive()` now clears `pinnedAt = nil` alongside the status/archivedAt updates.
- **#3 (empty pane label)** — Intentional per design — terminal ID was removed by spec. Future UX work if labels are needed.

PR #34, opened 2026-03-27
Fix: Cmd+click on PR links in Claude Code status bar does nothing

cheapsteak:
## Summary
- Cmd+clicking OSC 8 hyperlinks (like "PR #32" in Claude Code's status bar) did nothing — the underline appeared on hover but clicking had no effect
- SwiftTerm's native link click chain wasn't firing, so the URL never reached `requestOpenLink`
- Added `extractHyperlinkURL()` to read OSC 8 payloads directly from terminal buffer cells, wired into the existing cmd+click monitor as a fallback after file path extraction

## Test plan
- [ ] Run Claude Code in a TBD terminal, create a PR so the status bar shows "PR #XX"
- [ ] Cmd+hover over the PR link — should show underline
- [ ] Cmd+click the PR link — should open the GitHub PR in the browser
- [ ] Cmd+click a relative file path in terminal output — should still open as code viewer split pane

### DIFF
--- a/Sources/TBDApp/Panes/PanePlaceholder.swift
+++ b/Sources/TBDApp/Panes/PanePlaceholder.swift
@@ -140,6 +140,7 @@ struct PanePlaceholder: View {
                 tmuxWindowID: terminal.tmuxWindowID,
                 tmuxBridge: appState.tmuxBridge,
                 worktreePath: worktree.path,
+                remoteURL: appState.repos.first(where: { $0.id == worktree.repoID })?.remoteURL,
                 onFilePathClicked: { path in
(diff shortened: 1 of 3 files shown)


cheapsteak:
Both items addressed in 30efe33:

- **Cache `gitHubRepoURL()`** — Added `cachedGitHubRepoURL` + `gitHubRepoURLResolved` flag so the git subprocess only runs once per terminal view.
- **`bounds.height` vs `frame.height`** — Fixed `gridPosition()` to use `bounds.height`, matching the coordinate space from `convert(_:from:)`.

PR #35, opened 2026-03-27
Fix: Scrolling only affects middle teammate pane regardless of cursor position

cheapsteak:
## Summary
- With Claude Code agent teams in split-pane mode, scrolling always affected the middle teammate's pane no matter where the cursor was hovering
- The scroll wheel monitor was sending events at hardcoded coordinates `(cols/2, rows/2)` — the center of the terminal grid — so tmux always routed them to whatever pane occupied the middle
- Now converts the actual mouse position to grid coordinates so tmux routes scroll events to the pane under the cursor

## Test plan
- [ ] Restart TBD and spawn a Claude Code agent team with 3+ teammates (split-pane mode)
- [ ] Hover over the leftmost pane and scroll — verify it scrolls that pane's history
- [ ] Hover over the rightmost pane and scroll — verify it scrolls that pane's history
- [ ] Verify scrolling still works normally in single-pane terminals (no agent teams)

### DIFF
--- a/Sources/TBDApp/Terminal/TerminalPanelView.swift
+++ b/Sources/TBDApp/Terminal/TerminalPanelView.swift
@@ -162,18 +162,21 @@ struct TerminalPanelView: NSViewRepresentable {
                 guard deltaY != 0 else { return event }
 
                 let consumed = MainActor.assumeIsolated {
-                    guard let tv = ref.view else { return false }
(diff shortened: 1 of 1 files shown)


cheapsteak:
Addressed — replaced the inline cell math with `tv.gridPosition(atWindowLocation: location)` in a6d8402. Gets the bounds check for free too.

PR #44, opened 2026-03-30
Feat: Show frozen terminal snapshot instead of blank screen during session restore

cheapsteak:
## Summary
- When switching to a worktree with a suspended Claude terminal, users previously saw a blank/dead pane while Claude restarted. Now the terminal buffer is captured with ANSI colors before suspend and displayed as a read-only snapshot with a "Restoring session…" indicator until the live terminal takes over.
- Snapshot is stored in SQLite via a new `suspendedSnapshot` column (migration v7) and cleared once Claude is detected running in the new pane.
- The suspended state is held until Claude actually starts (polling up to 10s), ensuring the app's 2s polling loop has time to observe and display the snapshot.

## Test plan
- [ ] Switch away from a worktree with an idle Claude terminal — verify it suspends and the snapshot is captured (check `/tmp/tbd-suspend.log`)
- [ ] Switch back to the worktree — verify the frozen snapshot is displayed with "Restoring session…" pill, then transitions to the live terminal once Claude resumes
- [ ] Verify colors/formatting in the snapshot match the original terminal output
- [ ] Test with a pinned terminal — should not be suspended
- [ ] Test rapid switching between worktrees — snapshot should appear even on quick round-trips

### DIFF
--- a/Sources/TBDApp/Panes/PanePlaceholder.swift
+++ b/Sources/TBDApp/Panes/PanePlaceholder.swift
@@ -157,7 +157,8 @@ struct PanePlaceholder: View {
 
     @ViewBuilder
     private func terminalContent(terminalID: UUID) -> some View {
-        if let terminal = terminal(for: terminalID),
+        let term = terminal(for: terminalID)
+        if let terminal = term,
            terminal.suspendedAt != nil,
            let snapshot = terminal.suspendedSnapshot {
             ZStack(alignment: .bottomTrailing) {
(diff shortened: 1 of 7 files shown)


cheapsteak:
Addressed all three items from the automated review:

1. **Inner Task race condition** (Medium) — `01f4ee7` — Post-resume task is now stored in `inFlight` so rapid re-suspend cancels it. Added `Task.isCancelled` checks at each stage and the task cleans up its own `inFlight` entry on completion.

2. **ISO8601DateFormatter allocation** (Minor) — `01f4ee7` — Cached as a `nonisolated(unsafe) private let`.

3. **Snapshot geometry mismatch** (Minor) — `acb2458` — `SnapshotTerminalView` now uses `.zero` frame and defers `feed(text:)` to `updateNSView` after layout provides real dimensions.

PR #45, opened 2026-03-31
Fix: Stop "Restoring session" from getting stuck after suspend/resume (10-17s → 3s)

cheapsteak:
## Summary

The "Restoring session…" snapshot indicator stayed visible for 10-17s (sometimes indefinitely) after Claude had already resumed, requiring users to navigate away and back to see the live terminal.

**Why it happens:** The post-resume task polled tmux 50 times (200ms sleep + tmux round-trip + actor contention per iteration) to detect Claude before clearing the suspended state. Actor scheduling delays inflated the expected 10s to 17s+, and in some cases the view never re-rendered even after clearing.

**What this PR does:**
- Replaces the 50-iteration polling loop with a simple 3s fixed delay before clearing — long enough for the app's 2s poll to observe the snapshot, short enough to not feel stuck
- Adds an auto-suspend toggle button in the toolbar (replaces the add-repo button) so users can quickly disable suspension without opening Settings
- The toggle only gates suspend — resume always runs for already-suspended terminals

## Test plan
- [ ] Suspend a Claude terminal by switching worktrees, then switch back — "Restoring session…" should show for ~3-5s then transition to the live terminal
- [ ] Toggle auto-suspend off via toolbar button — switching worktrees should no longer suspend idle Claude instances
- [ ] Toggle auto-suspend back on — suspension should resume on next worktree switch
- [ ] VoiceOver reads the toggle button label correctly

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -10,6 +10,7 @@ Use the `tbd-project` skill for architecture, conventions, and file reference.
 - Use conventional commit messages: `feat:`, `fix:`, `docs:`, `refactor:`
 - Verify your changes compile (`swift build`) before committing.
 - Run `swift test` if you changed daemon or shared code.
+- When adding a branching conditional that gates behavior (feature flags, toggles, mode switches), add a test for each branch. Verify the gated behavior is off when the flag is off, and that ungated behavior still works.
 
 ## Critical Rules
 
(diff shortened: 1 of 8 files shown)


cheapsteak:
Fixed in 0eb521f — the `autoSuspendClaude` guard was wrapping the entire `worktreeSelectionChanged` RPC call, blocking resume too.

Now the call is always sent with a `suspendEnabled` flag that flows through:
- `WorktreeSelectionChangedParams.suspendEnabled` (optional, defaults to true for backwards compat)
- `RPCRouter+SelectionHandlers` passes it to the coordinator
- `SuspendResumeCoordinator.selectionChanged` only calls `scheduleSuspend` when `suspendEnabled` is true
- Resume always runs regardless

Also moved `refreshTerminals` outside the guard as noted in the minor feedback.

PR #48, opened 2026-04-01
Fix: Eliminate blank screen gap during terminal resume by feeding snapshot into the same view

cheapsteak:
## Summary

When switching to a worktree with a suspended Claude terminal, users saw **snapshot → blank screen → live content**. The blank gap lasted 1-3s while the TerminalPanelView connected to tmux.

**Why it happens:** The previous approach used a separate SnapshotTerminalView that was swapped out for TerminalPanelView when `suspendedAt` cleared. The new TerminalPanelView started with a black screen while the tmux grouped session connected.

**What this PR does:**
- Feeds the ANSI snapshot directly into the same TerminalView that will connect to tmux (`initialSnapshot` parameter with `\n` → `\r\n` fix for SwiftTerm)
- Live tmux output overwrites the snapshot seamlessly — no view swap, no blank gap
- Removes `SnapshotTerminalView.swift` and `RestoringIndicator` (no longer needed)
- Clears `suspendedAt` immediately after resume (no timer), keeps snapshot in DB for the view
- Adds just-in-time idle seeding when response_complete hook was missed
- Delays idle hook re-seed by 30s after resume to prevent rapid re-suspension
- Adds pause icon in sidebar for suspended worktrees

## Test plan
- [ ] Suspend a Claude terminal by switching worktrees — verify snapshot is captured (check `/tmp/tbd-suspend.log`)
- [ ] Switch back — should see the last terminal content immediately, then live content overwrites it with no blank gap
- [ ] Verify worktrees with missed response_complete hooks still suspend (just-in-time seeding)
- [ ] After resume, verify the terminal is not immediately re-suspended on brief departures (30s re-seed delay)
- [ ] Sidebar shows pause icon next to suspended worktrees
- [ ] Run `swift test` — 173 tests pass


cheapsteak:
Addressed in 855912b — added `isSuspendedSnapshot` flag to `TerminalPanelView`. When the terminal is suspended at view creation time, the snapshot is fed but `startTmuxClient` is skipped, preventing the old window's shell from overwriting it. The view recreates naturally when `tmuxWindowID` changes after resume completes.

Re: stale snapshot note — agreed this is the right tradeoff, already documented.

cheapsteak:
Addressed the race condition in ba930c7 — added `suspendedAt` to the view `.id`:
```swift
.id("\(terminal.id)-\(terminal.tmuxWindowID)-\(terminal.suspendedAt != nil)")
```
This ensures the view recreates when `suspendedAt` clears even if `tmuxWindowID` already updated from the first DB write.

Re: JIT idle seeding test — acknowledged, deferring as it requires making `ClaudeStateDetector` injectable.

PR #49, opened 2026-04-01
Feat: Manually suspend and resume Claude terminals from context menu and tab bar

cheapsteak:
## Summary
- Adds manual suspend/resume controls for Claude terminals — right-click a worktree in the sidebar ("Suspend Claude" / "Resume Claude") or use the pause/play button in individual terminal tab headers
- Manual suspend overrides pin state and waits up to 10s for Claude to go idle before sending `/exit`; aborts if still busy
- Suspended terminals show a frozen ANSI snapshot with a "Suspended" badge; resume creates a new tmux window with `claude --resume`

## Test plan
- [ ] Right-click an active worktree with a Claude terminal → "Suspend Claude" appears, clicking it suspends after idle
- [ ] Right-click same worktree → "Resume Claude" appears, clicking resumes the session
- [ ] Hover a Claude terminal tab → pause button appears, clicking suspends that terminal
- [ ] Suspended tab shows moon icon, dimmed label, and play button to resume
- [ ] Terminal content shows frozen snapshot with "Suspended" badge overlay
- [ ] Suspending a pinned Claude terminal works (pin is overridden)
- [ ] Suspending while Claude is mid-response waits up to 10s then aborts (no forced exit)
- [ ] `swift test` passes (176 tests)

### DIFF
--- a/Sources/TBDApp/DaemonClient.swift
+++ b/Sources/TBDApp/DaemonClient.swift
@@ -434,6 +434,38 @@ actor DaemonClient {
         )
     }
 
+    /// Manually suspend a single Claude terminal.
+    func terminalSuspend(terminalID: UUID) throws {
+        try callVoid(
+            method: RPCMethod.terminalSuspend,
+            params: TerminalSuspendParams(terminalID: terminalID)
+        )
+    }
+
+    /// Manually resume a single suspended terminal.
+    func terminalResume(terminalID: UUID) throws {
+        try callVoid(
+            method: RPCMethod.terminalResume,
+            params: TerminalResumeParams(terminalID: terminalID)
+        )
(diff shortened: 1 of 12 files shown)


cheapsteak:
Addressed the Medium:

**`inFlight` cancellation propagation** — Wrapped the inner suspend task with `withTaskCancellationHandler` that explicitly cancels the inner task when the wrapper is cancelled. Added `Task.isCancelled` checks in the idle-wait loop so cancellation stops the work promptly. (c101bd4)

The two Minors (silent `.busy` failure, `suspendLog` file I/O) are deferred — silent error handling is consistent with existing patterns, and the log I/O is pre-existing code.

PR #56, opened 2026-04-03
Feat: Add conductor pattern for meta-agent orchestration of Claude terminals

cheapsteak:
## Summary
- Adds a **conductor** system — persistent Claude Code sessions that monitor and coordinate other Claude terminals across worktrees and repos. Conductors can read agent conversation history via session JSONL, auto-respond to waiting agents, and escalate to the user via macOS notifications.
- Implements Phase 1a (minimal viable conductor): `WorktreeStatus.conductor`, DB migration v9 (conductor table + synthetic repo), `terminal.output` RPC (debug), `terminal.conversation` RPC (reads Claude session JSONL), `ConductorManager` lifecycle (setup/start/stop/teardown), CLI commands (`tbd conductor setup/start/stop/teardown/list/status`), and a CLAUDE.md template that teaches Claude how to be a conductor.
- Includes design spec for the full conductor pattern (Phase 1a + deferred Phase 1b + Phase 2) with flexible scoping (per-repo, per-worktree, per-terminal-label) and interaction locking.

## Test plan
- [ ] `swift build` succeeds
- [ ] `swift test` — all 195 tests pass (new: ConductorStore, ConductorManager, terminal.output, worktree/repo filtering)
- [ ] `tbd conductor setup test-conductor` creates directory at `~/.tbd/conductors/test-conductor/` with CLAUDE.md
- [ ] `tbd conductor list --json` shows the conductor
- [ ] `tbd conductor start test-conductor` creates tmux window in `tbd-conductor` server
- [ ] `tbd conductor stop test-conductor` kills the window
- [ ] `tbd conductor teardown test-conductor` removes everything
- [ ] `tbd terminal output <id>` captures terminal viewport (debug)
- [ ] `tbd terminal conversation <id>` reads Claude session JSONL
- [ ] Conductor worktrees don't appear in `tbd worktree list`
- [ ] Synthetic conductors repo doesn't appear in `tbd repo list`

### DIFF
--- a/.claude/skills/tbd-project/references/architecture.md
+++ b/.claude/skills/tbd-project/references/architecture.md
@@ -39,15 +39,19 @@ Stateless. Connects to daemon socket, sends RPC, prints result, exits.
 
 ## Data Model (SQLite)
 
-Four tables: `repo`, `worktree`, `terminal`, `notification`. See `Sources/TBDShared/Models.swift` for struct definitions and `Sources/TBDDaemon/Database/` for GRDB record types.
+Five tables: `repo`, `worktree`, `terminal`, `notification`, `conductor`. See `Sources/TBDShared/Models.swift` and `Sources/TBDShared/ConductorModels.swift` for struct definitions and `Sources/TBDDaemon/Database/` + `Sources/TBDDaemon/Conductor/` for GRDB record types.
 
 ### Migrations
 - **v1** — initial schema: repo, worktree, terminal, notification tables
 - **v2** — adds `gitStatus` column to worktree (defaults to "current")
 - **v3** — adds `hasConflicts` bool column to worktree (defaults to false), replacing the gitStatus enum
+- **v4-v8** — adds pinnedAt, claudeSessionID, suspendedAt, suspendedSnapshot, archivedClaudeSessions columns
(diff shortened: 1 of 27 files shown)


cheapsteak:
Fixed in 2ad09e2. `start()` now guards against double-start:

- If `terminalID` is set and the window is alive → throws `alreadyRunning`
- If `terminalID` is set but the window is dead → cleans up the stale record, then proceeds

Went with the "check + clean up stale" approach rather than just a simple guard, since it also handles the case where a conductor was started, its tmux window died externally, and the DB wasn't cleaned up.

cheapsteak:
Addressed the Medium in 2b75605 — replaced `JSONSerialization` + dynamic dict-casting with typed Codable structs (`SessionEntry`, `SessionMessage`, `ContentBlock`). Parse failures now surface as decode errors instead of silently returning empty.

Re: the minors:
- **`stop()` broadcast** — as noted, conductor terminal deltas are already suppressed by the label check. If conductors become visible in the UI (Phase 2), this path would need the broadcast added.
- **Label string convention** — agreed it's soft, but sufficient for Phase 1. A type-level guarantee would require adding a `isConductor` bool to the terminal model, which is overkill right now.
- **Test plan `--interact`** — good catch, that's stale from before permissions were removed. Updated the PR description.

PR #57, opened 2026-04-03
Feat: Let users create Claude sessions and notes from the tab bar

cheapsteak:
## Summary
- Replace the single `+` button with a menu offering three pane types: Shell (same as before), Claude (managed session with suspend/resume), and Note (freeform text editor with auto-save)
- Claude terminals created this way get proper labels and session IDs, so they automatically integrate with the existing suspend/resume coordinator — no coordinator changes needed
- Notes are a new entity (DB table, CRUD RPC, `NoteStore`) with debounced auto-save, inline-renameable titles, and daemon-side persistence

## Test plan
- [ ] Click `+` in tab bar → menu shows Shell, Claude, Note
- [ ] Create Shell → behaves exactly as before
- [ ] Create Claude → label shows "claude", suspend/resume button appears on tab, coordinator manages it on worktree switch
- [ ] Create Note → text editor appears with "Note 1" title, typing auto-saves after 500ms
- [ ] Click note title in pane header → inline edit, commit on Enter/blur
- [ ] Close note tab → note deleted from DB
- [ ] Create Note 1, Note 2, delete Note 1, create new → titled "Note 3" (not duplicate "Note 2")
- [ ] `swift test` → 180 tests pass

### DIFF
--- a/Sources/TBDApp/AppState+Notes.swift
+++ b/Sources/TBDApp/AppState+Notes.swift
@@ -50,36 +50,4 @@ extension AppState {
         }
     }
 
-    /// Fork a Claude terminal by resuming from an existing session ID.
-    func forkClaudeTerminal(worktreeID: UUID, sessionID: String) async {
-        do {
-            let terminal = try await daemonClient.createTerminal(
-                worktreeID: worktreeID,
-                resumeSessionID: sessionID
-            )
-            terminals[worktreeID, default: []].append(terminal)
-            let tab = Tab(id: terminal.id, content: .terminal(terminalID: terminal.id))
-            tabs[worktreeID, default: []].append(tab)
-        } catch {
-            logger.error("Failed to fork Claude terminal: \(error)")
-            handleConnectionError(error)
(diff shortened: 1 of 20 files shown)


cheapsteak:
Addressed the review feedback:

**Medium — NotePaneView data loss on tab switch** → Fixed in 199f881. `onDisappear` now flushes an immediate save instead of just cancelling the debounced task. If the note was deleted, the "Note not found" error is already silently absorbed by `handleConnectionError`.

**Minor — Terminal methods in wrong file** → Fixed in db534f3. Moved `forkClaudeTerminal` and `createClaudeTerminal` from `AppState+Notes.swift` to `AppState+Terminals.swift`.

**Minor — Force-unwrap UUIDs in NoteRecord.toModel()** → Not addressing. This is consistent with all other stores in the codebase (`TerminalRecord`, `WorktreeRecord`, `RepoRecord`, `NotificationRecord`). If we change this pattern, it should be a separate cleanup across all stores.

PR #58, opened 2026-04-04
Feat: Guake-style conductor overlay with one-click setup and navigation suggestions

cheapsteak:
## Summary
- Adds a toggleable terminal overlay (Opt+.) that renders the conductor session inline — drops from the top of the content area, resizable via drag handle, keeps tmux alive when hidden
- One-click conductor setup from toolbar button: no config dialogs, auto-derives name from repo. Right-click for stop/remove.
- Navigation suggestions: conductor calls `tbd conductor suggest` to show a clickable "Go to worktree" pill, letting users jump to relevant worktrees without leaving the conductor conversation

## Test plan
- [ ] Click conductor toolbar button with no conductor → auto-setup + start + overlay appears
- [ ] Opt+. toggles overlay visibility; terminal stays alive when hidden
- [ ] Drag handle resizes overlay height; height persists across app restarts
- [ ] Right-click conductor button → Stop/Remove actions work
- [ ] `tbd conductor suggest <name> --worktree <id> --label "test"` shows suggestion pill
- [ ] Clicking "Go" navigates to worktree and clears suggestion server-side
- [ ] Clicking "✕" dismisses suggestion without navigation
- [ ] Switching repos shows/hides conductor for that repo
- [ ] Content below overlay remains clickable; clicking a terminal below steals focus from conductor
- [ ] `swift test` — all 204 tests pass

### DIFF
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -33,6 +33,12 @@ final class AppState: ObservableObject {
     /// Archived worktrees keyed by repo ID, fetched on demand.
     @Published var archivedWorktrees: [UUID: [Worktree]] = [:]
 
+    /// The first selected worktree, if any.
+    var selectedWorktree: Worktree? {
+        guard let id = selectedWorktreeIDs.first else { return nil }
+        return worktrees.values.flatMap { $0 }.first { $0.id == id }
+    }
+
     /// All pinned terminals across all worktrees, sorted by pinnedAt.
     var pinnedTerminals: [Terminal] {
         terminals.values.flatMap { $0 }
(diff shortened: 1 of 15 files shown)


cheapsteak:
Addressed all three items:

1. **Overlay invisible on first launch** → Added `await refreshWorktrees()` before `await refreshConductors()` in `ensureConductorRunning()` so the terminals dict is populated before we look up the conductor's terminal. Also moved the hyphen trim to after `prefix(64)` (fix 3).

2. **Double-install guard** → Added `guard monitor == nil else { return }` at the top of `install()`.

3. **Trailing hyphen after truncation** → Moved `.trimmingCharacters` to after `prefix(64)`.

PR #60, opened 2026-04-05
Feat: File drag-and-drop into terminals + background session notifications

cheapsteak:
## Summary
- **Drag-and-drop files into terminals** — dropping files onto a terminal inserts their shell-quoted absolute paths as keyboard input, working with any shell or Claude Code
- When a Claude session finishes responding in a non-visible worktree, TBD now plays a configurable sound and/or shows a macOS notification banner with the worktree name and truncated response
- Adds real-time state subscription (persistent socket streaming deltas from daemon to app) so notifications fire instantly instead of waiting for the 2s polling cycle
- Settings UI with independent toggles for sound and notifications, system sound picker (default: Blow), custom file picker, and test button

## Drag-and-drop
Registers `TBDTerminalView` as a file URL drop target. On drop, absolute paths are shell-quoted (handling spaces and special characters) and sent to the PTY — the same channel as keyboard input. Multiple files are space-separated.

## Idle notifications
```
Claude Stop hook → extracts last_assistant_message via jq
  → tbd notify --type response_complete --message "..."
  → daemon stores notification, broadcasts StateDelta
  → app receives delta via persistent subscription socket
  → checks worktree visibility (same logic as unread badge)
  → if not visible: plays NSSound + posts UNUserNotificationCenter banner
```

## Changes
- **Terminal:** File drag-and-drop support on `TBDTerminalView` with shell quoting
- **Daemon:** `state.subscribe` streaming RPC, dead-subscriber auto-cleanup, `StateDelta` types moved to TBDShared
- **App:** Subscription client in DaemonClient, NotificationSoundPlayer (`@MainActor` + NSSound), MacNotificationManager (UNUserNotificationCenter), expanded Settings UI
- **CLI:** Stop hook now passes `--message` with Claude response, migration updates existing hooks, `TBD_WORKTREE_ID` env var for fast worktree resolution
- **Tmux:** Terminal creation exports `TBD_WORKTREE_ID` into shell environment

## Test plan
- [ ] Drag a file from Finder onto a terminal — verify absolute path appears at cursor
- [ ] Drag a file with spaces in the name — verify path is single-quoted
- [ ] Drag multiple files — verify space-separated paths
- [ ] Run `tbd setup-hooks --global` and verify hook in `~/.claude/settings.json` includes `--message "$MSG"`
- [ ] Open TBD app, start Claude in two worktrees, switch to one — verify the other triggers sound + notification when Claude finishes
- [ ] Settings → toggle sound off → verify no sound plays but notification still appears
- [ ] Settings → change sound to different system sound → verify "Test" button plays it
- [ ] Settings → pick custom sound file → verify it plays
- [ ] Verify no sound/notification when the finishing worktree is currently visible
- [ ] `swift test` — all 204+ tests pass


cheapsteak:
Addressed all three items:

- **`subscribe` blocking actor** — marked as `nonisolated` so the blocking `recv` loop runs off the actor executor (697189d)
- **Force-unwraps on UTType** — replaced with `compactMap` (2d30cba)
- **`handleTerminalRecreateWindow` missing `TBD_WORKTREE_ID`** — added clarifying comment; intentionally omitted since recreated windows run a plain shell (8dfa4dd)

cheapsteak:
Addressed round 2 feedback:

- **Subscription reconnect** — nil out `subscriptionTask` after `subscribe()` returns so the poll loop detects disconnection and restarts (9f4b988)
- **Socket duplication** — extracted `makeConnectedSocket()` helper, used by both `sendRaw()` and `subscribe()` — net -22 lines (68c3d40)
- **Notification pile-up** — use `worktreeID` as notification identifier so subsequent completions from the same worktree replace the previous banner (1af8a95)

cheapsteak:
Round 3:

- **Logger vs print()** — replaced `print()` with `Logger(subsystem:category:)` in MacNotificationManager (761c9bf)
- **Blocking recv() on cooperative thread pool** — acknowledged as non-blocker, will address in a follow-up

cheapsteak:
Round 4:

- **`__custom__` sentinel** — removed; `pickCustomSound()` no longer writes sentinel to `soundName`. The Picker binding derives display from `customPath.isEmpty`, so clearing customPath falls back to the last real system sound (3a9cad6)
- **Silent send failure** — added `daemonClientLogger.warning` on partial send (7edb669)
- **Ack-skipping comment** — clarified why the pattern is unique to subscription RPC (7edb669)
- **`subscribe()` cooperative thread** — deferred to follow-up (discussed in round 3)
- **Badge downgrade** — skipping; cosmetic and self-correcting within one poll cycle

cheapsteak:
Round 5:

- **Notification banners when TBD is frontmost** — implemented `UNUserNotificationCenterDelegate.willPresent` with `.banner` option, set delegate in `requestPermissionIfNeeded()` (8f4a8fe)
- **Conductor delta suppression test** — added test covering worktreeCreated with `.conductor` status and terminalCreated with `"conductor:"` label prefix, plus a non-conductor delta that should pass through (dc0a61d)

PR #61, opened 2026-04-05
Feat: Per-repo custom instructions injected into Claude sessions via --append-system-prompt

cheapsteak:
## Summary
- Adds configurable per-repo instructions (rename prompt + general instructions) that TBD injects into spawned Claude Code sessions via `--append-system-prompt`, without modifying CLAUDE.md files
- Three instruction layers: conditional rename prompt (fires when worktree hasn't been renamed yet), built-in TBD context (always injected, tells Claude about available CLI commands), and user-editable general instructions
- New repo detail view with segmented control (Archived | Instructions) replaces the archived worktrees view when clicking a repo name in the sidebar

## Test plan
- [ ] Verify `swift test` passes (224 tests including new RPC, SystemPromptBuilder, and shell escaping tests)
- [ ] Create a worktree, verify Claude session receives the rename prompt and TBD context in system prompt
- [ ] Rename the worktree, create a new Claude terminal — verify rename prompt no longer fires
- [ ] Click a repo name in sidebar, verify segmented control appears with Archived and Instructions tabs
- [ ] Edit rename prompt and general instructions, verify debounced auto-save with "Saved" indicator
- [ ] Reset rename prompt to default, verify it saves as nil (uses built-in default)
- [ ] Clear rename prompt entirely, verify it disables rename injection

### DIFF
--- a/Sources/TBDApp/AppState+Repos.swift
+++ b/Sources/TBDApp/AppState+Repos.swift
@@ -30,4 +30,17 @@ extension AppState {
             handleConnectionError(error)
         }
     }
+
+    /// Update per-repo instruction fields.
+    func updateRepoInstructions(repoID: UUID, renamePrompt: String?, customInstructions: String?) async {
+        do {
+            _ = try await daemonClient.repoUpdateInstructions(
+                repoID: repoID, renamePrompt: renamePrompt, customInstructions: customInstructions
+            )
+            await refreshRepos()
+        } catch {
+            logger.error("Failed to update instructions: \(error)")
+            handleConnectionError(error)
+        }
+    }
 }
--- a/Sources/TBDApp/AppState+Repos.swift
+++ b/Sources/TBDApp/AppState+Repos.swift
@@ -31,16 +31,19 @@ extension AppState {
         }
     }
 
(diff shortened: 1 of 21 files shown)


cheapsteak:
Re: **[Medium] `onDisappear` flush silently discards errors**

The flush path calls `appState.updateRepoInstructions(...)` which returns `Bool` — errors are logged internally via `handleConnectionError` (which logs to os_log and updates connection state). Since `onDisappear` fires when the view is already leaving the hierarchy, there's no visible UI surface to show an error indicator to the user.

The scheduled save path (which fires while the view is visible) already gates the "Saved" indicator on success — so users get confirmation when saves work, and no false confirmation when they don't. The `onDisappear` flush is a best-effort backstop for the sub-1s window between last keystroke and navigation.

Re: **Redundant save on view disappear when no edits made** — fair point but low impact (one extra no-op RPC on navigation). Adding a dirty flag would add state to track for marginal benefit.

Re: **No state broadcast after `repo.updateInstructions`** — correct, this is intentional for now. The only client that writes instructions is the same one that reads them (single app window), and it calls `refreshRepos()` after the write. A broadcast delta would be the right move if/when we support multiple windows.

PR #63, opened 2026-04-07
Feat: Run multiple Claude accounts side-by-side in TBD

cheapsteak:
## Summary
- Adds a multi-token Claude switcher: store multiple OAuth tokens, set a global default with per-repo overrides, and swap tokens mid-conversation by forking the session into a new tab.
- Tokens stored as 0600 files at `~/.tbd/claude-tokens/<uuid>.token` (Keychain rejected — code-signing-free SPM daemon can't get a stable ACL across rebuilds without a GUI auth prompt).
- Token-aware spawn paths: `WorktreeCreate`, `TerminalCreate`, fork, swap, and **suspend/resume** all route through `ClaudeSpawnCommandBuilder` and persist `claudeTokenID` on the terminal row, so the override survives restarts.
- UI: Settings → Claude Tokens manages the list; menu bar shows the active default; Claude tabs display the token name (numbered "Michel 1 / Michel 2" when forking the same token); right-click → "Swap token" opens a fresh forked tab on the chosen account.

## Test plan
- [ ] Add a token via Settings → Claude Tokens; verify the row appears with a "Saved" badge (and an "Unverified" orange badge if `/api/oauth/usage` 403s — expected for `claude setup-token` tokens lacking `user:profile` scope).
- [ ] Set a global default; create a new worktree; confirm the spawned Claude tab is labeled with the token name and the process is using that token (`echo \$CLAUDE_CODE_OAUTH_TOKEN` in a shell tab).
- [ ] Set a per-repo override on a different repo; create a worktree there; confirm it uses the override (not the global default).
- [ ] In a Claude tab, right-click → "Swap token → \<other token\>"; verify a NEW tab opens (old tab untouched), is auto-selected, and the label shows the new token name.
- [ ] Suspend a Claude tab with a non-default token, then resume; verify the resumed process is still on the correct token (covered by `SuspendResumeCoordinatorTests.resumeInjectsTokenWhenResolverProvided`).
- [ ] Restart the daemon; reopen the worktree; verify token-labeled tabs reload with their token name (not "Terminal N").
- [ ] `swift test` — 319/319 passing locally.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,6 +17,13 @@ The main chat session agent should not write code directly. Delegate all impleme
 - Run `swift test` if you changed daemon or shared code.
 - When adding a branching conditional that gates behavior (feature flags, toggles, mode switches), add a test for each branch. Verify the gated behavior is off when the flag is off, and that ungated behavior still works.
 
+### Restart must use the worktree's own script
+Always run `scripts/restart.sh` (relative, from the worktree cwd), never an absolute path to the main project's copy. Using `/Users/chang/projects/tbd/scripts/restart.sh` builds and starts binaries from the main branch, leaving old worktree processes running and causing "Unknown method" RPC errors. After any restart, verify with:
+```
+ps aux | grep -E "\.build/debug/TBD" | grep -v grep
+```
+There should be exactly one `TBDDaemon` and one `TBDApp`, both from the worktree path. If stale processes exist: `pkill -f TBDDaemon; pkill -f TBDApp` then re-run `scripts/restart.sh`.
+
 ## Critical Rules
 
 ### NEVER delete ~/.tbd/state.db
--- a/Package.swift
+++ b/Package.swift
@@ -38,7 +38,7 @@ let package = Package(
                 "TBDDaemonLib",
             ],
             path: "Sources/TBDDaemon",
-            exclude: ["Database", "Git", "Hooks", "Tmux", "Lifecycle", "Server", "SSH", "PR", "Conductor", "Daemon.swift", "PIDFile.swift"],
+            exclude: ["Database", "Git", "Hooks", "Tmux", "Lifecycle", "Server", "SSH", "PR", "Conductor", "Keychain", "Daemon.swift", "PIDFile.swift"],
             sources: ["main.swift"]
         ),
         .executableTarget(
--- a/Package.swift
+++ b/Package.swift
@@ -38,7 +38,7 @@ let package = Package(
                 "TBDDaemonLib",
             ],
             path: "Sources/TBDDaemon",
-            exclude: ["Database", "Git", "Hooks", "Tmux", "Lifecycle", "Server", "SSH", "PR", "Conductor", "Keychain", "Daemon.swift", "PIDFile.swift"],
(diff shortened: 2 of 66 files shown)


cheapsteak:
Addressed the review in 5 commits (af64a9c, 9275d55, 2207ee0, dffe639, 3abb9d9):

**High** — `precondition` → `guard` + `Logger.error` + return base command (no env injection). Daemon stays alive on unexpected token format.

**Medium — token in `ps aux`** — `ClaudeSpawnCommandBuilder.build` now returns `(command, sensitiveEnv)`. `TmuxManager.createWindow` accepts `sensitiveEnv:` and emits `tmux new-window -e KEY=VALUE` flags. The token never appears inside the shell command body. All token spawn tests now positively assert the `-e` flag and negatively assert the secret is absent from the shell body.

**Medium — silent global default fallback** — added `FileHandle.standardError.write(...)` warning matching the repo-override path.

**Minor — `print()` in daemon** — replaced with `Logger.warning` in `ClaudeUsagePoller`.

**Minor — `http429` vs `http_429`** — standardized on `http_429` everywhere; dropped legacy fallback in `ClaudeTokensSettingsView`.

**Deferred (one Minor)** — "revived secondary sessions lose their token assignment". The reviewer flagged this as a pre-existing design gap (`archivedClaudeSessions` only stores session UUIDs, no per-terminal token IDs). Not blocking; tracking separately.

`swift test`: 320/320 passing.

cheapsteak:
Round 2 addressed in 4 commits (55e4a88, ea2e8a4, 5a49fca, 9dbd1b3):

**Medium 1 — stale session ID after swap** — added a deferred Task in `handleTerminalSwapClaudeToken` mirroring the existing post-resume re-capture pattern: sleep 5s → `ClaudeStateDetector.captureSessionID` → `db.terminals.updateSessionID`. Suspend on a swapped tab now persists the real forked session ID, not the parent's.

**Medium 2 — char allowlist drops API keys** — good catch on the obsolete restriction. Since commit af64a9c we pass tokens via `tmux -e KEY=VALUE` (no shell escaping), so the `[A-Za-z0-9\-_]` allowlist is no longer needed. Removed the silent fallback in `ClaudeSpawnCommandBuilder` entirely. Storage-time validation in `handleClaudeTokenAdd` now only rejects `\n`, `\r`, `\0` (the chars that would actually break tmux argv parsing), with an explicit user-facing error. New test `addRejectsNewline` covers this.

**Minor 3 — silent keychain delete failure** — `try?` → do/catch + `logger.warning`. RPC still succeeds (DB row is the source of truth) but failures are now visible in `log stream`.

**Minor 4 — stderr → os.Logger** — converted all `FileHandle.standardError.write` calls in `Sources/TBDDaemon/` to `Logger.warning`. Found one extra site in `WorktreeLifecycle+Create.swift` while grepping; fixed too. The file-based `suspendLog` (`/tmp/tbd-suspend.log`) was left alone — it's a separate diagnostic channel, not stderr.

`swift test`: 321/321 passing (1 new test for the newline rejection).

cheapsteak:
Picked up Minor #1 in f7d5e6d (one-liner). Skipping Minor #2 (fire-and-forget Task) — practical impact is near-zero per your own note, and matching the SuspendResumeCoordinator's `inFlight` cancellation pattern would require pulling that bookkeeping out of the coordinator into shared state, which isn't worth it for a 5-second window guarded by `try?`. Thanks for the thorough review.

PR #66, opened 2026-04-08
Feat: Recover from moved repos and switch to visible ~/tbd home

cheapsteak:
## Summary

- **Recover from moved/missing repos.** TBD now validates repo health on startup and after reconcile, marks missing repos as such (dimmed in the app, `[missing]` in CLI), and offers a `Locate…` affordance + `tbd repo relocate` command to point them at the new location. Worktrees survive moves intact.
- **Canonical worktree location + visible home dir.** New worktrees now go to `~/tbd/worktrees/<repo>/<wt-name>/` instead of being scattered under each repo's `.tbd/worktrees/`. The daemon's home moved from hidden `~/.tbd` to visible `~/tbd`. Existing worktrees stay where they are and decay naturally — the legacy read path is marked `LEGACY-WORKTREE-LOCATION: remove after 2026-06-01`.
- **Migration was attempted then abandoned.** A `tbdd --migrate-worktrees` flow was built (with crash-recoverable journal, fcntl lock, manifest, db backup) but it broke every running Claude session whose cwd was inside a moved worktree (kernel held the dead inode → `posix_spawn` ENOENT). All migration code is deleted in this PR. The forward-only canonical path handles new worktrees; old ones age out.
- Adds `scripts/move-home-dir.sh` to safely do the one-shot `~/.tbd → ~/tbd` rename + transitional symlink + restart.

## Test plan

- [ ] `swift build` clean, `swift test` 260 passing
- [ ] After merge: run `scripts/move-home-dir.sh` once
- [ ] Verify `~/tbd/state.db`, `~/tbd/sock`, `~/tbd/tbdd.pid` exist; `~/.tbd` is a symlink to `tbd`
- [ ] Existing worktrees still appear in TBDApp and tmux panes work
- [ ] Create a new worktree → confirm it lands at `~/tbd/worktrees/<repo>/<wt-name>/`
- [ ] Move a repo on disk → TBDApp dims it and `Locate…` repairs it

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -26,7 +26,7 @@ There should be exactly one `TBDDaemon` and one `TBDApp`, both from the worktree
 
 ## Critical Rules
 
-### NEVER delete ~/.tbd/state.db
+### NEVER delete ~/tbd/state.db
 The database stores worktree display names, custom config, and notification history. Deleting it orphans tmux servers (repo UUID changes → tmux server name changes → old sessions become unreachable). If you encounter DB issues, diagnose and fix the schema/code — don't wipe the DB.
 
 ### Database migrations must update the shared model
--- a/Sources/TBDApp/AppIcon.swift
+++ b/Sources/TBDApp/AppIcon.swift
@@ -7,10 +7,30 @@ import CoreText
 /// Returns the worktree name (without date prefix) if running from a worktree build, nil for main.
 func detectWorktreeName() -> String? {
     let path = ProcessInfo.processInfo.arguments[0]
-    guard let range = path.range(of: ".tbd/worktrees/") else { return nil }
-    let afterPrefix = path[range.upperBound...]
-    guard let slashIndex = afterPrefix.firstIndex(of: "/") else { return nil }
-    let name = String(afterPrefix[afterPrefix.startIndex..<slashIndex])
+    let home = FileManager.default.homeDirectoryForCurrentUser.path
+    // Canonical lives under the user's home dir
+    // (~/tbd/worktrees/<slot>/<name>/...) so the worktree name is the SECOND
+    // path component after the marker.
+    let canonicalMarker = "\(home)/tbd/worktrees/"
+    let isCanonical: Bool
+    let afterPrefix: Substring
+    if let range = path.range(of: canonicalMarker) {
+        isCanonical = true
+        afterPrefix = path[range.upperBound...]
+    } else {
+        // LEGACY-WORKTREE-LOCATION: remove after 2026-06-01
+        // Reads worktrees from <repo>/.tbd/worktrees/ for backward compatibility with
+        // worktrees created before the canonical-location switch. New worktrees are
+        // always created under ~/tbd/worktrees/<repo>/<name>. After 2026-06-01, all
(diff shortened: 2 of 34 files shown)


cheapsteak:
Addressed all four items from the bot review:

- **TOCTOU race in `RepoStore.create`** — slot allocation and insert now run in a single `writer.write` transaction. (72b3bb5)
- **Periodic git fetch hits missing repos** — added `where repo.status != .missing` filter to the fetch loop. (dd40581)
- **No user feedback when `Locate…` picks an invalid dir** — set `alertMessage` / `alertIsError` on the relocate failure path so the user sees why the repo stayed dimmed. (be9b0a3)
- **Silent `try?` for main worktree path update** — now does `try await`, records failures in `worktreesFailed`, and logs. (7fbb1d6)

cheapsteak:
Round 2:

- **Same `try?` bug in the legacy-worktree branch** — applied the same do/catch fix; on DB failure we now log, append to `worktreesFailed`, and `continue` rather than running `git worktree repair` against an unpersisted path. (aebcddc)
- **`detectDefaultBranch` as HEAD-resolution probe** — added a comment explaining the result is intentionally discarded and flagging the network-call concern. (cadd19d)
- **Stale v14 migration comment** — trimmed to drop the "origin/main shipped v13" rationale, kept the forward-looking parallel-branch reason. (5dec940)
- The startup-reconcile spurious-warning note is informational — left as-is.

cheapsteak:
Round 3 (post-approval, addressing the two non-blocking minor items):

- **`gitStatusTask` doesn't skip `.missing` repos** — added `where repo.status != .missing` to match `gitFetchTask`. (9ccf7d5)
- **`precondition` in `WorktreeLayout.basePath`** — replaced with a `logger.fault` + UUID-derived fallback so a missing slot can't hard-kill the daemon. (de0fd4f)

Skipping the `repoAdded` → `repoUpdated` delta one — that needs a new RPCDelta variant which is bigger scope and the existing comment in code already flags it as future work. Will wire it up when the variant lands.

cheapsteak:
Round 4: fixed the remaining `try?` on the worktree status update — now logs and the UI status stays in sync with worktreesFailed. (361b449)

cheapsteak:
Round 5: broadcast a delta from the post-reconcile health transition so the sidebar dims immediately when reconcile flips a repo to .missing. (next sha)

cheapsteak:
Round 6: log if the worktree base directory wasn't actually created so permission errors surface here instead of as misleading git worktree add failures downstream. The remaining minor item (relocate doesn't verify same-repo via origin URL) is explicitly intentional per design §6 — left as-is.

cheapsteak:
Round 7 (post second approval, addressing 3/4 minor items):

- **A. Hoist relocate handler logger to file scope** (15fad90)
- **B. Single-pass dash collapse in `WorktreeLayout.sanitize`** (be1ab9a)
- **C. Log on reconcile health-transition DB failure** (11f9261)
- **D. `detectDefaultBranch` → `git rev-parse HEAD`** — skipped. The existing comment already documents the network-call concern as future work, and swapping to rev-parse needs a new GitManager method. Will pair it with the next health-check evolution.

PR #70, opened 2026-04-29
Feat: Add browser-style back/forward to revisit recently selected worktrees

cheapsteak:
## Summary
- Adds chevron-left / chevron-right buttons to the toolbar (also bound to Cmd-[ / Cmd-]) that walk a per-session history of worktree and repo selections, so you can quickly jump back to a worktree you just had open.
- History is suppressed while applying a back/forward entry, deduped against the current head, and capped at 100 entries.
- Preserves cmd-click pane order for multi-select split view by storing the ordered `[UUID]` instead of an unordered Set.

## Test plan
- [ ] Click worktree A, then B, then C. Cmd-[ walks back to B, then A; Cmd-] walks forward.
- [ ] Mid-history, click a new worktree D — forward becomes disabled (forward history truncated).
- [ ] Click a repo header to view archived worktrees, then a worktree — back returns to the archived view.
- [ ] Cmd-click two worktrees in a specific order to get split view, switch away, then back — pane order matches the original cmd-click order.
- [ ] Buttons disable correctly at the ends of history.

### DIFF
--- a/Sources/TBDApp/AppState+Navigation.swift
+++ b/Sources/TBDApp/AppState+Navigation.swift
@@ -3,7 +3,7 @@ import Foundation
 /// A single navigable view state — either a worktree selection (one or more)
 /// or a repo selection (showing archived worktrees in the detail pane).
 enum NavigationEntry: Equatable {
-    case worktrees(Set<UUID>)
+    case worktrees([UUID])
     case repo(UUID)
 }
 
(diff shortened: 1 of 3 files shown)


cheapsteak:
Addressed both notes in 85d8590: added the ordering invariant comment in applyNavigationEntry and a doc comment on the navigation-state block noting the mutation contract.

PR #73, opened 2026-04-29
Fix: Stop unviewed terminals from being frozen at 80×24

cheapsteak:
## Summary

**The bug.** When TBD spawns a terminal programmatically — `tbd worktree create --prompt-file -`, autonomous Claude tabs from another worktree, anything where the user isn't currently looking at the new tab — the daemon called `tmux new-session` / `tmux new-window` **without `-x`/`-y` flags**. Tmux defaulted to 80×24, `claude` started rendering hard-wrapped output into scrollback at that width, and tmux's own reflow can never undo it (Claude Code emits literal newlines for diff/box rendering, not soft-wrapped lines). When the user later opened the tab at the real width, the early scrollback was permanently stuck narrow.

**The fix.**
- App now publishes the live "main terminal area" size (excluding pinned dock + file panel) into `AppState.mainAreaSize` via a `GeometryReader` preference key on the main slot of `DockSplitView`.
- App passes `cols`/`rows` (computed from that size + SwiftTerm's font cell metrics) on `worktreeCreate`, `terminalCreate`, and `terminalRecreateWindow` RPCs.
- Daemon threads the size into `tmux new-session -x N -y M` and `tmux new-window -x N -y M`. Falls back to 220×50 when params are nil (CLI path / no app connected).
- New `setMainAreaSize` RPC, debounced 300ms after the user stops resizing the main window, broadcasts the new size to the daemon, which iterates all tracked terminals and runs `tmux resize-window` so detached panes stay in sync if the user resizes before viewing them.

Optional fields, no DB migration, fully backward-compatible with older daemons/CLIs.

## Test plan

- [ ] `swift build` clean
- [ ] `swift test` — TmuxManager tests pass (9 new tests cover both branches of the cols/rows conditional + the new resize-window helper)
- [ ] `scripts/restart.sh` (full restart so the daemon picks up new RPC fields), then:
  - [ ] From an existing Claude tab, run `tbd worktree create --prompt-file -` with a prompt that produces wide output. Open the new tab — early scrollback should already be at the main window's width, not 80 cols.
  - [ ] Resize the main window noticeably wider after a worktree was created but before viewing its tab. Open the tab — scrollback should match the new width.
  - [ ] Existing terminals continue to render correctly when the main window is resized.
  - [ ] CLI `tbd worktree create` (no app size info) still creates terminals; daemon falls back to 220×50.

### DIFF
--- a/Sources/TBDApp/AppState+ClaudeTokens.swift
+++ b/Sources/TBDApp/AppState+ClaudeTokens.swift
@@ -98,7 +98,8 @@ extension AppState {
     /// and selects it so the UI switches immediately.
     func swapClaudeTokenOnTerminal(terminalID: UUID, newTokenID: UUID?) async {
         do {
-            let newTerminal = try await daemonClient.swapClaudeTokenOnTerminal(terminalID: terminalID, newTokenID: newTokenID)
+            let size = mainAreaTerminalSize()
+            let newTerminal = try await daemonClient.swapClaudeTokenOnTerminal(terminalID: terminalID, newTokenID: newTokenID, cols: size.cols, rows: size.rows)
             let worktreeID = newTerminal.worktreeID
             terminals[worktreeID, default: []].append(newTerminal)
             let newTab = Tab(id: newTerminal.id, content: .terminal(terminalID: newTerminal.id))
--- a/Sources/TBDApp/AppState+Terminals.swift
+++ b/Sources/TBDApp/AppState+Terminals.swift
@@ -10,7 +10,8 @@ extension AppState {
     /// Create a terminal in a worktree and add a new tab for it.
     func createTerminal(worktreeID: UUID, cmd: String? = nil) async {
         do {
-            let terminal = try await daemonClient.createTerminal(worktreeID: worktreeID, cmd: cmd)
+            let size = mainAreaTerminalSize()
+            let terminal = try await daemonClient.createTerminal(worktreeID: worktreeID, cmd: cmd, cols: size.cols, rows: size.rows)
             terminals[worktreeID, default: []].append(terminal)
             let tab = Tab(id: terminal.id, content: .terminal(terminalID: terminal.id))
             tabs[worktreeID, default: []].append(tab)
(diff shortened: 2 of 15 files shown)


cheapsteak:
Both review items addressed:

- **Medium** (token swap at 80×24): `3bdbded` — `TerminalSwapClaudeTokenParams` now carries optional `cols`/`rows`, threaded through `DaemonClient.swapClaudeTokenOnTerminal` → `AppState+ClaudeTokens` (call site lives there, not `+Terminals`) → `handleTerminalSwapClaudeToken` with the same 220×50 fallback as `handleTerminalCreate`.
- **Minor** (revive size): `730cee7` — `WorktreeReviveParams` now has optional `cols`/`rows`, app passes `mainAreaTerminalSize()`, `handleWorktreeRevive` forwards to the lifecycle method (which already accepted these in this PR).

`swift build` clean, `TmuxManagerTests` (26) all pass. No new tests added — the gating conditional in tmux argv emission is already covered by `testNewWindowCommandWithExplicitSize`/`testNewWindowCommandWithoutSize`; the new wires here are pure pass-through.

cheapsteak:
Second-round review items addressed:

**Fixed:**
- **`handleSetMainAreaSize` archived-worktree iteration**: `1eb3613` — switched to `db.worktrees.list(status: .active)`. Confirms the same pattern `WorktreeLifecycle+Reconcile` and `RPCRouter+RepoHandlers` use. Dead servers from archived worktrees are no longer hit during the resize debounce.
- **220×50 magic number duplication**: `88745fc` — added `TmuxManager.defaultCols` / `defaultRows` siblings to the existing `minCols` / `minRows` constants; replaced the four pane-size fallback sites in `RPCRouter+TerminalHandlers.swift` and `WorktreeLifecycle+Create.swift`. (Did not touch the unrelated `params.lines ?? 50` in `handleTerminalOutput` — that's output line count, not pane size.)
- **`MainAreaSizeKey.reduce` single-source invariant**: `8200cad` — added a comment noting that exactly one view in the hierarchy is expected to post this key.

**Skipped:**
- **tmux `new-window -x/-y` version compat (Medium)**: not addressing in this PR. The flags being honored even on older tmux is a reasonable assumption for our user base, and as the bot itself notes, `new-session -x/-y` is well-established and `new-window` panes inherit session size — so the worst-case fallback path (if it ever became relevant) is to drop the flag from `new-window` only, not a deeper change. Happy to revisit if a real-world install actually fails.

`swift build` clean, `TmuxManagerTests` (26/26) pass.

PR #75, opened 2026-04-29
DX: Cache SwiftPM artifacts to stop CI rebuilding 7 deps every run

cheapsteak:
## Summary

- Every CI run on `macos-15` was re-cloning and recompiling all 7 SwiftPM dependencies (GRDB, swift-nio, SwiftTerm, Highlightr, swift-markdown-ui, swiftui-introspect, swift-argument-parser) from scratch — most of the ~5 min wall time was dependency compilation.
- Add `actions/cache@v4` for `.build`, `~/Library/Caches/org.swift.swiftpm`, and `~/Library/org.swift.swiftpm` with a two-tier `restore-keys`:
  - Exact match on `Package.resolved` + source hashes → fully warm `.build/`, near-instant.
  - `Package.resolved`-only fallback → deps stay built, only changed sources recompile (typical PR case).
  - OS-only fallback → at least the package source cache is reused on cold branches.
- Drop the standalone `swift build` step. `swift test` already builds everything it does plus test targets, so the separate step roughly doubled compile time for no gain.

First run after merge will be cold (same as today). Subsequent runs should land closer to ~1.5–2 min.

## Test plan

- [ ] CI on this PR is the cold run — confirm it still passes end-to-end.
- [ ] After merge, confirm a follow-up PR's CI is materially faster (cache hit visible in the `Cache SwiftPM` step log).
- [ ] If the cache key ever needs to bust (e.g. corrupted `.build`), bumping a comment in the workflow or Package.resolved invalidates it cleanly.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,6 @@ jobs:
           path: |
             .build
             ~/Library/Caches/org.swift.swiftpm
-            ~/Library/org.swift.swiftpm
           key: spm-${{ runner.os }}-${{ env.SWIFT_VERSION }}-${{ hashFiles('Package.resolved') }}-${{ hashFiles('Sources/**/*.swift', 'Tests/**/*.swift') }}
           restore-keys: |
             spm-${{ runner.os }}-${{ env.SWIFT_VERSION }}-${{ hashFiles('Package.resolved') }}-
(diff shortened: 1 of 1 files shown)


cheapsteak:
Addressed both items in 40f9c5d:

- **Toolchain in cache key**: Added a `Capture Swift toolchain version` step that writes `xcrun swift --version | head -1` to `$GITHUB_ENV`, and threaded `${{ env.SWIFT_VERSION }}` through the primary key and both restore-keys so an Xcode update on the runner image won't restore a `.build/` compiled by a different toolchain.
- **Step naming**: Renamed the cache step to `Cache SwiftPM` to match what the PR description references.

Also restored `swift build` (after `swift test`) on a separate point — `swift test` only builds targets reachable from test targets, which excludes `TBDCLI` and the bare `TBDDaemon` executable wrapper (no test target depends on them), so without that step a compile error in either could merge with green CI. Cache carries over from the test step, so it only recompiles those two.

cheapsteak:
Tested empirically on this branch (cold run 40f9c5d, warm run 206b7ba). Closing the loop on the bot's FYI:

**Cache behavior**
- Cold: miss → save 1.23 GB tarball (~26s upload). 4m34s wall clock.
- Warm (no source changes): primary key **hit** → restore 1.23 GB (~24s). 3m15s wall clock (~29% faster).
- Remaining time is test execution itself; not compile-bound.

**Path sizes captured on warm run** (`du -sh` after `swift build`):

| Path | Size |
|---|---|
| `.build` | 1.6 G |
| `~/Library/Caches/org.swift.swiftpm` | 409 M |
| `~/Library/org.swift.swiftpm` | 48 K (empty `configuration/` + `security/` dirs only) |

So the non-`Caches/` path holds only empty scaffolding — confirming the FYI. Dropped it in 60f574d.

PR #84, opened 2026-04-30
Fix: Stop tmux windows from oversizing past the visible terminal pane (Claude's input row was rendering below the viewport)

cheapsteak:
## Summary

- **What's broken:** In some worktrees the terminal looked too wide and Claude's input box rendered below the visible area — typing worked but the prompt was invisible.
- **Why it happens:** Recent commits (`bd549d0`, `c80dfa9`) hoisted the `mainAreaSize` `GeometryReader` to `ContentView`'s entire NavigationSplitView detail slot and forced it to `.frame(maxWidth: .infinity, maxHeight: .infinity)`. That measurement now includes the tab bar, dock column, and file panel — strictly larger than the actual SwiftTerm pane. The daemon feeds it to `tmux resize-window`, which puts the window into **manual size mode**, and SwiftTerm's later `TIOCSWINSZ` from the truly-smaller pane can't shrink it back. Tmux thinks there are more rows than the view actually shows, so the bottom rows clip below.
- **What this PR does:** Reverts those two commits. The `GeometryReader` lives back inside `TerminalContainerView` where it measures the actual pane area. Kept `13c6dfc` (`resize-window` after `new-window`), which independently fixes the original "unviewed terminals stuck at 80×24" problem — new windows get sized to `mainAreaTerminalSize()`, which uses the sensible 1120×776 default (~145×48 cells) when nothing's been measured yet.

## Test plan

- [ ] Open a worktree where the issue reproduced (e.g. "🔀 Reorder Worktrees") — Claude's input row should be visible at the bottom.
- [ ] Resize the app window — terminal should reflow, not over- or under-fill.
- [ ] Show/hide the file panel and dock — terminal should resize sensibly without clipping rows.
- [ ] Create a new worktree with no main-area measurement yet — pane should not be 80×24.
- [ ] `swift test --filter TmuxManagerTests` still green.

### DIFF
--- a/Sources/TBDApp/ContentView.swift
+++ b/Sources/TBDApp/ContentView.swift
@@ -22,58 +22,44 @@ struct ContentView: View {
             NavigationSplitView {
                 SidebarView()
             } detail: {
-                Group {
-                    if !appState.isConnected {
-                        disconnectedView
-                    } else if appState.repos.isEmpty {
-                        emptyStateView
-                    } else if let repoID = appState.selectedRepoID {
-                        RepoDetailView(repoID: repoID)
-                    } else if appState.selectedWorktreeIDs.isEmpty {
-                        Text("Select a worktree or click + to create one")
-                            .foregroundStyle(.secondary)
-                    } else {
-                        HStack(spacing: 0) {
-                            TerminalContainerView()
-                            if showFilePanel, let worktree = selectedWorktree, !worktree.path.isEmpty {
-                                FilePanelDivider(panelWidth: Binding(
-                                    get: { CGFloat(filePanelWidth) },
-                                    set: { filePanelWidth = Double($0) }
-                                ))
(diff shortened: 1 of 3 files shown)


cheapsteak:
Addressed the tab-bar height note in 83e00d2 — pushed the `MainAreaSizeKey` producer down to `SingleWorktreeView.layoutContent` and `MultiWorktreeView`'s grid, so the published size now excludes the tab bar and divider. `.onPreferenceChange` stays at the outer `mainContent` level since preferences propagate up the hierarchy.

cheapsteak:
Addressed in a38a8ce — `MultiWorktreeView` now publishes per-cell size (`cellWidth × cellHeight`) instead of the full grid. Also added a `// MARK: -` separator between the `TerminalContainerView` and `MainAreaSizeKey` doc blocks per the minor note.

PR #87, opened 2026-04-30
Feat: Browse archived worktree conversations and revive into the one you want

cheapsteak:
## Summary

- Archived worktrees now show a nested master-detail history view: archived list on the left, session list and transcript on the right (the existing `HistoryPaneView` reused as-is via a new `transcriptAction` parameter).
- The transcript header gains a "Revive with this session" action — revives the worktree and resumes the chosen Claude session in its primary terminal, in one click. Previously you had to revive blind and never picked up exactly the conversation you wanted.
- Auto-selects the most-recently-archived row on entry and the first session on load, so the view is useful without any clicking. (The same first-session auto-select also applies to active worktrees' history pane.)
- Filter dropdown in the archived header (default on) hides worktrees with no conversations on disk. Daemon scans each worktree's Claude project dir at list time and returns a real `liveClaudeSessionCount`, since the stored `archivedClaudeSessions` field can be stale.
- Just-revived rows linger with a "Reviving…" spinner and "Revived ✓" pill until you navigate away, so the action feels acknowledged.

Spec: `docs/superpowers/specs/2026-04-30-archived-worktree-history-design.md`
Plan: `docs/superpowers/plans/2026-04-30-archived-worktree-history.md`

## Test plan

- [ ] Open a repo's "Archived" sidebar entry. Most-recently-archived row is selected; right pane shows its session list with the first session's transcript loaded.
- [ ] Click a different archived row → sessions update; first session auto-selects and loads.
- [ ] With a session selected, click "Revive with this session". Row shows "Reviving…" spinner, then "Revived ✓"; the new active worktree's primary terminal is resuming that exact conversation.
- [ ] After revive, navigate to any other sidebar entry and back to Archived → lingering row is gone.
- [ ] Right-click a row → context menu Revive works without picking a session.
- [ ] Archived worktree with zero on-disk sessions: right pane shows "No archived sessions" + a plain Revive button.
- [ ] Filter dropdown toggles "Hide worktrees with no conversations". Default on. Worktrees whose stored session IDs point to deleted JSONL files (e.g. `lungfish`, `mollusk` in my workspace) are correctly hidden when the filter is on.
- [ ] Active worktree history pane: opening it auto-selects the first session and renders the transcript.
- [ ] Existing revive flow (without a session pick) still works: takes you to the new active worktree with sessions resumed in stored order.

### DIFF
--- a/Sources/TBDApp/AppState+History.swift
+++ b/Sources/TBDApp/AppState+History.swift
@@ -1,5 +1,8 @@
 import Foundation
 import TBDShared
+import os
+
+private let logger = Logger(subsystem: "com.tbd.app", category: "AppState+History")
 
 // MARK: - HistoryLoadState
 
@@ -114,6 +117,7 @@ extension AppState {
             .flatMap({ $0 })
             .first(where: { $0.id == worktreeID })
         else {
+            logger.warning("reviveWithSession: no archived snapshot for \(worktreeID, privacy: .public)")
             return
         }
         revivingArchived[worktreeID] = .inFlight(snapshot: snapshot)
--- a/Sources/TBDApp/AppState+History.swift
+++ b/Sources/TBDApp/AppState+History.swift
@@ -103,4 +103,56 @@ extension AppState {
             handleConnectionError(error)
         }
     }
+
+    /// Revive an archived worktree and resume the selected Claude session.
+    /// Marks the row as `inFlight` immediately so the archived view can show
+    /// a status pill, then flips to `.done` on success or clears on failure.
+    func reviveWithSession(worktreeID: UUID, sessionId: String) async {
+        // Find the snapshot in archivedWorktrees so we can keep the row visible
+        // after the daemon reconciles the worktree out of the archived list.
+        guard let snapshot = archivedWorktrees.values
+            .flatMap({ $0 })
+            .first(where: { $0.id == worktreeID })
+        else {
+            return
+        }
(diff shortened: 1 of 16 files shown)


cheapsteak:
Addressed all three review items:

**Main: `rightPane` gating on stale field** — fixed in 4b19ecd. Now uses `row.effectiveSessionCount == 0` (which falls back through live count → stored count) instead of `archivedClaudeSessions.isEmpty`. Older archives with nil `archivedClaudeSessions` but real on-disk JSONLs now correctly show the history view.

**Minor 1: `reviveWithSession` doesn't navigate** — aligned in dcbd7d0 by changing the OTHER direction. Made `reviveWorktree` (context-menu path) also stay in the archive view with the lingering 'Revived ✓' pill, instead of auto-jumping to the now-active worktree. The lingering UX was the intended design and applying it consistently to both revive paths is cleaner than splitting.

**Minor 2: `select()` blocks `.done` rows** — fixed in 4b19ecd. Narrowed the guard to `.inFlight` only; users can now browse a just-revived worktree's sessions without navigating away first.

cheapsteak:
Fixed in 47e4b53 — added `guard revivingArchived[id] == nil else { return }` to both `reviveWithSession` and `reviveWorktree` so re-entrant calls are no-ops.

Skipping the minor item about 0-byte JSONL stubs — agreed it's low probability and the existing comment on `countSessionFiles` already flags the limitation.

PR #88, opened 2026-04-30
fix: prevent TBD_WORKTREE_ID leak from daemon env into tmux servers and recreated panes

cheapsteak:
## Root cause

Recreated tmux panes (after daemon restart, via reconcile, or via the `terminal.recreateWindow` RPC) inherited a stale `TBD_WORKTREE_ID` from their tmux server's **global** environment instead of the worktree they actually live in. Notifications from sub-worktrees showed up under the main worktree's row.

`scripts/restart.sh` is normally invoked from inside a TBD-spawned terminal that exports `TBD_WORKTREE_ID=<main UUID>`. The daemon inherits it; the first `tmux new-session -L tbd-<repo>` bakes that var into the tmux server's global env permanently. Fresh terminals were fine because `handleTerminalCreate` inlines `export TBD_WORKTREE_ID='<correct UUID>';` into the spawn command — but recreate paths didn't.

## Fix

**Fix 1 — `Daemon.scrubInheritedTBDEnv()`**: at daemon startup, before any tmux server is spawned, `unsetenv` the four TBD_* vars (`TBD_WORKTREE_ID`, `TBD_PROMPT_CONTEXT`, `TBD_PROMPT_INSTRUCTIONS`, `TBD_PROMPT_RENAME`). Stops the poison at the source for any fresh `tmux new-session`.

**Fix 2 — recreate paths**: explicitly export `TBD_WORKTREE_ID` on every recreated pane regardless of tmux server env state.
- `WorktreeLifecycle.recreateAfterReboot` (all three branches: claude, codex, shell/cmd) — env assignment hoisted to the top of the function.
- `RPCRouter.handleTerminalRecreateWindow` — previously had a comment claiming plain shells don't need it; that's wrong because users run `tbd` CLI and `claude` from those shells.

## Tests

`Tests/TBDDaemonTests/TBDWorktreeIDLeakTests.swift` (new, 6 tests):
1. `scrubInheritedTBDEnv` clears all four sentinel vars.
2. `recreateAfterReboot` claude branch exports the worktree UUID.
3. `recreateAfterReboot` codex branch exports the worktree UUID (and still exports `CODEX_HOME`).
4. `recreateAfterReboot` shell branch exports the worktree UUID.
5. `handleTerminalRecreateWindow` exports the worktree UUID.
6. Regression: `handleTerminalCreate` still exports `params.worktreeID.uuidString`.

`swift test` — all 426 tests pass.

## Verification on a running system

After restarting the daemon (`scripts/restart.sh`) and recreating any window, every claude process under any worktree should report its own worktree's UUID:

```bash
for pid in $(pgrep -f '^.*claude '); do
  env=$(ps eww -o command= -p $pid | grep -oE 'TBD_WORKTREE_ID=[A-F0-9-]+')
  cwd=$(lsof -p $pid 2>/dev/null | awk '$4=="cwd"{print $NF; exit}')
  echo "$pid  $cwd  $env"
done
```

The `TBD_WORKTREE_ID=...` should match the worktree directory in `cwd` for every row.

## Note on existing state

Existing tmux servers spawned before this change still have the poisoned `TBD_WORKTREE_ID` baked into their global env. Users must manually clear them (e.g. `tmux -L tbd-<repo> kill-server`) or reboot to fully drop the leaked state — done deliberately so this fix doesn't disrupt live sessions.

## Test plan

- [x] `swift build` succeeds
- [x] `swift test` succeeds (426/426)
- [ ] Restart daemon, recreate a window in a sub-worktree, verify with the snippet above that the new pane reports the sub-worktree's UUID (not the main one)
- [ ] Trigger a reconcile reboot path (kill tmux server, restart daemon) and verify recreated panes also report the correct UUID

### DIFF
--- a/Sources/TBDDaemon/Daemon.swift
+++ b/Sources/TBDDaemon/Daemon.swift
@@ -23,6 +23,22 @@ public final class Daemon: Sendable {
         self.startTime = Date()
     }
 
+    /// Remove inherited TBD_* environment variables from the daemon's own
+    /// process environment. Called at startup before any tmux server is spawned.
+    ///
+    /// Rationale: tmux servers persist the env they were spawned with as their
+    /// global environment, and that env is then injected into every new window
+    /// (including reboot-recovery recreations). If the daemon inherits e.g.
+    /// `TBD_WORKTREE_ID=<main-uuid>` from a TBD-spawned launcher shell, every
+    /// recreated pane in every sub-worktree would report itself as the main
+    /// worktree, causing notifications to be misattributed.
+    public static func scrubInheritedTBDEnv() {
+        unsetenv("TBD_WORKTREE_ID")
+        unsetenv("TBD_PROMPT_CONTEXT")
+        unsetenv("TBD_PROMPT_INSTRUCTIONS")
+        unsetenv("TBD_PROMPT_RENAME")
+    }
+
     /// Start the daemon: create config directory, clean up stale state,
     /// initialize database and all managers, start servers, reconcile worktrees.
     public func start() async throws {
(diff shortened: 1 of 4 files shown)


cheapsteak:
Addressed both nits in 171f06f:
- Renamed the duplicate `4b` SSH-refresh comment to `4c` in Daemon.swift.
- Added `defer { unsetenv(...) }` to `testScrubInheritedTBDEnv` so the four sentinel TBD_* vars get cleaned up even if `#expect` failures cause us to skip the scrub call. Belt-and-suspenders for the concurrent-test concern.

PR #89, opened 2026-05-03
Feat: Add WIP flag to disable terminal auto-resize (off by default)

cheapsteak:
## Summary

The terminal auto-resize broadcast (#73 → #84 → #85) has been buggy enough in practice that the overall experience is worse than before it landed — chrome-precision arms race for what gets broadcast, plus tmux's `window-size manual` lock-in causing textbox clipping that's persisted through multiple follow-up fixes. Adding an escape hatch so users can opt out while we keep iterating.

- New `enableTerminalAutoResize` AppStorage flag, **default off**.
- When off: `mainAreaTerminalSize()` returns `(0, 0)`. Both the daemon-side `createWindow` resize gate and the `setMainAreaSize` broadcast handler already no-op below the 80×24 minimum, so a single (0, 0) signal cleanly disables the per-create resize *and* the post-create broadcast.
- `scheduleMainAreaSizeBroadcast()` also early-returns when off — belt-and-suspenders that avoids spinning up the debounce `Task`.
- New "Experimental" section in **Settings → General** with a toggle, help text explaining the WIP state and the known clipping bug, and a footer noting that toggling off and restarting is the bailout.

## Test plan

- [ ] Open Settings → General; "Experimental" section appears at the bottom with the WIP toggle off. Help/footer text reads sensibly.
- [ ] With flag off (default): create a terminal in a worktree, resize the app window. tmux windows do not get sized via `tmux resize-window` from TBD — they keep tmux's session/inherited size and react to attached SwiftTerm clients via `TIOCSWINSZ` only. No textbox clipping.
- [ ] With flag on: behavior matches current main (broadcast fires, per-create resize fires). Existing precision/unfreeze logic from #84 / #85 stays in effect.
- [ ] Flip the flag mid-session, restart the app, verify behavior changes.
- [ ] `swift test --filter TerminalAutoResizeFlagTests` clean (2/2 pass — flag-off returns (0, 0); flag-on returns real cell counts).

### DIFF
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -827,10 +827,28 @@ final class AppState: ObservableObject {
         return truncated.isEmpty ? "conductor" : truncated
     }
 
+    /// UserDefaults key for the WIP terminal-auto-resize feature. Off by
+    /// default — the feature broadcasts main-area pixel size to the daemon
+    /// and resizes every tracked tmux window on app resize / terminal
+    /// create. See `mainAreaTerminalSize()` and `scheduleMainAreaSizeBroadcast()`
+    /// for the two enforcement points. Settings UI toggle lives in the
+    /// "Experimental" section of the General settings tab.
+    static let terminalAutoResizeKey = "enableTerminalAutoResize"
+
+    /// Whether the WIP main-area resize broadcast is enabled. Default false.
+    private var terminalAutoResizeEnabled: Bool {
+        UserDefaults.standard.bool(forKey: Self.terminalAutoResizeKey)
+    }
+
     /// Convert the current `mainAreaSize` (pixels) into tmux cell dimensions
     /// using SwiftTerm's font metrics. Floors at the tmux minimum (80x24) so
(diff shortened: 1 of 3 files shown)


cheapsteak:
Addressed in 5442d36:

- **Medium (the real bug):** changed `mainAreaTerminalSize()` to return `(cols: Int?, rows: Int?)`. Returns `(nil, nil)` when flag is off, so the daemon's `?? TmuxManager.defaultCols/defaultRows` (220×50) fallback fires instead of being bypassed by `Optional.some(0)`. All callers already accept `Int?`, so the change is a clean type tightening — no call-site updates needed except the broadcast path which now unwraps via `guard let cols, let rows`.
- **Minor — three restart mentions:** dropped the in-section `Text("Restart the app after toggling…")`. Restart guidance now lives in the toggle's `.help()` tooltip and the section footer, no duplicate.
- **Minor — early-return guard:** added a comment on `scheduleMainAreaSizeBroadcast`'s guard explaining it's belt-and-suspenders for a mid-session toggle, not redundant.

Test updated to assert `nil` instead of `0` and unwrap with `#require` on the flag-on branch.

PR #90, opened 2026-05-03
Feat: Auto-install tbd CLI to ~/.local/bin so commands work from any shell

cheapsteak:
## Summary
- TBDApp's system prompt tells Claude sessions about `tbd worktree create`, `tbd terminal create`, etc. — but the project never put `tbd` on PATH, so those commands failed with "command not found" in any fresh shell.
- Adds a launch-time check and an "Install Command-Line Tool…" menu item that symlinks `~/.local/bin/tbd` at the running daemon's sibling `TBDCLI`. Re-resolves on every launch, so it follows whichever worktree's daemon is currently active.
- After install, if `~/.local/bin` isn't on the user's login-shell PATH, a follow-up dialog shows the exact export line for their shell (zsh / bash / fish). No sudo, no rc-file editing.

## Test plan
- [x] `swift build` clean
- [x] `swift test` — all 446 tests pass (14 new in `CLIInstallerTests`)
- [ ] `rm -f ~/.local/bin/tbd`, run `scripts/restart.sh` from this worktree, verify the launch prompt appears and clicking Install creates the symlink
- [ ] Use the "Install Command-Line Tool…" menu item from a healthy state — verify the "already installed" dialog reports the correct target
- [ ] In a fresh terminal (with `~/.local/bin` on PATH), confirm `tbd worktree list` works

### DIFF
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -225,6 +225,7 @@ final class AppState: ObservableObject {
 
     let daemonClient = DaemonClient()
     let tmuxBridge = TmuxBridge()
+    lazy var cliInstallerCoordinator = CLIInstallerCoordinator(daemonClient: daemonClient)
     private var pollTimer: Timer?
     private var pollCycle = 0
     private var subscriptionTask: Task<Void, Never>?
@@ -457,6 +458,15 @@ final class AppState: ObservableObject {
             pendingDeepLinkID = nil
             navigateToWorktree(pendingID)
         }
+        Task { [weak self] in
+            guard let self else { return }
+            await self.cliInstallerCoordinator.checkOnLaunch()
+        }
+    }
+
+    /// Menu entry point — install or refresh the `tbd` CLI symlink.
+    func installCLITool() async {
+        await cliInstallerCoordinator.runFromMenu()
     }
 
     /// Launch the daemon process and connect.
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -458,9 +458,11 @@ final class AppState: ObservableObject {
             pendingDeepLinkID = nil
             navigateToWorktree(pendingID)
         }
-        Task { [weak self] in
-            guard let self else { return }
-            await self.cliInstallerCoordinator.checkOnLaunch()
+        if didConnect {
+            Task { [weak self] in
+                guard let self else { return }
+                await self.cliInstallerCoordinator.checkOnLaunch()
+            }
         }
     }
 
--- a/Sources/TBDApp/CLIInstallerCoordinator.swift
+++ b/Sources/TBDApp/CLIInstallerCoordinator.swift
@@ -1,0 +1,174 @@
+import Foundation
+import AppKit
+import TBDShared
+import os
+
+private let logger = Logger(subsystem: "com.tbd.app", category: "cli-installer")
+
+@MainActor
+final class CLIInstallerCoordinator {
+    private let daemonClient: DaemonClient
+    private let installer = CLIInstaller()
+
+    init(daemonClient: DaemonClient) {
+        self.daemonClient = daemonClient
+    }
+
+    /// Called once at end of `connectAndLoadInitialState`. Surfaces a one-click
+    /// prompt if the symlink is missing or stale. Silent if everything's healthy.
+    func checkOnLaunch() async {
+        guard let target = await fetchExpectedTarget() else { return }
+        let state = installer.currentState(expectedTarget: target)
+        switch state {
+        case .installed:
+            logger.debug("CLI symlink installed and current at \(self.installer.symlinkPath, privacy: .public)")
+            return
+        case .notInstalled:
+            logger.info("CLI symlink missing — prompting to install")
+            await presentLaunchPrompt(target: target, kind: .missing)
+        case .stale(let current):
+            logger.info("CLI symlink stale: current=\(current, privacy: .public) expected=\(target, privacy: .public) — prompting to refresh")
+            await presentLaunchPrompt(target: target, kind: .stale(current: current))
+        }
+    }
+
+    /// Called from the menu item. Confirms, runs install, presents post-install dialog.
+    func runFromMenu() async {
+        guard let target = await fetchExpectedTarget() else {
+            presentTargetUnavailableAlert()
+            return
+        }
+        let state = installer.currentState(expectedTarget: target)
+        switch state {
+        case .installed:
+            presentAlreadyInstalledAlert(target: target)
+        case .notInstalled, .stale:
+            await performInstall(target: target, confirm: true)
+        }
+    }
+
+    // MARK: - Private
+
+    private enum LaunchPromptKind {
+        case missing
+        case stale(current: String)
+    }
+
+    private func fetchExpectedTarget() async -> String? {
+        let status: DaemonStatusResult
+        do {
+            status = try await daemonClient.daemonStatus()
+        } catch {
+            logger.warning("daemon.status failed during CLI install check: \(error.localizedDescription, privacy: .public)")
+            return nil
+        }
+        guard let exec = status.executablePath, !exec.isEmpty else {
+            logger.info("Daemon did not report executablePath (likely older daemon); skipping CLI install check")
+            return nil
+        }
+        let cli = CLIInstaller.cliPath(forDaemonExecutable: exec)
+        guard FileManager.default.fileExists(atPath: cli) else {
+            logger.warning("Expected TBDCLI binary not found at \(cli, privacy: .public)")
+            return nil
+        }
+        return cli
+    }
+
+    private func presentLaunchPrompt(target: String, kind: LaunchPromptKind) async {
+        let alert = NSAlert()
+        alert.alertStyle = .informational
+        switch kind {
+        case .missing:
+            alert.messageText = "Install the tbd command-line tool?"
+            alert.informativeText = "TBD can add a `tbd` command at ~/.local/bin/tbd so you can launch and control TBD from the terminal. No sudo required."
+        case .stale(let current):
+            alert.messageText = "Refresh the tbd command-line tool?"
+            alert.informativeText = "Your `tbd` symlink at ~/.local/bin/tbd points at \(current), which doesn't match this TBD's CLI. Update it?"
+        }
+        alert.addButton(withTitle: "Install")
+        alert.addButton(withTitle: "Not Now")
+
+        let response = alert.runModal()
+        if response == .alertFirstButtonReturn {
+            await performInstall(target: target, confirm: false)
+        }
+    }
+
+    private func performInstall(target: String, confirm: Bool) async {
+        if confirm {
+            let alert = NSAlert()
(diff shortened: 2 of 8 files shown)


cheapsteak:
Addressed both review items in 7f3981e:

- **Medium (main-thread block)**: `installer.install(target:)` is now offloaded to a `Task.detached(priority: .userInitiated)` inside `performInstall`, so the up-to-2s login-shell PATH probe can't freeze AppKit. `CLIInstaller`/`CLIInstallResult` gained `Sendable` conformance and the `pathProbe` closure became `@Sendable` to make the cross-actor handoff clean.
- **Minor (confusing "stale → itself" message)**: added `CLIInstallState.nonSymlink` for the regular-file-at-symlink-path case. The launch and menu prompts now show "Replace the file at ~/.local/bin/tbd?" instead of the previous self-referential text. New test `currentStateNonSymlinkWhenRegularFileExistsAtPath` covers the branch.

Full suite: 447 tests passing.

cheapsteak:
Addressed all three items:

- **Medium (cooperative pool block)** in fc54d85: rewrote \`defaultLoginShellPathProbe\` as \`async\` using \`withCheckedContinuation\` + \`Process.terminationHandler\` + a \`DispatchQueue\` timeout. \`install()\` and \`pathStatus()\` are now async; the coordinator drops \`Task.detached\` since the async probe yields cooperatively.
- **Minor (didConnect guard)** in 957265e: \`checkOnLaunch\` is now only spawned when \`didConnect\` is true.
- **Minor (fish quoting)** in fc54d85: non-default \`binDir\` now produces \`set -gx PATH '<path>' \$PATH\` so paths with spaces don't word-split. Added \`fishExportLineQuotesNonDefaultBinDirToHandleSpaces\` test.

448/448 tests passing.

cheapsteak:
- **Symlink resolution** fixed in 2070240 — \`resolvedExecutablePath()\` now uses \`resolvingSymlinksInPath().standardizedFileURL.path\`.
- **\`readToEnd()\` orphan-pipe note**: keeping \`-ilc\` rather than dropping to \`-lc\`. The trade-off is real but the other direction: dropping \`-i\` means missing PATH set in \`.zshrc\`/\`.bashrc\` (very common on macOS where users often skip the \`.zprofile\` convention), which would surface as a false "not on PATH" warning. The current 2-second timeout already unblocks the continuation, so the worst case is one parked Foundation dispatch thread until orphans exit — bounded and not on the cooperative pool. Acknowledging the note as informative but not changing.

448/448 tests still passing.

cheapsteak:
Both addressed:

- **Logger subsystem** in ac81451: switched to \`com.tbd.app\` (with \`cli-installer\` category) per the taxonomy in CLAUDE.md / docs/diagnostics-strategy.md.
- **Hardcoded alert paths** in 7df872d: all five alert strings now derive from \`installer.symlinkPath\` (and its parent dir for the "Install tbd to ..." / PATH-warning lines), so the messages stay accurate if the default ever changes.

448/448 tests still passing.

cheapsteak:
All three addressed:

- **Button label** in 620c491: primary button now reads Install / Refresh / Replace to match the prompt's verb.
- **\`ManagedAtomicBool\` rename** in 93feb6a: renamed to \`OnceFlag\`. Avoids the swift-atomics name collision and describes what the flag actually does (gate single-resume on the probe continuation).
- **Tilde-not-expanded in export lines** in 829f3eb: replaced \`displayPath\` (tilde form) with a new \`shellExpandable\` helper that uses \`\$HOME/...\` for under-home paths and the absolute path otherwise. Switched fish from single quotes to double quotes so \`\$HOME\` expands while spaces stay safe. Updated the fish-spaces test to match and added \`bashExportLineUsesHomeForPathsOutsideHome\` for the absolute-path fallback.

449/449 tests passing.

cheapsteak:
Addressed in 7ab9482: switched the PATH probe to take the last non-empty line of stdout. With `-ilc`, interactive shell rc files can emit banners before our `echo $PATH`, so `.first` was at risk of picking up the banner; `.last` is correct since `echo $PATH` is always the final line. No regression for clean shells where there's only one line. 449/449 tests still passing.

cheapsteak:
Fixed in 3af2531: clicking "Not Now" on the install-CLI launch prompt now sticks.

Behavior:
- \`.notInstalled\` + dismissed previously → silent on launch (menu still works)
- \`.notInstalled\` + not dismissed → prompt as before
- \`.stale\` / \`.nonSymlink\` → always prompt (broken-install states the user opted into)
- \`.installed\` → silent, AND clears any prior dismissal so future deletions re-arm the prompt

Persistence is in UserDefaults at \`com.tbd.app.cliInstaller.notInstalledDismissed\`. The dismissal clears on:
1. Successful install via menu or launch prompt
2. Observing \`.installed\` at launch (covers manual installs)

Added 5 \`launchPromptKind\` tests covering each branch of the state × dismissal matrix per the CLAUDE.md "test each branch of a behavior gate" rule. 454/454 tests passing.

cheapsteak:
Both addressed:

- **Dead enum cases** in bd4bdf1: removed \`daemonExecutablePathUnavailable\` and \`cliBinaryNotFound(searched:)\` — never thrown, leftovers from an earlier iteration.
- **Confirm-dialog wording** in 16ea062: routed \`runFromMenu\` through \`presentLaunchPrompt\` so the verb matches state (Install / Refresh / Replace). Menu flow passes \`recordDismissalOnDecline: false\` so a Not-Now click in the menu-invoked dialog doesn't suppress future auto-prompts. The redundant secondary "Install tbd to …?" confirm step is gone, and \`performInstall\` no longer takes a \`confirm\` parameter.

454/454 tests still passing.

cheapsteak:
Both addressed:

- **Static \`resolvedExecutablePath\`** in ee6bd4f: converted from a function to a \`static let\` evaluated once at module load. CWD is now captured at startup rather than on each \`daemon.status\` call, defending against a hypothetical \`chdir\` after init when \`argv[0]\` is relative.
- **Once-per-session check** in 9626eb8: added \`hasCheckedThisSession\` so daemon reconnects from the poll loop don't re-prompt for stale/nonSymlink. The flag latches only *after* a usable target resolves, so a first-call failure (older daemon, missing executablePath) doesn't lock out a healthy reconnect's retry.

454/454 tests still passing.

cheapsteak:
Both addressed:

- **Copy** in d2928c8: \`.nonSymlink\` alert now says \"A file already exists at …\" instead of \"A regular file …\", so the wording is accurate for directories too (technically reachable, even if practically improbable).
- **Sync \`presentAlreadyInstalledAlert\`** in a42f9b0: made the helper \`async\` and dropped the fire-and-forget \`Task\` around the Reinstall path. \`runFromMenu\` now awaits cleanly through to the post-install dialog, and any future error propagation in \`performInstall\` won't be silently dropped.

454/454 tests still passing.

cheapsteak:
Both addressed:

- **`CLIInstallState` Sendable** in 78498fd: matches `CLILaunchPromptKind`'s conformance.
- **Unused `displayPath`** in 69a6232: removed — replaced by `shellExpandable` earlier and never deleted.

454/454 tests still passing.

cheapsteak:
Both addressed:

- **`LocalizedError` conformance** in 3a2f97d: `CLIInstallerError` now provides `errorDescription` so the actual install-failure reason ("create parent: …", "remove existing: …", or underlying FileManager error) reaches both the alert and the log line, instead of the NSError-bridge boilerplate.
- **Unused `absolutize` default** in 92a5932: dropped the dead `homeDir: String = NSHomeDirectory()` default; all callers pass an explicit value.

454/454 tests still passing.

cheapsteak:
Two of three addressed; one declined:

- **Unreadable symlink** in 51814e5: when \`destinationOfSymbolicLink\` throws on a confirmed-present symlink, \`currentState\` now returns \`.stale(currentTarget: symlinkPath)\` instead of \`.notInstalled\`. \`.stale\` always surfaces a prompt regardless of dismissal, so a broken symlink can't be silently swallowed by the dismissal flag.
- **Directory recursive-delete safety** in 8895bb0: \`install()\` now \`lstat\`s the existing entry and throws if it's a directory ("remove it manually before installing"), instead of letting \`FileManager.removeItem\` silently rm -rf based on a "Replace the file" confirmation. Added \`installRefusesToReplaceDirectory\` test that creates a populated directory and confirms install throws and contents remain.
- **Case-insensitive PATH matching** declined: bot acknowledged "in practice this is harmless" — shell rc files write \`\$HOME/...\` and \`\$HOME\` expands to whatever \`NSHomeDirectory()\` returns, so the case will match. Skipping to avoid adding lowercased-comparison complexity for a non-issue.

455/455 tests passing.

cheapsteak:
**\"Replace\" promise vs install() throw for directories** addressed in 45debce: \`presentLaunchPrompt\` now pre-flight \`lstat\`s the symlink path when the kind is \`.nonSymlink\` and shows a separate single-OK warning ("won't replace a directory automatically; remove it manually and try again from the menu") instead of the misleading Replace dialog. \`install()\`'s directory guard stays as defense in depth.

**\"PR bundles unrelated changes\"** is a hallucination — I confirmed by inspecting \`git diff origin/main...HEAD\`. The actual diff for the three files cited is:

- \`AppState.swift\` (+12 lines): adds the \`cliInstallerCoordinator\` lazy var, the \`checkOnLaunch()\` task in \`connectAndLoadInitialState\`, and the \`installCLITool()\` menu entry point. Nothing else.
- \`RPCProtocol.swift\` (+15 lines): adds the optional \`executablePath\` field to \`DaemonStatusResult\`. Nothing else.
- \`RPCRouter+TerminalHandlers.swift\` (+21 lines): adds \`resolvedExecutablePath\` static and wires it into the \`daemon.status\` response. Nothing else.

No \`modelProfile* → claudeToken*\` rename, no \`dismissedProxyWarnings\` removal, no \`autoSuspendClaudeEnabled\` / \`suspendEnabled\` changes anywhere on this branch. Those concerns can be disregarded.

455/455 tests still passing.

cheapsteak:
This round:

- **\"Unrelated diffs bundled into the PR\"** — confirmed hallucination, second round in a row. I ran \`gh pr diff 90\` directly: it correctly returns three-dot diffs against merge-base, just the 8 PR-touched files, no \`modelProfile*\`/\`dismissedProxyWarnings\`/\`autoSuspendClaudeEnabled\` content. main is 5 commits ahead of the merge-base — including \`da487fd Feat: Per-terminal model profiles\` which legitimately added \`modelProfile*\` types — so a sub-agent in the review skill is most likely reading HEAD's full file contents via \`Read\` and conflating pre-existing-on-HEAD types with PR additions. Workflow plumbing is fine; this looks like a sub-agent prompt issue in the upstream \`code-review@claude-code-plugins\` skill.
- **\`readToEnd()\` orphan-pipe note** — already a known/acknowledged tradeoff (kept \`-ilc\` so we capture \`.zshrc\`/\`.bashrc\` PATHs; timeout unblocks the continuation; only impact is one parked Foundation thread until orphans exit).
- **\`absolutize\` \`relativeTo:\` is inert when expectedTarget is absolute** — addressed in a59e4ae with a brief comment explaining the combination.

455/455 tests still passing.

PR #94, opened 2026-05-04
Feat: Per-terminal model profiles — run Codex or a passthrough proxy

cheapsteak:
## Summary
- Generalize Claude tokens into **model profiles** so any TBD tab can target any Anthropic-compatible endpoint — Claude direct (today's path), Codex/GPT-5 via a user-managed [claude-code-router](https://github.com/musistudio/claude-code-router), or a logging passthrough proxy that forwards unchanged to api.anthropic.com. TBD never spawns or supervises the proxy; it just injects `ANTHROPIC_BASE_URL` / `ANTHROPIC_MODEL` (and the auth env var) per terminal via tmux's `-e` channel.
- Two-preset add-profile UI (Claude direct / Anthropic-compatible proxy). **`baseURL` is required, `model` is optional** so pass-through proxies forward whatever model Claude Code negotiates rather than have a hardcoded one override it. Quiet tab badge surfaces only when a tab's profile differs from its worktree default. Mid-session swap menu lists profiles. Per-repo override picker generalizes today's per-repo Claude-token override.
- TCP-connect health probe (HTTP GET would 404 on api.anthropic.com and most proxies → false-positive warnings). Save-time non-blocking inline warning + post-spawn dismissable terminal-panel banner; dismissal persists in `AppState.dismissedProxyWarnings` so it survives view reconstruction. v15 GRDB migration renames the table + id columns and adds `base_url`/`model`; existing rows behave identically (nil endpoint = today's path).

## Test plan
- [ ] `swift build` clean
- [ ] `swift test` green
- [ ] `scripts/restart.sh` brings up exactly one TBDDaemon + TBDApp from this worktree
- [ ] Existing Claude OAuth profile spawns terminals normally; no badge
- [ ] Add proxy profile pointing at a closed port → save shows non-blocking ⚠️ warning, save proceeds
- [ ] Add proxy profile pointing at a real passthrough (model blank) → terminal forwards Opus/Sonnet selection unchanged through the proxy
- [ ] Tab using proxy profile shows the badge when its worktree default is Claude-direct
- [ ] Banner appears ~500ms after spawn for unreachable proxy; dismissing it sticks across tab switches
- [ ] Resume an existing Claude session → pinned profile preserved (no model swap)
- [ ] No CCR-specific text in TBD UI (proxy-agnostic)

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=AB34701E-AC35-42F7-91CC-FF1DA47CDE8D)


cheapsteak:
Thanks for the review! Addressed in this branch (not yet pushed at time of writing):

**Medium**
- M1 — OAuth verification warning swallowed at line 357: fixed in `b204573`. The sheet now surfaces the warning inline via `errorMessage` and stays open instead of dismissing, mirroring the proxy-probe `.warn` path. Cancel still dismisses.
- M2 — Empty token accepted for proxy profiles: fixed in `151c295`. Added an explicit `guard !trimmed.isEmpty` before the kind-selection block in `handleModelProfileAdd`, plus two daemon tests (empty and whitespace-only token with `baseURL` set).

**Minor**
- MIN1 — `UUID(uuidString: id)!` in `ModelProfileRecord.toModel()`: respectfully pushing back on this one. Silently substituting a fresh `UUID()` on bad data would be strictly worse than crashing — it would corrupt the record graph silently (orphaning keychain entries, breaking foreign-key joins) and produce mysterious downstream bugs. The force-unwrap matches the project-wide pattern across every `*Record.toModel()` site in `Sources/TBDDaemon/Database/`, where IDs are always written via `uuidString` and therefore guaranteed parseable. If we ever want to harden this, it should be a single repo-wide change with a proper invariant-violation log + abort, not a one-off `?? UUID()`.
- MIN2 — `EditEndpointSheet.save()` dismissed on error: fixed in `0d9f4b7`. Mirrors the `priorAlert`/`newAlert` pattern from `AddModelProfileSheet`; the alert is rerouted into the inline `probeStatus = .warn` slot and the sheet stays open.
- MIN3 — `probeWarningMessage` return type: fixed in `2f9a06d`. Verified both branches return non-empty; tightened to `String` and dropped the `if let` binding (and dead `.idle` else) at both call sites.
- MIN4 — `terminal!` force-unwrap in `TabBar`: fixed in `74071e2`. Bound `terminal` in the `if let` head so the rest of the block uses the non-optional reference.

Verified via `swift build` and `swift test --filter "ModelProfile|ProbeWarning|ProfileBadge"` (81 tests passing).

PR #96, opened 2026-05-05
Feat: Rename projects from the sidebar with :emoji: autocomplete

cheapsteak:
## Summary
- Right-click a project header in the sidebar → **Rename...** to edit the display name inline, with the same `:emoji:` autocomplete (frecency-ranked) used by worktree rename.
- Pulled the inline-edit + emoji picker out of `WorktreeRowView` into a reusable `RenameableLabel` component so both worktree and project rename share one implementation.
- Added matching `tbd repo rename <id> <new-name>` CLI for parity with `tbd worktree rename`.

Display-name only — no on-disk path, git remote, or worktree slot changes. Reuses the existing `repo.displayName` column, no migration.

Spec: `docs/superpowers/specs/2026-05-04-rename-projects-design.md`

## Test plan
- [ ] Right-click a repo header → "Rename..." appears
- [ ] Click it → inline text field with current name; type `:rocket:` → emoji autocomplete shows; arrow keys navigate; Enter inserts
- [ ] Enter without active picker commits; Esc cancels; clicking elsewhere blurs and commits
- [ ] New name persists across `scripts/restart.sh`
- [ ] `tbd repo rename <uuid> "🎯 New Name"` works end-to-end
- [ ] Worktree rename still behaves identically (RenameableLabel refactor regression check)

## Notes
- Tests for `RepoStore.rename` were written but not executed in this branch — pre-existing `WorktreeReviveReorderTests.swift` (from #87) imports XCTest, which won't compile on Command Line Tools alone. Worth a separate cleanup to migrate to Swift Testing.
- Noticed a pre-existing crash in `tbd repo list` (non-`--json`): `String(format: "%-36s ...", uuidString as NSString, ...)` — `%s` expects a C string, passing `NSString` segfaults in `_platform_strlen`. Unrelated to this branch (lives since `de3b521`); flagging for a separate fix.

open in TBD: \`tbd://open?worktree=281EF37E-8278-4BBD-952F-0048BD831247\`

### DIFF
--- a/Sources/TBDApp/AppState+Repos.swift
+++ b/Sources/TBDApp/AppState+Repos.swift
@@ -51,6 +51,21 @@ extension AppState {
         }
     }
 
+    /// Rename a repo's display name. Updates local state optimistically before issuing the RPC.
+    func renameRepo(id: UUID, displayName: String) async {
+        if let idx = repos.firstIndex(where: { $0.id == id }) {
+            var repo = repos[idx]
+            repo.displayName = displayName
+            repos[idx] = repo
+        }
+        do {
+            try await daemonClient.renameRepo(id: id, displayName: displayName)
+        } catch {
+            logger.error("Failed to rename repo: \(error)")
+            handleConnectionError(error)
+        }
+    }
+
     /// Update per-repo instruction fields. Returns true on success.
     @discardableResult
     func updateRepoInstructions(repoID: UUID, renamePrompt: String?, customInstructions: String?) async -> Bool {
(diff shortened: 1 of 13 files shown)


cheapsteak:
Addressed minor #2 in d1051c6: added the same not-found guard `handleRepoRemove` / `handleRepoUpdateInstructions` use, returning a descriptive `RPCResponse(error:)` instead of letting GRDB's `DatabaseError` propagate. Skipped minor #1 (`repoRenamed` delta not handled in `AppState.handleDelta`) — the spec scoped cross-window sync out, and the existing `.worktreeRenamed` is in the same state.

PR #97, opened 2026-05-05
Remove ExpandingRow custom tooltip; use native .help() tooltip

cheapsteak:
## Summary
- Deletes `Sources/TBDApp/Helpers/ExpandingRow.swift` entirely (the `NSPanel + NSHostingView + addChildWindow` hover popover).
- Replaces the single `.expandingRow { ... }` callsite in `WorktreeRowView.swift` with SwiftUI's built-in `.help(worktree.displayName)`.
- Net: -319 / +1 lines.

## Why
Crash forensics shipped in PR #92 confirmed the 232x175 hover panel as the source of the recurring AppKit constraint-budget crash. From `~/Library/Logs/TBD/last-exception.txt`:

> `<SwiftUI.AppKitWindow: ...> {{0, -196}, {232, 175}}`

Three prior commits tried to patch the panel without removing it:
- `acb057f` — fix MainActor isolation warnings in exception preprocessor
- `b770ef3` — break AutoLayout loop in sidebar hover-expand panel
- `3fa2cc6` — persist NSHostingView in expanding row panel

None worked, because `addChildWindow` makes the panel participate in the parent split-view's layout cycle, exhausting AppKit's per-window constraint-update budget. No amount of patching inside the same design fixes it.

## Design change, not a refactor
This deliberately swaps the seamless inline expansion for a native macOS tooltip (the standard yellow rounded-rect AppKit renders for `NSView.toolTip`). Trade-off accepted: visual fidelity for crash-free behavior. `.help` always shows the full text on hover regardless of truncation, which matches the original intent of "show the entire worktree name".

## Test plan
- [x] `swift build` clean
- [ ] App launches via `scripts/restart.sh`
- [ ] Native tooltip appears when hovering a truncated worktree row
- [ ] No more AppKit constraint-budget crashes (will confirm by running for a while)

**Do not merge** — Chang reviews and merges.


cheapsteak:
Addressed the rename-tooltip nit in 706c549 — `.help(isEditing ? "" : worktree.displayName)`. The non-truncated-row tooltip behavior remains as documented in the PR description.

PR #99, opened 2026-05-05
Feat: Read Claude conversations as a chat, not a terminal grid

cheapsteak:
## Summary
- New "Transcript" button on every Claude tab toolbar opens a chat-style read-only pane next to the terminal. User prompts and Claude prose render in chat bubbles with markdown (inline emphasis, links, fenced code blocks); tool calls, thinking, system reminders, and slash commands render as activity rows. Polls every 1.5s and re-targets on `/clear` / `/compact` / suspend-resume rollovers.
- Curated tool renderers for the high-frequency ones (Read / Edit / MultiEdit / Write / Bash / Grep / Glob / Agent) plus a generic JSON-inspector fallback for everything else (MCP servers, custom skills, …). Edit hunks render as syntax-highlighted GitHub-style diffs via Highlightr (~30 languages). Bash output sits in a height-capped scroll container with independent expand-container and load-full-output affordances. Body content is truncated daemon-side (2 KB / 20 lines) with a `terminal.transcriptItemFullBody` RPC for on-demand expansion.
- Subagent (`Task`) dispatches expose a "▶ Show N subagent activities · <agentType>" disclosure beneath the parent card; expanding renders the inner conversation through the same `TranscriptItemsView`, recursively. Cycle detection prevents pathological inputs from looping. The History pane (archived sessions) gets the same renderer via a single shared dispatcher — `[ChatMessage]` is gone from the wire format, replaced by a structured `[TranscriptItem]` parsed daemon-side.

## Test plan
- [ ] Click the Transcript button on a Claude tab → pane opens to the right, populates within ~1.5s. Shell tabs don't show the button.
- [ ] Send a message in the terminal → bubble appears in the transcript within ~1.5s. Run `/clear` → transcript resets.
- [ ] Markdown rendering: `**bold**`, `*italic*`, `` `inline code` ``, `[link](url)`, fenced ` ```swift ` blocks all render correctly in chat bubbles.
- [ ] Tool cards by family: Read (collapsed by default, click to expand), Edit / MultiEdit (red/green diff hunks with syntax highlighting), Write (file path + line count + first 15 lines preview), Bash (command + scroll-capped output, expand-container button + `Show full output` footer), Grep / Glob (pattern + path chip + matches), Agent (description + agent_type chip + prompt preview).
- [ ] A `Task` dispatch shows `▶ Show N subagent activities · <agentType>`; expanding reveals an indented inner timeline. Nested subagents expand recursively.
- [ ] Truncated tool output shows `… N more chars · Show full output` footer; clicking fetches the un-truncated body.
- [ ] History tab on an archived worktree renders the same way (no truncation footers since there's no live terminal context).

specs:
- `docs/superpowers/specs/2026-05-04-claude-tab-live-transcript-design.md`
- `docs/superpowers/specs/2026-05-05-rich-transcript-rendering-design.md`

open in TBD: \`tbd://open?worktree=E8BD758F-F18E-4403-B5FE-50434995FEAD\`


cheapsteak:
Addressed bot feedback. Each in its own commit:

- **M1 — `messagesEqual` drops in-flight→complete transitions** → `2f32a83` Made `TranscriptItem` / `Subagent` / `ToolResult` `Equatable` (auto-synthesized) and use array `==` in the poll-loop check. Spinners now flip when results land.
- **M2 — `sessionTranscripts` unbounded growth** → `1b64210` Added a 50-entry LRU via a parallel access-order array and `touchSessionTranscript()`; called from both write sites (`AppState+History.swift` and `LiveTranscriptPaneView.swift`).
- **m1 — `BashCard.headerSummary` always appends `…`** → `80c3b17` Only append when `trimmed.count > 60`.
- **m2 — `DateFormatter` allocated on every call** → `c20d2b7` Cached static formatters for time / weekday / month-day / month-day-year. Behavior unchanged.
- **m3 — `nonisolated(unsafe)` on Highlightr lacks justification comment** → `1f3b4a4` Added a comment explaining the main-actor-only constraint and the internal serialization.

`swift test` 575/575 green; `swift build` clean.

cheapsteak:
Addressed second round:

- **Medium — subagent 'Show full output' silently fails** → \`98ac492\` Extended \`TranscriptParser.lookupFullBody\` with a multi-path overload; the handler now enumerates \`<sessionID>/subagents/agent-*.jsonl\` and tries the parent JSONL plus every subagent file in order. Subagent-resident tool_use_ids now resolve.
- **Minor — \`ReadCard\` shows \`lines X–X\` when limit absent** → \`8b80b97\` Renders \`from line N\` for the offset-without-limit case (read-to-EOF).

\`swift test\` 576/576; \`swift build\` clean.

cheapsteak:
Round 3 addressed:

- **Medium — nested ScrollView traps wheel + breaks LazyVStack laziness** → \`2da779c\` Hoisted the \`ScrollView\` out of \`TranscriptItemsView\`; the two top-level call sites (\`LiveTranscriptPaneView\`, \`SessionTranscriptView\`) now own it, and \`SubagentDisclosure\`'s recursive \`TranscriptItemsView\` call is bare.
- **Medium — \`agentIDByToolUseID\` mis-attributes parallel Task dispatches** → \`df47440\` Only set the mapping when the user line has exactly one \`tool_result\` block. Multi-block lines (rare today; parallel Tasks in one assistant turn) skip the mapping rather than silently corrupt — the parent Task tool_use just renders without a subagent disclosure.
- **Minor — \`ReadCard\` line range off-by-one** → \`736107c\` \`lim\` is a count, so the last line is \`off + lim - 1\`.
- **Minor — \`depth\` threaded but not gating render** → \`c211763\` Render a small \`… nested too deep\` placeholder beyond depth 8.

Skipped per the bot's own "cosmetic and mostly passable" framing: dark-mode Highlightr theme switch (deferred).

\`swift test\` 576/576; \`swift build\` clean.

cheapsteak:
Round 4 addressed:

- **Medium — diff hunks invisible in dark mode** → \`63876e7\` Two cached \`Highlightr\` instances (light + dark themes); pick at call time via \`NSApp.effectiveAppearance\`. Symmetric luminance clamp now also catches overly-dark colors on dark backgrounds (luminance < 0.15).
- **Medium — \`LazyVStack\` recycles per-card \`@State\`** → \`c3bb9f7\` Switch \`TranscriptItemsView\` to plain \`VStack\` so expand state and full-fetch results survive scroll. Acceptable for transcripts up to ~200 items.
- **Minor — unreachable "Jump to latest" pill** → \`71138e2\` Removed the dead UI block; kept the \`autoscrollEnabled\` state with a TODO so the deferral is explicit.
- **Minor — nondeterministic subagent file order** → \`0cac571\` Sort subagent JSONLs by name before building the lookup paths.
- **Minor — \`EditCard\` "all" chip too eager** → \`ff602e2\` Use \`allSatisfy\` with \`!hunks.isEmpty\`.
- **Minor — UUID() fallback breaks \`messagesEqual\`** → \`97de8ca\` Stable per-line index fallback (\`"line-\(i)"\`) instead of fresh \`UUID()\` per parse.

\`swift test\` 576/576; \`swift build\` clean.

cheapsteak:
Round 5 addressed:

- **Medium — daemon re-parses entire JSONL every poll** → \`cb3957e\` Added \`actor TranscriptParseCache\` keyed by file path with mtime+size fingerprint. Applied to both \`terminal.transcript\` and \`session.messages\` handlers; cache misses re-parse and write through.
- **Medium — \`highlightLines\` uncached on every render** → \`c7bc03d\` Process-wide LRU keyed by \`(isDark, language, content)\` with 500-entry cap. The plain-line fallback paths bypass the cache (they're already cheap). Caps Highlightr/JSCore cost to one call per unique snippet per appearance per session.
- **Minor — \`.thinking\` silently dropped** → \`3b49aa1\` Added a comment explaining the deliberate suppression in \`TranscriptItemsView\`.
- **Minor — \`Date\` extensions buried in \`HistoryPaneView\`** → \`01d5d2b\` Moved to a dedicated \`Sources/TBDApp/Panes/Transcript/DateFormatting.swift\`.

\`swift test\` 576/576; \`swift build\` clean.

cheapsteak:
Round 6 addressed:

- **Medium — depth guard fires N times** → \`7d973e6\` Hoisted the \`depth >= 8\` short-circuit out of \`rowFor()\` and into \`TranscriptItemsView.body\`. One placeholder for an over-deep subagent, not N.
- **Medium — \`TranscriptParseCache\` unbounded** → \`158668b\` Added a 50-entry LRU (\`order: [String]\` + \`cap\`) mirroring the app-side \`AppState.touchSessionTranscript\` pattern. Both \`terminal.transcript\` and \`session.messages\` populate it; eviction is now bounded.
- **Minor — \`WriteCard\` silently truncated to 15 lines** → \`20dfd5b\` Replaced the preview with a BashCard-style height-capped scroll container (120pt) plus an expand toggle. The full content is already in \`inputJSON\` so no RPC needed — just unfold the view.

\`swift test\` 576/576; \`swift build\` clean.

cheapsteak:
Round 7 addressed:

- **Medium — \`session.messages\` accepts arbitrary file path** → \`f208185\` Added a standardized-prefix check that the path is under \`~/.claude/projects/\`, matching \`handleTerminalTranscript\`'s trust boundary. \`standardizedFileURL\` resolves \`..\` and symlinks before comparison.
- **Medium — \`SlashCommandRow\` is dead code** → \`b39cf8d\` Deleted the file; replaced the dispatcher arm with a commented \`EmptyView()\`. The \`.slashCommand\` enum case stays in shared types for Codable safety; the parser continues to fold envelopes into \`.userPrompt\`.
- **Minor — vestigial \`ZStack\` after pill removal** → \`c7f0de7\` Inlined the lone \`ScrollViewReader\` child in \`transcriptWithAutoscroll\`. If/when scroll-freeze detection lands, the overlay structure comes back with it.

\`swift test\` 576/576; \`swift build\` clean.

cheapsteak:
Round 8 addressed:

- **Medium — \`SubagentDisclosure\` count includes hidden items** → \`ee8ac2c\` Centralized the visibility rule in a free \`isHiddenInTranscript(_:)\` helper alongside (now-internal) \`hiddenToolNames\`. The disclosure label and the \`.disabled\` check both filter through it, so "Show N activities" reflects what will actually render.
- **Minor — user-typed XML misclassified as system** → \`3043d2f\` Added an \`isRealUserMessage\` bail-out before \`classify\`'s unknown-tag fallback. \`<html>my page</html>\` from a user now flows to a chat bubble instead of an "info" activity row. Future unknown injections gracefully degrade to plain user prompts. Two new tests + retitled the previous "returns .other" expectation.

\`swift test\` 578/578; \`swift build\` clean.

cheapsteak:
Round 9 addressed (5 commits, see SHAs in commits). Pushing back on **M2** (live subagent progress) — the suggested cache-bypass-during-in-flight-Tasks doesn't actually surface progress earlier:

The parser links a Task `tool_use` to its `agent-<agentID>.jsonl` via `toolUseResult.agentId` from the matching `tool_result` line. Until that tool_result is written (i.e. the Task completes), there's no agentID, so the parser emits `.toolCall(result: nil, subagent: nil)` and no `SubagentDisclosure` is rendered. Re-parsing the unchanged parent yields the same item — the cache bypass costs work without revealing anything.

Surfacing live subagent progress would need a separate heuristic — e.g., scanning `subagents/` for the most-recently-modified `agent-*.jsonl` and tentatively associating it with the in-flight Task tool_use. That's a feature addition with its own design questions (multiple parallel Tasks? mis-attribution risk?), not a cache fix. Deferring to a follow-up.

PR #100, opened 2026-05-06
Perf: Stop big Write inputs from duplicating MBs across two transcript caches

cheapsteak:
## Summary

Two follow-ups from PR #99's bot review.

- **Tool-call inputs were unbounded.** Tool *results* in the transcript are size-capped, but inputs were serialized verbatim — a single `Write` of a real source file pushed hundreds of KB through the daemon→app `terminal.transcript` channel and stuck in both the daemon's `TranscriptParseCache` (50 sessions) and the app's `sessionTranscripts` LRU (50 sessions), getting re-sent on every 1.5s poll. Several `Write`s in one session = several MB duplicated, on every tick. Now the parser walks each tool input dict and truncates large string values using the existing `bodyCharCap`/`bodyLineCap`. The `inputJSON` stays valid JSON so cards keep parsing; a new `inputTruncatedTo: Int?` mirrors `ToolResult.truncatedTo`. The existing `terminal.transcriptItemFullBody` RPC gains a `<tool_use_id>#input` itemID form for on-demand fetch. `WriteCard`/`EditCard`/`AgentCard`/`BashCard` show the same `TruncationFooter` they already use for results.
- **`JSONDecoder` was being re-allocated on every render** for six tool cards. Hoisted to `private static let decoder = JSONDecoder()`, matching `EditCard`'s existing pattern.

Spec: [`docs/superpowers/specs/2026-05-05-transcript-input-cap-design.md`](docs/superpowers/specs/2026-05-05-transcript-input-cap-design.md)

## Test plan

- [x] `swift build` clean
- [x] `swift test` — 580 tests pass, including 4 new parser tests (`inputTruncatesLargeStringField`, `inputNotTruncatedWhenSmall`, `multiEditNestedStringIsTruncated`, `lookupFullBodyWithInputSuffix`)
- [x] `scripts/restart.sh` — daemon and app come up cleanly
- [ ] Drive a real `Write` of a large file in the app; confirm the truncation footer appears below the file body and tapping it loads full content
- [ ] Drive a `Task`/Agent invocation with a multi-KB prompt; confirm same behavior on `AgentCard`

open in TBD: \`tbd://open?worktree=9ADCC356-D87B-45D5-83FA-2C4B5D10F7CD\`

### DIFF
--- a/Sources/TBDApp/Panes/Transcript/AgentCard.swift
+++ b/Sources/TBDApp/Panes/Transcript/AgentCard.swift
@@ -8,12 +8,14 @@ import TBDShared
 struct AgentCard: View {
     let id: String
     let inputJSON: String
+    let inputTruncatedTo: Int?
     let result: ToolResult?
     let timestamp: Date?
     let terminalID: UUID?
 
     @State private var expanded = false
     @State private var fullResultText: String? = nil
+    @State private var fullInputJSON: String? = nil
     @EnvironmentObject var appState: AppState
 
     private struct Input: Decodable {
@@ -25,7 +27,7 @@ struct AgentCard: View {
     private static let decoder = JSONDecoder()
 
     private func decodeInput() -> Input? {
-        guard let data = inputJSON.data(using: .utf8) else { return nil }
+        guard let data = (fullInputJSON ?? inputJSON).data(using: .utf8) else { return nil }
         return try? Self.decoder.decode(Input.self, from: data)
     }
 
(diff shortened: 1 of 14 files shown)


cheapsteak:
Addressed both follow-ups from the review:

- **GenericToolCard plumbing** — `inputTruncatedTo` now flows through `TranscriptItemsView` to `GenericToolCard` (9fb4327). MCP tools and unrecognized tools get the same truncate-and-fetch treatment as the curated cards.
- **EditCard error path** — lifted the input `TruncationFooter` out of the non-error branch so it shows alongside both the diff hunks and the error text (9812355).

Skipped the dead-assignment note in `TranscriptParser.lookupFullBody` per the bot's "no action needed" framing.

cheapsteak:
Addressed the hot-path perf nit (ae11e40) — `originalJSON` is now only computed when truncation actually happened, saving one serialization + one String allocation on every non-truncated tool call.

cheapsteak:
Addressed the line-count UX miss (a6c35b9) — when content is truncated, the header now shows e.g. "≥20 lines" until the user expands.

Skipping the `os.Logger.debug` on the `inputTruncatedTo` re-serialization failure path: the bot's own analysis flags it as not happening in practice (`rawInput` is a just-deserialized Foundation object), and adding a log line for an impossible case is more clutter than insight.

PR #107, opened 2026-05-08
Feat: Read AskUserQuestion as a Q&A, not a wall of JSON

cheapsteak:
## Summary
- Previously every `AskUserQuestion` Claude makes fell through to `GenericToolCard` — a wrench-icon row that, when expanded, dumped the questions, options, and result text as raw nested JSON. Once Claude started asking multi-choice questions, transcripts stopped reading like a conversation.
- New `AskUserQuestionCard` renders each question as an assistant-styled chat bubble, paired with the user's answer in a regular user-style bubble. Tapping the question bubble expands it to show the original options with the picked one highlighted (radio + checkmark).
- Letter-shorthand answers like `"C"` are resolved to the full label (`"C. Loop until satisfied or budget"`) so the user bubble carries the meaning, not just the letter. Free-form "Other" answers render verbatim.
- Tests cover the result.text scanner and option matcher: letter-shorthand match, full-label match, free-form fallback, and multi-question parsing.
- `Theme.chatBubble` and `Color.chatBubbleInlineCode` had `private` removed in `ChatBubbleView.swift` so the answer bubble can reuse the exact same Markdown theme as real user messages.

## Test plan
- [ ] Open the maven-dashboard history thread with the `paab-agent-pipeline-design` session — multiple `AskUserQuestion` calls render as Q&A blocks (not wrench rows)
- [ ] Tap a question bubble; chevron flips and options appear inline with the picked option highlighted
- [ ] Free-form answer (e.g. user typed prose under "Other") renders verbatim in the answer bubble; expanding the question shows no option highlighted
- [ ] Multi-question call (rare) renders one Q + one A pair per question
- [ ] Pending state (no answer yet, before user responds) shows the question and a `Waiting for response…` placeholder on the right

open in TBD: \`tbd://open?worktree=CEED87C2-E1AD-4C97-B451-3EF7E88CC6CF\`

### DIFF
--- a/Sources/TBDApp/Panes/Transcript/AskUserQuestionCard.swift
+++ b/Sources/TBDApp/Panes/Transcript/AskUserQuestionCard.swift
@@ -95,8 +95,21 @@ struct AskUserQuestionCard: View {
                         match: match,
                         rawAnswer: answer,
                         options: q.options,
+                        pendingAnswer: result == nil,
                         timestamp: timestamp,
                         bubbleID: "\(self.id)#answer\(idx)"
+                    )
+                }
+                // If we got a result back but the parser produced nothing
+                // (e.g. Claude Code changed the canonical prefix), surface
+                // the raw text once so the user can still see what came back.
+                if result != nil, parsedAnswers.isEmpty {
+                    ChatBubbleView(
+                        item: .userPrompt(
+                            id: "\(self.id)#raw-result",
+                            text: resultText ?? "",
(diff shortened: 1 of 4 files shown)


cheapsteak:
Addressed all three observations:
- Single match() computation per question — a291653
- Stop the spinner when result is unparseable; show raw result as fallback — 4eb13be
- Case-insensitive letter shorthand + test — dac3272

PR #108, opened 2026-05-08
Feat: One viewer pane, not five — and it auto-refreshes when the file changes

cheapsteak:
## Summary
- Cmd-clicking N relative file paths in a terminal used to open N viewer panes, each one squeezed smaller than the last. Now there is at most one code-viewer pane per tab — subsequent clicks swap its content (preserving the pane's id, size, and position) instead of splitting again. After an explicit close, the next click splits a fresh pane as before.
- Whichever file the viewer is showing is now watched on disk and live-refreshes within ~150ms of any external write — debounced, atomic-save aware (vim `:w`, prettier, etc.), with FD and watcher cleanup tied to the SwiftUI view's lifetime.
- Two fixes for crashes the new lifecycle work surfaced: a pre-existing latent reentrancy in `PanePlaceholder.onPreferenceChange` (synchronous `@State` writes inside the preference-graph callback) deferred to the next runloop tick; and a Combine-on-non-main teardown crash from `@StateObject` + `@MainActor ObservableObject` + `@Published` resolved by making `FileWatcher` a plain `Sendable` class with a callback, owned via `@State` (which matches the pattern used by GitHub Copilot for Xcode and cmux for the same problem).

Spec at `docs/superpowers/specs/2026-05-07-single-viewer-file-watch-design.md` walks the design and the leak-prevention reasoning. Unit tests added for `LayoutNode.firstPaneID` / `replacingContent`, `routeFileClick`, and `FileWatcher` lifecycle balance.

## Test plan
- [x] Open a worktree, terminal, cmd-click 5 different `.md` files in sequence → exactly one viewer pane, size unchanged across clicks.
- [x] With a file in the viewer, externally edit it (`echo >>` and `vim :w` to cover atomic saves) → viewer updates within ~150ms each time.
- [x] Close the viewer pane via its X button → no SIGTRAP. Cmd-click another file → fresh viewer splits open.
- [x] `swift build` clean. `swift test` runs the new unit suites in CI (local runner is broken on this machine due to a CommandLineTools / Testing-framework runtime quirk, unrelated to this PR).

open in TBD: \`tbd://open?worktree=C81B77A1-5276-454E-AFA2-6095700E435C\`

### DIFF
--- a/Sources/TBDApp/Panes/CodeViewerPaneView.swift
+++ b/Sources/TBDApp/Panes/CodeViewerPaneView.swift
@@ -258,6 +258,11 @@ struct CodeViewerPaneView: View {
                 selectedFiles = [path]
             }
         }
+        .onChange(of: path) { _, newPath in
+            if !newPath.isEmpty && FileManager.default.fileExists(atPath: newPath) {
+                selectedFiles = [newPath]
+            }
+        }
     }
 
     private var emptyState: some View {
--- a/Sources/TBDApp/Panes/CodeViewerPaneView.swift
+++ b/Sources/TBDApp/Panes/CodeViewerPaneView.swift
@@ -312,15 +312,22 @@ private struct FilePreviewView: View {
     let filePath: String
     let showSourceCode: Bool
 
+    @StateObject private var watcher = FileWatcher()
+
     var body: some View {
-        if !showSourceCode && isRenderableFile(filePath) {
-            RenderedContentView(filePath: filePath)
-        } else if isImageFile(filePath) {
-            ImagePreviewView(filePath: filePath)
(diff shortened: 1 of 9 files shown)


cheapsteak:
Addressed in `ce7e936` — `open()` and the dispatch source configuration now run outside the lock. Three-stage flow: optimistically claim `watchedPath` under the lock, do the slow syscalls unguarded, then re-acquire the lock to install only if `watchedPath` still matches. A concurrent `stop()` or `observe(otherPath)` overwrites `watchedPath` and the in-flight setup is cleanly abandoned (the source's cancel handler closes the FD). Same fix applied to the rename/delete reopen task.

PR #110, opened 2026-05-08
Refactor: FileWatcher emits changes as AsyncStream<Void>

cheapsteak:
## Summary

- Replaces `FileWatcher`'s `onChange: (@Sendable () -> Void)?` callback (shipped in #108) with a single API: `changes(for:) -> AsyncStream<Void>`. Each call opens its own FD, builds its own dispatch source, and returns a fresh stream. There is no shared `observe`/`stop` state, no `onChange` to wire, no view-owned lifetime to track.
- Cleanup is the stream's `continuation.onTermination`: when the consuming `.task` is cancelled (path change, view teardown) or the iterator is dropped, the dispatch source is cancelled, which closes the FD exactly once via its cancel handler. That's the only place `close(fd)` runs.
- Atomic-save handling unchanged: a `.delete` / `.rename` / `.revoke` event on the watched inode triggers a re-open of the same path after ~50ms (and yields once on success); failure to re-open finishes the stream cleanly. 150ms trailing-edge debounce on `.write` / `.extend` unchanged.

## Why this refactor

Follow-up to #108. The callback design that shipped there is the correct simplification of an earlier (crashing) `@StateObject` + `@MainActor ObservableObject` + `@Published` design. It works — it's the same pattern used by GitHub Copilot for Xcode's `SingleFileWatcher` and cmux's `FileExplorerDirectoryWatcher`.

This PR is the next refinement, modeled on Point-Free's `FileStorage` PersistenceKey: wrap a `DispatchSourceFileSystemObject` in an `AsyncStream` and tie the FD's lifetime to `continuation.onTermination`. For our specific case (one watcher per `FilePreviewView`, consumed via SwiftUI's `.task(id:)`), AsyncStream is a natural fit:

```swift
.task(id: filePath) {
    for await _ in watcher.changes(for: filePath) {
        revision &+= 1
    }
}
```

The cancellation story is bulletproof — there's no `deinit`-on-wrong-thread risk to even worry about, because cleanup runs from `onTermination` on whatever thread released the iterator, and `source.cancel()` is documented as safe to call from anywhere.

## What changed

- `Sources/TBDApp/Panes/FileWatcher.swift` — full rewrite. `FileWatcher` is now a `Sendable` factory with a single method `changes(for:)`; per-stream state lives in a private `StreamState` box guarded by `OSAllocatedUnfairLock`. DEBUG-only `liveStreamCount` replaces `liveCount`.
- `Sources/TBDApp/Panes/CodeViewerPaneView.swift` — `FilePreviewView` swapped to `for await _ in watcher.changes(for: filePath)` inside its `.task(id: filePath)`. The `watcher` itself is now a stateless `private let` (no `@State` needed since it has no per-instance state).
- `Tests/TBDAppTests/FileWatcherTests.swift` — `observeIsIdempotentOnSamePath` removed (no `observe`/`onChange` API). New stream-lifecycle tests covering: stateless construction, iterator-drop cleanup, consuming-task-cancel cleanup, non-existent-path finishes immediately. All assert lifecycle balance via `FileWatcher.liveStreamCount` (DEBUG only).

## Behaviorally equivalent

- 150ms trailing-edge debounce: preserved.
- 50ms atomic-save reopen: preserved.
- `O_EVTONLY` FD, single `close(fd)` from cancel handler: preserved.
- Weak captures in event handlers: preserved.
- The closing tests from the existing suite (lifecycle balance after observe-on-temp-file) remain conceptually intact — the new `liveStreamCountReturnsToBaselineAfterIteratorDrops` is the same property in stream form.

No other behavioral changes. Routing (`LayoutNode.firstPaneID` / `replacingContent` / `routeFileClick` / `PanePlaceholder` wiring) and the preference-change deferral in `PanePlaceholder.onPreferenceChange` are untouched.

## Test plan

- [x] `swift build` clean (warnings unchanged from main).
- [ ] CI runs the unit tests on this PR (local `swift test` is broken on this machine due to a CommandLineTools / Testing-framework runtime quirk, unrelated to this PR — same situation as #108).
- [ ] Smoke: open a worktree, click a `.md` file, `echo >> file.md` and `vim :w` externally → viewer updates within ~150ms.
- [ ] Smoke: cmd-click a different file → viewer swaps content; old stream's onTermination cancels the previous source.
- [ ] Smoke: close the viewer pane → no SIGTRAP.

Follow-up to #108.

open in TBD: `tbd://open?worktree=C81B77A1-5276-454E-AFA2-6095700E435C`

### DIFF
--- a/Sources/TBDApp/Panes/CodeViewerPaneView.swift
+++ b/Sources/TBDApp/Panes/CodeViewerPaneView.swift
@@ -316,16 +316,17 @@ private func isTextFile(_ path: String) -> Bool {
 ///
 /// File-watching plumbing intentionally avoids `@StateObject` /
 /// `ObservableObject` / `@Published` — see the doc-comment on
-/// `FileWatcher` for the SIGTRAP that taught us why. The watcher lives in
-/// `@State` (which stores reference types stably across re-renders) and
-/// pokes the view via a callback that bumps `revision`, an Int observed
-/// by SwiftUI as ordinary `@State`.
+/// `FileWatcher` for the SIGTRAP that taught us why. The watcher exposes
+/// changes as an `AsyncStream<Void>` consumed inside `.task(id: filePath)`;
+/// when the task is cancelled (path change, view teardown) the stream's
+/// `onTermination` cancels the dispatch source, closing the FD via its
+/// cancel handler. SwiftUI just observes the `revision` Int.
 private struct FilePreviewView: View {
     let filePath: String
     let showSourceCode: Bool
 
-    @State private var watcher = FileWatcher()
     @State private var revision: Int = 0
+    private let watcher = FileWatcher()
 
     var body: some View {
         Group {
@@ -340,15 +341,16 @@ private struct FilePreviewView: View {
             }
         }
         .task(id: filePath) {
-            // Wire the callback fresh each time filePath changes. The
-            // callback bumps revision on the main actor; SwiftUI re-renders
-            // the leaf view, whose .task(id: "<path>#<rev>") reloads.
-            watcher.onChange = {
-                Task { @MainActor in
-                    revision &+= 1
-                }
+            // Each filePath change starts a fresh stream. Iterating it
+            // inside `.task` ties FD lifetime to this view: when the task
+            // is cancelled (path change, view teardown), the stream's
+            // iterator is dropped, `onTermination` fires, and the dispatch
+            // source is cancelled — which closes the FD via its cancel
+            // handler. SwiftUI re-renders the leaf view on each yield, and
+            // its `.task(id: "<path>#<rev>")` reloads the file contents.
+            for await _ in watcher.changes(for: filePath) {
+                revision &+= 1
             }
-            watcher.observe(filePath)
         }
     }
 }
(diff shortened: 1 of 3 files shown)


cheapsteak:
Addressed in `f13fd4e` (fix) and `8a3b932` (cleanup), force-pushed after rebasing onto current `main`.

**Medium — FD leak race in `startWatching`** (`f13fd4e`): dropped the second `withLock(terminated)` check; now unconditionally `previousSource?.cancel(); src.resume()` at the bottom. Per GCD docs, resuming an already-cancelled source dispatches the cancel handler and delivers no events — so `close(fd)` runs on every path:
- `terminated` already at install time → `previousSource = src` → cancel + resume → handler fires.
- Race between the two locks (`terminate()` runs after install) → `terminate()` cancels `src` while suspended, our `src.resume()` is what makes its cancel handler actually fire.
- Normal install path → cancel previous epoch, resume new `src`.

**Minor — redundant `path` parameter on `startWatching`** (`8a3b932`): `StreamState.path` was always the same value at every call site; dropped the parameter and use `self.path` directly.

Local `swift build` clean. CI re-running.

PR #111, opened 2026-05-08
Feat: Claude transcripts no longer freeze on /clear or /compact

cheapsteak:
## Summary

- TBD's transcript pane was source-of-truthing off Claude Code's session JSONL. When the user ran `/clear` (or `/compact`), Claude rolled to a new sessionId in a new file, but TBD never noticed — the pane stayed glued to the now-frozen pre-clear file. Confirmed empirically: the PID-keyed `~/.claude/sessions/<pid>.json` doesn't update on `/clear` either, so polling that wouldn't have helped.
- Routes the canonical signal (Claude's `SessionStart` hook with the `clear`/`compact`/`startup`/`resume` matchers) into TBD via a daemon-owned `--settings` overlay at `~/tbd/runtime/claude-overlay.json`. A new `tbd session-event` CLI bridges the hook payload into a `terminal.sessionEvent` RPC, which updates the DB and broadcasts a delta. The transcript pane already re-targets on sessionID change, so the existing UI plumbing carries the rest.
- New terminals also persist Claude's `transcript_path` directly on the terminal record, sidestepping the project-dir encoding rules that we observed Claude Code retarget unpredictably (e.g., into a totally different project dir when its internal worktree tool kicks in).
- Bonus consolidation: the existing notification `Stop` hook moves into the same overlay, so notifications now work out of the box for any TBD-spawned Claude — no manual `tbd setup-hooks` required. `tbd setup-hooks` is deprecated to a no-op message; `tbd hooks status` is the new inspector.
- Existing users with a legacy `tbd setup-hooks --global` install can opt into removing those entries via a new "Migrate Claude Hooks…" app menu (mirrors the CLI installer's interactive pattern). Settings.json mutations go through atomic write + roundtrip-validation + a one-time pristine `~/.claude/settings.json.tbd-backup`.

## Test plan

- [ ] Spawn a fresh Claude terminal in a TBD worktree, send a message, run `/clear`, send another message. Transcript pane should re-target to the new session within the next poll cycle (~1.5s).
- [ ] `tbd hooks status` shows the overlay path with both `SessionStart` and `Stop` hooks listed, and surfaces any legacy entries it detects in user/repo settings.
- [ ] If you have legacy entries: TBD app menu → "Migrate Claude Hooks…" → verify dialog copy, the `Show File…` button, and the `Remove` flow (writes pristine backup, atomic write, validation).
- [ ] `swift test` passes for the new suites: `MigrationV17Tests`, `TerminalSessionEventHandlerTests`, `ClaudeHookOverlayTests`, `LegacyHookScannerTests`, `SettingsJSONSafetyTests`, `ClaudeSpawnCommandBuilderTests`.

## Known limitations

- Claude processes spawned before this PR's overlay was installed won't pick up the new behavior — they were launched without `--settings`. Respawn to test against them.
- The hook commands in the overlay use bare `tbd`, so a stale `tbd` symlink on PATH (especially `/usr/local/bin/tbd` from older installs) will silently fail with exit 64. The CLI installer prompt at app launch addresses `~/.local/bin/tbd`; if you have a higher-priority root-owned symlink, `sudo rm /usr/local/bin/tbd` and let `~/.local/bin/tbd` take over.
- `swift test` full suite hangs on exit (pre-existing — `AppState.startPolling()`'s timer keeps the runloop alive). Targeted suites all run cleanly. Worth a separate cleanup ticket.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=B634D5A9-C7C7-428E-85DA-56AED40E5E54)

### DIFF
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -273,6 +273,7 @@ final class AppState: ObservableObject {
     let daemonClient = DaemonClient()
     let tmuxBridge = TmuxBridge()
     lazy var cliInstallerCoordinator = CLIInstallerCoordinator(daemonClient: daemonClient)
+    lazy var legacyHooksCoordinator = LegacyHooksCoordinator(daemonClient: daemonClient)
     private var pollTimer: Timer?
     private var pollCycle = 0
     private var subscriptionTask: Task<Void, Never>?
@@ -533,12 +534,22 @@ final class AppState: ObservableObject {
                 guard let self else { return }
                 await self.cliInstallerCoordinator.checkOnLaunch()
             }
+            Task { [weak self] in
+                guard let self else { return }
+                await self.legacyHooksCoordinator.checkOnLaunch()
+            }
         }
     }
 
     /// Menu entry point — install or refresh the `tbd` CLI symlink.
     func installCLITool() async {
         await cliInstallerCoordinator.runFromMenu()
+    }
+
+    /// Menu entry point — review and (optionally) remove TBD's legacy
+    /// hook entries from the user's `~/.claude/settings.json`.
+    func migrateClaudeHooks() async {
+        await legacyHooksCoordinator.runFromMenu()
     }
 
     /// Launch the daemon process and connect.
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -442,7 +442,13 @@ final class AppState: ObservableObject {
             return
         }
         terminals[delta.worktreeID]?[idx].claudeSessionID = delta.sessionID
-        terminals[delta.worktreeID]?[idx].transcriptPath = delta.transcriptPath
+        // Mirror TerminalStore.updateSession's preserve-on-nil: a delta with
+        // nil transcriptPath means the SessionStart payload didn't carry a
+        // path even though sessionID rolled. Keep the previous value so the
+        // in-memory model doesn't drift from the DB.
+        if let tp = delta.transcriptPath {
+            terminals[delta.worktreeID]?[idx].transcriptPath = tp
+        }
     }
 
     /// Update the in-place usage entry for a single profile. If no match,
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -425,9 +425,23 @@ final class AppState: ObservableObject {
             applyModelProfileUsageDelta(usage)
         case .modelProfilesChanged:
             Task { [weak self] in await self?.loadModelProfiles() }
+        case .terminalSessionUpdated(let d):
+            applyTerminalSessionDelta(d)
         default:
             break
         }
+    }
+
+    /// Apply a Claude session rollover (post-`/clear` / `/compact` / startup)
+    /// directly to the in-memory Terminal so LiveTranscriptPaneView re-targets
+    /// without waiting for the next 2s `terminal.list` poll. Silently ignores
+    /// terminals we don't know about — the next refresh will reconcile.
+    private func applyTerminalSessionDelta(_ delta: TerminalSessionDelta) {
+        guard let idx = terminals[delta.worktreeID]?.firstIndex(where: { $0.id == delta.terminalID }) else {
+            return
+        }
+        terminals[delta.worktreeID]?[idx].claudeSessionID = delta.sessionID
+        terminals[delta.worktreeID]?[idx].transcriptPath = delta.transcriptPath
     }
 
     /// Update the in-place usage entry for a single profile. If no match,
--- a/Sources/TBDApp/DaemonClient.swift
+++ b/Sources/TBDApp/DaemonClient.swift
@@ -928,4 +928,25 @@ actor DaemonClient {
             params: AppSetForegroundStateParams(isForeground: isForeground)
         )
     }
+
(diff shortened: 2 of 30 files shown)


cheapsteak:
Addressed the two minor items from the review:

- **`clearRecreated` doesn't reset `transcriptPath`** → fixed in 647cc93. Added `record.transcriptPath = nil` to the existing nil block.
- **`HooksStatusCommand` comment promises a local fallback that isn't implemented** → fixed in 534a6c8 by updating the comment. Implementing the actual local scan would require either moving `LegacyHookScanner` into `TBDShared` or adding a `TBDDaemon` dep on `TBDCLI` (the latter feels architecturally wrong — the CLI shouldn't import the daemon target). Comment now accurately describes the daemon-required behavior, leaving the refactor as future work if a daemon-down workflow becomes important.

cheapsteak:
Addressed the four minor items from the second review pass:

- **`SessionEventCommand` stdin loop** → 91f2e3c. Switched to `FileHandle.standardInput.readDataToEndOfFile()`; the 1 MiB guard now lives in the same `guard` chain.
- **`updateSession` clobbers transcriptPath when payload omits it** → c07a7a8. `updateSession` now leaves `transcriptPath` untouched when called with `nil`, preserving the previously-captured path. Added branch test `nilPathPreservesExisting` covering: first event sets `/abs/s1.jsonl`, second event with rejected path keeps `/abs/s1.jsonl` while still updating sessionID.
- **`stripEntries` reads oddly** → 955c8f9. Dropped the `var matchers / matchers = newMatchers` rebinding and use `newMatchers` directly in the empty-check.
- **Race window in `LegacyHooksCoordinator.checkOnLaunch()`** → ca5498f. Moved `hasCheckedThisSession = true` to immediately after the guard, matching `CLIInstallerCoordinator`'s pattern.

All targeted suites still green (13 tests across the two affected suites).

cheapsteak:
Addressed both items from the third pass:

- **Medium: `applyTerminalSessionDelta` clobbers transcriptPath on nil delta** → db54125. Mirrored the DB layer's preserve-on-nil semantics so the in-memory model stays consistent with the DB before the next 2s terminal.list reconcile.
- **Minor: `stripEntries` leaves `hooks: {}`** → 1007296. Drop the key entirely when the dict is empty after stripping.

cheapsteak:
Addressed the four minor items from the latest review:

- **`session-event` shows in `tbd --help`** → f741aa1. Added `shouldDisplay: false`, matching the deprecated `setup-hooks` pattern.
- **`readDataToEndOfFile` comment overstates bound** → f741aa1. Reworded to clarify the cap limits processing, not the read itself.
- **`atomicWriteValidated` hardcodes FileManager.default** → b7b7a19. Threaded a `fileManager:` parameter through, matching `ensureBackup`'s signature.
- **`ClaudeSpawnCommandBuilder.build()` no longer pure** → 64466bf. Added a `fileExists: (String) -> Bool` closure parameter (defaulting to `FileManager.default.fileExists(atPath:)`) so tests can keep the function hermetic and the docstring stays honest.

cheapsteak:
Addressed the documentation minor; punting on the shell-injection one:

- **`fileManager` parameter scope** → aa76920. Added a docstring paragraph clarifying that fileManager only covers directory probes; the final write is always real I/O via `Data.write(to:)`.
- **`stopCommand` shell injection in `$MSG`** → not in this PR. The bot itself flagged it as pre-existing/informational (the same string predates this PR — it shipped via `tbd setup-hooks --global` and is now relocated to the overlay). The `|| true` ensures any failure is a silent dropped notification, not a security/safety issue. Worth a follow-up PR but out of scope for the transcript-on-clear fix.

cheapsteak:
Addressed both minors from the latest pass:

- **NSAlert Return-default on destructive button** → 97e59ea. Moved the keyEquivalent off Remove and onto Not Now per HIG. Pristine-backup safety net stands; this just adds proper friction.
- **Test name promised broadcast coverage** → 61b790b. Renamed to `updates sessionID + transcriptPath in DB on a fresh SessionStart` to match what the body actually asserts. Adding a fake subscription capture would have required restructuring the router init; the broadcast path is exercised end-to-end by LiveTranscriptPaneView's polling cycle.

cheapsteak:
Addressed item 1, pushing back on items 2 and 3:

- **`runFromMenu` shows Remove button when only repo entries exist** → 3eeef95. Detect the repo-only case and surface an info-only alert listing the affected repo files; no dangling button.
- **Migration naming convention** → respectfully disagreeing. The bot's claim that all 16 prior migrations follow `v1`/`v2`/`v3` is inaccurate — `v14_worktree_location`, `v15_model_profiles`, and `v16_archived_head_sha` all use descriptive suffixes. `v17_terminal_transcript_path` matches the actual recent precedent. Happy to revisit if there's a deliberate intent to revert to bare numbers, but the pattern set by the last three migrations is descriptive suffixes.
- **`handleDaemonLegacyHooksStatus` swallows DB errors** → leaving as-is. The bot itself frames this as 'Reasonable degradation for a diagnostic endpoint... Minimal impact in practice.' Adding error surfacing here would mean either failing the whole RPC (worse UX — global-only would be invisible too) or returning a partial-error result type (architectural overhead for a single edge case). The diagnostic endpoint's role is informational, not authoritative.

PR #113, opened 2026-05-08
Fix: Cmd+click on OSC 8 hyperlinks no longer opens two viewer panes

cheapsteak:
## Summary

- **Symptom**: cmd+clicking a relative file link in a TBD terminal sometimes opens **two** code-viewer panes instead of one. Reproducible with any OSC 8 hyperlink whose visible text is also a real path (Claude Code emits these for tool results).
- **Root cause**: a single cmd+click was getting handled by two paths in `Sources/TBDApp/Terminal/TerminalPanelView.swift`. Our local `NSEvent` mouseDown monitor matched the visible text via `extractFilePath()` and fired `onFilePathClicked` (consuming mouseDown). But our monitor only intercepts mouseDown — `mouseUp` flowed through to SwiftTerm, whose `mouseUp` saw the OSC 8 payload + cmd modifier and called `requestOpenLink`, firing `onFilePathClicked` a second time. The "sometimes" comes from `linkHighlightMode = .hoverWithModifier` (Mac default): the second fire only happens when the link gets hover-highlighted before mouseUp.
- **Fix**: when the click cell carries an OSC 8 payload, the mouseDown monitor returns without consuming so SwiftTerm owns that click via `requestOpenLink`. Plain-text paths and `PR #123` matches still go through our monitor unchanged. Also wires `os.Logger` diagnostics on every fire path and on `routeFileClick`'s split-vs-swap branches so future occurrences are traceable.

## Test plan

- [ ] In any TBD terminal pane, run `printf '\e]8;;file://%s/Sources/TBDApp/Panes/ViewerRouting.swift\e\\%s\e]8;;\e\\\n' "$(pwd)" "Sources/TBDApp/Panes/ViewerRouting.swift"` to print a clickable link.
- [ ] Hover the cursor over the link with cmd held until it underlines, then click. Confirm exactly **one** viewer pane opens (pre-fix this would open two).
- [ ] Repeat without prior hover (fast cmd+click). Confirm one pane still opens (no regression for fast clickers — SwiftTerm's `mouseUp` calls `updateHoverLink` before `linkForClick`, so the highlight range is set in time).
- [ ] Cmd+click a plain-text path printed by `ls` or similar. Confirm it still opens (mouseDown monitor path).
- [ ] Cmd+click a `PR #123` reference. Confirm it still opens GitHub (PR-pattern path).
- [ ] Tail the logs to verify single-fire: `log stream --level debug --predicate 'subsystem BEGINSWITH "com.tbd" AND (category == "TerminalPanel" OR category == "ViewerRouting")'` — should show one `requestOpenLink/file` and one `routeFileClick[split]` (or `[swap]` on subsequent clicks) per click.

open in TBD: `tbd://open?worktree=E9FD18C8-A443-44A7-80DD-E6AC3B049ECA`

### DIFF
--- a/Sources/TBDApp/Panes/ViewerRouting.swift
+++ b/Sources/TBDApp/Panes/ViewerRouting.swift
@@ -1,4 +1,7 @@
 import Foundation
+import os
+
+private let logger = Logger(subsystem: "com.tbd.app", category: "ViewerRouting")
 
 /// Decides where a terminal-link click should land.
 ///
@@ -20,9 +23,11 @@ func routeFileClick(into layout: LayoutNode, terminalID: UUID, path: String) -> 
            with: .codeViewer(id: viewerID, path: path)
        )
     {
+        logger.debug("routeFileClick[swap]: viewerID=\(viewerID, privacy: .public) path=\(path, privacy: .public)")
         return updated
     }
 
+    logger.debug("routeFileClick[split]: terminalID=\(terminalID, privacy: .public) path=\(path, privacy: .public)")
     return layout.splitPane(
         id: terminalID,
         direction: .horizontal,
--- a/Sources/TBDApp/Terminal/TBDTerminalView.swift
+++ b/Sources/TBDApp/Terminal/TBDTerminalView.swift
@@ -215,12 +215,16 @@ class TBDTerminalView: TerminalView {
     /// payload. SwiftTerm's `mouseUp` will dispatch these via
     /// `requestOpenLink`, so our local mouseDown monitor must not also handle
     /// them — otherwise a single cmd+click opens two viewer panes.
+    ///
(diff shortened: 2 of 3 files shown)


cheapsteak:
Addressed both notes in c94b338:

1. **`hasOSC8Payload` cast** — switched to `getPayload() as? String != nil`. Good catch — sixel/inline-image cells would otherwise silently no-op for a file path overlaying graphics. Doc-comment now spells out the rationale.
2. **Dead OSC 8 branch in `extractHyperlinkURL`** — deleted (and trimmed the now-unused `col` binding). Updated the doc-comment so the function's stated purpose matches what's left (residual non-OSC-8 patterns, currently just `PR #123`).

Skipped the filtered double-`gridPosition` note — agree it's negligible, and merging the two functions to share a single lookup would add coupling not worth the cycles saved.

PR #117, opened 2026-05-09
Feat: Hide sidebar chevron and + behind hover to reclaim horizontal space

cheapsteak:
## Summary
- Repo header row: chevron moved from leading to trailing edge, only visible on hover of the header. The leading chevron and trailing `+` are gone, so project names get the full width of the row.
- Main worktree row: `+` moved here at the trailing edge, visible whenever any row in the repo section is hovered (80ms debounce so it doesn't flicker as the cursor crosses adjacent List cells).
- Project name font shrunk from `.headline` (13pt semibold) to `.system(size: 11, weight: .semibold)` to read more like a sidebar group header.
- Removed our explicit `.padding(.vertical, 2)` and `.padding(.horizontal, 4)` from worktree rows, and zeroed our `listRowInsets` leading from 4 to 0. (System `.listStyle(.sidebar)` row padding is unchanged — that's a separate, harder problem.)
- Hover-revealed buttons sit in a 20×20 placeholder so toggling visibility doesn't shift row content.

## Test plan
- [ ] Hover a project header row — chevron appears at the trailing edge; clicking it expands/collapses the section.
- [ ] Move the cursor off the project header — chevron disappears; project name is unindented and uses the full row width.
- [ ] Hover any row inside a project (header, main, child worktree) — `+` appears on the main worktree row's trailing edge.
- [ ] Move the cursor between adjacent rows in the same section — `+` stays visible (no flicker).
- [ ] Move the cursor out of the section entirely — `+` disappears after ~80ms.
- [ ] Click `+` on the main row — creates a new worktree (same behavior as the old leading `+`).
- [ ] Right-click a project header — context menu still offers "Rename...".
- [ ] Click a worktree row — selection still works (single, cmd-click multi-select, click-to-rename when already selected).
- [ ] `+` is disabled if the repo is `.missing`.

open in TBD: \`tbd://open?worktree=E83C5596-3226-43B8-A8DA-2F8BADA7DDF9\`

### DIFF
--- a/Sources/TBDApp/Sidebar/RepoSectionView.swift
+++ b/Sources/TBDApp/Sidebar/RepoSectionView.swift
@@ -30,6 +30,28 @@ struct RepoSectionView: View {
 
     @State private var isExpanded = true
     @State private var isEditing = false
+    @State private var isHeaderHovered = false
+    @State private var isSectionHovered = false
+    @State private var hoverDebounceTask: Task<Void, Never>?
+
+    private func onSectionHoverChange(_ hovering: Bool) {
+        if hovering {
+            hoverDebounceTask?.cancel()
+            hoverDebounceTask = nil
+            if !isSectionHovered {
+                isSectionHovered = true
+            }
+        } else {
+            hoverDebounceTask?.cancel()
+            let task = Task { @MainActor in
+                try? await Task.sleep(nanoseconds: 80_000_000)
+                if !Task.isCancelled {
+                    isSectionHovered = false
+                }
+            }
+            hoverDebounceTask = task
+        }
+    }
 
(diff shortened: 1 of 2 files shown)


cheapsteak:
Both addressed in 1bf5dc6 and 26dcc26:

- **Tap dead zone**: replaced `Spacer()` with `.frame(maxWidth: .infinity, alignment: .leading)` on `WorktreeRowView` so its `contentShape` covers the full row width — click-to-select works across the empty area between the worktree name and the `+` button again.
- **Task.sleep cleanup**: switched to `try await` and let cancellation propagate; dropped the manual `!Task.isCancelled` guard. Property type changed from `Task<Void, Never>?` to `Task<Void, Error>?`.

PR #119, opened 2026-05-09
Reliability: Stop the sidebar from collapsing on wake-from-sleep

cheapsteak:
## Summary

The worktree-list sidebar would randomly hide itself, especially after the Mac woke from sleep, with no UI to bring it back. Root cause: SwiftUI's `NavigationSplitView` (with `.prominentDetail` style and `.toolbar(removing: .sidebarToggle)`) auto-collapses the sidebar at narrow widths. On wake, macOS reports a transient ~200×200 window size; SwiftUI eagerly collapses, then never restores when the window grows back. Auto-collapse is sticky.

This PR is staged in two commits so each is independently revertable:

- **Stage 1** (`6a550c7`): Hard minimum window size — `.windowResizability(.contentMinSize)` on the scene + `.frame(minWidth: 800, minHeight: 500)` on the root content. The bogus tiny size now can't land in the first place. Defensive fix that may by itself eliminate the wake symptom; worth keeping regardless.
- **Stage 2** (`09d32fe`): Replace `NavigationSplitView` with an `NSViewControllerRepresentable` wrapping `NSSplitViewController`. The sidebar split item gets `canCollapse = false`, `minimumThickness = 220`, `maximumThickness = 400`. Layout is now deterministic — SwiftUI cannot auto-collapse what AppKit refuses to collapse. Also removes a redundant `.toolbar { Picker }` from `SidebarView` that was already declared in `ContentView` and would have stopped propagating from inside an `NSHostingController` anyway.

A parallel sister-session investigation (worktree `20260509-friendly-chicken`) independently landed on the same diagnosis and AppKit answer — this PR consolidates both threads.

**Deferred follow-up (not this PR):** Replacing the SwiftUI `List(.sidebar)` inside `SidebarView` with a raw `NSOutlineView` to reach Tower-grade row density. PR #117 already shipped the chevron / "+" / font-size wins on the SwiftUI side; the row-height ceiling needs AppKit. The `NSSplitViewController` wrap from this PR is the natural foundation. Estimated 4–8h when there's appetite.

## Test plan

After pulling, run `scripts/restart.sh` from the worktree, then verify:

- [ ] Sidebar is visible on launch.
- [ ] Drag the window narrow — it stops at 800×500, sidebar stays put.
- [ ] Sleep / wake the Mac — sidebar is still visible.
- [ ] Worktree selection works: single-click, Cmd-click multi-select.
- [ ] Drag-to-reorder worktrees within a repo (`.onMove` in `RepoSectionView`).
- [ ] Hover-revealed chevron on the repo header (PR #117 work).
- [ ] Hover-revealed "+" on the main worktree row (PR #117 work).
- [ ] Inline rename via double-click and the context menu.
- [ ] "Add Repository" button in the bottom safe-area inset still triggers the open panel.
- [ ] Repo-filter Picker in the window toolbar still filters the sidebar.
- [ ] Toolbar items unchanged: back/forward, auto-suspend, conductor, filter, PR button (when applicable), file-panel toggle (⌘⇧E).
- [ ] Terminal panes resize correctly when the window is resized (PTY reflow).
- [ ] Sidebar source-list vibrancy still applies.

## Notes for reviewers

- `EnvironmentObject` does not auto-cross the SwiftUI → AppKit → SwiftUI boundary. `AppKitSplitView` re-injects `appState` as an environment object on every `NSHostingController` rootView, in both `makeNSViewController` and `updateNSViewController`. Skipping either would crash at runtime.
- `updateNSViewController` reassigns `rootView` via `AnyView` erasure. The concrete-generic alternative compiles but produces casts that silently break if the wrapping `.environmentObject(...)` modifier chain ever changes. Negligible diffing cost; large readability win.
- Did not override `splitView.dividerStyle` — `NSSplitViewItem(sidebarWithViewController:)` configures source-list-appropriate defaults.

open in TBD: \`tbd://open?worktree=CDB7B3FA-55F9-4E46-A65E-CE435AB168F5\`

### DIFF
--- a/Sources/TBDApp/ContentView.swift
+++ b/Sources/TBDApp/ContentView.swift
@@ -180,6 +180,7 @@ struct ContentView: View {
 
             StatusBarView()
         }
+        .frame(minWidth: 800, minHeight: 500)
         .onChange(of: appState.selectedWorktreeIDs) { oldSelection, newSelection in
             markSelectedWorktreesAsRead(newSelection)
             let newlySelected = newSelection.subtracting(oldSelection)
--- a/Sources/TBDApp/ContentView.swift
+++ b/Sources/TBDApp/ContentView.swift
@@ -19,50 +19,50 @@ struct ContentView: View {
 
     var body: some View {
         VStack(spacing: 0) {
-            NavigationSplitView {
-                SidebarView()
-            } detail: {
-                if !appState.isConnected {
-                    disconnectedView
-                } else if appState.repos.isEmpty {
-                    emptyStateView
-                } else if let repoID = appState.selectedRepoID {
-                    RepoDetailView(repoID: repoID)
-                } else if appState.selectedWorktreeIDs.isEmpty {
-                    Text("Select a worktree or click + to create one")
-                        .foregroundStyle(.secondary)
-                } else {
-                    HStack(spacing: 0) {
-                        TerminalContainerView()
-                        if showFilePanel, let worktree = selectedWorktree, !worktree.path.isEmpty {
-                            FilePanelDivider(panelWidth: Binding(
-                                get: { CGFloat(filePanelWidth) },
-                                set: { filePanelWidth = Double($0) }
-                            ))
-                            FileViewerPanel(worktree: worktree)
-                                .frame(width: CGFloat(filePanelWidth))
-                                .id(worktree.id)
+            AppKitSplitView(
+                appState: appState,
+                sidebar: { SidebarView() },
+                detail: {
+                    if !appState.isConnected {
+                        disconnectedView
+                    } else if appState.repos.isEmpty {
+                        emptyStateView
+                    } else if let repoID = appState.selectedRepoID {
+                        RepoDetailView(repoID: repoID)
+                    } else if appState.selectedWorktreeIDs.isEmpty {
+                        Text("Select a worktree or click + to create one")
+                            .foregroundStyle(.secondary)
+                    } else {
+                        HStack(spacing: 0) {
+                            TerminalContainerView()
+                            if showFilePanel, let worktree = selectedWorktree, !worktree.path.isEmpty {
+                                FilePanelDivider(panelWidth: Binding(
+                                    get: { CGFloat(filePanelWidth) },
+                                    set: { filePanelWidth = Double($0) }
+                                ))
+                                FileViewerPanel(worktree: worktree)
+                                    .frame(width: CGFloat(filePanelWidth))
+                                    .id(worktree.id)
+                            }
                         }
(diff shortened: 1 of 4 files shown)


cheapsteak:
Thanks for the careful review.

**Addressed**: `autosaveName = "main.sidebar"` added in 7920cdf (well, the next commit hash — see push above). Divider width now persists across launches.

**On the `AnyView` identity concern**: real point, and I weighed the Coordinator-typed-property alternative (storing the `NSHostingController<Sidebar>` / `NSHostingController<Detail>` on a Coordinator class so `updateNSViewController` can reassign `rootView` without erasing or casting). Choosing to defer that refactor for now — the typed-generic chains involve `ModifiedContent<…, _EnvironmentKeyWritingModifier<…>>` which is annoying to spell, and the underlying `TerminalContainerView` uses an `NSViewRepresentable` with a Coordinator-held PTY/scroll handle that doesn't depend on parent SwiftUI identity for its session state. Worst-case practical symptom is scroll-position reset on AppState mutation, not PTY disruption. If we observe that in practice, the Coordinator refactor is the right fix and the call site doesn't change. Filing a mental note.

cheapsteak:
Update — the bot's substantive concern wasn't theoretical, it was actively breaking the sidebar.

After restarting against the AnyView build (the original commit), `SidebarView`'s `List(.sidebar) { ForEach(filteredRepos) { ... } }` rendered empty on first state population — the worktree list was blank, only the bottom `safeAreaInset` ("Add Repository" button) was visible. Detail pane simultaneously showed "Select a worktree" (so `appState.repos` was non-empty), confirming the data was there but SwiftUI had failed to reconcile it into the table view. Reentrant `NSTableView` delegate warnings were piling up in the log.

Cause: every `updateNSViewController` reassigned `rootView = AnyView(sidebar().environmentObject(appState))`. The `AnyView` wrap erases structural identity, so SwiftUI's reconcile fell back to a heuristic that, for the source-list `List`, dropped its rows.

Fix in 9d9078c (well, the new commit — see push): introduce a `Coordinator` holding typed `NSHostingController<EnvironmentInjector<Sidebar>>` and `NSHostingController<EnvironmentInjector<Detail>>`, plus a tiny private `EnvironmentInjector<Content>` view so the rootView's static type is spellable without dipping into private `_EnvironmentKeyWritingModifier` internals. `updateNSViewController` now reassigns `rootView` with the same concrete wrapper type each time → SwiftUI runs its normal structural reconcile → the sidebar populates correctly.

Verified by restart on the new build: sidebar lists all repos and worktrees, detail pane shows the active terminal, file panel renders. Thanks for the precise diagnosis.

PR #120, opened 2026-05-09
Fix: 17s transcript hang from per-row text-selection storm

cheapsteak:
## Summary

Two macOS spindumps in 17 minutes (`crash.log`, `crash.2.log` — both ~17 s main-thread stalls) showed `TBDApp` wedged in `GraphHost.flushTransactions → AG::Subgraph::update`, with the heaviest leaves landing in `SelectionOverlay.updateNSView → -[NSControl setFont:] → -[NSTextFieldCell _invalidateEffectiveFont]`. Every visible transcript row pays for `.textSelection(.enabled)` — SwiftUI materializes a `SelectionOverlay` (`NSViewRepresentable` wrapping `NSTextField`) per Text. On any environment/state change all visible cells re-run the font/cell-invalidation cascade plus per-cell `objc_getAssociatedObject` env propagation. With ~15 selection sites across 12 cards × N visible rows, layout passes go super-linear and the main thread stalls.

This PR has two commits:

1. **`fix:` Gate transcript text selection on hover.** New `transcriptSelectableText()` view modifier reads a `transcriptTextSelection` env value and applies `.textSelection(.enabled)` only when true. `TranscriptItemsView` tracks one `hoveredItemID` and pushes `true` only into the hovered row's subtree. Latched on hover-enter (not cleared on exit) so drag-select survives the cursor briefly leaving the row, and so cross-card selection doesn't tear down the `NSTextField` mid-drag. Replaced 18 call sites across 12 transcript cards; `CodeViewerPaneView` and `FileViewerPanel` left as-is (single-view panes, not lazy lists).

2. **`feat:` Add main-thread hang watchdog.** Background `DispatchSourceTimer` at `.utility` QoS heartbeats main every 250 ms via `mach_absolute_time`; on healthy→hung transition (default 1500 ms threshold) emits one `os.Logger` `.warning` line with stall ms, focused terminal short-ID, transcript item count, and pane label. Edge-triggered, so a sustained hang logs once, not every tick. `LiveTranscriptPaneView` feeds the snapshot on `.onAppear` and on `messages.count` change — exactly the path that produced these two stalls. Other panes log with `pane=-` for now (TODO comment in source).

Both pieces have unit tests covering the env=true/env=false branches (`TranscriptSelectableTextTests`) and all four `(wasHung, isHungNow)` transitions of the watchdog's pure `evaluate(...)` helper (`HangWatchdogTests`), per the project's branch-test rule in `CLAUDE.md`.

## Test plan
- [ ] Open a transcript with >100 items; scroll/interact and confirm no Beach Ball
- [ ] Hover any row → text becomes selectable; drag a selection that briefly exits the row → selection survives
- [ ] Hover a different row → selection moves with hover (latched, not live)
- [ ] Force a synthetic stall (e.g. `Thread.sleep(forTimeInterval: 2)` on MainActor in a debug build) → exactly one `category == "hang-watchdog"` `.warning` line on the stall, exactly one `.info` recovery line after
- [ ] Verify log fields populate when stall happens on the live transcript pane: `terminalID=<short>`, `itemCount=<n>`, `pane=liveTranscript`
- [ ] `swift build` and `swift test` pass in CI (Swift Testing — local CLT-only machines will skip)

open in TBD: \`tbd://open?worktree=6665CA74-58F8-43DA-B353-860023123A69\`


cheapsteak:
Addressed both minor items in d91b694:

- **Label rename**: `evaluate(nowNs:)` → `evaluate(stallNs:)`. Agreed it read like "now in nanoseconds" — the rename matches the internal parameter name and makes the helper self-documenting at every call site.
- **`stop()` state reset**: now clears `wasHung` and `hangStartedAtMach`. Good catch on the latent edge — `stop()` isn't wired today but if it ever is, the spurious "recovered" log would have been a head-scratcher.

Skipping the informational item about depth>0 nested rows by design — env inheritance from the parent row is the intended behavior for the `DisclosureGroup`-wrapped sub-trees, since they live inside a single visual row. If a future card type produces long nested content I'll revisit.

cheapsteak:
Addressed both items in a488afc:

- **AgentCard migration**: legitimate miss in the initial pass. The first grep listed first-20 results and AgentCard didn't make the cut — only a full sweep would have caught it. Both lines now go through `.transcriptSelectableText()`. Worth noting AgentCard is a hot card during multi-agent sessions, so this was a real correctness gap.
- **`machDeltaToNanos` comment**: rewrote to describe the actual reasoning (small numer, bounded delta) instead of the imaginary UInt128 path.

cheapsteak:
Both addressed in 5d72d63:

- **`recordTick()` @MainActor**: contract was load-bearing (silently corrupting `lastTick` from a background thread would mask hangs), so worth enforcing at compile time. Both call sites are inside `DispatchQueue.main.async` already; added `@MainActor` to the closure capture so Swift concurrency proves call-site-on-main.
- **Preview text selection**: set `.environment(\.transcriptTextSelection, true)` once on the outer ScrollView in `ChatBubbleParityPreviews` rather than per-bubble — applies to both `oldBubble` and `newBubble` and matches the spirit of the fix (env-driven gating).

cheapsteak:
Both addressed in 1111715:

- **Stale snapshot context**: `LiveTranscriptPaneView` now nils `paneLabel`, `transcriptItemCount`, and `focusedTerminalIDShort` on `.onDisappear`. The bot is right that this was a real misattribution risk — a hang in the terminal pane would have been tagged `pane=liveTranscript` after the live transcript was visited even once. Empty-state now matches the documented contract for both "never visited" and "left".
- **Stale TODO**: rewrote the comment above `transcriptItemCount`/`paneLabel` — the "leave as TODO until a natural call site emerges" rationale doesn't apply once the call sites exist. Comment now describes actual nil semantics so a future reader doesn't think the fields are dead.

cheapsteak:
Last cosmetic addressed in 3e28594: `ChatBubbleView_Previews` now matches `ChatBubbleParityPreviews` — both inject `transcriptTextSelection = true` for review ergonomics.

Confirming the three intentional-design notes:
- **Heartbeat accumulation during hangs** (~68 pending blocks for a 17 s stall): considered, fine for a diagnostic tool. Each block is two lock ops; they drain on recovery and the `.toHealthy` edge fires correctly on the next timer tick after the first heartbeat lands.
- **`@unchecked Sendable` with queue-isolated state**: intentional. `wasHung` / `hangStartedAtMach` are touched only in `tick()` and `stop()`, both dispatched on the watchdog's serial `queue`.
- **`ThinkingRow` migration is a no-op today**: intentional. `rowFor` returns `EmptyView()` for `.thinking`, but migrating the modifier keeps the file consistent for if/when extended-thinking content ever ships inline.

PR #121, opened 2026-05-09
Feat: Polish sidebar — emoji-as-icon, tighter rows, project hierarchy

cheapsteak:
## Summary
- **Emoji as repo icon.** When a repo's display name starts with an emoji, the folder icon is hidden so the emoji itself anchors the row. When there's no emoji yet, the folder icon becomes a button that opens the existing `EmojiPickerView` floating panel and prepends the chosen emoji to the display name.
- **Tighter, more uniform row heights.** Switched the sidebar from `.listStyle(.sidebar)` (which is an opaque NSOutlineView wrapper that ignores SwiftUI's row-sizing knobs) to `.listStyle(.plain)`. The translucent backdrop still comes from `NavigationSplitView`, so the visual surface is preserved. Worktree rows are now an explicit 28pt; the repo header bottom-aligns content in a 22pt frame to tighten the gap to the first worktree row.
- **Visible parent/child hierarchy.** Worktree rows are inset 12pt under their repo header so it's easy to scan where each project starts and ends. Removed the "main" branch icon (redundant with the new repo-level visual).

## Test plan
- [ ] Open the sidebar — confirm repo headers without an emoji show a folder icon, repos with an emoji prefix show only the emoji
- [ ] Click the folder icon on a repo with no emoji — emoji picker floats below the icon; pick one — name updates and the folder icon disappears
- [ ] Confirm worktree rows have ~28pt of vertical room and sit indented 12pt under their repo header
- [ ] Drag-reorder worktrees within a repo — order persists
- [ ] Cmd+click multi-select still works
- [ ] Inline rename via right-click → "Rename..." still works (including `:emoji:` autocomplete)
- [ ] Hover over a repo header — chevron and "+" buttons still appear
- [ ] Selection highlight still renders on the selected worktree row

open in TBD: \`tbd://open?worktree=4851AC97-E2D4-4E13-8F97-4EB205CD699B\`

### DIFF
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -111,12 +111,6 @@ final class AppState: ObservableObject {
     /// before the daemon RPC has returned.
     @Published var isInitialStateLoaded: Bool = false
 
-    /// ID of the repo whose emoji picker is currently presented. Holding this in
-    /// shared state (rather than per-`RepoSectionView` `@State`) ensures only one
-    /// `FloatingPanel` is ever visible — opening picker B while A is open
-    /// implicitly dismisses A.
-    @Published var activeEmojiPickerRepoID: UUID?
-
     /// Buffers a deep-link target UUID when `.onOpenURL` fires before the
     /// initial state load completes. Drained at the end of
     /// `connectAndLoadInitialState()`. Internal-only — never written from
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -111,6 +111,12 @@ final class AppState: ObservableObject {
     /// before the daemon RPC has returned.
     @Published var isInitialStateLoaded: Bool = false
 
+    /// ID of the repo whose emoji picker is currently presented. Holding this in
+    /// shared state (rather than per-`RepoSectionView` `@State`) ensures only one
+    /// `FloatingPanel` is ever visible — opening picker B while A is open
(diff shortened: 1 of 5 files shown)


cheapsteak:
Addressed all three points in `ac10001`:

1. **Multi-picker** — moved `showEmojiPicker` state into `AppState.activeEmojiPickerRepoID`; only one picker can be active at a time, opening another implicitly dismisses the first
2. **Light-dismiss** — `EmojiPanelAnchor` now takes an optional `onOutsideClick` callback; when set, the coordinator installs local + global `NSEvent` mouse-down monitors and dismisses if the click isn't routed to the panel's NSWindow. Monitors are torn down on hide/dismantle. `RenameableLabel` keeps its focus-driven dismissal (passes nil)
3. **Native row background** — added `.listRowBackground(Color.clear)` to every row so the custom `RoundedRectangle` highlight in `WorktreeRowView` is the single source of truth for selection visuals

cheapsteak:
Pivoted on the picker entirely (commit \`f1a9582\`): rather than fix the multi-picker and light-dismiss behavior, removed the click-to-pick path. The folder icon is a plain \`Image\` again. Adding/changing an emoji is done via inline rename + \`:foo:\` autocomplete, which \`RenameableLabel\` already supports cleanly.

Kept the \`.listRowBackground(Color.clear)\` from the previous fix — it's a real independent issue with \`.plain\` list style.

PR #127, opened 2026-05-10
Fix: Stop transcript pane from blanking after new messages

cheapsteak:
## Summary

- After a new message arrived in the live transcript pane (often right after the user submitted a prompt in the terminal), the pane intermittently went blank — scrolling up a short distance brought content back.
- Root cause: the `ScrollView` relied solely on `.defaultScrollAnchor(.bottom)` paired with a `LazyVStack`. There was no `.scrollPosition(id:)` binding and no explicit autoscroll on append, so on content growth the lazy stack's row realization fell out of sync with the scroll offset. The viewport landed past the realized rows; manual scroll forced re-realization.
- Fix follows two existing specs that were never landed (`docs/superpowers/specs/2026-05-06-transcript-scroll-position-design.md` and `2026-05-06-transcript-reentry-fix-design.md`): a `@State visibleID` + `.scrollPosition(id: $visibleID, anchor: .bottom)` on `LiveTranscriptPaneView` and `HistoryPaneView`'s `SessionTranscriptView`. New messages drive autoscroll via `.onChange(of: messages.last?.id)` with a `guard let _ = oldID` nil→value guard (avoids a spurious snap when `messages` transiently empties during nav churn). An `.onAppear` `proxy.scrollTo(visibleID, anchor: .bottom)` acts as a position-no-op realize-poke so the lazy stack realizes rows around the bound id on re-entry.

## Test plan

- [ ] `scripts/restart.sh`, open a Claude terminal, switch to the live transcript pane.
- [ ] Submit a new message in the terminal; transcript should keep showing content (no blank flash) and follow to the new last row.
- [ ] Scroll up to older messages, switch to another tab, switch back — scroll position is preserved, no blank panel.
- [ ] Open the History pane on a worktree with prior sessions, scroll a session up, switch sessions and back — position preserved.

open in TBD: \`tbd://open?worktree=70B550D1-79E7-4489-BB01-D1522F645A5F\`

### DIFF
--- a/Sources/TBDApp/Panes/HistoryPaneView.swift
+++ b/Sources/TBDApp/Panes/HistoryPaneView.swift
@@ -288,10 +288,6 @@ struct SessionTranscriptView: View {
     /// shown only when the user has consciously scrolled up.
     @State private var atBottom: Bool = true
 
-    /// Tracks the row at the bottom edge of the viewport. Drives re-entry
-    /// restoration via `.scrollPosition(id:)`.
-    @State private var visibleID: String?
-
     private var messages: [TranscriptItem] {
         appState.sessionTranscripts[sessionId] ?? []
     }
@@ -365,7 +361,6 @@ struct SessionTranscriptView: View {
                         TranscriptItemsView(items: messages, terminalID: nil)
                     }
                     .defaultScrollAnchor(.bottom)
-                    .scrollPosition(id: $visibleID, anchor: .bottom)
                     .onScrollGeometryChange(for: Bool.self) { geometry in
                         let viewportBottom = geometry.contentOffset.y + geometry.containerSize.height
                         let contentBottom = geometry.contentSize.height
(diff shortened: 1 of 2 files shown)


cheapsteak:
Thanks for the careful review — both findings stand up.

**Medium (SessionTranscriptView dead code)**: confirmed. `visibleID` was never written, the shared `TranscriptItemsView` `LazyVStack` doesn't apply `.scrollTargetLayout()` so the framework can't write back on user scroll either, and `@State` resets across the view-identity change when the user picks a different session anyway — so even with a writeback the position wouldn't persist across switches without `AppState`-keyed storage. Took option B and reverted the `SessionTranscriptView` changes entirely in 769fb97. The original blank-flash bug only affects the *live* pane (which gets new messages); the history pane doesn't have the symptom. Test-plan claim about history-pane position preservation removed from scope.

**Minor (unconditional autoscroll in `LiveTranscriptPaneView`)**: addressed in 20cdf8d. Gated the `.onChange(of: messages.last?.id)` on the existing `atBottom` flag so a user scrolled up to read older messages is no longer yanked down by new content. The jump-to-bottom button keeps its meaning: at-bottom → follow; scrolled up → stay and let the user opt back in. The design spec deferred the broader `autoscrollEnabled` feature but this is narrower — reusing the pane's already-maintained `atBottom`, not introducing new state.

Issue #129, opened 2026-05-10
Tracking: recurring main-thread UI hangs in the transcript pane


cheapsteak:
## What we're tracking

A recurring class of main-thread hangs (sub-second to 17s) located in SwiftUI layout of `TranscriptItemsView` (`Sources/TBDApp/Panes/Transcript/`) under `LazyVStack` + variable-height markdown rows + nested `ScrollView`s inside row cards. Multiple landed fixes have peeled off individual contributors, but the underlying architecture is fragile.

This issue is the umbrella; specific fix PRs link back here.

## Recent evidence

- 2026-05-10: 1.05s hang sampled by macOS hang reporter (sample-only — not a fatal crash). Stack: `LazyVStack.place → ForEach → ChatBubble row → ScrollViewLayoutComputer.Engine.sizeThatFits → StyledTextLayoutEngine`. Suspect: `BashCard.swift:71/105` — `.frame(maxHeight: containerExpanded ? .infinity : 120)`. The `.infinity` branch removes the row's upper bound, forcing the LazyVStack to measure inner ScrollView content for sizing.

## Already done (not duplicates)

- PR #120 (`a41f716`) — gated `.textSelection(.enabled)` on hover; fixed 17s storm.
- `2da779c` — hoisted ScrollView out of recursive `TranscriptItemsView` calls.
- `5e10ba8` — `HangWatchdog` ships hang detection telemetry (threshold 1500ms).
- `29348d3` — defer `onPreferenceChange` writes to avoid reentrant layout.
- `3fa2cc6` — persist NSHostingView in expanding row panel to break layout-update loop.
- `dd6c1cf` — capped daemon body to 20 lines initial.
- #102 — LazyVStack-vs-per-card-@State tradeoff (separate concern, intentionally accepted).

## Next steps to investigate (in rough order)

1. **Replace `.frame(maxHeight: .infinity)` in `BashCard` and `WriteCard` expand state with a finite cap** (e.g., 600pt). Cheap, local, addresses today's hang signature. Validate with `perf-transcript` instrumentation.
2. **Tighten `HangWatchdog.defaultThresholdMs`** from 1500ms to ~1000ms so sub-second hangs like today's 1.05s sample are captured by our own logs, not just by macOS hang reports the user has to hand-collect.
3. **Audit the rest of the transcript row cards** for the same pattern (`.frame(maxHeight: .infinity)` on a nested ScrollView inside a LazyVStack row).
[...]

PR #130, opened 2026-05-11
Fix: Stop expanded transcript cards from freezing the UI

cheapsteak:
## Summary

The transcript pane intermittently freezes for ~1+ second after a new message lands or on scroll, especially in long sessions. A macOS hang report sampled today showed all 11 main-thread samples stuck in SwiftUI layout: `LazyVStack.place` → per-row `_FlexFrameLayout.sizeThatFits` → recursive `ScrollViewLayoutComputer.Engine.sizeThatFits` → `StyledTextLayoutEngine`.

Root cause: `BashCard` and `WriteCard` use `.frame(maxHeight: containerExpanded ? .infinity : 120)` on the inner output ScrollView. Inside a `LazyVStack` row, `.infinity` strips the upper bound, so `_FlexFrameLayout` cannot short-circuit — every row's inner ScrollView content gets fully measured during layout. Cost is super-linear with transcript length.

This PR:

- **Caps the expanded scroll height at 600 pt** (5x the collapsed 120) in `BashCard.swift` (command + result containers) and `WriteCard.swift`. Bounded enough for `_FlexFrameLayout` to short-circuit; generous enough for typical bash/write output. If 600 turns out to feel cramped we can tune up — but going back to `.infinity` reintroduces the hang.
- **Tightens `HangWatchdog.defaultThresholdMs` 1500 → 1000** so sub-second hangs like today's 1.05 s sample land in our own `os.Logger` (`subsystem com.tbd.app, category hang-watchdog`) instead of only in macOS hang reports the user has to hand-collect. Existing test's local constant + comment updated to match; stall values (100/3000/5000 ms) still bracket cleanly.

Issue #129 tracks the broader umbrella (more steps to consider, including a strategic move to `List` per `docs/superpowers/specs/research-2026-05-06-swiftui-long-list-perf.md`). This PR is the quick-win subset.

## Test plan

- [ ] `scripts/restart.sh` (worktree-relative, per CLAUDE.md) and verify `ps aux | grep TBDApp` shows the new binary
- [ ] Open a long bash-heavy transcript; expand a few BashCards; confirm no UI freeze on scroll or new message
- [ ] Confirm expanded BashCard/WriteCard caps at ~600 pt and scrolls internally past that
- [ ] Stream `log stream --level info --predicate 'subsystem == "com.tbd.app" AND category == "hang-watchdog"'` and confirm sub-second hangs (if any reproduce) now show up
- [ ] No regression in the existing 17 s text-selection-storm fix (PR #120) — text selection on hovered row still works

## Notes

- `swift build` clean.
- `swift test` not run — the existing test target fails in this toolchain with `no such module 'Testing'`, a pre-existing issue confirmed by stashing changes and re-running. The HangWatchdog test edits are doc/value-only so this PR doesn't widen the gap.
- Empirical perf measurement (re-sampling on a long transcript before/after) is left for the verification step above; the change is small and well-grounded in the spindump signature.

Refs #129

open in TBD: `tbd://open?worktree=D917BD15-077C-41E4-82E8-FFF69A960D3E`

### DIFF
--- a/Sources/TBDApp/Diagnostics/HangWatchdog.swift
+++ b/Sources/TBDApp/Diagnostics/HangWatchdog.swift
@@ -75,9 +75,11 @@ final class HangWatchdog: @unchecked Sendable {
     // MARK: Configuration
 
     /// Threshold in milliseconds beyond which a stall is reported as a hang.
-    /// 1500 ms is comfortably above the worst non-hang transcript layout pause
-    /// we've measured but well below the 17 s hangs in `crash.log`.
-    static let defaultThresholdMs: UInt64 = 1500
+    /// 1000 ms catches sub-second hangs like the 1.05 s sample on 2026-05-10
+    /// (issue #129) — at the previous 1500 ms threshold that event slipped
+    /// past our own diagnostics even though the macOS hang reporter caught
+    /// it. Still well above any expected steady-state main-thread pause.
+    static let defaultThresholdMs: UInt64 = 1000
 
     /// Background tick interval. 250 ms gives sub-half-second detection of
     /// onset and recovery without meaningful battery cost — the timer just
--- a/Sources/TBDApp/Panes/Transcript/BashCard.swift
+++ b/Sources/TBDApp/Panes/Transcript/BashCard.swift
@@ -68,7 +68,9 @@ struct BashCard: View {
                                 .padding(8)
                                 .frame(maxWidth: .infinity, alignment: .leading)
                         }
-                        .frame(maxHeight: commandContainerExpanded ? .infinity : 120)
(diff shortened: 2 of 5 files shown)


cheapsteak:
Addressed in 9cab305 — moved `600` (and the collapsed `120`) into a new `TranscriptCardLayout` namespace enum at `Sources/TBDApp/Panes/Transcript/TranscriptCardLayout.swift`. Three callsites in BashCard/WriteCard now read `TranscriptCardLayout.expandedMaxHeight` / `.collapsedMaxHeight`. Kept the per-site `// Finite cap…see #129.` comments — the *why* belongs at the callsite even though the *value* now lives in one place.

PR #134, opened 2026-05-11
Perf: flatten transcript ForEach body to kill estimatedCount recursion (#129)

### LINKED ISSUE #129: Tracking: recurring main-thread UI hangs in the transcript pane

cheapsteak:
## Summary

- Targets the recurring `_ViewList_Group.estimatedCount` recursion driving transcript-pane hangs by making the `LazyVStack { ForEach }` body satisfy Apple's documented constant-view-count rule. Tracking issue: #129 — **leave open after merging this PR**; longitudinal observation is the real validation and the bug class has survived prior fixes.
- Pre-computes `[TranscriptRenderNode]` outside `body`; filters hidden items upstream, inlines `ContextUsageBadge` as a row field, and collapses subagent disclosure into a single non-interactive summary row — eliminating the three cardinality violations in `rowFor`/`toolCardFor`/`SubagentDisclosure` recursion.
- Drops `.onScrollGeometryChange(... geometry.contentSize.height ...)` in both `LiveTranscriptPaneView` and `HistoryPaneView` (Gemini's hypothesis: reading content height every scroll event forced LazyVStack to compute total size on every transaction); replaces with a 1pt `Color.clear` sentinel at the bottom of the LazyVStack whose `.onAppear`/`.onDisappear` drives `atBottom`.
- Adds `OSSignposter` regions under `category: perf-transcript` (`transcript.swap`, `transcript.items.body`, `transcript.markdown.segment`, `transcript.scrollTo`) for before/after instrumentation.
- Subagent inline expansion is **descoped** — replaced with a single static summary row. A pop-out viewer for subagent activity is planned as a follow-up.

## Why this and not just "migrate to List"

`List` is the structural fallback if this doesn't hold (research doc + migration design doc are both in `docs/superpowers/specs/`). Today's investigation surfaced a more targeted fix that preserves the UX wins List can't match on macOS 15:

- `.defaultScrollAnchor(.bottom)` doesn't work on `List` ([Apple DTS forum #770682](https://developer.apple.com/forums/thread/770682)), so migration costs a first-paint flash. Empirically confirmed today.
- Drag-select across rows is contiguous in `LazyVStack`; ambiguous in NSTableView-backed `List`.
- `.scrollPosition(id:)` doesn't work on `List` either — would defer scroll-position-preservation as Phase-2 work.

Codex and Gemini independently converged on the flattening fix (see `docs/superpowers/specs/2026-05-11-transcript-hang-research-discussion.md`), grounded in Apple's [ForEach docs](https://developer.apple.com/documentation/swiftui/foreach) + WWDC23 ["Demystify SwiftUI Performance"](https://developer.apple.com/videos/play/wwdc2023/10160/) constant-view-count rule. The current code violated it in three documented ways; this PR addresses all three.

## Commits

| Hash | Description |
|---|---|
| `1281873` | `feat(transcript): add OSSignposter regions for perf baseline` — zero behavior change |
| `159b1bf` | `perf(transcript): drop onScrollGeometryChange contentSize read` — Gemini single-variable test |
| `4960199` | `feat(transcript): add TranscriptRenderNode model + TranscriptRow` — new types, not wired |
| `2d05c59` | `perf(transcript): flatten ForEach body via TranscriptRenderNode` — the actual fix; deletes SubagentDisclosure; also drops HistoryPaneView's contentSize reader |
| `82b492d` | `docs(transcript): briefs + impl spec for the rescue experiment` |

## Test plan

The hang is not reliably reproducible locally, so longitudinal observation is the real validation. After merging:

- [ ] No `hang detected stallMs=>1000` lines from `com.tbd.app/hang-watchdog` over normal multi-hour usage: `log show --last 24h --predicate 'subsystem == "com.tbd.app" AND category == "hang-watchdog"' --info --style compact`
- [ ] Visual sanity: transcript pane scrolls smoothly; subagent rows render as static `SubagentSummaryRow` (no chevron, no expand); context-usage badge appears under latest assistant turn; jump-to-bottom button hides/appears as expected on scroll in both `LiveTranscriptPaneView` and `HistoryPaneView`.
- [ ] If Instruments trace is captured under the SwiftUI template, `_ViewList_Group.estimatedCount` recursion should be ≤2 frames deep (was ≥40 in the 12:18 freeze).
- [ ] `swift build` clean (verified: 681 tests passing locally).

If hangs recur with the same `estimatedCount` signature, the flattening theory is wrong and we fall back to the List migration (design doc already in place: `docs/superpowers/specs/2026-05-06-transcript-list-migration-design.md`).

## Refs

- Tracking issue (keep open): #129
- Research: `docs/superpowers/specs/research-2026-05-06-swiftui-long-list-perf.md`
- Migration fallback: `docs/superpowers/specs/2026-05-06-transcript-list-migration-design.md`
- Impl spec: `docs/superpowers/specs/2026-05-11-transcript-render-node-design.md`
- Reviewer discussion (Codex + Gemini): `docs/superpowers/specs/2026-05-11-transcript-hang-research-discussion.md`

---

_Do not auto-close #129 on merge — this is an experimental fix in a recurring bug class. Will close manually after longitudinal observation confirms the hang has not recurred._


cheapsteak:
Addressed all three items from @claude's review.

**Medium — scroll-target mismatch on toolCall-with-subagent** → `558a8d8`. Added `lastRenderedNodeID(for:)` in `TranscriptRenderNode.swift` (walks `items` from the end; applies the same toolCall-with-subagent rule as `transcriptRenderNodes`, so it's O(1)-typical without materializing the full node array). Used it at all three scroll-target call sites — the `.onChange` autoscroll in `LiveTranscriptPaneView` and the jump-to-bottom button in both `LiveTranscriptPaneView` and `HistoryPaneView`. Preserved the existing `oldID`/`newID` guards on `.onChange`.

**Minor — misleading comment in `transcriptRenderNodes`** → `238962d`. Comment now reflects that only `.thinking` is filtered by `isHiddenInTranscript`; `.slashCommand` is silently dropped here (TranscriptParser folds it into `.userPrompt` in practice, so the branch is effectively dead).

**Minor — document at-bottom threshold trade-off** → `d1bf9e2`. Added a comment above the sentinel explaining the prior 50pt threshold → effective ~0pt, including the jump-button-flash-during-animation trade-off and the perf reason for the change.

`swift build` clean, no new warnings.

PR #139, opened 2026-05-13
Feat: Sidebar extends under the titlebar (Tower / Finder look)

cheapsteak:
## Summary
- Swap the custom `NSSplitViewController` wrapper (`AppKitSplitView`) for `NavigationSplitView` so the sidebar's vibrant material extends up under the titlebar — `NavigationSplitView` gets this treatment because the system adds an `NSToolbarItem.Identifier.sidebarTrackingSeparator` that AppKit keys off of; a SwiftUI `.toolbar { }` over a custom split controller cannot.
- Keep the deterministic sidebar visibility the custom wrapper was originally introduced for: `AppDelegate` walks the window's view hierarchy on launch and on the existing 5s heartbeat, finds SwiftUI's internal `NavigationSplitViewController`, and clears both `canCollapse` and `canCollapseFromWindowResize` on the sidebar item. SwiftUI actively resets `canCollapse = true` on every layout pass — the heartbeat re-application is load-bearing, not defensive.
- Window chrome (`titlebarAppearsTransparent`, `.fullSizeContentView`, `titlebarSeparatorStyle = .none`, `toolbarStyle = .unified`) applied from `AppDelegate` alongside the sidebar lock; both helpers log on state transitions only.

Source for the `canCollapseFromWindowResize` trick: [Apple Dev Forums thread 727968](https://developer.apple.com/forums/thread/727968).

## Test plan
- [ ] Sidebar's gray vibrant material extends up behind the traffic lights with no horizontal seam between titlebar and sidebar
- [ ] Shrink window to its minimum width — sidebar stays visible (does NOT collapse)
- [ ] Wake-from-sleep does not collapse the sidebar
- [ ] All toolbar items render correctly: back/forward, auto-suspend, conductor, repo filter, PR link, file-panel toggle
- [ ] Sidebar autosave (column width) still persists across launches

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=45DEFBC5-663C-4430-AC5A-0196609D56B5)


cheapsteak:
All three addressed:

- **Medium (window-targeting)** — d621acc: prefer `NSWindow.identifier == "main"`, fall back to the heuristic only if that finds nothing.
- **Minor (sidebar min-width regression)** — 27b3496: added `.navigationSplitViewColumnWidth(min: 220, ideal: 260, max: 400)` on `SidebarView()` to restore the 220pt floor.
- **Minor (non-functional sidebar toggle)** — cfde1ef: added `.toolbar(removing: .sidebarToggle)` to suppress the auto-inserted toggle since `canCollapse` is held false.

PR #141, opened 2026-05-13
Fix: Stop Codex tab from silently morphing into a shell on tab-switch

cheapsteak:
## Summary

- **What's broken:** Creating a Codex tab works — codex CLI renders, you can interact. Switch to another tab and back, and the tab is now a generic shell. No warning, no error.
- **Why it happens:** Codex panes run `codex --full-auto` as the only process (no shell wrapper). When the user briefly detaches, the codex pane dies. `TerminalPanelView`'s window-dead recovery auto-fires `recreateTerminalWindow`, which unconditionally respawns a `/bin/zsh` and stamps `label = "shell"` over the row — destroying the only thing identifying it as Codex.
- **What this PR does:** Adds a structural `kind: TerminalKind` enum (`.shell / .claude / .codex`) on `Terminal`, persisted via migration v22 with backfill from existing label heuristics. The recreate handler now branches on `kind`: codex respawns `codex --full-auto` with `CODEX_HOME` and keeps `kind`/`label` intact; claude/shell keep today's "becomes a shell" behavior unchanged. `WorktreeLifecycle+Reconcile` and the app's `isCodexTerminal` both check `kind` first, with the legacy `label == "Codex"` match as a fallback.

The daemon comment that read *"No structured type field exists; label matching is the only discriminator available"* is now obsolete and updated.

## Test plan

- [ ] Existing terminals (Claude / setup / shell) still spawn and run correctly after restart — migration backfills their `kind` from `label` + `claudeSessionID`.
- [ ] Create a new Codex tab → verify the codex prompt renders and you can interact.
- [ ] Switch tabs, switch back → tab still shows the chevron icon, still labeled "Codex", and the pane is still `codex --full-auto`.
- [ ] Click the "Recreate Window" button on a Codex tab → pane respawns as codex, identity preserved.
- [ ] Click "Recreate Window" on a Claude tab → still becomes a plain shell (existing escape-hatch behavior is unchanged).
- [ ] `MigrationV22Tests` covers kind round-tripping for all three variants on a fresh DB.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=0A9CE90C-9348-412D-9B89-89E095098E56)

### DIFF
--- a/Sources/TBDApp/TabBar.swift
+++ b/Sources/TBDApp/TabBar.swift
@@ -292,7 +292,7 @@ private struct TabBarItem: View {
     }
 
     private var isCodexTerminal: Bool {
-        terminal?.label == "Codex"
+        terminal?.kind == .codex || terminal?.label == "Codex"
     }
 
     private var isSuspended: Bool {
--- a/Sources/TBDDaemon/Conductor/ConductorManager.swift
+++ b/Sources/TBDDaemon/Conductor/ConductorManager.swift
@@ -176,7 +176,8 @@ public final class ConductorManager: Sendable {
             worktreeID: worktreeID,
             tmuxWindowID: window.windowID,
             tmuxPaneID: window.paneID,
-            label: "conductor:\(name)"
+            label: "conductor:\(name)",
+            kind: .shell
         )
 
         try await db.conductors.updateTerminalID(conductorID: conductor.id, terminalID: terminal.id)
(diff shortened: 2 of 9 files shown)


cheapsteak:
Addressed in b65cfd8 — `clearRecreated` now stamps `kind = .shell` alongside the existing `label = "shell"`, so the field stays consistent with the actual pane state on claude→shell recreates.

Issue #142, opened 2026-05-13
Sidebar: drag-to-nest worktrees (failed MVP, deferred)


cheapsteak:
## Goal

Let users manually nest worktrees in the sidebar via drag-and-drop. Drag worktree A onto worktree B → A becomes B's last child. This is the manual counterpart to the CLI auto-parent path (\`tbd worktree create\` from inside a TBD-managed terminal already nests via \`TBD_WORKTREE_ID\`).

Other forms (drag to un-parent / promote, cross-repo re-parent via drag, drag-to-reorder-children) are out of scope here — handled via context menu or CLI.

## Status

**Deferred.** All approaches we tried failed on macOS. The PR that introduced the rest of the nested-worktree feature ships without manual drag-to-nest. CLI auto-parent + the existing \`.onMove\` top-level reorder remain the supported manual paths.

## What's already in place (do NOT rebuild)

- \`Worktree.parentWorktreeID: UUID?\` (TBDShared) + DB migration \`v22_worktree_parent\`.
- Daemon \`worktree.move\` RPC with cycle / self / not-main / parent-existence validation (\`WorktreeStore.move\`).
- \`AppState.moveWorktree(id:newParentID:newSortOrder:)\` (optimistic + RPC + rollback).
- \`AppState.children(of:)\` + \`findWorktree(id:)\` helpers.
- Recursive sidebar render via \`WorktreeSubtreeView\` (each worktree + descendants).
- Visual hierarchy: 16pt-per-level indent + 1pt vertical guide line per ancestor depth in \`WorktreeRowView\`.
- Top-level reorder via SwiftUI \`List\`'s \`.onMove\` calling \`AppState.reorderTopLevelWorktrees\` which sends the bulk \`worktree.reorder\` RPC (daemon-side scoped to \`parentWorktreeID IS NULL\` so nested children's sortOrders aren't trampled).
- Archive menu item disabled on rows with active children, with tooltip \"Archive nested worktrees first\".
- CLI: \`--parent <id-or-name>\`, \`--sibling\`, \`--no-parent\` flags on \`tbd worktree create\`; default auto-parents to \`TBD_WORKTREE_ID\`'s worktree.

So a fresh implementer only needs to wire the drag/drop UI itself — all data plumbing exists.

### Original spec design

[...]

PR #143, opened 2026-05-13
Feat: Nest worktrees by their spawner so the sidebar shows lineage

cheapsteak:
## Summary

- **CLI auto-parent**: `tbd worktree create` from inside a TBD-managed terminal now nests the new worktree under its caller (using `TBD_WORKTREE_ID`). Multi-level supported. Cross-repo lineage works (child can live in a different repo from its parent, rendered under the parent with a muted `(repo-name)` suffix).
- **Manual nested-create**: each non-main worktree row gets a context-menu "Create Nested Worktree" item and a hover-revealed `+` button on the right.
- **CLI flags**: `--parent <id-or-name>`, `--sibling`, `--no-parent` on `tbd worktree create` to override the default auto-parent behavior.
- **Sidebar render**: recursive — children appear indented under their parent (16pt per level) with thin vertical guide lines at each ancestor depth. Top-level reorder via List's native `.onMove`.
- **Archive guard**: a worktree with active children can't be archived. The Archive menu item is disabled with the tooltip "Archive nested worktrees first"; the CLI surfaces the same error.
- **Data model**: new `parentWorktreeID` column on `worktree` (migration v23) + new `worktree.move` RPC + `worktreeMoved` state delta. `Worktree.sortOrder` is now scoped per sibling group (top-level OR children-of-parent), not per repo.

## Caveats

- **Drag-to-nest deferred to #142.** SwiftUI's `.onMove` (which we use for top-level reorder) and `.draggable` (which we'd need for nest) are mutually exclusive on the same List row — `.onMove` consumes the drag gesture before `.draggable` can engage. We tried 8 distinct API combinations across two debugging sessions before retreating. Issue captures the full failure log + a viable path forward for a future implementer (drop `.onMove`, do reorder + nest both via `.draggable` + drop targets).
- **Lost click-to-rename.** The custom `.onTapGesture` we had on each row was preempting `NSTableView`'s row-drag tracker (FB7367473 family), so `.onMove` only initiated from the empty right-hand gutter. Removing the gesture lets selection + drag flow through `List(selection:)` natively. Rename is still available via the context menu "Rename..." and the inline editor that auto-appears on worktree create.

## Test plan

- [ ] From a TBD-managed terminal: `tbd worktree create` → new row appears indented under the current worktree.
- [ ] `tbd worktree create --no-parent` → flat. `--sibling` → same level as caller. `--parent <name>` → explicit parent.
- [ ] Right-click a worktree row → "Create Nested Worktree" creates a child.
- [ ] Hover a row → `+` button on the right creates a child.
- [ ] Drag a top-level row to reorder → reorders and persists across restart.
- [ ] Right-click a parent with active children → Archive item disabled, tooltip explains why.
- [ ] Restart the daemon while nested → layout survives (`worktree.parentWorktreeID` persisted in v23).
- [ ] Stop daemon, set a bad `parentWorktreeID` in `~/tbd/state.db`, restart → orphan parent pointer is nulled on reconcile, child reverts to top-level.
- [ ] Selection: single-click selects, cmd-click toggles multi-select, context-menu "Rename..." works.

Spec: `docs/superpowers/specs/2026-05-12-nested-worktrees-design.md`
Follow-up: #142 (drag-to-nest)

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=4DE489CE-1ACE-4D88-9C14-1BE36B4716CC)

### DIFF
--- a/Sources/TBDApp/AppState+Repos.swift
+++ b/Sources/TBDApp/AppState+Repos.swift
@@ -5,6 +5,13 @@ import os
 private let logger = Logger(subsystem: "com.tbd.app", category: "AppState+Repos")
 
 extension AppState {
+    // MARK: - Repo Lookups
+
+    /// Display name of a repo by ID, if known.
+    func repoName(for repoID: UUID) -> String? {
+        repos.first(where: { $0.id == repoID })?.displayName
+    }
+
     // MARK: - Repo Actions
 
     /// Add a repository by path.
--- a/Sources/TBDApp/AppState+Worktrees.swift
+++ b/Sources/TBDApp/AppState+Worktrees.swift
@@ -267,6 +267,50 @@ extension AppState {
         }
     }
 
+    // MARK: - Reorder top-level
+
+    /// Reorder ONLY the top-level worktrees of a repo (parentWorktreeID == nil),
+    /// triggered by SwiftUI `.onMove` whose indices index the top-level ForEach.
+    /// Nested children stay attached to their parents — only top-level sortOrders change.
+    /// Updates locally first (optimistic), then persists via RPC; rolls back on error.
+    func reorderTopLevelWorktrees(repoID: UUID, fromOffsets source: IndexSet, toOffset destination: Int) {
+        let previous = worktrees[repoID]
+        var rows = (worktrees[repoID] ?? [])
+        // Snapshot the top-level order BEFORE the move (matches the ForEach).
+        var topLevel = rows
+            .filter { ($0.status == .active || $0.status == .creating) && $0.parentWorktreeID == nil }
+            .sorted { $0.sortOrder < $1.sortOrder }
+        logger.notice("reorderTopLevel BEFORE: \(topLevel.map(\.displayName).joined(separator: " | "), privacy: .public) source=\(Array(source), privacy: .public) destination=\(destination, privacy: .public)")
+        // Items the user is moving (typically one).
+        let movedIDs = source.map { topLevel[$0].id }
+        // Apply the swap to derive the new top-level order.
+        topLevel.move(fromOffsets: source, toOffset: destination)
+        logger.notice("reorderTopLevel AFTER: \(topLevel.map(\.displayName).joined(separator: " | "), privacy: .public)")
+
+        // Optimistic local update: reassign sortOrders for the new top-level order.
+        for (i, wt) in topLevel.enumerated() {
+            if let idx = rows.firstIndex(where: { $0.id == wt.id }) {
+                rows[idx].sortOrder = i
+            }
+        }
+        worktrees[repoID] = rows
+
+        // Persist via the bulk reorder RPC. Daemon renumbers all listed worktrees
+        // to contiguous sortOrders matching the new top-level order, avoiding
+        // gappy/non-contiguous values from prior individual moves.
+        let orderedIDs = topLevel.map(\.id)
+        _ = movedIDs  // suppress unused warning; kept for log readability earlier
+        Task {
+            do {
+                logger.notice("RPC worktree.reorder ids=\(orderedIDs.map { $0.uuidString.prefix(8) }.joined(separator: ","), privacy: .public)")
+                try await daemonClient.reorderWorktrees(repoID: repoID, worktreeIDs: orderedIDs)
+            } catch {
+                logger.error("reorderTopLevelWorktrees RPC failed: \(error.localizedDescription, privacy: .public)")
+                await MainActor.run { self.worktrees[repoID] = previous }
+            }
+        }
+    }
+
     // MARK: - Move (nested worktrees)
 
     /// Move a worktree to a new parent (or top-level) and sortOrder.
--- a/Sources/TBDApp/AppState+Worktrees.swift
+++ b/Sources/TBDApp/AppState+Worktrees.swift
@@ -259,8 +259,6 @@ extension AppState {
             .filter { ($0.status == .active || $0.status == .creating) && $0.parentWorktreeID == nil }
             .sorted { $0.sortOrder < $1.sortOrder }
         logger.debug("reorderTopLevel BEFORE: \(topLevel.map(\.displayName).joined(separator: " | "), privacy: .public) source=\(Array(source), privacy: .public) destination=\(destination, privacy: .public)")
-        // Items the user is moving (typically one).
-        let movedIDs = source.map { topLevel[$0].id }
         // Apply the swap to derive the new top-level order.
         topLevel.move(fromOffsets: source, toOffset: destination)
         logger.debug("reorderTopLevel AFTER: \(topLevel.map(\.displayName).joined(separator: " | "), privacy: .public)")
@@ -277,7 +275,6 @@ extension AppState {
         // to contiguous sortOrders matching the new top-level order, avoiding
         // gappy/non-contiguous values from prior individual moves.
         let orderedIDs = topLevel.map(\.id)
-        _ = movedIDs  // suppress unused warning; kept for log readability earlier
         Task {
             do {
                 logger.debug("RPC worktree.reorder ids=\(orderedIDs.map { $0.uuidString.prefix(8) }.joined(separator: ","), privacy: .public)")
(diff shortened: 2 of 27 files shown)


cheapsteak:
Addressed all three issues from the review:

- **`logger.notice` → `logger.debug`** in `reorderTopLevelWorktrees` (3 sites) — commit `a54f47c`. Drag traces no longer persist to the unified log store; activated via `log stream --level debug` per CLAUDE.md.
- **`public var` → `public let`** on the four new `WorktreeCreateParams` nesting fields — commit `ae38119`. Matches the existing pattern in the same struct.
- **Removed dead `AppState.reorderWorktrees`** — commit `806a05f`. The flat-list reorder was orphaned when `.onMove` switched to `reorderTopLevelWorktrees`, and (as you noted) its index math would have been wrong with nested children sharing the repo bucket. `DaemonClient.reorderWorktrees` stays — `reorderTopLevelWorktrees` still calls it for the bulk renumber.

Good catches on all three. The filtered-out items also looked right to skip (the pre-existing `print()` in RPCRouter is a separate cleanup, and the archive-guard race + cross-repo render + `moveWorktree` infra without a sidebar caller are all by-design as the followup work in #142 lands).

cheapsteak:
Addressed the three minors from the post-approval review:

- **Bounded cycle-detection walk** in `WorktreeStore.move` — commit `b9b93fb`. Added a visited Set so a pre-existing DB cycle can't hang the write transaction.
- **`print()` → `os.Logger`** in the background-create error path of `RPCRouter+WorktreeHandlers.swift` — commit `7853520`. Per CLAUDE.md.
- **SQL now matches its comment** — commit `77558dc`. The top-level sibling-shift SQL in `WorktreeStore.move` now actually excludes `main` (`AND status != 'main'`).

Skipping the forward-looking note on `WorktreeDelta` carrying `parentWorktreeID`: agree it's only a concern if/when we add a multi-client subscription model, which TBD doesn't have today. Captured for when we do.

`swift test --filter WorktreeMoveTests` still 10/10 green.

cheapsteak:
Addressed the medium finding:

- **`nullOrphanedParents` now detects + breaks cycles** at startup (`451e4c4`). For each row with a non-null parent, the walk uses a visited set; if a node is revisited (cycle), the original row's parent is NULL'd to break the chain. Minimal damage — one row's parent pointer changes per cycle.
- **Depth cap (50) on `WorktreeSubtreeView`** as a UI-layer backstop. If a cycle is ever introduced live (manual sqlite edit, future regression), the recursive view logs a warning at `.error` level and stops rendering instead of stack-overflowing.
- **Test added**: `cycleParentPointerIsBrokenByReconcile` injects a cycle via `writer.write` (the public API's cycle guard correctly refuses to create one), verifies `nullOrphanedParents` breaks it. 3/3 reconcile tests passing.

Filtered items + the forward-looking `WorktreeDelta` note remain noted-but-skipped; happy to revisit if a follow-up needs them.

cheapsteak:
Both addressed:

- **`allWorktreesOrdered` does a depth-first tree walk** — commit `bf23f51`. Walks main → top-level rows (by sortOrder) → recurse into each one's children (by sortOrder). Matches the sidebar's visual order so cmd+N keyboard nav lands on the right row. Cross-repo children come along with their parent via `children(of:)`.
- **Cycle-break is now startup-only** — commit `6abbe26`. Split `nullOrphanedParents` (cheap UPDATE for missing-parent orphans, still called per reconcile) from new `breakCyclicParents` (the O(N) walk, called once from `Daemon.start()` before the per-repo reconcile loop). Updated `cycleParentPointerIsBrokenByReconcile` to call the new function. 3/3 reconcile tests still pass.

Filtered items remain noted-but-skipped (CLI `--parent` case sensitivity is intentional — exact match avoids ambiguity; `children(of:)` O(N) scan is fine at TBD's worktree counts; the rest are pre-existing or by-design).

cheapsteak:
Addressed all four:

- **Replaced new `print()` in `Daemon.swift`** with `os.Logger.warning` — commit `da2cab8`. Added a file-scoped `daemonLogger` (subsystem `com.tbd.daemon`, category `startup`).
- **Removed dead `movedIDs`** in `reorderTopLevelWorktrees` — commit `db9ccb1`. The comment was right; the debug log it once fed had been simplified away.
- **Fixed stale comment** in `WorktreeSubtreeView`'s depth-cap docs to reference `breakCyclicParents()` (the actual cycle-break path post-split) instead of the old pre-split `nullOrphanedParents` — same commit.
- **`worktree.move` returns `.ok()`** for consistency with other write-only handlers — same commit. `DaemonClient.moveWorktree` uses `callVoidAsync`; `WorktreeMoveResult` deleted entirely.

Build clean, 11/11 daemon tests pass.

Sorry for the ironic regression — I'll start running `git grep "print(" Sources/` before pushing.

cheapsteak:
Addressed both findings:

- **`nullOrphanedParents` now also nulls pointers to archived parents** — commit `a195c06`. The revive-into-archived-parent scenario you described is real: archive C (leaf) → archive B (allowed, no active children) → revive C → C becomes invisible because `topLevelWorktrees` excludes children AND `WorktreeSubtreeView` skips archived parents' subtrees. Reconcile fires after revive, so the revived child is promoted to top-level automatically. The test that previously asserted "archived parent left alone" is renamed and inverted to verify the new behavior.
- **`appendSubtree` capped at depth 50** — commit `b51b91c`. Symmetric to `WorktreeSubtreeView.kMaxSubtreeDepth`. Latent today since DB-side cycle guards prevent in-memory cycles, but keeps the keyboard-nav path from blowing the stack while the renderer gracefully degrades.

Skipping the `moveWorktree` sibling-shift mirror — no UI caller for it today, and the daemon's `worktreeMoved` delta arrives within 2s on the next list refresh and corrects local state. Will fold that in when drag-to-nest (#142) lands.

Also: the previous Test workflow failure was a stale CI cache (cache restore-keys grabbed a pre-`TerminalKind` `.build/`, daemon then linker-failed against TBDShared). I re-ran the workflow and it should be on the latest commit now.

cheapsteak:
Addressed the P2:

- **`ParentResolver` now rejects archived parents** — commit `f580683`. Three call sites tightened with different policies:
  - `explicitParent` (`--parent <id>` or RPC field): throws new `WorktreeMoveError.parentIsArchived`. The user asked for a specific parent; refuse loudly.
  - `caller` (from `TBD_WORKTREE_ID` env): silent fallback to nil. Env may be stale from a previous session; quietly orphan beats failing.
  - `siblingOf`: silent fallback to nil. Same reasoning.
- 3 new tests covering each path. 13/13 parent-resolution + reconcile tests still pass.

This pairs with the reconcile fix (`a195c06`) — that one handles parents that were archived *after* the child was created; this one handles parents that are archived *at* create time.

cheapsteak:
Addressed both:

- **`WorktreeStore.move` now rejects archived parents** — commit `eb11a85`. Same guard as `ParentResolver` (the new `WorktreeMoveError.parentIsArchived` case), placed right after the `parentIsMain` check. When drag-to-nest (#142) eventually calls `worktree.move`, dropping on an archived row throws instead of producing an invisible child.
- **`assertArchivable` doc updated** to call out that the direct-children-only scope is intentional, with the partially-archived-chain edge case described and the reconcile-side self-heal explained.
- New test `movingUnderArchivedParentIsRejected` covers the move path. 22/22 tests across the four relevant suites pass.

Five rounds was a lot — appreciate the patience. With this, both create-time and move-time paths refuse archived-parent attachments, and reconcile cleans up anything that slipped through (or was created before this PR landed).

cheapsteak:
Picked up the two minors:

- **Reconcile.swift comment** updated to match the new `nullOrphanedParents` behavior (missing OR archived) — commit `bda5b76`.
- **`WorktreeMoveError`** got a docstring noting it's also raised by `ParentResolver` during creation, so future contributors don't wonder why a "MoveError" surfaces in a create path. Same commit.
- **PR body** updated: migration number is `v23_worktree_parent` (renamed from `v22` during the rebase against main's `v22_terminal_kind`).

Skipping the rest of the filtered minors — all either pre-existing patterns or fine at TBD's scale.

PR #145, opened 2026-05-13
Reliability: Stop renumbered migrations from bricking the daemon on startup (2 incidents in 2 weeks)

cheapsteak:
## Summary

- **What kept breaking.** Twice in the last two weeks (May 6 and May 13) the daemon crashed on startup with `SQLite error 1: duplicate column name`. Both times the schema change was already applied — only GRDB's bookkeeping was stale because two branches each registered the same additive migration under different string IDs (e.g. `v15_archived_head_sha` vs `v16_archived_head_sha`, `v22_worktree_parent` vs `v23_worktree_parent`). Each incident required manually running `INSERT INTO grdb_migrations (identifier) VALUES (...)` to unstick the daemon.
- **What this PR adds.** Three idempotent helpers on `GRDB.Database` (`addColumnIfMissing`, `createTableIfNotExists`, `addIndexIfMissing`) so future migration bodies become logged no-ops instead of fatal aborts when re-applied under a different ID. Plus a best-effort `VACUUM INTO` snapshot of `state.db` before any migration with pending work — a one-command rollback if a future migration goes sideways.
- **How we enforce it going forward.** A path-scoped `Sources/TBDDaemon/Database/CLAUDE.md` instructs future migrations to use the helpers and explains the snapshot. Existing migrations v1–v22 are intentionally left alone to preserve their applied-state semantics.

Out of scope (deferred after discussion): retrofitting v1–v22 (risk of altering applied-state), replacing the daemon's `Fatal` path with a degraded-mode RPC (separate change), and `dropColumnIfPresent` (no current need).

## Test plan

- [x] `swift build` clean
- [x] `swift test --filter MigrationHelpers` — 9/9 pass (includes the May-13 reproducer: raw \`t.add(column:)\` under one ID, then \`addColumnIfMissing\` under a different ID against the same DB)
- [x] `swift test --filter Migration` — 35/35 pass (existing migration tests still green)
- [x] `swift test --filter Database` — 27/27 pass
- [x] \`scripts/restart.sh\` produces exactly one TBDDaemon + one TBDApp from this worktree
- [x] Clean startup, no snapshot written (no pending migrations) — correct steady-state behavior

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=340E2F8F-1A1E-4CF8-84BA-33437B39F724)

### DIFF
--- a/Sources/TBDDaemon/Database/CLAUDE.md
+++ b/Sources/TBDDaemon/Database/CLAUDE.md
@@ -1,0 +1,39 @@
+# Database / Migrations
+
+## Use the idempotent helpers in new migrations
+
+New migrations that add columns, tables, or indexes MUST go through
+`MigrationHelpers.swift`:
+
+- `addColumnIfMissing(table:column:type:defaults:)`
+- `createTableIfNotExists(_:body:)`
+- `addIndexIfMissing(_:on:columns:unique:where:)`
+
+Do not call `t.add(column:)`, `db.create(table:)`, or raw `CREATE INDEX`
+directly in a new migration. Reason: parallel branches that renumber the same
+additive migration have bricked the daemon twice (May 6 and May 13 2026). When
+the schema already has the column but the renamed migration ID is unapplied,
+GRDB re-runs the body and SQLite throws `duplicate column name`. The helpers
+turn that scenario into a logged no-op.
+
+## Migration ID convention
+
+Continue `v<N>_descriptive_suffix` (e.g. `v23_repo_archived_at`). The numeric
+prefix preserves ordering; the suffix makes parallel branch collisions
+debuggable.
+
+## Don't edit existing migration bodies
+
+v1 through v22 are frozen. They have run on user machines; mutating them now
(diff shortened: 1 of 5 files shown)


cheapsteak:
Both findings addressed.

**Medium — `VACUUM INTO` was a no-op:** confirmed real. `pool.write` opens a deferred transaction; SQLite refuses `VACUUM INTO` outside autocommit; the catch-and-log path was swallowing the error so startup looked clean. Fixed by switching to `pool.writeWithoutTransaction` (a752ab8). Added `PreMigrationSnapshotTests` that would have caught this: asserts the snapshot file is created, non-empty, and round-trips a row from the source DB.

**Minor — v23 grandfathering:** v23 landed on main via #143 before this PR opened, so updated `Sources/TBDDaemon/Database/CLAUDE.md` to say v1–v23 are frozen (69c590d). Per the same doc, rewriting an applied migration body to use the new helpers would itself be a divergence-risk, so grandfathering was the right call here.

Thanks for the catch on the snapshot — that would have bitten badly the first time we actually needed a rollback.

PR #146, opened 2026-05-14
Compound: Stop print() from landing in Sources/ — SwiftLint at CI and pre-push

cheapsteak:
## Summary

- Adopts SwiftLint via [SimplyDanny/SwiftLintPlugins](https://github.com/SimplyDanny/SwiftLintPlugins) (the SwiftLint README's recommended fork — ships SwiftLint as a binary artifact bundle, so contributors need no `brew install`). Pinned `from: "0.63.2"`.
- Custom rule `no_print_in_sources` mechanically enforces the existing CLAUDE.md "No `print()` in `Sources/`" rule. Scoped via regex `included:` to `TBDShared`, `TBDDaemonLib`, `TBDDaemon`, `TBDApp` only — `TBDCLI` is excluded because its `print()` calls are intentional user-facing CLI output. Enforced at two layers: **CI** (`.github/workflows/test.yml` runs `swift package plugin swiftlint --strict`) and **pre-push hook** (`scripts/git-hooks/pre-push` lints changed files). The build-tool plugin attachments were attempted but removed (commit `f933a83`) because attaching the plugin to four targets exploded prebuild memory on the macos-15 GHA runner; CI + hook give the same enforcement floor without the cost.
- Replaces 17 `print()` calls in `TBDDaemon` with `os.Logger` to bring the existing tree to zero violations under the new rule, using the patterns PR #143 established (subsystem `com.tbd.daemon`, feature-shaped categories like `startup`, `reconcile`, `http`, `socket`, `database`, `worktreeHandlers`).

PR #143 had a chain of "Address review" commits hand-fixing exactly these violations after the fact (commits `7853520`, `da2cab8` on that PR). This PR is the [Compound](#) follow-up so the same rule never has to be enforced through review again.

## Scope decisions worth a second look

- **`.swiftlint.yml` disables 16 default rules** (e.g. `line_length`, `function_body_length`, `identifier_name`, `force_try`, `todo`). All fired on existing code; rather than reformat now, each is silenced with a one-line reason. Future PRs can pick rules off this list one at a time and clean up.
- **No `force_try`/`force_cast` severity bumps, no `weak_delegate`, no other rules added.** Punted to follow-ups per the brief's "minimum scope" guidance.
- **No CLIOutput shim refactor in TBDCLI.** Considered and rejected — the CLAUDE.md rule's rationale (use `os.Logger`) is about diagnostic logging, not stdout output. CLAUDE.md updated to clarify the scope explicitly.
- **Build-tool plugin attachments are NOT included.** They were in the initial commit (`70c8800`) but dropped in `f933a83` after they OOM'd the macos-15 runner during prebuild. Local enforcement is via the pre-push hook; CI enforcement is the dedicated `SwiftLint (strict)` job step. `swift build` does NOT run SwiftLint.

Full design rationale: `docs/superpowers/specs/2026-05-13-swiftlint-adoption-research.md`.

## Test plan

- [ ] `swift build` exits clean.
- [ ] Adding a `print("test")` to `Sources/TBDDaemon/Daemon.swift` is caught by **CI** (`SwiftLint (strict)` step) and by the **pre-push hook** — `swift build` itself does not run SwiftLint, so it will compile locally until the hook or CI rejects it.
- [ ] Adding the same `print("test")` to `Sources/TBDCLI/Utilities.swift` does NOT fail SwiftLint (TBDCLI is excluded).
- [ ] `swift test --parallel` is green.
- [ ] `swift package plugin --allow-writing-to-package-directory swiftlint --strict` matches the CI invocation and exits 0 on a clean tree.
- [ ] After running `scripts/install-hooks.sh`, attempting to push a commit that introduces a `print()` to a covered file is blocked at the pre-push stage.
- [ ] The pre-push hook runs cleanly under macOS bash 3.2 (no `mapfile`).
- [ ] CI test workflow shows the new `SwiftLint (strict)` step passing.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=4C32CBD0-0C98-499A-9BAE-91882DAED30C)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,6 +34,14 @@ jobs:
         # drives TmuxManager against a real tmux server on a unique socket.
         run: brew install tmux
 
+      # macos-15 runner defaults to Xcode 16.4 (Swift 6.1.2), which has a
+      # SILGen/Sema crasher exercised by combinations of `try? await` on
+      # Void-async-throws and @Sendable analysis. The crash is deterministic
+      # (`error: fatalError` while compiling TmuxManager.swift) and was fixed
+      # in the Swift 6.2 release cycle. Select Xcode 26.1.1 (Swift 6.2.1).
+      - name: Select Xcode 26.1.1 (Swift 6.2.x) to avoid 6.1.2 compiler crash
+        run: sudo xcode-select -s /Applications/Xcode_26.1.1.app
+
       - name: Capture Swift toolchain version
         run: echo "SWIFT_VERSION=$(xcrun swift --version | head -1)" >> $GITHUB_ENV
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,6 +51,9 @@ jobs:
           restore-keys: |
             spm-build-${{ runner.os }}-${{ env.SWIFT_VERSION }}-
 
+      - name: SwiftLint (strict)
+        run: swift package plugin --allow-writing-to-package-directory swiftlint --strict
+
       - name: Test
         run: swift test --parallel
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,14 +34,10 @@ jobs:
         # drives TmuxManager against a real tmux server on a unique socket.
         run: brew install tmux
 
-      # macos-15 runner defaults to Xcode 16.4 (Swift 6.1.2), which has a
-      # SILGen/Sema crasher exercised by combinations of `try? await` on
-      # Void-async-throws and @Sendable analysis. The crash is deterministic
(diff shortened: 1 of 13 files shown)


cheapsteak:
On the TBDCLI-filter suggestion: tested locally and SwiftLint already honors the top-level `excluded:` for explicitly-passed paths. From `swift package plugin swiftlint --strict -- Sources/TBDCLI/Utilities.swift`:

```
Linting Swift files at paths Sources/TBDCLI/Utilities.swift
Done linting! Found 0 violations, 0 serious in 0 files.
```

Note "in 0 files" — the file is excluded from linting even when passed explicitly, so the hook is already correct as-is. Leaving the hook unchanged for now; happy to revisit if a future SwiftLint version changes that behavior.

PR #152, opened 2026-05-15
Compound: Stop --sibling from making orchestrators spawn peers instead of children

cheapsteak:
## Summary

A sibling Claude session running `tbd worktree create --sibling` four times got bitten by the flag's ambiguity: it parsed "sibling" as "sibling of the workers I'm spawning" and got peers of *itself* instead. The default behavior (nest under caller) also had no name, so an agent skimming `--help` couldn't grep for it and had to infer it by elimination.

- Replace `--parent` / `--sibling` / `--no-parent` with a single `--position=child|sibling|root` enum (default `child`).
- Every value now names the *new worktree's* tree position relative to the caller — no reference-point ambiguity, no nameless default.
- Updated the bundled `tbd` skill with a worked example for the orchestrator-fan-out case.

Arbitrary-UUID parent attachment is no longer reachable via a dedicated CLI flag — the daemon RPC still supports it (used by the SwiftUI sidebar), and the env-spoof `TBD_WORKTREE_ID=<uuid> tbd worktree create --position=child ...` keeps it reachable from scripts. Adding a `--parent=<uuid>` flag later is ~30 lines if it turns out to matter.

## Test plan
- [ ] `swift test` passes (new `WorktreePositionTests` covers each position + caller-unset fallbacks)
- [ ] `tbd worktree create --help` shows the three positions with the reference point spelled out
- [ ] From inside a worktree: default (`--position=child` implicit) nests under caller
- [ ] From inside a worktree: `--position=sibling` becomes peer of caller
- [ ] `--position=root` creates top-level regardless of caller

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=412473E9-880C-464F-8CF1-BC79C5E7F439)

### DIFF
--- a/Sources/TBDCLI/Commands/WorktreeCommands.swift
+++ b/Sources/TBDCLI/Commands/WorktreeCommands.swift
@@ -42,14 +42,18 @@ struct WorktreeCreate: AsyncParsableCommand {
     @Option(name: .long, help: "Read initial prompt from a file (use - for stdin)")
     var promptFile: String?
 
-    @Option(name: .customLong("parent"), help: "Nest the new worktree under this parent (worktree ID or display name).")
-    var parent: String?
-
-    @Flag(name: .customLong("sibling"), help: "Spawn as a sibling of the caller (same parent as TBD_WORKTREE_ID).")
-    var asSibling: Bool = false
-
-    @Flag(name: .customLong("no-parent"), help: "Force the new worktree to be top-level (ignore TBD_WORKTREE_ID).")
-    var noParent: Bool = false
+    @Option(
+        name: .customLong("position"),
+        help: ArgumentHelp(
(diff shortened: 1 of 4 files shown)


cheapsteak:
Confirming the name-lookup removal is deliberate — the user is the sole consumer and chose a clean break over an escape hatch. The env-spoof `TBD_WORKTREE_ID=<uuid> tbd worktree create --position=child` covers the UUID case; the name case is now `tbd worktree list` → grep → env-spoof, which we're accepting.

Implemented the stderr warning for `--position=sibling` without a caller (4c48e55) — agreed that silently degrading to root is exactly the failure mode this PR is supposed to prevent. `--position=child` and `--position=root` without a caller stay silent because top-level is their documented graceful fallback.

PR #153, opened 2026-05-15
Isolate tests from the live daemon (~/tbd) and developer UserDefaults

cheapsteak:
## Summary

`TBDConstants` paths were hard-coded to `~/tbd` and `UserDefaults.standard` was reachable through `AppState`. Concurrent `swift test` runs across worktrees could collide on the live daemon socket / prod DB / developer's `TBDApp.plist`. PR #133 started the UserDefaults DI but only finished `autoSuspendClaude`; this PR finishes the rest and adds env overrides on `TBDConstants` so any future filesystem-touching test can opt in.

> **Note:** When this work was first scoped, the smoking gun was `ConductorManagerTests` writing into `~/tbd/conductors/` on every run. That test (and the entire Conductor feature) was retired on main in #147 while this PR was in flight, so the in-flight `IsolatedTBDHome` SwiftTesting trait was dropped at rebase time — it had no remaining consumer. The env-override layer is left in place for future filesystem-touching tests to wire up directly via `setenv`.

Production behavior is unchanged when env vars are unset and the default `.standard` is used.

## Commits

1. **`feat(shared): honor TBD_HOME and TBD_SOCKET_PATH env overrides`** — pure refactor of `TBDConstants` from `static let` → env-resolving `static var`. `TBD_HOME` redirects the base config dir (every derived path follows); `TBD_SOCKET_PATH` independently overrides the socket as an escape hatch for darwin's 104-char `sun_path` limit. New tests in `ConstantsTests.swift` cover both branches and the override precedence.
2. **`fix(app): finish routing AppState UserDefaults through the injected suite`** — closes the remaining `UserDefaults.standard` leaks in `AppState` (`dockRatio` reader/writer, `persistLayouts` / `restoreLayouts`) and adds a `userDefaults:` init param to `CLIInstallerCoordinator` + `LegacyHooksCoordinator` so their dismissal keys flow through the same suite. Regression tests assert the developer's `.standard` is never touched when a non-standard suite is injected.
3. **`docs: document TBD_HOME and TBD_SOCKET_PATH env overrides`** — adds a "Tests must not touch ~/tbd" section to `CLAUDE.md`.

## Design tradeoffs

- **Scope: test isolation only.** Production multi-instance daemons stay out of scope — LaunchServices, port allocation, and menu-bar coordination would explode the diff with no immediate payoff.
- **Two env vars, not one.** `TBD_SOCKET_PATH` is independently overridable so a deep `TBD_HOME` (sandboxed CI tmpdir) can still bind a short `/tmp/` socket without overflowing `sun_path`.
- **`@AppStorage` views are NOT refactored.** Today's tests don't render those views, so the `@AppStorage(...)` → `.standard` leak is latent. Refactoring all of them would require threading a UserDefaults through the SwiftUI environment — bigger blast radius, no concrete bug to motivate it.

## Out-of-scope follow-ups noticed (not in this PR)

- `registerFocusObservers()` + `startMemoryPressureMonitor()` still run unconditionally under tests. Benign today; the #149 reviewer flagged it.
- Existing tests (`GitStatusTests`, `WorktreeLifecycleTests`, `TBDWorktreeIDLeakTests`, `ArchivedReviveBranchRenameTests`) write `tbd-test-<UUID>` dirs into `$TMPDIR` without cleanup — 1800+ residue dirs accumulated locally. Pre-existing, not addressed here.
- The view-level `UserDefaults.standard` / `@AppStorage` call sites (`EmojiPickerView`, `OpenInEditorButton`, `ContentView`, `SettingsView`, etc.) remain unwrapped. Refactor when a test path needs them.

## Test plan

- [x] `swift build` — clean.
- [x] `swift test` — 792 tests in 82 suites, all passing.
- [x] `swift test --filter TBDSharedTests` — includes 6 new env-override tests.
- [x] `swift test --filter AppStateUserDefaultsIsolationTests` — 3 new tests confirming AppState never touches `.standard` when given an injected suite.

[isolate-test-paths](https://cheapsteak.github.io/tbd/open/?worktree=515CE532-6D1A-43F2-A595-9204E38CA7AF)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -29,6 +29,14 @@ There should be exactly one `TBDDaemon` and one `TBDApp`, both from the worktree
 ### NEVER delete ~/tbd/state.db
 The database stores worktree display names, custom config, and notification history. Deleting it orphans tmux servers (repo UUID changes → tmux server name changes → old sessions become unreachable). If you encounter DB issues, diagnose and fix the schema/code — don't wipe the DB.
 
+### Tests must not touch ~/tbd
+`TBDConstants.configDir` and `socketPath` honor two env vars so tests, CI, and concurrent `swift test` runs across worktrees never collide with the live daemon or the developer's real config:
+
+- `TBD_HOME` redirects the base config directory (`~/tbd` by default). Every derived path (`socketPath`, `databasePath`, `pidFilePath`, `portFilePath`, `reposDir`) follows.
+- `TBD_SOCKET_PATH` overrides only the socket — an escape hatch for darwin's ~104-char `sun_path` limit when `TBD_HOME` is a deep tmp path.
+
+Tests that need filesystem isolation should `setenv("TBD_HOME", "/some/tmp/dir", 1)` (and optionally `TBD_SOCKET_PATH`) before any `TBDConstants.<path>` access. Tests that mutate `UserDefaults` should construct `AppState(userDefaults: UserDefaults(suiteName:))` and tear the suite down with `removePersistentDomain(forName:)` — `UserDefaults.standard` on this unbundled executable is the developer's real `TBDApp.plist`.
+
 ### Database migrations must update the shared model
 When adding a DB column in `Sources/TBDDaemon/Database/Database.swift`:
 1. Add the column with a `.defaults(to:)` value in the migration
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -142,7 +142,7 @@ final class AppState: ObservableObject {
     }
 
     @Published var dockRatio: CGFloat = 0.3 {
-        didSet { UserDefaults.standard.set(Double(dockRatio), forKey: Self.dockRatioKey) }
+        didSet { userDefaults.set(Double(dockRatio), forKey: Self.dockRatioKey) }
     }
     /// Pixel size of the main terminal area (the SingleWorktreeView slot
     /// inside DockSplitView, excluding the pinned dock and file panel).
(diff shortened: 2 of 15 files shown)


cheapsteak:
Addressing the two items raised:

### Medium — PR description / branch mismatch

The bot review fired against the original 4-commit body before I rebased on current `origin/main`. PR #147 retired the entire Conductor feature mid-flight, which made commit 2 (`feat(tests): add IsolatedTBDHome trait + migrate ConductorManagerTests`) consumer-less, so I dropped it during the rebase. The PR description above was rewritten at the same time to reflect the actual 3-commit content (env override + UserDefaults DI + docs) and explicitly call out the trait drop and the `conductorHeight` removal. The Minor sub-item about `conductorHeight` is also resolved by that edit.

### CI link failure — `unsafeMutableAddressor` on `TBDConstants.configDir`

CI under Xcode 26.3 / Swift 6 failed at the link step with:

```
Undefined symbols for architecture arm64:
  "TBDShared.TBDConstants.configDir.unsafeMutableAddressor : Foundation.URL"
   referenced from: ArchivedReviveBranchRenameTests.swift.o
```

The test never touches `TBDConstants` directly — it calls `HookResolver()` with no args. The compiler re-emits `globalHooksDir:`'s default expression (`= TBDConstants.configDir.appendingPathComponent(...).path`) at the test's call site, which under Xcode 26.3 generates a reference to the property's `unsafeMutableAddressor` — a symbol that exists for `static let` but not for the computed `static var` this PR introduced.

Fix (commit b2b0cf6): move every `init(x: T = TBDConstants.<computed>)` to the nil-sentinel pattern `init(x: T? = nil) { self.x = x ?? TBDConstants... }`. The `TBDConstants` access now happens inside the defining module's compilation context where the compiler emits a plain getter call. Same eight initializers across daemon, app, and CLI. Semantics unchanged — each `init()` call still re-reads the env at runtime through the `TBDConstants` getter. Local clean rebuild + `swift test` runs all 792 tests green.

PR #161, opened 2026-05-17
Feat: One-script migration from Conductor to TBD

cheapsteak:
## Summary

- New `tbd worktree adopt <path>` CLI primitive that registers an existing git worktree directory into TBD's database. No `git worktree add`, no tmux spawning — just a row insert with idempotent revive-if-archived semantics. Useful beyond Conductor (raw `git worktree add`, other tools).
- `scripts/import-conductor.sh` reads Conductor's SQLite DB at `~/Library/Application Support/com.conductor.app/conductor.db`, auto-registers any missing repos in TBD (deduplicating by `root_path`), and adopts each `ready` workspace in place. Files stay at `~/conductor/workspaces/<repo>/<name>` — Conductor keeps working alongside (dual-use is the design intent).
- Inherited for free via in-place adoption: Claude session transcripts (TBD reads `~/.claude/projects/<encoded-cwd>/`) and `conductor.json` hooks (TBD's `HookResolver` already reads them with a deprecation warning).
- Flags: `--all` (include archived Conductor workspaces), `--repo <name>` (limit to one), `--dry-run`.
- 7 unit tests for the lifecycle method; verified end-to-end against a real Conductor DB (3 workspaces adopted, idempotent re-run, 818+1 → 819 tests pass, SwiftLint clean).

Design doc: `docs/superpowers/specs/2026-05-16-conductor-import-design.md` (includes a "What's inherited for free" section explaining why we don't migrate `session_messages` or hook configs — TBD already finds them).

## Known issues (documented in design doc)

- **Archive is destructive for adopted worktrees.** `tbd worktree archive` runs `git worktree remove` in its background phase, which deletes the underlying Conductor directory. Revive recovers committed files but loses uncommitted changes, `.envrc`/`.direnv`/`node_modules`/IDE state. Workaround: don't archive adopted Conductor worktrees. Future fix: add a `worktrees.adopted` column so archive can skip the destructive step.
- **Detached-HEAD worktrees adopt with an empty branch column.** `GitManager.parseWorktreeList` doesn't surface the `HEAD <sha>` line for detached worktrees. Not hit by Conductor (always branches as `cw/<name>`). Pinned by `testAdoptDetachedHeadStoresEmptyBranch` so a future parser fix deliberately breaks it.

## Test plan

- [ ] `./scripts/import-conductor.sh --dry-run` prints a plan listing your `ready` Conductor workspaces marked `+ adopt`, with parent repos marked `~ reuse` (if already registered) or `+ add` (if not)
- [ ] `./scripts/import-conductor.sh --dry-run --repo <name>` filters to one Conductor repo
- [ ] `./scripts/import-conductor.sh --dry-run --all` includes archived workspaces (most will skip as "path missing" since Conductor cleans up archived dirs)
- [ ] `./scripts/import-conductor.sh` runs end-to-end, prints `[N/M] adopting … ok` lines, summary with `0 failed`
- [ ] Re-running is idempotent (no duplicate rows; second run still shows `+ adopt` in the plan but the `adopt` RPC returns the existing row)
- [ ] Adopted worktrees appear in the TBD sidebar with correct branch and name, indistinguishable from native worktrees
- [ ] Right-click → "Show in Finder" on an adopted worktree opens the actual `~/conductor/workspaces/...` directory
- [ ] For a worktree with prior Claude history (e.g. `atlanta`), the session count surfaces automatically
- [ ] `swift test` (819 tests pass) and `swift package plugin --allow-writing-to-package-directory swiftlint --strict` (0 violations) both clean

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=1702371D-C8EC-4EB8-B600-0A75F1C9C327)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,24 @@ scripts/restart.sh --quick  # skip build
 swift test
 ```
 
+## Migrating from Conductor
+
+Adopt your existing [Conductor](https://conductor.build) worktrees into TBD in place — no files moved, branches untouched, Conductor keeps working alongside. By default, only active (`ready`) Conductor workspaces are adopted, and any repos they reference are auto-registered in TBD.
+
+```sh
+./scripts/import-conductor.sh --dry-run    # preview
+./scripts/import-conductor.sh              # run
+```
+
+Flags:
+- `--all` — also adopt archived Conductor workspaces.
+- `--repo <name>` — limit to one Conductor repo (e.g. `--repo longeye-app`).
+- `--dry-run` — print the plan, don't write anything.
+
+Idempotent — safe to re-run as you create new Conductor worktrees.
+
+Existing Claude session transcripts and `conductor.json` hooks are picked up automatically — nothing extra to migrate.
+
 ## License
 
 Private / All rights reserved.
--- a/Sources/TBDCLI/Commands/WorktreeCommands.swift
+++ b/Sources/TBDCLI/Commands/WorktreeCommands.swift
@@ -249,7 +249,7 @@ struct WorktreeAdopt: AsyncParsableCommand {
     var json = false
 
     mutating func run() async throws {
-        let absolute = resolvePath(path)
+        let absolute = try canonicalizePath(path)
         guard FileManager.default.fileExists(atPath: absolute) else {
             throw CLIError.invalidArgument("Path does not exist: \(absolute)")
         }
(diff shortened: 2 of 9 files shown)


cheapsteak:
Addressed all four issues from the review:

**Medium 1 — symlink canonicalization** (`50f415e`): CLI now uses C `realpath(3)` before passing the path to the daemon. Used `realpath` rather than the suggested `URL.resolvingSymlinksInPath()` because Foundation doesn't follow `/var → /private/var` on macOS (verified — that was why the test originally needed its own `realpath` workaround).

**Medium 2 — spurious `worktreeCreated` broadcast** (`1167621`): `adoptWorktree` now returns `WorktreeAdoptOutcome` (`.inserted`, `.revived`, `.unchanged`). Handler picks the matching delta per case — `.worktreeCreated` for new rows, `.worktreeRevived` for archived→active flips (mirrors the existing revive RPC), nothing for already-active.

**Minor 1 — `displayName` ignored on revive** (`1167621`, same commit as #2 since they touch the same lifecycle method): revival now applies the override via `db.worktrees.rename`. Added `testAdoptRevivalAppliesDisplayNameOverride`.

**Minor 2 — step counter mismatch** (`949a53a`): orphaned workspaces now consume a step slot, so `[N/M]` reaches M.

820 tests pass (added 1), SwiftLint clean. Branch is now at `949a53a`.

cheapsteak:
Round 2 addressed (`674a1ad`):

- **Medium — summary counts inflated**: tracks `n_repo_actually_added` / `n_ws_actually_adopted`, incremented only on successful `tbd` exit. Summary now reflects what happened, not what was planned.
- **Medium — python3 undeclared dependency**: added to the precondition block alongside `tbd` and `sqlite3`. Reasoning: macOS Xcode Command Line Tools ship it (and a TBD dev env already needs CLT for Swift), so a hard fail is honest rather than silently degrading the plan preview.
- **Minor — `--repo` without value crashes with `shift` error**: guard prints `Error: --repo requires a Conductor repo name` + usage and exits 2.

Bundled into one commit since they're tightly co-located in the same script. 820 tests still pass, SwiftLint clean.

PR #163, opened 2026-05-17
Feat: Customize terminal font, colors, cursor, and thin strokes

cheapsteak:
## Summary
- New **Terminal** tab in Settings with live-applied controls: font family + size (NSFontPanel + inline stepper, Monaco 12 default), 8 bundled color schemes (Tango default — Tango palette on pure black), cursor style with blink variants, and a thin-strokes toggle that maps to SwiftTerm's new `fontSmoothing` property so text renders at iTerm-like weight on Retina displays.
- Every setting applies live to every open pane via a Combine subscription on `AppearanceSettings.objectWillChange`. Font changes also propagate cell-dim updates to the underlying tmux pane.
- Surfaces an info ⓘ tooltip next to the Scheme picker when the user's `~/.tmux.conf` defines `window-style`/`window-active-style`/`pane-style`/`default-style`, since those tmux options paint over the scheme's colors. We don't override the user's tmux config — we explain it.
- Bumps SwiftTerm from v1.12.0 to a main commit (`dae32cc8`) to pick up the public `fontSmoothing` property (PR #531) plus the v1.13.0 changes that came in between.

## Test plan
- [ ] Open Settings → Terminal: verify Font (with Choose Font… button + Size stepper), Color Scheme picker (8 entries), Cursor Style picker (6 entries), Thin Strokes toggle, and the relocated Auto-resize toggle in Experimental.
- [ ] Cycle each Color Scheme — every open terminal pane updates live; `ls --color` shows the new ANSI palette.
- [ ] Cycle each Cursor Style — caret shape/blink updates live.
- [ ] Change font (e.g. Menlo 14) via Choose Font… — every open pane updates live, no clipping at the bottom; tmux pane resizes to match.
- [ ] Toggle Thin Strokes — text visibly thins on Retina.
- [ ] Quit + relaunch — settings persist.
- [ ] Add `set -g window-style 'fg=colour247,bg=colour236'` to `~/.tmux.conf`, kill the TBD tmux server, relaunch — info ⓘ appears on the Scheme picker with the explanatory tooltip.
- [ ] Poison UserDefaults (`defaults write com.tbd.app terminal.scheme.id bogus`, `terminal.font.name NotARealFont`) — no crash, falls back to TBD Default + system mono.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=4E8CF284-2A26-4BC6-852D-DAAF7E71D107)

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "98d5c07c3f7f007250c8a918b5602f2c33fe71b469babf32f3dfb2ba1c2fff13",
+  "originHash" : "54e8d1aacfe50bff048354e01819161f554066fe1dfca9c8d088e22b86c67500",
   "pins" : [
     {
       "identity" : "grdb.swift",
@@ -105,8 +105,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/migueldeicaza/SwiftTerm",
       "state" : {
-        "revision" : "3c45fdcfcf4395c72d2a4ee23c0bce79017b5391",
-        "version" : "1.12.0"
+        "revision" : "dae32cc8f9bcda15713f4091bb2ea7e11f6dd57c"
       }
     },
     {
--- a/Package.swift
+++ b/Package.swift
@@ -10,7 +10,7 @@ let package = Package(
         .package(url: "https://github.com/groue/GRDB.swift", from: "7.0.0"),
         .package(url: "https://github.com/apple/swift-argument-parser", from: "1.3.0"),
         .package(url: "https://github.com/apple/swift-nio", from: "2.65.0"),
-        .package(url: "https://github.com/migueldeicaza/SwiftTerm", from: "1.0.0"),
+        .package(url: "https://github.com/migueldeicaza/SwiftTerm", revision: "dae32cc8f9bcda15713f4091bb2ea7e11f6dd57c"),
         .package(url: "https://github.com/raspu/Highlightr", from: "2.2.1"),
         .package(url: "https://github.com/siteline/swiftui-introspect", from: "1.0.0"),
         .package(url: "https://github.com/gonzalezreal/swift-markdown-ui", from: "2.4.0"),
--- a/Package.swift
+++ b/Package.swift
@@ -10,6 +10,10 @@ let package = Package(
         .package(url: "https://github.com/groue/GRDB.swift", from: "7.0.0"),
         .package(url: "https://github.com/apple/swift-argument-parser", from: "1.3.0"),
         .package(url: "https://github.com/apple/swift-nio", from: "2.65.0"),
+        // Pinned to a main-branch revision because the `fontSmoothing` public
+        // property (PR #531) hasn't shipped in a tagged release yet. Switch back
+        // to `from: "1.14.0"` (or whichever) once SwiftTerm cuts a release that
+        // includes commit dae32cc.
         .package(url: "https://github.com/migueldeicaza/SwiftTerm", revision: "dae32cc8f9bcda15713f4091bb2ea7e11f6dd57c"),
         .package(url: "https://github.com/raspu/Highlightr", from: "2.2.1"),
         .package(url: "https://github.com/siteline/swiftui-introspect", from: "1.0.0"),
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -22,6 +22,14 @@ enum ReviveState: Equatable {
 
 @MainActor
 final class AppState: ObservableObject {
+    /// Reference to the global appearance settings, wired by `TBDAppMain`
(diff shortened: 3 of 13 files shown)


cheapsteak:
Addressed all three:
- SwiftTerm pin rationale documented in Package.swift (c5d0ab9)
- TBDTerminalView gets appearance.font at construction (7f44427)
- detectTmuxStyleOverrides now runs off the main thread (4528746)

cheapsteak:
Addressed the displayFontName note as well in `8c00776` — now uses `appearance.font.displayName` so the Settings label tracks the actual resolved font, including the fallback path.

cheapsteak:
Addressed both new minor comments:
- Stale 'Task 5' planning references rewritten to describe behavior (`76cce30`)
- 'Debounce' wording corrected — `receive(on:)` is async dispatch to read post-commit property values, not deduplication (`a31d63c`)

cheapsteak:
Addressed both new minor comments:
- FontPickerCoordinator clears `completion` after firing (`07afb92`)
- Unknown `schemeID` is normalized to the default at init so the Picker always has a matching selection (`2a0d3d6`), with a new test pinning the behavior

cheapsteak:
Addressed both new minor comments:
- `applyFontSmoothing` inversion now has an inline explanation tying `fontSmoothing = false` to iTerm-style thin strokes (`d620860`)
- `scheme(forID:)` fallback and `AppearanceSettings.Defaults.schemeID` now share `ColorSchemes.defaultScheme` as their single source of truth — they can't drift again (`71da201`)

cheapsteak:
Addressed: `nsColor(from:)` now uses `NSColor(srgbRed:)` so bundled sRGB-spec palettes (Solarized, Dracula, Nord, etc.) render correctly on Display P3 displays (`8d5b01a`).

cheapsteak:
Addressed two of the three new minors:
- `hasTmuxStyleOverrides` is now `private(set)` (`099e005`)
- `tango` moved to the head of `bundled` so the default tops the Picker (`56b5e08`)

Skipped the `FontPickerCoordinator` in-flight-completion note — the bot itself framed it as awareness-only and the existing `completion = nil` after firing already guards the realistic case (NSFontPanel is modal-style during interaction).

cheapsteak:
Reviewed both new minors but going to **push back on both** this round:

**`applyAll()` full repaint on every settings change** — the simplicity is doing real work here. Every settings change re-applies *everything*; no missed updates, no synchronization between which subset to apply for which property. Splitting into per-property paths trades a few μs of repaint cost on user-driven (rare) actions for an enum + a switch + multiple Combine subscriptions instead of one. The bot itself flagged it as "polish opportunity, not a blocker" — I think the polish is net-negative against the existing design. Happy to revisit if real perf becomes a concern.

**`@retroactive @unchecked Sendable` module-wide** — the existing one-line comment ("SwiftTerm.Color is a class but our usage is read-only post-construction") already states the invariant. A scoped wrapper struct would add a layer of indirection for the abstract concern of "some future caller might mutate Color" — a future caller mutating an externally-provided class instance would have the same thread-safety question whether they go through our wrapper or hit `SwiftTerm.Color` directly. The conformance is the right escape hatch; the comment carries the rationale.

If either of these turns out to bite us, the change to revisit is small.

cheapsteak:
Addressed all three from this round:
- **Medium:** `detectTmuxStyleOverrides` now skips lines whose flag group contains `u` — `set -gu window-style` (the fix our tooltip recommends!) no longer false-positives (`bbe5ea8`)
- Renamed `lookupDefault` → `lookupTBDDefaultByID` to reflect what it actually tests (`37e7595`)
- `rgb()` helper now takes `UInt8` so the 0–255 channel constraint is enforced by the compiler instead of being a latent overflow trap (`3e2dae8`)

cheapsteak:
Fixed the regression my last commit introduced (`e2de630`): the guard was treating an absent flag group (`set window-style ...` with no flags at all) as 'skip', silently dropping legitimate overrides. Split the guard so only flag groups *explicitly containing* `u` are skipped; absent group → override.

Good catch, that was a real bug I should have caught with a test. The function isn't currently unit-tested because it reads hardcoded $HOME paths — would need a refactor to inject paths. Not doing that in this round.

cheapsteak:
Acknowledging this finding but **deferring to a follow-up PR**: the regex's `*` flag group only retaining the last iteration is exactly the class of bug the planned `detectTmuxStyleOverrides` refactor will catch with proper test coverage (split into a pure `tmuxConfigDeclaresStyleOverride(in: String) -> Bool` + thin I/O wrapper, extracted to its own file with a corpus of `tmux.conf` snippets as test fixtures).

Three regex-edge-case bugs across the last four review rounds (`set -gu` false positive, absent-flag-group regression, and now `set -u -g` unusual ordering) is the function telling us it wants real test coverage rather than another one-line patch. Will open the refactor PR same-day after this lands.

Merging when CI re-greens.

PR #169, opened 2026-05-17
Refactor: Make tmux style detection testable, lock in regex regressions

cheapsteak:
## Summary
- Extracts the tmux config-style detection from `AppearanceSettings` (private static method, hardcoded `\$HOME` paths, untestable) into a new `TmuxConfigStyleDetector` namespace at `Sources/TBDApp/Terminal/TmuxConfigStyleDetector.swift`.
- Splits the logic into a pure `declaresStyleOverride(in: String) -> Bool` (now unit-testable) plus a thin `detectFromUserConfig()` I/O wrapper that walks candidate paths.
- Adds a 15-case test corpus that locks in **all three regex regressions** caught during #163 review:
  - `set -gu window-style` → false (the unset form, which is exactly what our own tooltip tells users to run)
  - bare `set window-style ...` (no flags at all) → true
  - `set -u -g window-style` (unusual flag ordering, `-u` not last) → false (still an unset directive)
- Folds in `\$XDG_CONFIG_HOME/tmux/tmux.conf` lookup that was flagged but deferred during #163.

## Background

During #163 review, the original `detectTmuxStyleOverrides` had three subtle regex bugs across four review rounds. Each was a one-line patch on the same function. The pattern was clear: this function wanted real test coverage, not another one-line patch. The design was scoped during #163 (deferred to this PR) — see the design notes in that thread for the trade-offs considered (content vs path injection, file placement, scope creep guards).

## Non-goals

- **Not** parsing `source-file`, `if-shell`, `%if`/`%endif` directives. Best-effort regex only.
- **Not** tracking last-write-wins semantics. We detect *presence* of override directives, not effective state. An override followed by an unset later in the file still reports \`true\` — documented in the multi-line test case as an accepted false positive for an unusual case.
- **Not** handling line continuations.
- **Not** introducing a \`FileReader\` protocol or other DI scaffolding beyond the content/path split.

## Test plan
- [ ] \`swift test --filter TmuxConfigStyleDetectorTests\` — 15/15 pass
- [ ] \`swift test --filter AppearanceSettingsTests\` — 8/8 still pass (caller updated, behavior unchanged)
- [ ] \`swift test --filter ColorSchemesTests\` — 7/7 still pass (untouched)
- [ ] \`scripts/restart.sh\`, open Settings → Terminal: info ⓘ icon still appears on the Scheme picker (your \`~/.tmux.conf\` has \`window-style\` set, so detection should fire)

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=4E8CF284-2A26-4BC6-852D-DAAF7E71D107)

### DIFF
--- a/Sources/TBDApp/Terminal/AppearanceSettings.swift
+++ b/Sources/TBDApp/Terminal/AppearanceSettings.swift
@@ -41,7 +41,7 @@ final class AppearanceSettings: ObservableObject {
 
     /// True if the user's tmux config sets a cell-painting style option that
     /// would override the chosen color scheme's foreground/background. Best-
-    /// effort detection at init — see `detectTmuxStyleOverrides()`.
+    /// effort detection at init — see `TmuxConfigStyleDetector`.
     @Published private(set) var hasTmuxStyleOverrides: Bool
 
     init(defaults: UserDefaults = .standard) {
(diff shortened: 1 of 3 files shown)


cheapsteak:
Addressed both minors in `49486a0`:
- Hoisted both regex compilations to file-scope `static let`s (avoid recompiling the unset-flag pattern on every outer match)
- Empty `$XDG_CONFIG_HOME` now falls back to `~/.config` per the XDG Base Directory spec, not to the literal `/tmux/tmux.conf`

PR #173, opened 2026-05-18
DX: Per-row transcript signposts to identify which row hangs the main thread

cheapsteak:
## Summary

The 2026-05-17 transcript-pane hang (`freeze.log` at repo root, 19.21s, stack recurses through `LazyVStack.place` → `_FlexFrameLayout.sizeThatFits`) told us *that* a row hung but not *which* row. This PR adds zero-cost `OSSignposter` intervals around each `TranscriptRow.body` evaluation so the next freeze captures actionable evidence.

- **`transcript.row.body`** — one interval per `TranscriptRow.body`, tagged with `id=`, `kind=` (e.g. `tool:Bash`, `assistantText`, `subagentSummary`), and `len=` (payload size in chars). Look for the longest interval near the hang marker.
- **`transcript.markdown.build`** — wraps the `MarkdownSegments.split` + MarkdownUI view-tree assembly inside `ChatBubbleView.bubbleBody`, so "row slow because markdown" is distinguishable from "row slow because outer layout".
- **`hang.detected` event** — `HangWatchdog` now emits a one-shot signpost event on the same `perf-transcript` lane when it fires, so the hang marker visually aligns with the row intervals in Instruments.

Diagnostic-only change: signposts are zero-cost when no tracer is attached, so there is no behavior change and no perf change in non-traced runs. The existing `HangWatchdog` structured log line remains the primary text-based diagnostic.

Capture recipe (Instruments → SwiftUI template → subsystem `com.tbd.app` / category `perf-transcript`) added to [`docs/diagnostics-strategy.md`](./docs/diagnostics-strategy.md) under "Capturing a transcript-perf trace".

Refs #129.

## Test plan

- [x] `swift build` clean.
- [x] `scripts/restart.sh` (worktree copy) — app launches, transcript pane renders without crashing.
- [x] `Tests/TBDAppTests/TranscriptSignpostsTests.swift` covers `kindLabel` / `contentLength` per render-node kind plus a smoke test that the new `beginInterval` / `endInterval` / `emitEvent` calls don't crash. (CI will run them — local `swift test` can't load the `Testing` module without Xcode in this sandbox.)
- [ ] Next time the transcript pane hangs, capture an Instruments trace per the docs section and confirm the longest `transcript.row.body` interval near the `hang.detected` event identifies the offending row.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=30E6C3DB-4375-4A49-8656-615F6C6D50E0)

### DIFF
--- a/Sources/TBDApp/Diagnostics/HangWatchdog.swift
+++ b/Sources/TBDApp/Diagnostics/HangWatchdog.swift
@@ -227,6 +227,17 @@ final class HangWatchdog: @unchecked Sendable {
             let stallMs = stallNs / 1_000_000
             let snap = snapshotLock.withLock { $0 }
 
+            // Drop a one-shot event signpost on the shared transcript-perf
+            // timeline so a captured Instruments trace shows the hang marker
+            // aligned with the surrounding `transcript.row.body` /
+            // `transcript.markdown.build` intervals. The structured log line
+            // below remains the primary diagnostic; this is purely a marker
+            // for visual correlation in the os_signpost lane.
+            TranscriptSignposts.signposter.emitEvent(
+                "hang.detected",
+                "stallMs=\(stallMs, privacy: .public) terminalID=\(snap.focusedTerminalIDShort ?? "-", privacy: .public) itemCount=\(snap.transcriptItemCount ?? -1, privacy: .public) pane=\(snap.paneLabel ?? "-", privacy: .public)"
+            )
+
             // Sample the main thread stack and write to disk.
             let frames = MainThreadSampler.sample()
             if let fileURL = HangStackWriter.shared.recordHangStart(stallMs: stallMs, snapshot: snap, frames: frames) {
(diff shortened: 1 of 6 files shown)


cheapsteak:
Addressed the review feedback:

- **Medium (exhaustive `kindLabel` inner switch)** — fixed in 0465bdf. Replaced the `default:` arm in `case .chatBubble(let item):` with explicit cases for `.thinking`, `.systemReminder` (`chatSystemReminder`), `.slashCommand`, and `.toolCall` (`chatTool:<name>`). You were right that the old `"chat"` fallback erased precision for exactly the rows most likely to be large/slow — and now adding a new `TranscriptItem` case will break the build instead of silently collapsing.
- **Minor 1 (test coverage)** — same commit, 0465bdf. Added one `@Test` per new label following the existing node-builder style.
- **Minor 2 (docs clarification)** — a208ecf. Added a sentence in `docs/diagnostics-strategy.md` noting `defer { signposter.endInterval(...) }` as an equivalent alternative to `withIntervalSignpost { ... }` for non-throwing computed properties like SwiftUI `body`.

`swift build` clean locally. CI will validate `swift test` + SwiftLint.

PR #174, opened 2026-05-18
Reliability: Stop concurrent worktree creates from stalling 2-4min

cheapsteak:
## Summary

Orchestrator agents that fan out multiple `tbd worktree create` calls against the same repo were taking 2–4 minutes per call wall-clock — long enough to exceed Claude Code's 120s Bash timeout. The agent then loses direct stdout and has to poll output files to verify the create succeeded.

**Root cause:** the daemon's Phase-2 worktree work (`git fetch` + `git worktree add`) ran as `Task.detached` with no coordination, so concurrent same-repo creates contended on `.git/index.lock` and serialized at the git level anyway — just messily, and without bound.

**What this PR does:**
- Adds a `RepoSerializer` actor that chains Phase-2 work per `repoID`. Different repos still run in parallel; same-repo creates queue cleanly.
- Phase 1 (DB row insert) still returns immediately, so the CLI sees no added latency on the RPC round-trip.
- Bumps the CLI's `waitForActive` deadline 60s → 180s and adds a 2s stderr heartbeat (`waiting for worktree to become active… 12s`) so a watching agent (or human) sees liveness instead of silent polling.

## Test plan

- [x] `swift build` clean
- [x] New `RepoSerializerTests` (3 tests) pass: same-repo serial ordering, cross-repo overlap, non-blocking submit
- [x] `RPCRouterWorktreeTests | WorktreeLifecycleTests | WorktreeAdoptTests` — 31/31 pass, no regressions
- [ ] Manual verification: spawn 3+ worktrees back-to-back in same repo from an orchestrator agent; confirm each `tbd worktree create` completes inside the 120s Bash window and stderr heartbeat is visible

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=27AB895B-80EE-47AE-AF32-772C3E531C94)

### DIFF
--- a/Sources/TBDCLI/Commands/WorktreeCommands.swift
+++ b/Sources/TBDCLI/Commands/WorktreeCommands.swift
@@ -139,7 +139,9 @@ struct WorktreeCreate: AsyncParsableCommand {
     }
 
     private func waitForActive(pending: Worktree, client: SocketClient) throws -> Worktree {
-        let deadline = Date().addingTimeInterval(60)
+        let start = Date()
+        let deadline = start.addingTimeInterval(180)
+        var nextHeartbeat: TimeInterval = 2
         while Date() < deadline {
             let worktrees: [Worktree] = try client.call(
                 method: RPCMethod.worktreeList,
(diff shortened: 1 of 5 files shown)


cheapsteak:
Addressed both minor notes in 87f12ae:
- `lanes` is now pruned via a follow-up cleanup task that nils the entry only if the current tail still matches (so a later submit's entry isn't clobbered).
- `waitForRepo` dropped to internal to match `trackedRepoCount`.
- Added a 4th test covering lane pruning after work completes.

PR #193, opened 2026-05-24
Fix: Stop 35s transcript hangs (71-item freeze) by moving long content into a click-to-open overlay

### LINKED ISSUE #129: Tracking: recurring main-thread UI hangs in the transcript pane

cheapsteak:
## Summary

- Recurring main-thread hangs in the live transcript pane traced to `BashCard` / `WriteCard` wrapping content in a nested `ScrollView` bounded by `.frame(maxHeight:)`. The 2026-05-22 capture hung a **71-item** transcript for 35+ seconds (force-quit) — not a "long list" problem. Root cause: `.frame(maxHeight:)` produces a `_FlexFrameLayout` (a range), so SwiftUI must measure the child's ideal size before clamping. The child is a `ScrollView` whose ideal size is its full content — every layout invalidation triggers a full `ScrollViewLayoutComputer.Engine` pass inside each row. The 600 pt cap from PR #137 clamped the result but not the measurement.
- Rather than patching the cap further, this PR removes the structural trap entirely: all 11 tool row cards become **header-only** (one line, no inline body). Click any row → overlay opens over the row's bound terminal pane (looked up via `terminalID` in the layout tree), centered-modal fallback for History view / closed terminal. Dismiss via Esc, click-outside, or click-same-row. Swap on different-row click. AgentCard has a one-step back-stack for nested subagent transcripts.
- Adds a SwiftLint custom rule `no_scrollview_in_transcript_cards` + per-directory `Sources/TBDApp/Panes/Transcript/CLAUDE.md` so a future change can't reintroduce the trap.
- Documents in `Sources/TBDApp/Terminal/CLAUDE.md` that `TerminalPanelView` installs app-wide `NSEvent` monitors for scroll/click that run *before* SwiftUI's responder chain — discovered during this PR's manual testing when overlay scrolls leaked to the AppKit terminal underneath. The fix threads a \`shouldSuppressEvents\` closure into the monitors so any view layered over a terminal can opt out of having its events stolen.

Closes #129.

## Test plan

- [x] \`swift build\` clean, \`swift test\` 1042/1042 passing, SwiftLint \`--strict\` clean (verified locally)
- [ ] Open a worktree — transcript pane shows header-only rows with no chevrons; hover reveals a subtle ⌖ glyph trailing
- [ ] Click each card type (Bash, Write, Read, Edit, Skill, Grep, Glob, GenericTool, Agent, Thinking, SystemReminder) → overlay opens over the bound terminal with correct body content
- [ ] Esc / click-outside / click-same-row all dismiss; click-different-row swaps content while staying open
- [ ] Scroll inside an overlay does NOT scroll the terminal underneath
- [ ] AgentCard: click agent row, click a nested row inside the overlay, verify ← back returns to the subagent view
- [ ] History pane (closed session): click rows → centered modal opens
- [ ] Resize window with overlay open — overlay tracks the terminal pane's bounds
- [ ] No \`MainThreadSampler\` hang on a ≥100-item transcript during scroll or overlay open/close

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=F93102F0-9CAD-4D85-A241-4686248176AD)


cheapsteak:
### Code review

Found 1 issue:

1. The window-root fallback overlay in `ContentView` can never resolve History-pane transcript items, so every History row click shows "Item not found." `TranscriptOverlayView` looks up items by reading `@Environment(\.historyTranscriptItems)`, but that key is only set inside `SessionTranscriptView`'s `ScrollView` subtree, while the fallback overlay is rendered as `.overlay { ... }` on the detail container — a sibling of the History view, not a descendant. SwiftUI environment values flow down the view tree, so the overlay reads the `[]` default and `deepFind` returns `nil`. This breaks the spec's stated "History pane (closed session): click rows → centered modal opens" path.

   - Fallback overlay site (no `\.historyTranscriptItems` ancestor): https://github.com/cheapsteak/tbd/blob/cbd97e4359640325d0e00e335978599b0c6a5a94/Sources/TBDApp/ContentView.swift#L61-L74
   - Environment only injected deep inside `SessionTranscriptView`: https://github.com/cheapsteak/tbd/blob/cbd97e4359640325d0e00e335978599b0c6a5a94/Sources/TBDApp/Panes/HistoryPaneView.swift#L360-L367
   - Lookup that falls back to the (empty) env value when `terminalID == nil`: https://github.com/cheapsteak/tbd/blob/cbd97e4359640325d0e00e335978599b0c6a5a94/Sources/TBDApp/Panes/Transcript/TranscriptOverlayView.swift#L189-L203

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

cheapsteak:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

cheapsteak:
Addressed [@claude's review](https://github.com/cheapsteak/tbd/pull/193#issuecomment-4527033657):

**Medium — SwiftLint rule on `ChatBubbleView.swift`: dismissed as false positive.**

The rule's regex is `\bScrollView\s*\(` — it requires a literal `(` after `ScrollView`. `ChatBubbleView.swift` lines 285 and 375 use the trailing-closure form `ScrollView {` (no parens), so the regex does not match. Verified locally:

```
$ grep -n "ScrollView" Sources/TBDApp/Panes/Transcript/ChatBubbleView.swift
285:        ScrollView {
375:        ScrollView {

$ grep -n -P '\bScrollView\s*\(' Sources/TBDApp/Panes/Transcript/ChatBubbleView.swift
# (no matches)

$ swift package plugin swiftlint --strict
Done linting! Found 0 violations, 0 serious in 193 files.
```

This is by design — the rule deliberately targets only the `ScrollView(.vertical)` parenthesized form that produced the issue #129 freeze. The review's diagnostic notes that it could not run SwiftLint on a Linux runner, so this looks like a regex-by-inspection misread.

**Minor 1 — 4× `lookupItem()` per render: fixed in [`a2519cc`](https://github.com/cheapsteak/tbd/commit/a2519cc).** `var body` now computes the resolved item once and threads it through `header(item:)`, `headerIcon(item:)`, `headerLabel(item:)`, `bodyContent(item:)` — one lookup per render instead of four.

**Minor 2 — Redundant Esc handling: fixed in [`e1d10cb`](https://github.com/cheapsteak/tbd/commit/e1d10cb).** Dropped `.onKeyPress(.escape)` on the overlay body since the close button's `.keyboardShortcut(.cancelAction)` already binds Esc globally.

`swift build`, `swift test` (15/15 on `TranscriptOverlayCoordinatorTests`), and SwiftLint `--strict` all clean.

cheapsteak:
Addressed [@claude's 02:14 review](https://github.com/cheapsteak/tbd/pull/193#issuecomment-4527033657):

**Medium — SwiftLint coverage gap: fixed in [`26caeb3`](https://github.com/cheapsteak/tbd/commit/26caeb3).** You're right — the trailing-closure `ScrollView { ... }` form produces the same `_FlexFrameLayout` trap. Widened the regex to `\bScrollView\b` and silenced the two known `PreviewProvider` lines in `ChatBubbleView.swift` with per-line `// swiftlint:disable:next` comments. Also added `match_kinds: [identifier]` so the regex doesn't false-positive on the word "ScrollView" appearing in documentation comments (mirrors the existing `no_print_in_sources` pattern). Rule still excludes `TranscriptOverlayView.swift` as before.

Verified locally: `swift package plugin swiftlint --strict` → 0 violations across 193 files. The widened rule now correctly fires on any production `ScrollView` in `Sources/TBDApp/Panes/Transcript/` except `TranscriptOverlayView.swift`, regardless of init form.

**Minor — `AgentCardBody` `terminalID: nil` intentionality: documented in [`8c3494b`](https://github.com/cheapsteak/tbd/commit/8c3494b).** Added a comment explaining that subagent items are model-resident (embedded in the parent toolCall) and not indexed in the live transcript store, so the "Show full output" buttons must be disabled there — passing the parent's `terminalID` would surface the buttons but the daemon fetch path can't resolve those IDs.

**Minor — `lookupItem()` deepFind per render: dismissed as future optimization.** Your note explicitly tagged it as "not a correctness issue and orders of magnitude better than the original hang — mentioning as a future optimization target." Leaving for a follow-up when transcript size justifies it.

CI fresh on `8c3494b`.

cheapsteak:
Addressed [@claude's latest review](https://github.com/cheapsteak/tbd/pull/193#issuecomment-4527033657):

**Medium — Preview file button silently missing in overlay: fixed in [`edfdf27`](https://github.com/cheapsteak/tbd/commit/edfdf27).** You're right — this was a real regression. The `openFilePreview` env was injected only at the `.liveTranscript` case in `PanePlaceholder` (where the bodies used to render inline). After the rework, the bodies live in the overlay attached to the `.terminal` case, which wasn't getting the injection. Added the same closure at the overlay site, splitting off the terminal pane's `content.paneID` (so the new code-viewer pane appears next to the terminal — same place the existing transcript-pane closure would put it).

Skipped injecting in the `ContentView` fallback overlay as you suggested — that path covers History pane / no-terminal scenarios where there's no `layout` binding to mutate.

**Minor — overlay persists across worktree switch: fixed in [`5ed53af`](https://github.com/cheapsteak/tbd/commit/5ed53af).** Added `overlayCoordinator.close()` to the existing `.onChange(of: appState.selectedWorktreeIDs)` handler at the top of the ContentView VStack.

**Minor — overlay bleeds to ContentView fallback after terminal pane close: fixed in [`00564b0`](https://github.com/cheapsteak/tbd/commit/00564b0).** Added a conditional `overlayCoordinator.close()` inside the existing `.onDisappear` on `TerminalPanelView`, so closing the bound terminal pane clears the overlay rather than drifting it to centered-modal.

`swift build` clean, SwiftLint `--strict` clean across all three commits.

PR #199, opened 2026-05-25
Feat: Markdown + clickable file paths in Agent transcript overlay

cheapsteak:
## Summary

- The Agent tool overlay now renders the **prompt** and **result** as markdown (via MarkdownUI, reusing the existing `chatBubble` theme) instead of monospaced plain text — headings, lists, bold, inline code all light up.
- Bare absolute POSIX paths in either panel (e.g. `output_file: /private/tmp/claude-501/…`) are detected, filtered against `FileManager.fileExists`, and rewritten as `tbd-file:` markdown links. Clicking one opens that file **inside the overlay** — `.md` rendered as markdown, other text as monospaced, large/binary files get a "Reveal in Finder" placeholder.
- Replaces the coordinator's single-step parent-frame back-pointer with a real `[OverlayFrame]` stack (`.item` or `.file` cases) so `agent → subagent-item → file → another-file` all back-navigate one step at a time.
- New: `LocalFileLinker` (pure helper, 15 tests covering punctuation, code fences, inline code, percent-encoding, nonexistent-path filtering), `OverlayFileView` (file frame body), and an extended `TranscriptOverlayCoordinator` (18 tests covering mixed item/file stack sequences).
- Scope is just the Agent overlay — Read/Edit/Bash specialised bodies are unchanged.

## Test plan

- [x] `swift test` — all 1092 tests pass (15 new `LocalFileLinkerTests`, 18 rewritten `TranscriptOverlayCoordinatorTests`).
- [x] `swiftlint --strict` — clean (added `OverlayFileView.swift` to the `no_scrollview_in_transcript_cards` exclusion list; it lives outside the LazyVStack so the `ScrollView` is safe).
- [x] Manual: opened the Agent overlay on a real subagent invocation, confirmed markdown formatting renders in both panels, clicked a bare path in the result, confirmed the modal swapped to the file's markdown rendering with a back chevron, popped back to the agent.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=C5DC73B9-C863-4D3B-AA63-F422DE27408A)

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -60,6 +60,8 @@ custom_rules:
     excluded:
       # TranscriptOverlayView renders outside the LazyVStack — ScrollView is safe here (#129).
       - "Sources/TBDApp/Panes/Transcript/TranscriptOverlayView\\.swift"
+      # OverlayFileView is the body of a file frame in the overlay — sibling of LazyVStack, real ScrollView is safe.
+      - "Sources/TBDApp/Panes/Transcript/OverlayFileView\\.swift"
     regex: '\bScrollView\b'
     match_kinds:
       - identifier
--- a/Sources/TBDApp/ContentView.swift
+++ b/Sources/TBDApp/ContentView.swift
@@ -60,12 +60,12 @@ struct ContentView: View {
                     })
                     .onPreferenceChange(ContentHeightKey.self) { contentAreaHeight = $0 }
                     .overlay {
-                        if let frame = overlayCoordinator.openOverlay,
-                           frame.terminalID.map({ !visibleTerminalIDs.contains($0) }) ?? true {
+                        if let frame = overlayCoordinator.current,
+                           overlayFrameIsWindowRoot(frame, visibleTerminalIDs: visibleTerminalIDs) {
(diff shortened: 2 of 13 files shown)


cheapsteak:
Thanks for the review. Addressed minors #2, #3, #4 in 2897186 and fc7f940 — dropped the no-op `removingPercentEncoding`, added `=` as a path predecessor with a test for the `output_file=/tmp/foo` shape, and added a NOTE that fence-protection only covers 3-backtick fences. Skipping #1 (sync `fileExists` in view body) per your own note that it isn't currently a problem; will track as a follow-up if it becomes one.

cheapsteak:
Fixed in af505e8 + 50dbee1: file frames now suppress terminal key events via the extracted `shouldSuppressEvents(in:forTerminalID:)` helper (5 tests covering all branches including the history-pane case), and dropped the redundant `MainActor.run` hops in `OverlayFileView.load()`.

cheapsteak:
Picked up both minors as long as I'm here: extracted `overlayFileLinkAction` into a shared helper (912aa69) and added the trim-back regression test (b9c9231).

PR #203, opened 2026-05-26
Split Codex stop hooks to preserve stdin

Jephuff:
## Summary
- split the Codex `Stop` hook into separate `tbd notify` and `tbd hooks stop-rename-check` entries
- preserve stdin for `stop-rename-check` so Codex can still read the Stop payload and block for worktree rename prompts
- tighten `CodexHookOverlayTests` so the hook shape must stay split and cannot regress to a combined command

## Test Plan
- `swift test`


cheapsteak (approved):
🙏 🙏

PR #204, opened 2026-05-26
fix: make tab padding clickable

Jephuff:
## Summary
- move the tab body's leading padding and minimum height onto the select button so the padded tab area is clickable
- keep the simpler `HStack` tab layout instead of overlaying the close control with a `ZStack`
- add a regression test that asserts the tab hit target extends beyond the label-only width

## Test Plan
- [x] `swift test`
- [x] `swift test --filter TabBarHitAreaTests`
- [x] `swift test --filter AddTabMenuTests`
- [x] `swift test --filter TabReorderTests`

### DIFF
--- a/Sources/TBDApp/TabBar.swift
+++ b/Sources/TBDApp/TabBar.swift
@@ -435,6 +435,9 @@ private struct TabBarItem: View {
                             .foregroundStyle(isSuspended ? .tertiary : (isSelected ? .primary : .secondary))
                     }
                 }
+                .padding(.leading, 8)
+                .frame(minHeight: 28, alignment: .leading)
+                .contentShape(Rectangle())
             }
             .buttonStyle(.plain)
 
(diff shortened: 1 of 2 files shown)


Jephuff:
Fixed the `TabBarHitAreaTests` suite-teardown issue.

Changes:
- extracted the per-test `UserDefaults` suite name to a local variable
- wrote a probe value so the postcondition is meaningful
- added `defer { defaults.removePersistentDomain(forName: suiteName) }`
- asserted the suite's persistent domain is gone after the scoped test work finishes

Fresh verification:
- `swift test --filter TabBarHitAreaTests`
- `swift test --filter AddTabMenuTests`
- `swift test --filter TabReorderTests`
- `swift test`

Jephuff:
Addressed the two follow-up review notes in `038bf8f`.

Changes:
- replaced the tight absolute hit-width threshold with a comparison against a label-only baseline, so the regression stays tied to the actual bug instead of a fragile magic number
- added an explicit comment documenting the `KeyViewProxy` string match as a deliberate private AppKit test seam

Fresh verification:
- `swift test --filter TabBarHitAreaTests`
- `swift test`

cheapsteak (commented):
tyty 🙏

PR #206, opened 2026-05-26
Fix: Stop worktree cleanup from bricking the `tbd` CLI on PATH

cheapsteak:
## Summary

`~/.local/bin/tbd` was a symlink into a worktree's `.build/.../TBDCLI`. When that worktree got archived (`tbd worktree archive` → `git worktree remove`) or pruned (`tbd cleanup` → `git worktree prune`), the symlink dangled → user's `tbd` command 404'd on PATH until the next TBDApp launch silently re-installed it. Caught when a Claude Code skill couldn't run `tbd link` to append a worktree deep-link to a PR body.

**Primary fix**: switch `CLIInstaller` to install a **hard link** instead of a symlink. The install path shares the inode with `.build/.../TBDCLI`, so deleting the worktree only drops the file's link count by 1 — the inode stays alive and `tbd` keeps working. Same trick #195 applied to `/Applications/TBD.app/Contents/MacOS/TBDApp` for the same worktree-binding fragility.

`currentState` now compares inodes via `stat(2)` instead of reading a symlink destination. Legacy symlink installs are reported `.stale` so they self-heal into hard links via the existing launch-time refresh prompt.

**Safety nets** (defense in depth — won't fire under normal hardlink flow, but cover legacy installs and the gap before a user's app restart picks up the new install code):

- **`tbd doctor`** subcommand — connect to daemon, report install state, repair if dangling/stale. Works without TBDApp running.
- **Archive self-heal** in `WorktreeLifecycle+Archive`: after `git worktree remove`, re-point the install path if a legacy symlink now dangles into the removed worktree. Path-prefix check uses a trailing-slash boundary so `/path/wt-2` doesn't false-match while archiving `/path/wt`.
- **Cleanup self-heal** in `handleCleanup`: same check after `git worktree prune`.

**Renames** (post-hardlink the old names are misleading):
- `CLIInstallState.nonSymlink` → `.unexpectedFileType`
- `CLIInstallerError.symlinkCreationFailed` → `.linkCreationFailed`
- `CLIInstaller.symlinkPath` → `.installPath`

**Docs**: file-level doc comment at top of `CLIInstaller.swift` explains the hardlink rationale and references #195. Breadcrumb in `Sources/TBDShared/CLAUDE.md` points back to it.

## Test plan

- [x] `swift build` clean.
- [x] `swift test --filter CLIInstaller` — 31/31 pass, including new tests: hardlink-survives-source-deletion, inode-differs → stale, legacy-symlink → stale, repair-if-dangling helper.
- [x] `swiftlint --strict` — 0 violations.
- [x] Manual: ran `scripts/restart.sh` from this worktree; CLIInstallerCoordinator saw the legacy dangling symlink at `~/.local/bin/tbd`, presented the Refresh prompt, accepted → fresh hard link installed.
- [x] Verified `ls -la ~/.local/bin/tbd` is a regular file with link count 2, sharing inode `319476383` with this worktree's `.build/arm64-apple-macosx/debug/TBDCLI`.
- [x] `~/.local/bin/tbd --version` returns `0.1.0`.
- [ ] Reviewer: archive this worktree (or any worktree whose `.build/.../TBDCLI` your install path was linked to) and confirm `tbd` on PATH still works after `git worktree remove`.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=52A1A8F4-2EA1-4A1A-B024-736EA271F075)


cheapsteak:
Addressed both <details> minor items from the bot review:

- `1deee78` — fixed stale "symlink" wording in `CLIInstallerCoordinator.checkOnLaunch` docstring (s/symlink/CLI install/).
- `0acc8f7` — `tbd doctor` now exits early with code 5 on `.unexpectedFileType` instead of falling through to `install()` and surfacing a low-level error. Prints a clear "remove the directory and re-run `tbd doctor`" next step.

Both verified — `swift build` clean, `DoctorCommand` tests pass.

cheapsteak:
Addressed the "nearly dead code" observation from the bot review:

- `beddb1e` — removed the archive + cleanup self-heal hooks (~254 lines of impl + tests). With hardlinks as the default, the hook only fired for legacy symlink installs archived before the next TBDApp launch — and `CLIInstallerCoordinator`'s launch-time stale prompt already migrates those. `tbd doctor` remains as the long-term diagnostic + manual repair path.

The canonicalization nit is mooted along with the branch.

Full `swift test` now 1105 passing (was 1111, -6 self-heal tests). swiftlint clean.

cheapsteak:
Two more review follow-ups for the new top-level Minor items:

- `52d036c` — reverted `RPCRouter.resolvedExecutablePath` to `private static let`. It was promoted to internal so the self-heal in `WorktreeLifecycle+Archive` could read it; self-heal is gone in `beddb1e`, so the only caller is back to inside the same type.
- `ccd6801` — surface EXDEV (cross-filesystem hard-link failure) with an actionable typed error `CLIInstallerError.crossDeviceLink`. `install()` now calls `link(2)` directly via an injectable `linkFunc` so we can route on `errno` cleanly (EXDEV → typed case; everything else → `.linkCreationFailed` with `strerror` text). `tbd doctor`'s discussion text also warns about the same-volume requirement upfront. Two new tests cover EXDEV → typed case and EACCES → linkCreationFailed.

Test count 1105 → 1107. Build clean, swiftlint clean.

PR #207, opened 2026-05-26
fix: autofocus terminal after worktree switch

Jephuff:
## Summary
- restore terminal focus when switching to or restoring a single selected worktree
- track live terminal views so the active layout's first terminal can become first responder again
- add autofocus coverage for terminal tabs, split layouts, history mode, and non-terminal tabs

## Test Plan
- [x] swift test


cheapsteak (approved):
🙏 🙏

PR #212, opened 2026-05-26
fix: make PR status sync upstream-aware

Jephuff:
## Summary
- run targeted PR refresh from the worktree path so `gh pr view` can see worktree-local upstream config
- match batch PR status sync against either the local branch name or the tracked upstream head branch when they differ
- add focused tests for upstream branch resolution and branch-candidate matching

## Test Plan
- `swift test --filter GitManagerTests`
- `swift test --filter PRStatusManagerTests`

### DIFF
--- a/Sources/TBDDaemon/Git/GitManager.swift
+++ b/Sources/TBDDaemon/Git/GitManager.swift
@@ -67,6 +67,25 @@ public struct GitManager: Sendable {
         return trimmed.isEmpty ? nil : trimmed
     }
 
+    /// Returns the upstream head branch name configured for the current worktree branch.
+    public func upstreamBranchName(worktreePath: String, branch: String) async -> String? {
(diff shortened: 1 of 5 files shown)


Jephuff:
Addressed the targeted-refresh gap.

Changes:
- `PRStatusManager.refresh(...)` now tries the same local/upstream branch candidates as batch sync
- `handlePRRefresh` resolves `upstreamBranch` from the worktree and passes it into `refresh`

Verification:
- `swift test --filter GitManagerTests`
- `swift test --filter PRStatusManagerTests`
- reproduced the real mismatch case locally: `gh pr view fix-codex-profile-v2` fails in the worktree while `gh pr view tbd/20260525-cuddly-gerbil` succeeds; refresh now follows that candidate order

cheapsteak (approved):

PR #214, opened 2026-05-26
fix: clarify PR status semantics

Jephuff:
## Summary
- add richer PR status semantics for GitHub sync, including draft and failing-check states while preserving upstream-branch PR lookup
- centralize PR icon/color presentation in the app and align the sidebar/toolbar with the new semantics
- show a yellow conflict fallback marker in the sidebar only when a worktree has conflicts, no PR, and no notification badge

Note: this changes the existing conflict fallback from a yellow pr icon, to a yellow X to consistently differentiate between PRs and no PRs
image

## Test Plan
- [x] swift test --filter PRStatusManagerTests
- [x] swift test --filter PRStatusPresentationTests
- [x] swift test --filter WorktreeRowIndicatorTests
- [x] swift test

### DIFF
--- a/Sources/TBDApp/ContentView.swift
+++ b/Sources/TBDApp/ContentView.swift
@@ -309,33 +309,16 @@ private func overlayFrameIsWindowRoot(
 private struct PRButtonLabel: View {
     let prStatus: PRStatus
 
-    private var iconName: String {
-        switch prStatus.state {
-        case .open, .changesRequested, .mergeable: return "git-pull-request"
-        case .merged:                              return "git-merge"
-        case .closed:                              return "git-pull-request-closed"
-        }
-    }
-
-    private var iconColor: Color {
-        switch prStatus.state {
-        case .open:             return .secondary
-        case .changesRequested: return .red
-        case .mergeable:        return .green
-        case .merged:           return .purple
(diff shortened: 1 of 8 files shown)


cheapsteak (approved):
🙏 🙏

appreciate the screenshot

thoughts on changing the icon to something like hand.raised.slash or hand.raised.slash.fill? 

also the yellow looks a bit low contrast with the surrounding gray

Jephuff:
I can do thoughts on changing the icon to something like hand.raised.slash or hand.raised.slash.fill! my main goal was to differentiate. I think maybe I lost the original yellow which was better contrast

Jephuff:
Follow-up pushed in `459c762`.

- simplified `WorktreeRowView` so the PR presentation is computed once per row
- changed the conflict-only fallback to a red `hand.raised.slash.fill` while keeping notification precedence
- added coverage for `CLEAN + PENDING` and the unknown-future `mergeStateStatus` fallback paths
- `closed` remains red intentionally to keep the "known non-mergeable" semantics aligned with GitHub's closed-unmerged iconography

cheapsteak (approved):

Jephuff:
image

PR #216, opened 2026-05-26
Add webview find interface

Jephuff:
## Summary
- add a Chrome-style find bar wrapper around the webview pane and route text-finder actions through it
- wire keyboard shortcuts for find, next match, and previous match to the active webview pane
- cover the new behavior with focused TBDApp tests for command routing and incremental find behavior

## Test Plan
- swift test

### DIFF
--- a/Sources/TBDApp/Helpers/KeyboardShortcuts.swift
+++ b/Sources/TBDApp/Helpers/KeyboardShortcuts.swift
@@ -3,12 +3,15 @@ import AppKit
 
 enum TextFinderCommand {
     static let action = #selector(NSResponder.performTextFinderAction(_:))
-    static let tag = NSTextFinder.Action.showFindInterface.rawValue
+    
+    static func tag(for action: NSTextFinder.Action) -> Int {
+        action.rawValue
+    }
 
     @MainActor
-    static func perform() {
+    static func perform(_ finderAction: NSTextFinder.Action = .showFindInterface) {
         let sender = NSMenuItem()
-        sender.tag = tag
+        sender.tag = tag(for: finderAction)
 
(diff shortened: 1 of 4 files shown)


Jephuff:
Addressed the review points in `b89d64c`.

- added explicit `Find Next` / `Find Previous` commands on `Cmd-G` and `Cmd-Shift-G`, routed through `TextFinderCommand`
- narrowed `WebviewPaneHostView` so unsupported text-finder actions no longer open the find bar, and only supported actions validate as enabled
- added focused tests for the new action tags and the supported/unsupported finder behavior

Verification: `swift test`

cheapsteak (approved):
🙏 🙏 🙏

Jephuff:
Addressed the remaining review point in `ca5a600`.

- handle `cancelOperation:` via `WebviewFindBarView`'s text-editing delegate path so Escape closes the find bar while the field editor is active
- add a focused regression test that exercises the delegate command route directly

Verification:
- `swiftlint --strict`
- `swift test`

Jephuff:
Addressed the latest Claude feedback in `fa7c9e0`.

- removed the dead `WebviewFindSearchField` subclass and its unused `onSubmit` / `onEscape` wiring
- kept the live AppKit paths: search-field action for Return and `control(_:textView:doCommandBy:)` for Escape
- clipped the rounded visual effect with `layer?.masksToBounds = true`
- added a focused test that asserts the find bar now uses a plain `NSSearchField`

One nit from the review was not applied: `@retroactive` is rejected here because `WebviewFindClient` is declared in TBDApp, so this is not an external-protocol conformance case.

Verification:
- `swift test`
- `swiftlint --strict`

PR #217, opened 2026-05-26
Fix Cmd-W close tab behavior

Jephuff:
## Summary
- make plain `Cmd-W` close the active tab instead of falling through to window-close behavior
- scope tab closing to the focused terminal's owning tab so multi-worktree views do not close an arbitrary selected worktree tab
- remove the File menu override and keep `Close Tab` under the Terminal menu, enabled only when a focused closable tab exists

## Test Plan
- `swift test`


cheapsteak (approved):
🙏 🙏 🙏

PR #218, opened 2026-05-26
Fix Codex tab labels

Jephuff:
## Summary
- add a shared auto-label resolver so Codex terminals render as `Codex` instead of falling back to `Terminal N`
- keep terminal type labels derived at render time by removing the stored `terminal.label` override from Codex tab creation
- cover Codex, shell, and Claude profile label behavior with focused app tests

## Test Plan
- `swift test`


Jephuff:
image

cheapsteak (approved):

PR #222, opened 2026-05-26
Fix: Show merge-conflict marker on PRs that would fail to merge

cheapsteak:
## Summary
- Invert the sidebar's merge-conflict gate: the marker now appears when a PR exists AND the branch would conflict, instead of the opposite. Branches with conflicts but no PR no longer get the warning — there's no merge action pending.
- Detect conflicts against `origin/<defaultBranch>` instead of the local default branch. The daemon already fetches `origin/main` every 60s, so the comparison stays fresh independently of whether the user pulls. Previously a stale local `main` could mask conflicts that GitHub would actually reject.
- Restore the custom `git-merge-conflict` glyph (replacing `hand.raised.slash.fill` from #214), still rendered red, still using the same gating slot.

## Test plan
- [x] `swift test --filter WorktreeRowIndicatorTests` — gate semantics
- [x] `swift test --filter GitStatusTests` — reconcile against `origin/main` with a bare-origin fixture, plus a new no-conflict happy-path test
- [x] `swift test` full suite (one pre-existing unrelated failure in `TabBarHitAreaTests`)
- [ ] Visually: a worktree with an open PR whose branch would conflict shows the red merge-conflict glyph; the same conflict state on a branch with no PR shows nothing

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=70A5E099-B3E7-4D67-8640-D8F23AC2E476)

### DIFF
--- a/Sources/TBDApp/Sidebar/WorktreeRowView.swift
+++ b/Sources/TBDApp/Sidebar/WorktreeRowView.swift
@@ -10,7 +10,9 @@ enum WorktreeRowConflictFallback {
         hasNotification: Bool
     ) -> Bool {
         guard let prStatus, hasConflicts, !hasNotification else { return false }
-        return prStatus.state != .merged
+        // Closed and merged PRs have no pending merge action, so a conflict warning is noise.
+        // The PR-icon slot is mutually exclusive with the conflict glyph; suppress here to keep .closed/.merged visible.
+        return prStatus.state != .merged && prStatus.state != .closed
     }
 }
 
--- a/Sources/TBDApp/Sidebar/WorktreeRowView.swift
+++ b/Sources/TBDApp/Sidebar/WorktreeRowView.swift
@@ -2,14 +2,14 @@ import SwiftUI
 import TBDShared
 
 enum WorktreeRowConflictFallback {
-    static let iconName = "hand.raised.slash.fill"
+    static let iconName = "git-merge-conflict"
 
     static func shouldShow(
         prStatus: PRStatus?,
         hasConflicts: Bool,
         hasNotification: Bool
     ) -> Bool {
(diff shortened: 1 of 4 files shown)


cheapsteak:
Addressed the minor feedback in dd04b5c:

- Renamed `prStatusSuppressesConflictFallback` → `conflictFallbackAppearsWithPR` to match the inverted gate.
- Expanded the nil-return comment in `checkHasConflicts` to explicitly call out the local-only / pre-first-fetch case.

Skipping the test boilerplate extraction for now — that's a fair point but more of a refactor than a polish, happy to follow up if you'd like.

cheapsteak:
Cycle 2 updates in d2da935:

- **Duplicate icon fix.** The PR icon was rendering alongside the conflict glyph because the two `if` blocks in `rowIcons()` are independent — the gate inversion turned what used to be "either/or" into "both". Gated the PR-icon render with `!showsConflictFallback`.
- **Merged PR wins over conflict.** Squash-merge orphans the branch's commits from main's history, so `merge-tree origin/main <branch>` reports false-positive conflicts on shipped work. `.merged` now suppresses the fallback so the purple merged glyph shows. Added a regression test.
- **Extracted `shell` helper.** Addressed the top-level Minor from cycle 1 — the 10-line nested `shell` function is now a single file-scope `runShell(_:at:)` with three call sites.

CI'll re-run; restarted locally to confirm the duplicate is gone.

cheapsteak:
Addressed the `.closed` issue in 0042c68. `shouldShow` now exempts both `.merged` and `.closed` — neither has a pending merge action, so the conflict warning was just hiding the meaningful PR-state indicator. Added `closedPRSuppressesConflictFallback` test mirroring the merged case. 6/6 indicator tests pass.

PR #234, opened 2026-05-27
Fix: Profile swap gives blank session when negative project-dir cache is stale

cheapsteak:
## Summary

- **Bug:** Swapping a Claude terminal's model profile mid-session starts a blank session instead of resuming the conversation. The user sees a fresh prompt even though `/resume` shows the conversation is still available.
- **Root cause:** `ClaudeProjectDirectory.resolve()` caches negative lookups (project dir not found) permanently. When the TBD app polls the transcript pane before Claude writes its first JSONL line, a nil gets cached. Later, `isSessionBlank` hits the stale nil cache, concludes the session is blank, and spawns a fresh session instead of resuming.
- **Fix:** (1) Pass `transcriptFilePath` through to `isSessionBlank` so the swap handler bypasses the stale cache when the SessionStart hook has already recorded the real path. (2) Add a 30-second TTL to negative cache entries so they expire and get re-resolved, preventing the same class of bug in other code paths.

## Test plan
- [ ] Create a worktree, let Claude start, then swap to a different model profile — should resume the conversation, not start blank
- [ ] `swift test --filter ClaudeSessionScanner` — 17 tests covering blank detection, transcriptFilePath bypass, and cache TTL

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=545BE962-14F8-4768-850C-F27A6AFB0D22)

### DIFF
--- a/Sources/TBDDaemon/Claude/ClaudeSessionScanner.swift
+++ b/Sources/TBDDaemon/Claude/ClaudeSessionScanner.swift
@@ -10,13 +10,17 @@ private let logger = Logger(subsystem: "com.tbd.daemon", category: "session-scan
 /// Three-tier lookup: exact encoding → regex fallback → full content scan.
 /// Results are cached after first resolution.
 enum ClaudeProjectDirectory {
-    /// Cache entries: `.some(url)` for resolved hits, `.none` for confirmed
-    /// misses. Caching negatives is essential — the tier-3 fallback enumerates
-    /// every directory under `~/.claude/projects` and reads the first line of
-    /// every session JSONL. Without negative caching, every repeat call for a
-    /// worktree with no matching project dir re-runs that scan, which pegs the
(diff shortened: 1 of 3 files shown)


cheapsteak:
Addressed the Medium finding — the `transcriptPathBypassesCache` test was indeed testing the wrong scenario. `makeProjectDir` created the encoded directory before the first call, so `resolve()` cached a positive entry, not nil. The test only passed because the session _file_ was missing, not because of a stale negative cache entry.

Rewritten in 42f31d1:
- Uses an empty `projectsBase` where `resolve()` genuinely can't find the project dir (no encoded subdir exists), so `isSessionBlank` returns `true` via the nil resolve path
- Then shows `transcriptFilePath` bypasses `resolve()` entirely and finds the content
- Also fixed the redundant `createDirectory` in `positiveEntryRevalidation` and corrected misleading comments in `transcriptPathFallback`

PR #236, opened 2026-05-28
Perf: Stop fetching 348 archived worktrees to show a list

cheapsteak:
## Summary
- Archived worktree list fetched ALL rows on open — 348 for longeye-app, each enriched with a filesystem scan for session count. Now fetches the first 100 with a "Load More…" button for pagination.
- Adds `limit`/`offset` to the `worktree.list` RPC and sorts archived rows by `archivedAt DESC` in the DB query (matching the view's existing sort)
- Hardens against concurrency issues: guards double-tap on "Load More", prevents stale-offset appends after concurrent refresh, preserves loaded pages across revive/navigation refreshes, and fixes deep-link handler leaving pagination state stale

## Test plan
- [ ] Open archived view for a repo with >100 archived worktrees — should show first 100 with "Load More…" button
- [ ] Click "Load More…" — next 100 should append, button disappears when all loaded
- [ ] With `hideEmpty` filter on, count header shows "N of M+" when more pages exist
- [ ] Revive an archived worktree after loading 2+ pages — list should preserve loaded count, not reset to 100
- [ ] Deep-link to an archived worktree — "Load More…" should not appear (unbounded fetch)
- [ ] Rapid-click "Load More…" — should not duplicate rows (button disables during load)
- [ ] `tbd worktree list --status archived` CLI still returns all rows (no limit)

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=4086368E-CA0A-4DE3-9CB3-20A6A65ADEC6)

### DIFF
--- a/Sources/TBDApp/AppState+Worktrees.swift
+++ b/Sources/TBDApp/AppState+Worktrees.swift
@@ -198,6 +198,7 @@ extension AppState {
         selectedWorktreeIDs = []
         selectedRepoID = wt.repoID
         archivedWorktrees[wt.repoID] = archived.filter { $0.repoID == wt.repoID }
+        archivedWorktreesHasMore[wt.repoID] = false
         highlightedArchivedWorktreeID = id
         if NSApplication.shared.isRunning {
             NSApplication.shared.activate(ignoringOtherApps: true)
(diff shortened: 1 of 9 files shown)


cheapsteak:
Addressed the hasMore false-positive after refresh — added `knownExhausted` guard in e0496d0. A non-multiple-of-pageSize loaded count proves the last page was partial, so we preserve `hasMore = false` on refresh instead of re-triggering the `>=` heuristic.

PR #237, opened 2026-05-28
Fix Codex profile flag compatibility

Jephuff:
## Summary
- Detect the supported Codex profile flag from `codex --help`, preferring `--profile-v2` when available for pre-0.134 profile-v2 support.
- Fall back to `codex --version` with the 0.134.0 rename boundary, then default to current `--profile` behavior if probing fails.
- Update Codex spawn tests to cover old and new flag behavior without pinning higher-level tests to one installed CLI generation.

## Background
Codex CLI renamed the profile-v2 selector in the 0.134.0 release. The release notes include `cli: rename profile v2 flag to --profile` and list `--profile` as the primary selector for config profiles.

Relevant upstream links:
- Codex releases: https://github.com/openai/codex/releases
- Rename PR: https://github.com/openai/codex/pull/23883
- 0.133.0 source, before the rename: https://github.com/openai/codex/blob/rust-v0.133.0/codex-rs/utils/cli/src/shared_options.rs
- 0.134.0 source, after the rename: https://github.com/openai/codex/blob/rust-v0.134.0/codex-rs/utils/cli/src/shared_options.rs

## Compatibility Strategy
This uses capability detection first instead of pure version detection. Older Codex builds can expose both `--profile` and `--profile-v2`, but in those builds `--profile` refers to the legacy profile path while TBD needs the profile-v2 overlay. Checking for `--profile-v2` in `codex --help` is therefore the safer discriminator.

Fallback behavior:
- `codex --help` contains `--profile-v2` → use `--profile-v2 tbd`
- otherwise `codex --help` contains `--profile` → use `--profile tbd`
- if help probing fails, parse `codex --version`
- `>= 0.134.0` → use `--profile tbd`
- `< 0.134.0` → use `--profile-v2 tbd`
- if both probes fail, default to current `--profile` behavior

## Local Evidence
On the currently installed CLI:
- `codex --version` reports `codex-cli 0.134.0`
- `codex --profile tbd --help` exits 0
- `codex --profile-v2 tbd --help` exits 2 with `unexpected argument '--profile-v2'` and suggests `--profile`

## Test Plan
- `codex --profile tbd --help`
- `codex --profile-v2 tbd --help`
- `swift test --filter CodexSpawnCommandBuilderTests`
- `swift test --filter TBDWorktreeIDLeakTests`
- `swift test --filter CodexHookOverlayTests`
- `swift build`
- `swift test`

### DIFF
--- a/Sources/TBDDaemon/Codex/CodexHomeManager.swift
+++ b/Sources/TBDDaemon/Codex/CodexHomeManager.swift
@@ -253,8 +253,8 @@ enum CodexProfileWriter {
 
 enum CodexSpawnCommandBuilder {
     private static let detectedProfileFlag = profileFlag(
-        codexHelpOutput: codexCommandOutput(arguments: ["--help"]),
-        codexVersionOutput: codexCommandOutput(arguments: ["--version"])
+        codexHelpOutput: commandOutput(arguments: ["codex", "--help"], timeout: 3),
+        codexVersionOutput: commandOutput(arguments: ["codex", "--version"], timeout: 3)
     )
 
     static var command: String {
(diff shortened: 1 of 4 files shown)


Jephuff:
Addressed Claude review feedback in d1e500b.

Changes:
- Replaced the unbounded waitUntilExit probe path with commandOutput(..., timeout:).
- Codex --help / --version probes now use a 3-second timeout and terminate the subprocess on timeout.
- stdout and stderr are drained through readability handlers into locked storage, avoiding sequential pipe reads.
- Added tests for stdout+stderr capture and timeout behavior.

Verification run locally:
- swift test --filter CodexSpawnCommandBuilderTests
- swift build
- swift test

Jephuff:
Addressed the lazy fallback feedback in 52bb469.

Changes:
- Replaced eager argument evaluation with detectProfileFlag(commandOutput:), so runtime detection probes codex --help first.
- The codex --version probe now only runs when help output is missing or does not identify either profile flag.
- Added tests proving version probing is skipped when help is sufficient and used only when help is inconclusive.

Verification run locally:
- swift test --filter CodexSpawnCommandBuilderTests
- swift build
- swift test

cheapsteak (approved):

PR #242, opened 2026-06-01
feat: label worktree setup-script terminal tab as "Setup"

Jephuff:
## Summary
image

The worktree setup-script terminal now displays its tab label as **"Setup"** instead of the generic "Terminal N".

The setup terminal was already created with the internal label `"setup"` (in `WorktreeLifecycle+Create.swift`), but the UI's `AutoTabLabelResolver` never read that field for shell terminals, so it fell through to the generic "Terminal N" fallback. This teaches the resolver to recognize it.

## Changes

- `Sources/TBDApp/AutoTabLabelResolver.swift` — return `"Setup"` when `terminal.label == "setup"`, before the generic numbered fallback.
- `Tests/TBDAppTests/AutoTabLabelResolverTests.swift` — added `setupTerminalUsesSetupLabel`; the existing shell test (label `nil` → "Terminal 3") still passes.

## Testing

- `swift build` ✅
- `swift test --filter AutoTabLabelResolver` ✅ (all 4 tests pass)

### DIFF
--- a/Sources/TBDApp/AutoTabLabelResolver.swift
+++ b/Sources/TBDApp/AutoTabLabelResolver.swift
@@ -37,6 +37,10 @@ enum AutoTabLabelResolver {
             return "Codex"
         }
 
+        if terminal?.label == "setup" {
+            return "Setup"
+        }
+
         return "Terminal \(fallbackIndex + 1)"
     }
 }
--- a/Tests/TBDAppTests/AutoTabLabelResolverTests.swift
+++ b/Tests/TBDAppTests/AutoTabLabelResolverTests.swift
@@ -46,6 +46,26 @@ struct AutoTabLabelResolverTests {
         #expect(label == "Terminal 3")
     }
 
+    @Test func setupTerminalUsesSetupLabel() {
+        let terminal = Terminal(
+            worktreeID: UUID(),
+            tmuxWindowID: "@1",
+            tmuxPaneID: "%1",
+            label: "setup",
+            kind: .shell
+        )
+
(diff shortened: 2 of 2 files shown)


cheapsteak (approved):
💭 not 100% sure on this one

after the setup phase is done it'll still show setup as the name right? would that feel weird?

Jephuff:
sounds like we want that terminal to different things so this probably isn't as cut and dry! 
I think in my head it is always the setup tab, and I would probably start another terminal to do work (i.e. if that tab was not interactive I'd be a-ok with that)
But you want it to be the first terminal, and the first terminal opened runs the setup script. 

So maybe i need to think about this more to try to make us both happy haha

cheapsteak:
>But you want it to be the first terminal, and the first terminal opened runs the setup script.

It's complicated, I don't know what exactly I want 😂  

What it does right now happened to be the minimum amount of decisions to have gotten something working, but it's probably not the ideal state, but I'm also not sure what an ideal state might look like (and legitimately open to the idea that this might be a step towards ideal)

I think the core tension is -- is it better for setup to be a "special dedicated thing" or "something that happened to run in a terminal"

If it should be a special dedicated thing, then this PR is a step in the right direction, even if "special dedicated thing that can also be used like a terminal" feels not quite right. In the future, maybe it shouldn't be an interactive terminal panel, but should be more like a system dialog with read-only output?

But there's also something attractive about it not being a separate kind of thing

Curious to get your braindump/stacktrace/thoughts/gut feels/kneejerk reactions 🙏 
(What motivated the PR?)

PR #243, opened 2026-06-01
feat: show full live URL with copy button in webview top bar

Jephuff:
image   

## What

The webview pane's top bar showed only the host (e.g. `github.com`). It now shows the **full current URL**, updating live as you navigate within the page, and adds a **copy-to-clipboard button**.

## Changes

- **`WebviewPaneView.swift`** — Added a `WebviewState` observable model holding the live `currentURL`. The coordinator KVO-observes `webView.url`, so the bar tracks the actual page as you browse, including SPA/pushState changes that don't fire navigation delegates.
- **`PanePlaceholder.swift`**
  - Top-bar label now shows the full URL (tail-truncated to one line, with the full URL as a hover tooltip).
  - Added a copy button (`doc.on.doc`, flips to a `checkmark` for ~1.5s after copying) that writes the current URL to `NSPasteboard`.

The tab title (in `TabBar.swift`) is intentionally left as the host — a full URL would be too long for a tab.

## Behavior decisions
- **Live current URL** (not just the initial URL) — reflects in-page navigation.
- **Tail truncation** — keeps the domain/start visible; copy always copies the full URL.

## Testing
- `swift build` clean.
- View-layer (AppKit/SwiftUI) change; not exercised by `swift test`. Manual verification of live update + copy in the running app still recommended.

### DIFF
--- a/Sources/TBDApp/Panes/PanePlaceholder.swift
+++ b/Sources/TBDApp/Panes/PanePlaceholder.swift
@@ -37,6 +37,8 @@ struct PanePlaceholder: View {
     @State private var isHeaderHovering = false
     @State private var showSourceCode = false
     @State private var hasRenderableContent = false
+    @StateObject private var webviewState = WebviewState()
+    @State private var didCopyURL = false
 
     /// Find the Terminal model matching a terminal ID across all worktree terminals.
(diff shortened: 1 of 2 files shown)


cheapsteak (approved):
🙏🙏🙏

PR #245, opened 2026-06-02
Hide unavailable agent menu items

Jephuff:
## Summary
- Detect whether `claude` and `codex` executables are available before opening the add-tab menu.
- Hide Claude/profile menu items when Claude is unavailable, and hide Codex when Codex is unavailable.
- Add focused tests for executable detection and menu filtering.

## Test Plan
- [x] `swift test --filter AddTabMenu`

### DIFF
--- a/Sources/TBDApp/TabBar.swift
+++ b/Sources/TBDApp/TabBar.swift
@@ -170,6 +170,7 @@ private struct AddTabButton: View {
     let onAddCodex: () -> Void
     let onAddNote: () -> Void
     @State private var isHovering = false
+    @State private var availability = AgentExecutableAvailability.detect()
 
     var body: some View {
         Image(systemName: "plus")
@@ -185,9 +186,13 @@ private struct AddTabButton: View {
             .onHover { hovering in isHovering = hovering }
             .onTapGesture { showMenu() }
             .help("New Tab")
+            .task {
+                availability = AgentExecutableAvailability.detect()
+            }
     }
 
     private func showMenu() {
+        availability = AgentExecutableAvailability.detect()
         let coordinator = MenuCoordinator(
             onShell: onAddShell,
             onClaude: onAddClaude,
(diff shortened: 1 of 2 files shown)


cheapsteak (commented):
## Review: when does the check run, and how to tighten it

The availability check runs in **three** places, but only one of them actually drives the menu:

| Location | When | Observed? |
|---|---|---|
| `@State private var availability = .detect()` | once at view-state init | no |
| `.task { availability = .detect() }` | once when view appears | no |
| `showMenu()` → `availability = .detect()` | **every click**, right before `build(...)` | yes — the menu is built from this value |

Because `showMenu()` re-detects and passes the fresh value straight into `AddTabMenu.build`, behavior is correct and stays fresh: install/remove `claude` while the app runs and the next menu open reflects it. 👍

### Suggestion: collapse to the click-time detection

Nothing in `body` ever reads `availability`, so the `@State` default and the `.task` block are dead I/O — removing them changes zero behavior. Recommend a single local detection:

```swift
private func showMenu() {
    let availability = AgentExecutableAvailability.detect()
    // ...build as before...
}
```

…and drop `@State private var availability` and the `.task { ... }`. One detection path, no stored state, and the existing tests already cover `detect()` and `build(availability:)` directly.

### Not recommended: caching

Tempting to memoize to avoid the per-click `stat`s on the main thread, but it's a handful of stats on a local path (sub-ms, only on explicit click), and a cache would reintroduce the freshness/invalidation complexity we'd otherwise be deleting. Keep it synchronous-per-click.

### Minor

- Add a one-line comment on `detect()` explaining the hardcoded fallback dirs are load-bearing — GUI apps inherit a stripped `PATH`, not the interactive shell's — so nobody "cleans them up" later.
- Worth a quick grep for wherever `claude`/`codex` binaries get resolved at spawn time; if that logic exists, consider sharing a helper so "where do we look for the agent binary" has one answer.

Otherwise the feature + tests look good.

cheapsteak:
(it added a "Not recommended: caching" section because I asked if if we should add caching XD)

cheapsteak (approved):

PR #253, opened 2026-06-05
feat: let Claude-direct (OAuth) profiles set a model

Jephuff:
## What

Lets **Claude (direct) / OAuth** model profiles carry an optional model, so `ANTHROPIC_MODEL` is emitted for them when spawning a `claude` session.

## Why

The daemon already supported a per-profile model end-to-end — `handleModelProfileAdd` stored it for OAuth profiles, `ModelProfileResolver` passed it through, and `ClaudeSpawnCommandBuilder` emitted `ANTHROPIC_MODEL` for non-bedrock profiles. The only missing piece was the UI: Claude-direct profiles had no field to set or edit a model, so the value was always `nil`. This is therefore a **UI-only** change — no RPC, daemon, or DB-migration work.

## Changes

- **Create sheet** (`AddModelProfileSheet`): added an optional **Model** field to the Claude-direct preset; `save()` now persists it (previously forced to `nil`).
- **Edit path**: new `EditClaudeDirectSheet` (name + optional model), reachable via an **Edit** button on Claude-direct profile rows (`kind != .bedrock && baseURL == nil`, covering OAuth and direct API-key). It reuses the existing `updateModelProfileEndpoint` RPC with `baseURL: nil`, which updates only the `base_url`/`model` columns and leaves `kind` intact.
- **Tests**: two `ClaudeSpawnCommandBuilder` tests locking the contract — OAuth + model emits `ANTHROPIC_MODEL`; OAuth without a model does not.

## Verification

- `swift build` ✅
- `swift test --filter ClaudeSpawnCommandBuilder` ✅ (37 passed, including the 2 new)

### DIFF
--- a/Sources/TBDApp/Settings/ModelProfilesSettingsView.swift
+++ b/Sources/TBDApp/Settings/ModelProfilesSettingsView.swift
@@ -642,7 +642,9 @@ struct AddModelProfileSheet: View {
                 name: trimmedName,
                 token: preset == .claudeDirect ? nil : tokenValue,
                 baseURL: preset == .proxy ? trimmedBase : nil,
-                model: (preset == .proxy || preset == .claudeDirect) ? (trimmedModel.isEmpty ? nil : trimmedModel) : nil
+                // Bedrock returns early above, so only proxy/claudeDirect reach here —
+                // both carry an optional model.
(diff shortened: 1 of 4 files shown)


cheapsteak (approved):
🙏

PR #258, opened 2026-06-09
fix: sign TBD bundle with a stable identity so TCC grants persist

Jephuff:
## Problem

TBD users see recurring macOS permission prompts (Desktop, Documents, Downloads, Photos, Media library, network volumes, Full Disk) that **never stop** — clicking "Don't Allow" doesn't stick, re-prompting many times a day even without a rebuild.

### Why the prompts appear at all (expected)
TBD spawns `claude`/`codex`/shell/git. When a child touches a TCC-protected folder, macOS attributes it to the **responsible app** (TBD) and prompts under TBD's name. Claude Code probes the standard home folders at session startup, so every new tab / `/clear` / `/compact` triggers a burst. TBD's own Swift code never scans those folders.

### Why they never stop (the bug)
`scripts/restart.sh` signed the bundle **ad-hoc** (`codesign --sign -`), giving TCC only a bare `cdhash` with no stable anchor:

```
designated => cdhash H"967b10c8…"   flags=0x2(adhoc)   TeamIdentifier=not set
```

The responsible process resolved to an anchorless ad-hoc identity that didn't validate against the stored requirement. `tccd` logged this directly — 16× in one day, including between restarts:

```
tccd: Failed to match existing code requirement for subject
      com.github.cheapsteak.tbd and service kTCCServiceSystemPolicyAllFiles
```

No match → re-prompt, forever.

## Fix

Sign with a stable self-signed `TBD Dev Signing` identity instead of ad-hoc, producing an anchored, rebuild-stable designated requirement:

```
designated => identifier "com.github.cheapsteak.tbd"
               and certificate leaf = H"9934b5c2…"   flags=0x0(none)
```

The leaf-cert hash is constant, so one Allow/Deny decision persists across rebuilds and across spawned child sessions. `restart.sh` uses the identity when present and **falls back to ad-hoc** otherwise, so fresh clones / other contributors keep working.

## Changes
- `scripts/restart.sh`: stable-identity signing with ad-hoc fallback.
- `docs/tcc-signing.md`: root-cause writeup + one-time per-machine cert setup.

## Verification
- Confirmed via `codesign -d -r-` that the installed `/Applications/TBD.app` requirement flipped from ad-hoc `cdhash` to the anchored identifier+leaf form after `restart.sh`.
- `bash -n scripts/restart.sh` parses.
- Stale TCC rows cleared with `tccutil reset All com.github.cheapsteak.tbd`.

⚠️ Each developer must run the one-time setup in `docs/tcc-signing.md` to create the `TBD Dev Signing` identity; without it, signing falls back to ad-hoc (unchanged prior behavior). The final "deny truly persists" check requires clicking the GUI prompt once.

### DIFF
--- a/docs/tcc-signing.md
+++ b/docs/tcc-signing.md
@@ -1,0 +1,116 @@
+# Stable code signing for TCC permission persistence
+
+## Problem
+
+TBD spawns `claude`/`codex`/shell/git as child processes. When one of those
+touches a macOS TCC-protected location (Desktop, Documents, Downloads, Photos,
+Media library, network volumes, Full Disk), macOS attributes the access to the
+**responsible app** — TBD — and shows a consent prompt under TBD's name. In
+particular, Claude Code probes the standard home folders at session startup, so
+every new tab / `/clear` / `/compact` can trigger a burst of prompts.
+
+That much is expected. The real bug was that the prompts **never stopped** — a
+"Don't Allow" never stuck, re-prompting many times a day even without a rebuild.
+
+### Root cause
+
+`scripts/restart.sh` signed the assembled bundle **ad-hoc** (`codesign --sign -`).
+An ad-hoc signature gives TCC a designated requirement that is just a bare
+`cdhash` with no stable anchor:
+
+```
+designated => cdhash H"967b10c8…"     flags=0x2(adhoc)   TeamIdentifier=not set
+```
+
+TCC stores the user's decision keyed on the app's code requirement, but the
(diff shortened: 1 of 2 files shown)


cheapsteak (approved):

cheapsteak:
I wonder why I haven't hit this yet

Jephuff:
not sure! it seemed so random, but I have been running this for awhile and it shut it up! so thats good :)

PR #269, opened 2026-06-11
Feat: Run blocking setup hooks before the agent spawns, with live output instead of an empty panel

cheapsteak:
## Summary

**What's wrong today:** when a worktree is created, the Claude window and the `setup` hook window are spawned as two parallel fire-and-forget tmux windows — Claude always starts *before* the hook even launches, so hooks that copy `.env` files or install dependencies race against the agent. Meanwhile the app shows the "No terminals" empty state for the whole creation, and hook output/status is visible nowhere except its own pane.

**This PR adds a blocking `preSession` worktree hook** (`.worktree-hooks/preSession`, same resolution chain as the other events) that runs to completion *before* the primary agent terminal spawns — and surfaces it:

- The hook runs in a **visible terminal** created first, so the panel shows its live output instead of the empty state; the sidebar reads "Running setup…" and a slim banner explains the agent will start when setup completes.
- Completion is signaled by a marker file under `~/tbd/runtime/presession/` polled by the daemon (600s timeout). Non-zero exit, timeout, or a killed pane → daemon notification, and the agent **spawns anyway** — failure sequences the spawn, it never blocks it forever.
- The wait runs in a detached phase-3 task so the per-repo creation lane (RepoSerializer) is never blocked; revive gets the same non-blocking treatment (`archived → creating → active`).
- Startup recovery resumes orphaned `.creating` rows after a daemon restart mid-hook (the tmux server and hook survive restarts), including mid-revive rows whose archived Claude sessions are restored, not replaced.
- The existing `setup` hook is unchanged: still parallel and non-blocking. **No `preSession` hook → behavior identical to before** (covered by tests on both branches).
- App consumes `terminalCreated` deltas (previously ignored): live tab updates with dedupe, tab-order convergence with the daemon, and selection hand-off to the agent tab when it appears.

Hardened through three internal review rounds (commits `a50b9ea`, `ed114cd`): reconcile no longer kills live `.creating` windows, repo-remove mid-wait can't leak agent processes, marker race/leak fixes, hook env vars (`TBD_WORKTREE_NAME/PATH`, `TBD_REPO_PATH`, `TBD_BRANCH`, `TBD_EVENT`, `TBD_TERMINAL_ID`), static banner icon (no always-animating spinner), and the revive "✓" pill now waits for the worktree to actually be usable.

**Known minor follow-ups** (reviewed, judged shippable): the revive pill can strand at in-flight if a phase-3 DB write fails (self-heals on daemon restart); a worktree archived with zero Claude sessions is treated as mid-create by crash recovery (cosmetic, self-healing); terminal label strings could move to a shared constant; `TestSupport` could be a `.testTarget`.

## Test plan
- [ ] 51 new tests across daemon + app cover: hook resolution, gated spawn ordering, marker exit-code/timeout/killed-pane paths, daemon-restart recovery (mid-create and mid-revive), reconcile sparing `.creating` windows, broadcast exactly-once gates, app dedupe/selection/tab-order/banner/subtitle/revive-pill branches — `swift test` (1550 tests) green apart from the known pre-existing TabBarHitAreaTests failure on this machine
- [ ] Manual: add `.worktree-hooks/preSession` with `sleep 15; touch /tmp/presession-done` to a repo, create a worktree → panel shows the hook terminal running, sidebar says "Running setup…", Claude tab appears and takes focus only after 15s
- [ ] Manual: same with `exit 1` → notification fires, Claude still spawns
- [ ] Manual: repo without the hook → creation behaves exactly as before

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=B8D31DBB-8049-438F-A808-74F350CBFEFC)

### DIFF
--- a/Sources/TBDApp/AppState+History.swift
+++ b/Sources/TBDApp/AppState+History.swift
@@ -145,7 +145,9 @@ extension AppState {
 
     /// Revive an archived worktree and resume the selected Claude session.
     /// Marks the row as `inFlight` immediately so the archived view can show
-    /// a status pill, then flips to `.done` on success or clears on failure.
+    /// a status pill, then flips to `.done` once the worktree is usable —
+    /// immediately when no blocking preSession hook gates it, otherwise via
+    /// `promoteRevivedWorktrees` when the row turns `.active`. Clears on failure.
     func reviveWithSession(worktreeID: UUID, sessionId: String) async {
         // Idempotency: ignore re-entrant calls (e.g. rapid double-click)
         // so a concurrent invocation can't overwrite the .done state with
@@ -168,13 +170,13 @@ extension AppState {
 
         do {
             let size = mainAreaTerminalSize()
-            try await daemonClient.reviveWorktree(
+            let revived = try await daemonClient.reviveWorktree(
                 id: worktreeID,
                 cols: size.cols,
                 rows: size.rows,
                 preferredSessionID: sessionId
             )
-            revivingArchived[worktreeID] = .done(snapshot: snapshot)
+            settleReviveState(id: worktreeID, snapshot: snapshot, revived: revived)
             await refreshWorktrees()
             await refreshArchivedWorktrees(repoID: snapshot.repoID)
         } catch {
--- a/Sources/TBDApp/AppState+Terminals.swift
+++ b/Sources/TBDApp/AppState+Terminals.swift
@@ -17,6 +17,108 @@ extension AppState {
         terminal.kind == .codex || terminal.label == "Codex" ? terminal.label : nil
     }
 
+    // MARK: - Pre-session hook terminals
+
+    /// Label the daemon assigns to the blocking `preSession` hook terminal
+    /// (see WorktreeLifecycle+PreSession). The app keys "is this the
+    /// pre-session tab?" decisions off this label.
+    static let preSessionTerminalLabel = "pre-session"
+
+    /// True when a pre-session hook terminal exists in state for the worktree.
+    /// Drives the sidebar "Running setup…" subtitle while the worktree is
+    /// still `.creating`.
+    func hasPreSessionTerminal(worktreeID: UUID) -> Bool {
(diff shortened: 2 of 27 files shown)


cheapsteak:
Addressed the review findings in c6375bd:

- **Medium (label divergence)**: added a shared `TerminalLabel` namespace in `TBDShared` (`pre-session`, `setup`, `shell`, `Claude Code`, `Codex`) and migrated every label assignment/comparison site in the daemon and app to it, including the recovery classifier this finding called out. Test literals were intentionally kept to pin the on-disk contract.
- **Minor (recovery tab cleanup)**: the no-terminals recovery branch now deletes terminal + tab rows before the worktree row, matching the checkout-missing branch; test extended to seed and assert a stray tab row.
- **Minor (redundant refresh)**: dropped the explicit `refreshWorktrees()` in `reviveWorktree` — the `.worktreeRevived` delta handler already triggers it, same as `createWorktree`. Verified the revive settle/promotion logic doesn't depend on the synchronous call.

cheapsteak:
Addressed in 308c1e1:

- **Medium (DB error vs. deleted row)**: the phase-3 existence guard now uses explicit do/catch — a thrown DB error logs a warning and proceeds with the primary spawn; only a genuine missing row takes the teardown path. Covered by a test that closes the GRDB writer mid-wait (`dbErrorDuringRowCheckDoesNotTearDownPreSessionWindow`).
- **Minor (reconcile gate)**: `shouldReconcileTabOrderFromDaemon` is now scoped to worktrees with status `.creating`, so externally-created terminals in long-lived worktrees no longer trigger extra `listTabs` fetches. Branch tests added for active/not-found worktrees.
- **Minor (dead marker removal)**: removed the no-op second `removeItem` and corrected the comment.
- **Minor (repo-less .creating row)**: recovery now defensively deletes terminals + tabs + row when the repo record is missing instead of stranding it; test seeds the orphan via `PRAGMA foreign_keys = OFF`.

The `getTerminal(id:)` RPC suggestion is deferred as a follow-up per the review note (scope expansion).

PR #274, opened 2026-06-16
Feat: Configure environment variable overrides for spawned Claude/Codex sessions

cheapsteak:
## Summary

Lets you set arbitrary `KEY=VALUE` environment variables on the Claude/Codex sessions TBD spawns, at three scopes — **global**, **per-repo**, and **per-model-profile** — so things like routing a specific repo's Claude through Bedrock (`CLAUDE_CODE_USE_BEDROCK=1`, `AWS_REGION`, …) become configurable instead of hard-coded.

- **Free-form `[String: String]`** at each scope, persisted in a new JSON `env_overrides` column on the `config`/`repo`/`model_profiles` tables (migration `v31`). No validation/denylist/secret machinery by design — kept deliberately simple.
- **Precedence `global < repo < profile`** (union merge; more specific scope wins collisions), computed by a small pure `EnvOverrideResolver.merge`. The Claude model profile's structured **auth/routing env stays final** and can't be clobbered by a free-form var (OAuth dir, Bedrock `CLAUDE_CODE_USE_BEDROCK`/`AWS_REGION`/`AWS_PROFILE`, `ANTHROPIC_MODEL`/`ANTHROPIC_BASE_URL`).
- Applies to **both Claude and Codex** primary sessions and survives every respawn path (resume, reboot-recovery, manual create, recreate-window, mid-conversation profile swap). Plain shell and setup/pre-session hook terminals are intentionally excluded.
- Edited from three UI surfaces via a reusable `EnvOverridesEditor`: Settings → General (global), the per-repo settings pane, and the model-profile edit sheets.
- Docs: `docs/env-overrides.md` (scopes, precedence, auth-tier rule, Bedrock worked example) + design spec under `docs/superpowers/specs/`.

## Test plan

- [x] `swift build`, `swiftlint --strict` (0 violations), `swift test` (green except the pre-existing `TabBarHitAreaTests` headless-rendering flake)
- [x] Unit: `EnvOverrideResolver.merge` precedence/union/nil; `EnvOverridesCoding` round-trip/corrupt; per-scope store round-trips; RPC round-trips persist to DB; resolver passthrough
- [x] Spawn: Claude builder env wins over free-form (auth final); **Codex on/off branch tests** (merged global+repo vars reach `-e`; empty config injects nothing)
- [ ] Manual GUI smoke (needs app restart): set an override at each scope, spawn a Claude **and** a Codex worktree, confirm `printenv` shows the merged result, that a profile var beats a colliding repo var, and that an auth var (e.g. `AWS_REGION`) is not overridable from repo/global

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=97D6FD5E-1AE3-4376-A4FC-B4254AFF6C96)

### DIFF
--- a/.claude/skills/tbd-project/SKILL.md
+++ b/.claude/skills/tbd-project/SKILL.md
@@ -88,6 +88,8 @@ TBD instruments every `claude` (and `codex`) session it spawns so the app can ob
 
 The overlay and plugin are regenerated on every daemon startup; both writes are idempotent and non-fatal on failure (the session just loses instrumentation, it still runs).
 
+Spawned Claude/Codex sessions also receive free-form `KEY=VALUE` env overrides merged across three scopes (`global < repo < profile`), with the Claude builder's auth/routing env layered last so it can't be clobbered. See `docs/env-overrides.md`.
+
 ## Common Tasks
 
 ### Restart for testing
--- a/Sources/TBDApp/AppState+ModelProfiles.swift
+++ b/Sources/TBDApp/AppState+ModelProfiles.swift
@@ -25,6 +25,9 @@ extension AppState {
             if result.primaryAgentPreference != primaryAgentPreference {
                 primaryAgentPreference = result.primaryAgentPreference
             }
+            if result.globalEnvOverrides != globalEnvOverrides {
+                globalEnvOverrides = result.globalEnvOverrides
+            }
         } catch {
             logger.error("Failed to list model profiles: \(error, privacy: .public)")
             handleConnectionError(error)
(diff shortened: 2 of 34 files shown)


cheapsteak:
Thanks for the review. Disposition:

**Medium — missing Claude spawn integration test (profile branch + auth-wins):** Addressed in `e41dae6`. Added two tests to `ModelProfileSpawnTests.swift` driving the real `spawnPrimaryTerminals` Claude path via the lifecycle fixture (now wired with a `ModelProfileResolver`):
- `claudeReceivesMergedEnvOverrides` — global + repo + profile overrides all appear in the tmux `-e` flags.
- `claudeAuthEnvWinsOverFreeFormCollision` — a Bedrock profile with `awsRegion = us-west-2` and a colliding free-form `AWS_REGION = us-east-1` spawns with the builder's `AWS_REGION=us-west-2` (+ `CLAUDE_CODE_USE_BEDROCK=1`), not the free-form value. This locks the auth-final invariant at the spawn site — it would catch an inverted/removed merge that the isolated `Dictionary.merging` test cannot.

`swift test --filter ModelProfileSpawn` → 15/15 pass; `swiftlint --strict` clean.

**Minor — `EnvOverridesEditor` stale rows on same-repo external update:** Deferring, as noted. It's a pre-existing `@State`-seeded-once pattern, and there is currently no path (no CLI) that mutates `env_overrides` for a displayed repo out-of-band, so it's unreachable today. If a CLI for env overrides ever lands, the fix is a one-line `.onChange(of: initial)` reseed. Tracking as a follow-up rather than adding speculative code now.

PR #276, opened 2026-06-16
feat: add non-destructive `tbd worktree forget` command

zionts:
## What
Adds `tbd worktree forget <name-or-id>` — removes a worktree from TBD's tracking but **keeps its directory on disk**. Unlike `archive` (which runs `git worktree remove --force` and deletes the folder), `forget` deliberately skips the git removal, so the folder and all its files — including uncommitted and gitignored content like `.context/` — stay exactly in place.

## Why
Archiving is destructive: it deletes the working directory. Sometimes you want a worktree gone from the TBD sidebar but its files preserved on disk to reference later. `forget` provides that non-destructive path.

## Behavior
- Mirrors archive's cleanup: kills tmux windows, deletes terminals + tabs, clears pending questions, reclaims per-session `ClaudeHookOverlay`.
- **Hard-deletes** the worktree row → absent from both the active and archived lists.
- Does **not** call `git worktree remove`; does **not** run the `before_worktree_remove` archive hook.
- Broadcasts the `.worktreeArchived` StateDelta so the app updates live.

## Caveat (documented in code + CLI help)
`reconcile` re-adopts on-disk git worktrees whose path is under one of TBD's own prefixes (`~/tbd/worktrees/<slot>/`, `<repo>/.tbd/worktrees/`), so a forgotten worktree under those paths reappears on the next reconcile. Worktrees outside those prefixes (e.g. `~/conductor/workspaces/...`) are never re-adopted, so forget sticks for them. An `os.Logger` warning fires when forgetting a TBD-prefixed worktree; a tombstone/ignore-list to make it stick everywhere is a noted follow-up (not built here).

## Files (6)
- `Sources/TBDDaemon/Lifecycle/WorktreeLifecycle+Forget.swift` (new) — `forgetWorktree(worktreeID:)`
- `Sources/TBDShared/RPCProtocol.swift` — `worktree.forget` method + `WorktreeForgetParams`/`WorktreeForgetResult`
- `Sources/TBDDaemon/Server/RPCRouter+WorktreeHandlers.swift` + `RPCRouter.swift` — handler + dispatch case
- `Sources/TBDCLI/Commands/WorktreeCommands.swift` — `forget` subcommand (mirrors `archive`)
- `Tests/TBDDaemonTests/WorktreeForgetTests.swift` (new) — both gating branches (forget **keeps** the dir + removes from both lists + cleans terminals/tabs; archive **removes** the dir), plus not-found and main-worktree-refused guards

## Testing
- `swift build` passes locally.
- `swift test` could **not** be run in the dev environment (CommandLineTools only, no full Xcode → `import Testing` fails repo-wide; pre-existing and unrelated to this change). The new tests use verified APIs and should run under CI/Xcode.

## Usage
```
tbd worktree forget <name-or-id> [--json]
```

### DIFF
--- a/Sources/TBDCLI/Commands/WorktreeCommands.swift
+++ b/Sources/TBDCLI/Commands/WorktreeCommands.swift
@@ -11,6 +11,7 @@ struct WorktreeCommand: ParsableCommand {
             WorktreeList.self,
             WorktreeAdopt.self,
             WorktreeArchive.self,
+            WorktreeForget.self,
             WorktreeRevive.self,
             WorktreeRename.self,
             WorktreeReparent.self,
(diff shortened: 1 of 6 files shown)


cheapsteak (approved):
🎉

PR #278, opened 2026-06-18
Perf: Flatten transcript row layout so a full-list measure stops pinning the main thread (15s+ hover freeze at 143 items)

cheapsteak:
## Problem

The transcript freeze ( issue #129 ) is **not item-count-bound and not trigger-specific** — it's per-row layout-tree **depth**. The latest hang capture (`hang-20260618-131434Z-67437.txt`, archived `history` pane, **itemCount=143**, hung ≥15 s, force-quit) was sampled 4× across the full window. Every resample bottoms out in the same chain:

- **`NSHostingView.didRequestHoverUpdate` → `ViewGraph.responderNode` → `LazySubviewPlacements` invalidation** — a mouse **hover** (not `scrollTo`, not scroll — there is no scroll frame anywhere) forces SwiftUI to re-place every subview, which asks the `LazyVStack` for an ideal size.
- That descends into a deep **recursive per-row tower**: `_FlexFrameLayout → _FrameLayout → _PaddingLayout → StackLayout/VStackLayout`, with **`explicitAlignment` re-invoked at every nesting level** — the classic SwiftUI alignment-guide blow-up (each `.frame(…alignment:)` re-resolves its child's alignment by descending the whole subtree).

**~80 % of every resample is this per-row layout recursion.** Zero TBDApp frames appear in the hot path — it's pure SwiftUI layout-engine cost reacting to the **shape** of our per-row view tree. We can't stop SwiftUI from hit-testing on hover (hover/scroll/focus all force a measure), so the durable fix is to make each row's `sizeThatFits` **cheap**.

## Approach

Each row nested layout-affecting modifiers that each produce a layout-engine node — `.frame(maxWidth:.infinity, alignment:)` (a `_FlexFrameLayout`), stacked `.padding(.horizontal,_).padding(.vertical,_)` (two `_PaddingLayout`s), an extra `HStack` + `Spacer`. Flattening the two dominant inline archetypes (~85 % of a real transcript: chat bubbles ≈50 %, activity/tool/thinking/system headers ≈35 %) removes redundant nodes from the tower the full-list measure pays **once per row, on every measure pass**.

### `ChatBubbleView`
- Replaced the outer `HStack { Spacer ; column.frame(maxWidth:.infinity) ; Spacer }` with a **single column** that fills width and reserves the 52 pt opposite-side gutter via `.padding(.trailing/.leading, 52)` — drops a `StackLayout` + `Spacer` per bubble row.
- Collapsed **three** stacked `.padding(.horizontal,_).padding(.vertical,_)` pairs into single `.padding(EdgeInsets(…))` (one `_PaddingLayout` instead of two, each).

### `ActivityRowChrome`
- Dropped the **redundant** `.frame(maxWidth:.infinity, alignment:.leading)` — the inner greedy `Spacer(minLength:8)` already fills width, so the flex-frame did nothing for layout. Removing it deletes a `_FlexFrameLayout` + its `explicitAlignment` recursion from every tool/thinking/system row.
- Collapsed the padding pair to `EdgeInsets`.

## Evidence (measured, not asserted)

Per-row layout-engine node count / max depth, current → flattened:

| archetype | nodes | max depth |
|---|---|---|
| chat bubble (assistant **and** user) | ~16 → ~11 | 7 → 5 |
| activity chrome row | ~12 → ~10 | 5 → 4 |

A throwaway NSHostingView harness (built, run, then removed before this PR) measured a **cold full-list eager measure** at N ∈ {150, 300, 600}:

- **Chat bubble: consistent ~11 % wrapper-measure reduction** at every N (e.g. N=300: 447.8 ms → 397.0 ms).
- Activity row: flex-frame **proven redundant**, perf delta small (~4–6 %, noise at N=600) — kept for correctness/cleanliness.

**No visual regression — pixel-proven.** NSHostingView → `NSBitmapImageRep` bitmap diff of current vs flattened, single representative row:

| archetype | differing px / total | max channel Δ |
|---|---|---|
| chat bubble (assistant) | **0 / 341 600** | 0 |
| chat bubble (user, alignment flip) | **0 / 341 600** | 0 |
| activity chrome | **0 / 67 200** | 0 |

Byte-for-byte identical, including the user/trailing alignment-flip case.

## Scope & honesty

- This shaves the **wrapper** slice that is ours to control. A chat bubble's body is MarkdownUI-internal depth (out of our control short of the rejected Textual migration) — the ~11 % is a reduction on the wrapper tower, stacked on top of unchanged markdown cost, not 11 % of the whole row.
- The live freeze is **not reliably reproducible** (synthetic cursor warps don't fire SwiftUI hover, and the natural repro is intermittent), so the harness + bitmap-diff are the verification of record rather than a live before/after.
- Orthogonal to the in-flight `scrollTo`/measureEstimates work ("Track A1") — that kills a scroll trigger; this reduces per-row depth so any full-list measure (hover, scroll, focus) is cheaper. Both compose.
- Text selection (#244) unaffected. Added an `ActivityRowChrome` parity preview; `ChatBubbleView` parity previews already exist.

`swift build` clean; `swiftlint --strict` clean; transcript test suites green. The 3–5 failing tests on this run (`TabBarHitArea` geometry, `ThemeStore` FileWatcher, `AppearanceDebounce`) are pre-existing timing/geometry flakes in unrelated subsystems (the failing set varied run-to-run).

Refs #129.

---
🌳 [Open in TBD](tbd://open?worktree=501C9A1F-7C70-458E-B285-10E52BE1B3BB)

### DIFF
--- a/Sources/TBDApp/Panes/Transcript/ActivityRowChrome.swift
+++ b/Sources/TBDApp/Panes/Transcript/ActivityRowChrome.swift
@@ -48,9 +48,13 @@ struct ActivityRowChrome<Header: View>: View {
             }
         }
         .buttonStyle(.plain)
-        .padding(.horizontal, 12)
-        .padding(.vertical, 4)
-        .frame(maxWidth: .infinity, alignment: .leading)
+        // Flattened (issue #129 per-row layout-depth): the inner HStack's greedy
+        // `Spacer(minLength: 8)` already stretches the row to full width, so the
+        // prior `.frame(maxWidth: .infinity, alignment: .leading)` was a redundant
+        // _FlexFrameLayout (proven pixel-identical when removed). Dropping it
+        // removes a flex-frame + its explicitAlignment recursion from every
+        // tool/thinking/system row's measure pass. Padding pair → single EdgeInsets.
+        .padding(EdgeInsets(top: 4, leading: 12, bottom: 4, trailing: 12))
         .background(Color(nsColor: .windowBackgroundColor).opacity(hovering ? 0.65 : 0.4))
         .contentShape(Rectangle())
         .onHover { hovering = $0 }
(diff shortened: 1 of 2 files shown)


cheapsteak:
Incorporated the Minor finding directly (commit 23511df): folded the 52pt gutter into the chrome `EdgeInsets` (12 + 52 = 64), removing one more `_PaddingLayout` per bubble row. Nested uniform paddings compose additively, so it's layout-identical to the prior two-`.padding` chain — pixel parity preserved. Thanks for the catch.

PR #289, opened 2026-06-21
feat(transcript): opt-in virtualized transcript list to fix #129 freeze (Settings → Experimental, default off)

### LINKED ISSUE #129: Tracking: recurring main-thread UI hangs in the transcript pane

cheapsteak:
## Problem

Issue #129's transcript freeze is dominated by **non-virtualized `ForEach` list reconciliation**: every transcript content change re-runs `ForEachState.applyNodes` over **all N rows** (O(N), heavy per-item constant). `LazyVStack` makes view-*body* realization lazy but **not** list reconciliation — three macOS spindumps all bottom out in `applyNodes`/`DynamicContainerInfo.updateItems`, and a 500-row mount hard-freezes the main thread ~3.9s. PR #287 (per-row layout flatten) bought ~3%; this is the structural lever.

## What this adds (opt-in, default OFF)

An **experimental Settings → Experimental toggle, "Virtualized transcript (no text selection)"** that swaps the live-transcript pane from `LazyVStack { ForEach }` to an AppKit virtualizer — `NSScrollView` + view-based `NSTableView` (`usesAutomaticRowHeights`) hosting one SwiftUI `TranscriptRow` per visible row via `NSHostingView`. Only ~visible rows realize/reconcile → **O(visible)**, not O(N).

- `@AppStorage`-backed → flipping the toggle switches the pane live (no relaunch). Env var `TBD_VIRT_TRANSCRIPT=1` remains as a harness/testing override.
- **Default off; the `LazyVStack` path is byte-for-byte unchanged when off** (mirrors the existing `enableTranscript` experimental toggle).

## Measured (perf harness, 500 preseed + streaming)

| Metric | `LazyVStack` (current) | Virtualized + cache |
|---|---|---|
| Mount of 500 rows | **3.94 s hard freeze** | **fast — max stall 299 ms** |
| Rows realized | all N (O(N)) | **32 / 600** (O(visible)) |
| Re-realization churn | ~6× per row | **~1×** |

The mount/activation freeze that motivates #129 is eliminated. The per-update O(N) node rebuild is avoided by reusing the existing `TranscriptNodeCache` (the same memoization the `LazyVStack` path already uses).

## Trade-offs / known limitations (intentional)

- **In-row text selection is dropped.** AppKit's `NSTableView` owns the mouse for row selection; making hosted SwiftUI `.textSelection` win the drag required an event hack that infinite-recursed (SIGSEGV), and the clean non-recursive path doesn't deliver selection. A diagnostic (force-enabled selection) confirmed it's the table-vs-text event conflict, not the hover gate. Per product call, in-row drag-select is **removed** (cross-row selection never worked in either path anyway — production gates selection to a single hovered row). The full investigation is in `docs/spikes/2026-06-20-nscollectionview-transcript-spike.md`.
- **Click-through to hosted controls (tool-row overlay buttons, file links, AskUserQuestion) is the one thing to validate during soak** — the synthetic harness only emits chat bubbles, so taps on interactive cards are unverified live. `validateProposedFirstResponder`/`shouldSelectRow` are wired for it; flagged in the file header.
- **Pre-existing, NOT introduced here:** a sustained ~1/sec ~250–290 ms idle stall after large transcripts — the `LazyVStack` baseline shows the same tail (111 vs 98 stalls over ~94 s). Separate follow-up.

## Why merge behind a toggle

The freeze isn't reliably reproducible synthetically; it needs real-session soak. This lands the alternative view default-off so it can be flipped on per-session, compared against the `LazyVStack`, and fall back instantly — with production untouched until it's proven.

## Implementation notes (for review)

- Gate moved to `LiveTranscriptPaneView.transcriptWithAutoscroll` so the virtualizer *replaces* the SwiftUI `ScrollView` and gets a bounded viewport (a virtualizer nested in a SwiftUI `ScrollView` is proposed unbounded height → blank).
- Table mutations are deferred out of SwiftUI's `updateNSView` pass (coalesced `DispatchQueue.main.async`) to avoid the "reentrant operation in NSTableView delegate" → dropped-reload → blank.
- `AppState.virtualizedTranscriptEnabled(defaults:)` is the fail-closed accessor; `TranscriptItemsView.virtualizedTranscriptEnvOverride(_:)` is the env override.

## Tests

`TranscriptVirtualizationGateTests` (7) covers the env override (on/off) and the `AppState` setting accessor (unset/true/false) via injected `UserDefaults(suiteName:)`. `swift build` green, `swiftlint --strict` clean. Full `swift test` green except the pre-existing environment-sensitive `TabBarHitAreaTests` (fails identically on `main`).

[🧪 NSCollectionView Spike](https://cheapsteak.github.io/tbd/open/?worktree=82B7BC00-B218-4984-B2B4-25B1BFB5B8E7)

### DIFF
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -1433,6 +1433,20 @@ final class AppState: ObservableObject {
         defaults.object(forKey: enableTranscriptKey) as? Bool ?? false
     }
 
+    /// UserDefaults key for the `@AppStorage` toggle in Settings → Experimental
+    /// that swaps the live-transcript pane to the AppKit-virtualized list
+    /// (NSTableView + NSHostingView) — O(visible) reconciliation instead of the
+    /// LazyVStack's O(N) per change, eliminating the large-transcript mount/scroll
+    /// freeze (issue #129). Trade-off: in-row text selection is dropped. Opt-in,
+    /// fails closed (LazyVStack) so production is unchanged.
+    static let useVirtualizedTranscriptKey = "useVirtualizedTranscript"
+
+    /// Fail-closed read of the virtualized-transcript toggle for non-View callers.
+    /// Defaults false when untouched, matching the `@AppStorage` default.
+    static func virtualizedTranscriptEnabled(defaults: UserDefaults = .standard) -> Bool {
+        defaults.object(forKey: useVirtualizedTranscriptKey) as? Bool ?? false
+    }
+
     /// UserDefaults key for a Claude spawn-env setting, by registry ID.
     nonisolated static func claudeEnvKey(_ settingID: String) -> String {
         "claudeEnvSetting.\(settingID)"
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -1437,14 +1437,20 @@ final class AppState: ObservableObject {
     /// that swaps the live-transcript pane to the AppKit-virtualized list
     /// (NSTableView + NSHostingView) — O(visible) reconciliation instead of the
     /// LazyVStack's O(N) per change, eliminating the large-transcript mount/scroll
-    /// freeze (issue #129). Trade-off: in-row text selection is dropped. Opt-in,
-    /// fails closed (LazyVStack) so production is unchanged.
+    /// freeze (issue #129). Trade-off: in-row text selection is dropped.
+    /// Defaults ON: when the (separately opt-in) live-transcript pane is enabled,
+    /// the virtualized renderer is the default because the freeze fix matters
+    /// more than in-row text selection. Explicitly turning the toggle OFF selects
+    /// the LazyVStack renderer.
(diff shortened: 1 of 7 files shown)


cheapsteak:
Closing as a documented dead-end. The AppKit `NSTableView` + `NSHostingView` virtualizer eliminated the freeze (500-row mount 3.94s → 299ms) but hit two fundamental blockers — no in-row text selection (NSTableView owns the mouse; hosted SwiftUI `.textSelection` never gets the drag) and row clipping (NSHostingView under-reports MarkdownUI's height; correct heights need O(N) measurement → re-freeze). `gonzalezreal/Textual` was evaluated and inherits both (it's SwiftUI.Text-based). Full write-up in #129. Pivoting to pure-SwiftUI windowing (bound the live LazyVStack to the last ~N messages + Load earlier), which keeps native selection + correct rendering. Branch retained for reference.

PR #300, opened 2026-06-26
Perf: Long transcripts open instantly instead of freezing the app (13s → ~50ms)

cheapsteak:
## Summary

The live transcript pane (SwiftUI `LazyVStack`) hung the main thread for seconds on long conversations (#129) — it measured every row on each layout pass, and opening a large session blocked on parsing the whole transcript (+ recursively every subagent file). This rebuilds the pane on a **virtualized AppKit `NSTableView`**, now the **default renderer** when the transcript is enabled, with a **Settings toggle ("New transcript renderer", on by default)** to fall back to the legacy SwiftUI pane for comparison.

- **No more freeze + near-instant open.** Cold open of a ~2,000-item / 30 MB session went from **~13 s of blank/frozen UI to ~50 ms to first content**: the daemon returns just the last ~60 items (parsing only the file's *tail bytes*), the pane renders them at the bottom immediately, then backfills the full history behind it with the viewport pinned.
- **Block-based message rendering.** Each message is a stack of native blocks in one bubble — prose on TextKit 1 (exact, fast `usedRect`), GFM tables as their own native block view — so tall messages no longer clip and tables render cohesively inside the bubble (no TextKit2 over-measurement, no attachment-icon fallback). Per-message text selection + a "Copy message" command. Per-block heights are cached so scroll-reuse never re-allocates an `NSHostingController`.
- **Native tool/system cards.** Tool/system header rows render as one native `ActivityRowCellView` (icon + title + timestamp + badges + hover + click-to-overlay) instead of per-row SwiftUI hosting, removing the last hot-path `NSHostingController.sizeThatFits` cost.
- **Stable virtualization.** Row-height estimation disabled; bottom-window measured exactly + cheap per-kind estimates corrected on realize; opens and stays pinned at the bottom.
- **Subagent parsing removed from the transcript.** The daemon parses the parent file only (Task/Agent calls render as normal tool cards) — this was the bulk of the cold-open cost on subagent-heavy sessions.

Addresses #129. The new pane is the **default** (the transcript itself remains opt-in via the existing "Live transcript pane" setting); flip the "New transcript renderer" toggle off to compare against the legacy SwiftUI pane. The old SwiftUI pane (and the intermediate TextKit2 pane, defaults-write only) remain as fallback paths. Perf diagnostics are retained at `.debug` (`log stream --level debug --predicate 'category == "table-transcript"'`).

Known non-blocking follow-ups (separate PRs): make the 1.5 s poll incremental (it still re-fetches the full transcript each tick once loaded); optional viewport-locking for the minor height nudge when scrolling far up into never-measured history.

## Test plan

- [ ] With "Live transcript pane" enabled, open a long transcript → lands at the bottom near-instantly, no freeze; history fills in behind without the viewport jumping. (New renderer is on by default.)
- [ ] Settings → toggle "New transcript renderer" off → switches live to the legacy SwiftUI pane; back on → table pane.
- [ ] Tables / code blocks / token badges render cohesively inside bubbles; tall messages don't clip.
- [ ] Drag-select text in a message + ⌘C copies it; right-click → "Copy message" copies the whole message.
- [ ] Tool/Read/Edit/Bash/Agent + system rows render natively; click opens the detail overlay.
- [ ] `swift test` green (the pre-existing `TabBarHitAreaTests` / `ThemeStore` file-watcher headless flakes are unrelated).

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=82B7BC00-B218-4984-B2B4-25B1BFB5B8E7)


cheapsteak:
Thanks for the review — addressed all findings:

- **Medium (table-block re-measure on scroll reuse):** the Coordinator now caches per-block heights (keyed `id+contentVersion+width`), populated for free during the existing row-height measure, and passes them into the cell so `rebuildBlockStack` reuses them instead of allocating a fresh `NSHostingController` per table block on every dequeue. Common scroll-reuse path is now a pure cache hit (zero hosting-controller alloc); render==measure stays 0.00pt incl. the table bubble. (293fdf8)
- **Minor — unused `role` param** in `bodyWidth`: removed + all call sites updated. (1e3398d)
- **Minor — `tailLimit` `var`→`let`.** (7e676bd)
- **Minor — dead `navigateTargetID`:** removed the field + always-nil handling arm. (10c8863)

The 3 filtered findings (double `scrollToEnd`, per-file `RoundedBoxView` isolation, `tail-N` IDs) are intentional and left as-is.

(The earlier `test` CI failure was the `ThemeStore` file-watcher timing flake — `userThemes.count → 0 == 1` — unrelated to this PR; the fresh run should clear it.)

PR #309, opened 2026-06-29
fix(worktree): de-dup folder name against archived row paths when opening existing branch

Jephuff:
## Problem

Using **open existing branch** for a branch whose previous worktree was archived made the new worktree row appear in the sidebar and then immediately vanish ("finds it → starts creating it → disappears").

## Root cause

Archived worktrees keep their `path` in the DB even after their directory is deleted, and `worktree.path` is **globally `UNIQUE`** (archived rows included). But `uniqueFolderName` only checked the **filesystem** when picking a folder name. With the archived directory already gone, it returned the base name unchanged, so phase‑1 `db.worktrees.create(path:)` threw `UNIQUE constraint failed: worktree.path`. The create RPC errored and the app removed its optimistic placeholder — the row "disappeared."

This was deterministic and specific to any branch that has a colliding archived row (e.g. re-opening a branch you previously archived).

## Fix

Make folder naming DB‑aware: a candidate name is "taken" if its absolute path exists on disk **or** is already reserved by any worktree row (active, archived, creating, main) — mirroring the global UNIQUE constraint. The branch now de‑dupes to `<name>-2` instead of throwing.

- `Sources/TBDDaemon/Lifecycle/WorktreeLifecycle+Create.swift`: `uniqueFolderName` takes a `reserved: Set<String>`; `beginCreateWorktree` fetches all worktree paths regardless of status and passes them in. No schema change; single call site (the `useExistingBranch` path).

## Test

Added `testCreateWorktreeForExistingBranchDeDupesAgainstArchivedRowPath` — sets up an archived row reserving the computed path (no dir on disk), re-opens the branch, and asserts creation succeeds with a de-duped `-2` name and a real checkout. Fails with the UNIQUE error before the fix, passes after.

## Verification

- `swift build` clean
- Full daemon suite: **933 tests pass**

### DIFF
--- a/Sources/TBDDaemon/Lifecycle/WorktreeLifecycle+Create.swift
+++ b/Sources/TBDDaemon/Lifecycle/WorktreeLifecycle+Create.swift
@@ -103,8 +103,14 @@ extension WorktreeLifecycle {
                 : providedBranch
             let sanitized = WorktreeLayout.sanitize(localBranch)
             let baseFolder = sanitized.isEmpty ? "branch" : sanitized
+            // Mirror the global UNIQUE constraint on `worktree.path`: avoid
+            // paths already reserved by ANY row (active, archived, creating,
+            // main). An archived worktree keeps its `path` even after its
+            // directory is deleted, so a filesystem-only check would collide
+            // and the insert would throw `UNIQUE constraint failed`.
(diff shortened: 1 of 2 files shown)


cheapsteak (approved):

cheapsteak:
nice, this bit me a few times already 🙏

PR #312, opened 2026-06-29
feat(guardrails): extensible PreToolUse hook framework + background-job rule

cheapsteak:
## What

Adds a repo-scoped, extensible **PreToolUse guardrail framework** under `.claude/`, with one initial rule that blocks background-process commands lacking a reliable teardown.

## Why

An autonomous agent (running with permissions skipped) simulated CI load in a flaky-test loop and reaped its background jobs with `kill $(jobs -p)`. Job control is off in the non-interactive tool shell, so that reaper silently does nothing — 22 `yes` processes orphaned to `launchd` and burned ~850% CPU for a week. Advisory `CLAUDE.md` text can't prevent this: it's only a nudge, it isn't reliably in context, and it doesn't hold under `--dangerously-skip-permissions`. A PreToolUse hook **does** hold under bypass mode, so enforcement belongs there.

## Design (avoids ad-hoc hook growth)

Registration is separated from rules: **one** `PreToolUse` entry in `.claude/settings.json` (matcher `Bash|Edit|Write`) points at a single `dispatch.py`. Adding a future rule is dropping one file in `rules/` plus a test — `settings.json` is never touched again.

- `dispatch.py` — reads the hook JSON on stdin, routes by `tool_name`, runs every applicable rule, aggregates (any `deny` wins), emits the `hookSpecificOutput` decision. **Fails open**: any internal error logs and allows, so a buggy rule can never wedge an agent.
- `lib/` — `Rule` interface + `Decision`; `registry.py` auto-discovers `rules/*.py` (package path derived from `__file__`, so it's cwd-independent and can't go inert).
- `rules/background_jobs.py` — high-precision: denies a genuine backgrounding `&` combined with the broken `jobs -p` reaper or a load-spawning loop, when no robust teardown (`trap '...' EXIT` / `kill 0` / `pkill -P`) is present. Errs toward allow; ignores `&&`, `&>`, `2>&1`, and properly-cleaned single backgrounds. The deny reason is instructive and is fed back to the model so it self-corrects.
- `tests/` — `unittest` suite (12 tests) pinned to the three real leaked commands plus safe-command fixtures, dispatcher fail-open/routing/aggregation.
- `README.md` — architecture + "how to add a rule" contract.
- `Tests/CLAUDE.md` — short domain note documenting the correct load-test pattern (documentation, not the enforcement).

## Verification

- `python3 .claude/hooks/guardrails/dispatch.py --self-test` → 12/12 green.
- Deny/allow fixtures piped from a non-repo cwd behave correctly; works from a worktree checkout too (cwd-independent).
- `pyright` → 0 errors / 0 warnings.

## Scope / rollout

Repo-scoped and version-controlled, so it propagates to worktrees as they pick up `main` (new worktrees cut from `main` get it automatically). Runtime cost is ~36 ms per matched call (dominated by interpreter startup), only on `Bash`/`Edit`/`Write`.

## Notes

- The rule matches command **text**, so a command that merely embeds the pattern as a string literal is also blocked (rare; pass such payloads via files).
- This framework is the migration path for the advisory "NEVER ..." rules in `CLAUDE.md` (e.g. deleting `state.db`, absolute-path `restart.sh`) — each becomes a small rule + test later.

### DIFF
--- a/.claude/hooks/guardrails/README.md
+++ b/.claude/hooks/guardrails/README.md
@@ -18,7 +18,7 @@ could not prevent it; a `PreToolUse` hook can. See `rules/background_jobs.py`.
 ## Architecture
 
 ```
-.claude/settings.json            ONE PreToolUse entry (matcher "Bash|Edit|Write") → dispatch.py
+.claude/settings.json            ONE PreToolUse entry (matcher "Bash") → dispatch.py
 .claude/hooks/guardrails/
   dispatch.py                    entrypoint: read stdin JSON, run rules, aggregate, fail-open
   lib/rule.py                    Rule base class + Decision value
--- a/.claude/hooks/guardrails/README.md
+++ b/.claude/hooks/guardrails/README.md
@@ -1,0 +1,132 @@
+# Guardrails — repo-scoped PreToolUse enforcement
+
+A small, extensible framework that inspects every Claude Code tool call **before**
+it runs and can BLOCK it. Unlike advisory `CLAUDE.md` text, a `PreToolUse` hook
+holds even under `--dangerously-skip-permissions`, so it is the right enforcement
+layer for "NEVER do X" rules.
+
+This framework is committed to the repo so it propagates to every worktree via git.
+
+## Why this exists
+
+An autonomous agent once ran a CI-load simulation that backgrounded `yes`
+processes and reaped them with `kill $(jobs -p)`. Job control is OFF in the
+non-interactive tool shell, so `jobs -p` returned nothing, `kill` reaped nothing,
+and 22 `yes` processes orphaned to launchd at ~850% CPU for a week. Advisory text
+could not prevent it; a `PreToolUse` hook can. See `rules/background_jobs.py`.
+
+## Architecture
+
+```
+.claude/settings.json            ONE PreToolUse entry (matcher "Bash|Edit|Write") → dispatch.py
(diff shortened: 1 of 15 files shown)


cheapsteak:
Addressed the Claude review feedback in three commits on \`guardrails-framework\`:

- **Medium — registry per-module import isolation** (\`c4c679a\`): wrapped the per-module \`importlib.import_module\` in \`try/except\` so a rule file with a SyntaxError or missing import is skipped individually instead of propagating out of \`load_rules()\` and dropping *all* rules via the dispatcher fail-open. Added \`tests/test_registry.py\` pinning the isolation (drops a broken module into \`rules/\`, asserts \`background-jobs\` still loads).
- **Minor — trap regex misses EXIT among multiple signals** (\`1aa9c40\`): the teardown regex now allows other signal names between the closing quote and EXIT, so the idiomatic \`trap 'kill $pids' TERM EXIT\` / \`INT TERM EXIT\` no longer false-denies a safe load loop. Verified it still denies a loop whose EXIT trap does not kill. Added a multi-signal-trap allow test.
- **Minor — stale README matcher** (\`6dd6c04\`): synced the architecture diagram from \`"Bash|Edit|Write"\` to \`"Bash"\` (matcher was narrowed in \`505686d\`).

Self-test green (14 tests), Pyright clean (0 errors/warnings).

PR #316, opened 2026-07-01
DX: Spin up an isolated seeded TBD instance for UI work and staged screenshots

cheapsteak:
## Summary

Developing TBD's UI (or capturing screenshots) currently means fighting the one real daemon and live `~/tbd/state.db` — there's no way to conjure rich, varied worktree state on demand, and no safe throwaway instance to drive.

This adds a **mock harness**: an isolated, hand-seeded daemon+app pair you launch with one command, that renders committed fixtures and **never touches the real `~/tbd`**.

- **`scripts/mock.sh up|down|shot|restart`** — launches an isolated daemon+app pair under `TBD_HOME=/tmp/tbd-mock`, seeded from a committed scenario. `shot <name>` screenshots the mock window to `artifacts/mock/`. It direct-execs the worktree's own `.build/debug/TBD.app`, so it never touches `/Applications` or the real `tbd://` deep-link registration, and tearing down leaves `~/tbd` untouched.
- **Daemon mock mode** (`TBD_MOCK=1` + `TBD_MOCK_FIXTURE=<scenario.json>`) — seeds the fixture through the existing GRDB stores, then skips all live reconciliation (health validator, git fetch/status loops, PR poller, reaper, suspend/resume, usage poller) so authored state renders exactly as written. Servers still start, so the app renders through the real RPC path.
- **Hand-authored fixtures** under `Tests/Fixtures/mock-state/` (placeholder `acme`/`acme-prod` repos): worktrees with varied statuses, PR badges, conflicts, parent/child nesting, an archived row, a suspended terminal, plus transcript `.jsonl` fixtures for the transcript pane.
- **Isolated app settings** — the mock app routes `AppState` to a `com.tbd.app.mock` UserDefaults suite and suppresses notification prompts, so a mock run never writes window/layout state back into the real `TBDApp.plist`.

Scope is **Tier 1** (sidebar, transcript pane, dialogs, toolbars, PR badges). Live terminal *scrollback* (canned tmux replay) is a deferred Tier 2 follow-up; no live `claude`/`codex` is spawned.

Design + plan: `docs/superpowers/specs/2026-06-30-mock-data-harness-design.md`, `docs/mock-harness.md`.

## Test plan

- [x] `swift test` — full suite green (1874 tests / 212 suites), incl. new `MockModeTests`, `MockSeederTests`, `MockUserDefaultsTests`
- [x] `swift build` + `swiftlint --strict` clean
- [x] Live: `scripts/mock.sh up default` renders the full seeded scenario in the sidebar (repos, emoji names, PR badges, parent/child nesting, activity states, archived hidden); `~/tbd/state.db` mtime unchanged before/after; `down` removes the mock home and leaves no orphan processes
- [ ] Reviewer: on a mock instance, click a seeded worktree (e.g. *Add CSV Export*) to confirm the transcript pane renders the seeded `.jsonl` (requires a physical click — TBD's SwiftUI rows don't respond to synthetic clicks)

## Known follow-ups (not in this PR)

- Mock-gate the app's `refreshPRStatuses` so screenshots don't trigger live `gh` calls on the fake repos (log noise only; seeded sidebar badges are unaffected).
- Tier 2: canned terminal scrollback via mock tmux sessions.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=6D749030-6C8A-4298-90D8-FBA4C11CE3A4)


cheapsteak:
Both findings addressed in `0afef23`.

**Medium — missing test for the mock-mode gate.** You were right that `Daemon.start()` can't be unit-tested directly (it binds the real socket + HTTP servers and spawns background tasks), so I extracted the DB-mutating reconciliation into a testable seam, `Daemon.performStartupReconciliation(mockMode:database:git:lifecycle:)`, and moved the `guard mockMode == nil else { return }` inside it. `start()` now just calls that seam. New `Tests/TBDDaemonTests/DaemonMockGateTests.swift` covers both branches against an in-memory DB, no servers required:

- **mock OFF** (`mockOffRunsReconciliation`): seeds a repo with a non-existent path, calls `performStartupReconciliation(mockMode: nil, …)`, asserts the repo flips to `.missing` — proving `RepoHealthValidator` ran.
- **mock ON** (`mockOnSkipsReconciliation`): seeds the same stale-path repo plus a worktree, calls `performStartupReconciliation(mockMode: .enabled(…), …)`, asserts the repo stays `.ok` and the worktree survives — proving neither `RepoHealthValidator` nor the reconcile loop ran.

The design doc's "Testing" section is updated to describe the seam and to state explicitly that `Daemon.start()` itself is not unit-tested (it binds real servers) — the maintainer call you flagged.

**Minor — "Started successfully" trapped inside the `if mockMode == nil` block.** Moved back out; it now logs unconditionally on both paths.

Both mock-gate tests pass (0.07s). Full suite is green apart from the pre-existing `AppearanceDebounceTests` timer flake (#249), which my branch doesn't touch and which passes in isolation.

PR #317, opened 2026-07-02
Feat: Render terminals through one tmux control-mode connection per repo (opt-in Phases 1+2)

cheapsteak:
## Summary

- **Phase 1 (Foundation):** the daemon can open a `tmux -CC attach` control-mode connection per repo — pty-driven subprocess, dedicated reader thread, incremental protocol parser (`%output` octal decoding, `%begin/%end` blocks, lifecycle notifications), all behind the `TBD_TMUX_CONTROL_MODE` env gate + tmux ≥ 3.2 version check. Gate off (the default) = zero behavior change; grouped sessions remain the rendering path.
- **Phase 2 (FD vending + single-pane render):** one visible pane renders end-to-end through control mode. The daemon fans decoded `%output` bytes into a per-pane pipe **on the reader thread** (`PaneFanout`, composite `(server, paneID)` keys, partial-write handling), vends the pipe's read fd to the app over a sidecar Unix socket via `SCM_RIGHTS` (`FDChannel` in TBDShared — hand-rolled CMSG macros, since Swift won't import the C ones), and the app drains it into SwiftTerm from a view-independent reader registry. The attach handshake follows the spec's non-negotiable ordering: vend fd → app wires reader → `attach.ready` ack → write gate opens, with a generation-scoped 5s cancel for un-acked attaches.
- **Design/plan docs:** the v2 control-mode design (A-lite + FD vending, α replay, Policy B flow control) and both phase implementation plans, including a Rev 2 of the Phase 2 plan that folded in a pre-implementation review (FDChannel target placement, `daemon.capabilities` instead of env mirroring — the app is launched via `open`, which drops shell env — sidecar header demux with per-attach nonces, composite pane keys, reader-thread fanout).
- Also in this branch: auto-suspend and its brittle tmux status-bar text-matching are removed (the Stop hook is now the sole Claude-state signal); the `autoSuspendClaude` toggle survives, repointed at main's pre-sleep suspend; the dead `worktree.selectionChanged` RPC is deleted end-to-end.
- **Behavior change (deliberate):** manual Suspend no longer waits for Claude to go idle before sending `/exit` (previously a 10s capture-pane poll that could refuse with `.busy`). The idle detector was the brittle status-bar text-matching this branch removes — each Claude CLI release that tweaked the status display silently broke it. Mitigations: suspended sessions are always resumable via `--resume`, and Claude Code queues input typed mid-response rather than interrupting the stream. If a busy guard proves needed, the non-brittle signal is the control-mode `%output` activity tracker planned for a later phase — not text matching.

## What Phase 2 explicitly does NOT do (later phases)

No keystrokes (P3), no size arbitration (P4), no scrollback/α-replay — early bytes before `attach.ready` are dropped (P5), no flow control beyond nonblocking drop-and-count (P6), no crash recovery (P7). Known hardening notes for later: length-prefixed sidecar framing, and moving the fanout's per-chunk `[UInt8]` copy off the hot path when Phase 6 reworks the write loop.

## Test plan

- [x] `swift test` — 1937 tests / 230 suites green, including live-tmux integration tests (throwaway `-L` servers): control-mode connection events, teardown escalation + trailing-output preservation, and the full Phase 2 path (live tmux → reader thread → fanout pipe → `SCM_RIGHTS`-vended fd → marker read back)
- [x] `swiftlint --strict` — clean
- [x] Adversarial-review pass with fixes: per-attach nonce demux (two in-flight attaches for the same pane can't cross-deliver fds), generation-scoped ready timeout (a superseded attach's stale timer can't kill a fresh attach), capabilities fetched before `refreshAll()` so initially-rendered panes honor the gate
- [x] Manual, gate **off** (default): `scripts/restart.sh` → one daemon + one app, terminals render via grouped sessions, zero control-mode activity in logs, `daemon.capabilities` reports `controlModeEnabled: false`
- [x] Manual, gate **on**: `TBD_TMUX_CONTROL_MODE=1 scripts/restart.sh` → unified log shows the full sequence (`FD vending client connected` → `started tmux -CC connection` → `fanout attach …/%N gen=1` → `fanout ready`), one control client on the active server, readers exit cleanly across app restarts. This manual pass caught (and this branch fixes) attach against a server with no `-CC` connection — `attach.request` now ensures the connection idempotently.
- [x] Environment restored to gate-off defaults after verification

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=039AF536-0AC2-4EF7-977C-423B30D1BD47)


cheapsteak:
Re the **Medium** finding (manual Suspend no longer waits for idle): confirmed intentional — calling it out in the PR description now (see the new "Behavior change (deliberate)" bullet).

Rationale: the 10s idle wait was built on `ClaudeStateDetector`'s capture-pane text matching against Claude's status bar ("esc to interrupt" etc.), which is exactly the brittle surface this branch removes — it silently broke whenever a Claude CLI release tweaked the status display, and commit `a576332` documented dropping `.busy` as a consequence. Two mitigating factors for the interrupted-work concern: suspend marks the session resumable (`--resume` restores it), and Claude Code queues input submitted mid-response rather than killing the stream, so `/exit` while streaming takes effect at the response boundary rather than mid-generation. If a busy guard turns out to be wanted in practice, the right signal is the control-mode `%output` activity tracker planned for a later phase (typed byte-rate, no text matching) — noted in the description as the follow-up path.

On the minor: `AttachRequestParams.windowID` is deliberate scaffolding — the spec's attach message table carries `{paneID, windowID}` and Phase 4 (size arbitration) consumes the window target (`resize-window -t @<id>`); keeping it in the wire shape now avoids a params-struct migration then.

PR #321, opened 2026-07-03
fix(daemon): trim trailing blank pane rows before idle detection

zionts:
## The bug

`ClaudeStateDetector.checkIdle(output:)` takes the last 5 lines of tmux `capture-pane -p` output via `suffix(5)` **without trimming trailing blank rows first**. When a tmux pane is taller than Claude's drawn UI, the prompt (`❯`) and status-bar lines sit *above* a block of trailing blank rows. The `suffix(5)` window then captures only blanks, so neither the status-bar check nor the prompt check matches — a genuinely idle session is misread as busy, and `terminal.suspend` refuses to park it with "Claude is busy, try again later".

## Observed impact

During a memory-pressure incident (swap at 96%), **9 idle sessions** sitting at bare rest prompts were unparkable, blocking recovery — while sessions whose panes happened to be full parked fine. The differentiator was purely pane height vs. drawn-UI height, i.e. how many trailing blank rows the capture contained.

## The fix

Trim trailing empty/whitespace-only rows before `suffix(5)`:

```swift
var lines = output.split(separator: "\n", omittingEmptySubsequences: false)
while let last = lines.last,
      last.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty {
    lines.removeLast()
}
let lastLines = lines.suffix(5).map(String.init)
```

The prompt line is `❯` + a non-breaking space (`\u{00a0}`), which is not whitespace-empty and is therefore correctly preserved; only truly empty/whitespace-only rows are trimmed. Because the busy-indicator, status-bar, and prompt checks all derive from this same window, fixing it in one place corrects all three consistently.

## Tests

Added to `Tests/TBDDaemonTests/ClaudeStateDetectorTests.swift`:
- `idleWithTrailingBlankRows` — an idle pane (bare prompt + status bar) followed by a mix of empty and whitespace-only trailing rows now reads `== true` (the bug case).
- `busyWithTrailingBlankRowsStillNotIdle` — a busy pane (`esc to interrupt`) followed by trailing blanks stays `== false`, proving trimming doesn't flip a busy pane to idle.

`swift build` succeeds; the full `ClaudeStateDetector` suite (14 tests) passes. Two unrelated failures in the full suite (`agentExecutableAvailability_detectsHomeLocalBinFallback`, tab-bar hit-area layout) are pre-existing and environment-dependent — confirmed failing on the clean baseline without this change.

## Deployment note

This ships via PR only. I deliberately did **not** restart the live daemon (other sessions depend on it). Adam should run `scripts/restart.sh` when he chooses to pick up the fix.

### DIFF
--- a/Sources/TBDDaemon/Lifecycle/PluginDirWriter.swift
+++ b/Sources/TBDDaemon/Lifecycle/PluginDirWriter.swift
@@ -74,6 +74,39 @@ struct PluginDirWriter {
             encoding: .utf8
         )
 
+        // Bundled `nightwatch` skill (quota-lean fleet babysitter). Ships its
+        // SKILL.md + scripts/tick.py + config/* so it works against a TBD fleet
+        // out of the box. Dormant until its description matches a babysitting task.
+        try writeNightwatch(pluginDir: pluginDir)
+
         Self.logger.info("Wrote TBD plugin at \(pluginDir, privacy: .public)")
+    }
+
+    /// Install the `nightwatch` skill into the plugin's skills dir. `tick.py`
+    /// is written with the executable bit so cron/launchd can run it directly.
+    private func writeNightwatch(pluginDir: String) throws {
+        let fm = FileManager.default
+        let root = pluginDir + "/skills/nightwatch"
+        let scripts = root + "/scripts"
+        let config = root + "/config"
+        try fm.createDirectory(atPath: scripts, withIntermediateDirectories: true)
+        try fm.createDirectory(atPath: config, withIntermediateDirectories: true)
+
+        try NightwatchSkillContent.skillMd.write(toFile: root + "/SKILL.md", atomically: true, encoding: .utf8)
+        try NightwatchSkillContent.prioritiesTxt.write(toFile: config + "/priorities.txt", atomically: true, encoding: .utf8)
+        try NightwatchSkillContent.safeWedgesTxt.write(toFile: config + "/safe_wedges.txt", atomically: true, encoding: .utf8)
+        try NightwatchSkillContent.dontTouchTxt.write(toFile: config + "/dont_touch.txt", atomically: true, encoding: .utf8)
+
+        // Scripts ship executable. NOTE: the scheduler (scheduler.sh / tick-cron.sh)
+        // is installed but NEVER auto-loaded — durable scheduling is opt-in via
+        // `scheduler.sh enable`.
+        for (name, body) in [("tick.py", NightwatchSkillContent.tickPy),
+                             ("judge.py", NightwatchSkillContent.judgePy),
+                             ("tick-cron.sh", NightwatchSkillContent.tickCronSh),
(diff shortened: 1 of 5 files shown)


cheapsteak (approved):

cheapsteak (commented):
Thanks for chasing this down — the memory-pressure incident write-up is genuinely useful, and the root-cause reasoning (pane height vs. drawn-UI height determining whether `suffix(5)` lands on blank rows) is exactly right for the code as it stood when you branched.

The catch: **this fix targets code that's already been removed from `main`.** `ClaudeStateDetector.checkIdle(output:)` — along with `isIdle`, the busy/status/prompt pattern constants, and the whole auto-suspend path that consumed them — was deleted in #317 (the tmux control-mode work). This branch is based on `0c73f7ed` (#292) and is ~43 commits behind, so it never picked that up. On current `main` there's no `checkIdle` to patch, and merging would either resurrect the deleted detector or conflict.

That removal was deliberate, and for the same class of reason your bug demonstrates: **we're trying to move away from screen-scraping Claude's TUI.** Text-matching the status bar (`esc to interrupt`, `? for shortcuts`, the `❯` prompt) breaks every time the Claude CLI tweaks its display — the trailing-blank-rows bug is just another instance of that brittleness. The intended direction if idle-gating comes back is hook-based (the Stop hook's `response_complete` signal), not pane parsing.

So the "trim trailing blank rows" fix likely isn't needed anymore — there's no surviving caller to protect. Before spending more on it, worth confirming whether any current code path still parses pane text for idleness; I don't think one does.

Two more things, separately:
- The diff also bundles an unrelated stacked commit (`d98c2cc4` "Bundle the nightwatch fleet-babysitter skill (#1)") — the ~577-line `NightwatchSkillContent.swift`, `PluginDirWriter` changes, and their tests. That's inherited from the stale base, not part of this fix.

**Suggested path:** close this one out (the bug can't occur on `main` now that the detector is gone). If there's a real need for blank-row-robust pane parsing in some surviving path, let's rebase onto current `main` first and re-scope around the actual caller.

zionts:
Heads-up before this merges: this PR refines `checkIdle`, but #317 deliberately removed the entire idle text-matching system (`checkIdle`/`isIdle`/`isIdleConfirmed` + pattern constants) as brittle, and auto-suspend with it. Rebasing this onto main would mean reintroducing that removed machinery just to apply the blank-row trim — resurrecting code main intentionally dropped.

Recommend closing (or reframing against whatever replaces idle detection). Flagging rather than auto-merging despite approval — the approval predates #317 landing.

zionts:
Closing as superseded. This patched the pane-text idle heuristic (`checkIdle` / `busyIndicators` / `suffix(5)`), but `main` removed that whole approach in the idle-detection refactor (hibernation work, #344/#345). The fix targets code that no longer exists, and the trailing-blank-rows failure mode is moot under the new session/process-based idle detection. Nothing to rebase onto.

PR #324, opened 2026-07-03
Model profiles: login identity, one-click login, tbd profile CLI, keychain cleanup — plus daemon liveness/skew robustness

zionts:
## What's in here

Seven commits in three groups (all signed, reviewable commit-by-commit):

### Model-profile login UX (the multi-account story)
- **feat(profiles)**: the model-profile list RPC now reports each OAuth profile's logged-in identity (email from its isolated config dir) plus its config-dir path; deleting a profile also removes its Claude Code OAuth credential from the login keychain (service names are always path-hash-suffixed, so the user's default `~/.claude` credential is structurally untouchable).
- **feat(cli)**: `tbd profile list / set-default / login`. `login <name>` provisions the profile's config dir via a new `modelProfile.prepareConfigDir` RPC, scrubs poison env (`ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_BASE_URL`, `CLAUDE_CODE_USE_BEDROCK`…), and execs `claude` under the right `CLAUDE_CONFIG_DIR` — eliminating the logged-into-the-wrong-keychain trap.
- **feat(app)**: Settings rows show "Logged in as <email>" / "No login detected — run /login once" with a one-click "Open login session" action; the "+" tab menu and menu-bar switcher show per-profile identity; creating an OAuth profile ends with an "Open login session" step.

### Daemon lifecycle robustness
- **fix(app)**: after a reboot, a stale pid file whose pid got recycled made the app believe the daemon was alive (never spawning one, connecting to a dead socket forever). Liveness now verifies the pid's executable really is TBDDaemon (proc_pidpath), cleans stale pid/socket files, and the reconnect poll can route into the spawn path. Also: a build-identity handshake — the app compares the daemon's executablePath (already in the status RPC) against its own expected sibling and shows a persistent banner when the running daemon comes from a different build/worktree, instead of silent "Unknown method" weirdness.

### Earlier fixes riding along
- **feat(daemon)**: tombstone forgotten worktrees so reconcile stops re-adopting them.
- **fix(cli)**: replace crashing `%s` String(format:) table rows with a safe padding helper.
- Bundle the nightwatch fleet-babysitter skill into the TBD plugin.

## Testing
`swift test`: 2035 tests / 240 suites, 1 failure — the pre-existing environment-dependent `agentExecutableAvailability_detectsHomeLocalBinFallback` (fails on any machine with `/opt/homebrew/bin/claude` present; fails identically on upstream main here). `swiftlint --strict` clean. Rebased onto current main (one trivial init-order conflict with the new `controlMode` wiring, resolved by keeping both).

### DIFF
--- a/Sources/TBDApp/AppState+ModelProfiles.swift
+++ b/Sources/TBDApp/AppState+ModelProfiles.swift
@@ -286,6 +286,25 @@ extension AppState {
         }
     }
 
+    /// Open a Claude session pinned to `profileID` in the currently selected
+    /// worktree so the user can run `/login` there (the profile's isolated
+    /// config dir captures the credentials). Returns false and shows a
+    /// non-error alert when no worktree is selected — callers normally disable
+    /// the affordance in that case, so this is a fallback.
+    ///
+    /// Refreshes the profile list afterwards; the completed login itself is
+    /// picked up later via the daemon's `modelProfilesChanged` delta.
+    @discardableResult
+    func openLoginSession(profileID: UUID) async -> Bool {
+        guard let worktree = selectedWorktree else {
+            showAlert("Select a worktree first, then open a login session.", isError: false)
+            return false
+        }
+        await createClaudeTerminal(worktreeID: worktree.id, profileID: profileID)
+        await loadModelProfiles()
+        return true
+    }
+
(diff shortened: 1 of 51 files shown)


cheapsteak (approved):
🙏🙏

PR #337, opened 2026-07-04
Accounts: per-profile usage snapshots + spawn-time account picker with session tooltips

zionts:
Builds on #324 (merged). Two commits:

**feat(daemon): per-profile OAuth usage snapshots** — the daemon polls every logged-in profile's usage (~90s sweep, staggered) via the oauth usage endpoint (token read per profile through the security CLI; the daemon can't SecItemCopyMatching Claude Code's ACL'd items). Parses the `limits[]` buckets generically: `session` (5h window), `weekly_all`, `weekly_scoped` per model family (e.g. Fable) — unknown kinds flow through as data. Snapshots ride the model-profile list RPC (decode-compatible optional field) + a `modelProfile.usageRefresh` RPC for on-demand freshness. `tbd profile list` grows 5H/RESET/WK columns + dynamic per-family columns + `--refresh`.

**feat(app): spawn-time account picker** — the account is chosen per session, with the data in view, and each session stays pinned to its account for life (no more one-account clobbering → mass restart → uncached-cost burst):
- "+" menu profile items show compact usage (`Gmail — g@x.co · 5h 0% ↺23:10 · wk 76% · F 100%`) + a "Choose account…" dialog item
- plain Claude opens the picker (severity-colored bars per bucket, healthiest-window-expiring-soonest sort, refresh-on-open, "use default without asking" escape persisted in defaults)
- hover tooltips: Claude tabs show account/usage/spawn-time (ambient legacy sessions marked "may drift on token refresh"), worktree rows show a deduped account list

Testing: full suite 2,235 tests green locally (one known env-dependent failure, `agentExecutableAvailability_detectsHomeLocalBinFallback`, fails on machines with /opt/homebrew/bin/claude — also fails on main). swiftlint --strict clean. Live-verified on a 3-account setup: distinct per-account numbers, picker spawn pins correctly.

### DIFF
--- a/Sources/TBDApp/AccountPickerSheet.swift
+++ b/Sources/TBDApp/AccountPickerSheet.swift
@@ -1,0 +1,225 @@
+import SwiftUI
+import TBDShared
+
+/// Spawn-time account picker: the sheet shown when the user clicks the plain
+/// "Claude" action (unless "Use default without asking" is on) or the
+/// "Choose account…" item in the "+" menu.
+///
+/// Lists every model profile with full usage data — identity, 5h window bar +
+/// reset time, weekly-all bar, per-family bars — sorted
+/// healthiest-session-window first (display order only; nothing is
+/// auto-picked). Cached snapshots render immediately; a forced daemon-side
+/// usage sweep runs on open and updates rows in place ("refreshing…", never a
+/// blocking spinner). Selecting a row spawns a Claude session pinned to that
+/// profile for life.
+struct AccountPickerSheet: View {
+    @EnvironmentObject var appState: AppState
+    @Environment(\.dismiss) private var dismiss
+
+    /// Called with the chosen profile id; the caller owns the actual spawn.
+    let onPick: (UUID) -> Void
+
+    @State private var isRefreshing = false
+
+    private var sortedEntries: [ModelProfileWithUsage] {
+        ProfileUsagePresentation.sortedForPicker(appState.modelProfiles)
+    }
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 12) {
+            header
+
+            if sortedEntries.isEmpty {
+                Text("No model profiles configured — add one in Settings → Model Profiles.")
+                    .font(.callout)
+                    .foregroundStyle(.secondary)
+                    .frame(maxWidth: .infinity, minHeight: 80)
+            } else {
+                ScrollView {
+                    VStack(spacing: 4) {
+                        ForEach(sortedEntries, id: \.profile.id) { entry in
+                            AccountPickerRow(
+                                entry: entry,
+                                isDefault: entry.profile.id == appState.defaultProfileID,
+                                onPick: {
+                                    onPick(entry.profile.id)
+                                    dismiss()
+                                }
+                            )
+                        }
+                    }
+                }
+                .frame(maxHeight: 380)
+            }
+
+            Divider()
+
+            HStack {
+                Toggle("Use default without asking", isOn: $appState.skipAccountPicker)
+                    .font(.caption)
+                    .help("Skip this dialog: plain Claude spawns on the global default profile. Toggle back here or in Settings → Model Profiles.")
+                Spacer()
+                Button("Cancel") { dismiss() }
+                    .keyboardShortcut(.cancelAction)
+            }
+        }
+        .padding(20)
+        .frame(width: 480)
+        .task {
+            // Decision-grade data: force a sweep on open. Cached rows are
+            // already on screen; fresh snapshots merge in when they land.
+            isRefreshing = true
+            await appState.refreshUsageSnapshots()
+            isRefreshing = false
+        }
+    }
+
+    private var header: some View {
+        HStack {
+            Text("Choose account for Claude")
+                .font(.headline)
+            Spacer()
+            if isRefreshing {
+                HStack(spacing: 5) {
+                    ProgressView()
+                        .controlSize(.small)
+                    Text("refreshing…")
+                        .font(.caption)
+                        .foregroundStyle(.secondary)
+                }
+            }
+        }
+    }
+}
+
+// MARK: - Row
+
+private struct AccountPickerRow: View {
+    let entry: ModelProfileWithUsage
+    let isDefault: Bool
+    let onPick: () -> Void
+
+    @State private var isHovering = false
+
+    private var isSelectable: Bool {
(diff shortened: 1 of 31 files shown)


cheapsteak (approved):

zionts (commented):
## Review Summary

**Verdict:** APPROVE ✓

This is solid work. The daemon-side poller is well-architected with proper isolation, error handling, and broadcast semantics. The RPC surface maintains backward compatibility. Tests are comprehensive and properly isolated. No blocking issues.

---

## Daemon Patterns ✓

1. **Actor isolation** — OAuthProfileUsagePoller is an actor; all state access goes through it, preventing races. The broadcast closure is `@Sendable`, safe across boundaries.

2. **No print() statements** — Uses os.Logger with proper subsystem/category throughout (oauthUsagePoller, profileUsageFetcher, modelProfileHandlers). Correct privacy annotations on dynamic values.

3. **Lifecycle** — start() and stop() are idempotent (Sources/TBDDaemon/Daemon.swift:510–522 for start, 553–555 for stop). Called in correct order: after HTTP server starts, stopped before other background tasks. ✓

4. **No NIO thread-safety issues** — All RPCRouter handlers are `async throws`, always called from an async context. The modelProfileList handler correctly awaits the actor's allSnapshots() method.

5. **No database migrations** — Usage snapshots are in-memory only; ProfileUsageSnapshot and ClaudeUsageLimitBucket are pure data structures with no persistence. Correct design choice.

---

## Poller Correctness ✓

1. **Lifecycle & cancellation** — runLoop() checks Task.isCancelled at each iteration (line 87). Error from sleeper is caught and swallowed (line 89). Both start and stop are safe to call multiple times.

2. **Per-profile error isolation** — One profile's HTTP 401/429/network error doesn't stop the sweep (lines 139–146). The record() method handles each status independently, marking stale data with timestamp + reason.

3. **Stale data retention** — After a fetch failure, previous buckets are kept (line 174: `previous?.buckets ?? []`). The status field includes timestamps: "stale since 2026-07-04T12:34:56Z; fetch failed: HTTP 401" (line 169). Picker can show outdated-but-real numbers to the user.

4. **Rate hygiene** — 90s cadence is reasonable for the OAuth endpoint. Per-profile stagger (2s between fetches in one sweep, line 141) avoids thundering herd. No explicit backoff on 429s, but the 90s cadence and stale-data fallback mean hammering is bounded. If this becomes an issue in production, exponential backoff per profile on 429 could be added.

5. **Broadcast optimization** — hasMeaningfulChange() compares buckets, fetchedAt, and status, but NOT lastAttemptAt (line 192). So a failing profile in a state of repeated identical failures broadcasts once, not every 90s. ✓

6. **Targeted sweeps** — sweepNow(profileID:) fetches only the requested profile (line 108), full sweeps prune ineligible ghosts (lines 133–135). Tested thoroughly.

---

## RPC Surface & Backward Compat ✓

1. **Optional field in payload** — ModelProfileWithUsage.usageSnapshot is optional with default nil (Sources/TBDShared/Models.swift:507–509). Old daemons won't send it; new clients decode as nil. Old clients ignore the field. ✓

2. **New RPC method** — modelProfile.usageRefresh has proper params and result structs (RPCProtocol.swift lines 485–503). Method is cleanly dispatched in RPCRouter.swift:263–264.

3. **No hot-path bloat** — modelProfileList includes snapshots (ModelProfileHandlers.swift:32, 50), but ProfileUsageSnapshot is lightweight: just buckets array + two timestamps + status string. Not a problem.

4. **Strong typing** — ClaudeUsageLimitBucket.kind is String (to handle future API additions), but all enum-shaped values (CredentialKind, ModelProfile.kind, etc.) are proper enums. No raw strings past RPC boundaries. ✓

5. **RPC compatibility tests** — Verify old clients decode new payloads (OAuthProfileUsagePollerTests.swift:273–325). Unknown extra fields are ignored, missing fields default to nil. ✓

---

## Keychain & Credential Handling ✓

- **No direct SecItemCopyMatching** — Uses `/usr/bin/security find-generic-password` instead (ProfileUsageFetcher.swift:77–109). Claude Code's keychain items are ACL'd to Claude Code only; the security CLI is authorized. Verified empirically in the commit message (2026-07-03).

- **No secrets in argv** — Token is read via security CLI pipes, not passed to external processes. Fallback to .credentials.json is only used when keychain is unavailable (line 64).

- **Error messages scrub tokens** — ClaudeOAuthTokenReadError and ProfileUsageFetchStatus.failureReason are human-readable without leaking credential bytes. ✓

---

## Tests ✓

**OAuthProfileUsagePollerTests.swift** (340 lines):
- Sweep population: only OAuth + logged-in profiles ✓
- Per-profile failure isolation ✓
- Stale data retention on retry ✓
- Empty sweep when no profiles logged in ✓
- Broadcast-only-on-change ✓
- Targeted vs full sweeps ✓
- Profile pruning (dropped logins) ✓
- Provider error leaves snapshots intact ✓
- RPC decode compat (old clients, new payloads) ✓
- Result round-tripping ✓

**ProfileUsageFetcherTests.swift** (300+ lines):
- Captured real 2026-07-03 response parses correctly ✓
- Unknown bucket kinds flow through as-data ✓
- Null resetsAt → nil Date ✓
- Malformed entries skipped, valid ones survive ✓
- Fallback to legacy five_hour/seven_day windows ✓
- All using in-memory fixtures, no TBD_HOME touches ✓

All tests use proper isolation (DispatchQueue for thread-safety in doubles, in-memory profiles provider, fake fetcher). Zero reliance on ~/tbd.

---

## Non-Blocking Observations

1. **Broadcast is broad** — sweepNow() triggers `broadcast: .modelProfilesChanged`, which is a full model-profile-list delta. This resyncs all app windows showing profiles. Could be fine-grained per-profile (e.g., `.modelProfileUsageChanged(id)`), but the current approach is simpler and correct.

2. **No retry backoff on transient errors** — networkError and other non-401/429 failures still retry every 90s. Given the cadence and stale-data fallback, this is fine. If future monitoring shows thundering herd, add exponential backoff.

3. **Future: in-place account switch** — PR description mentions an additional commit may add transcript carry + tmux respawn-in-same-window. The current code doesn't block it. When that lands, verify RPCRouter+TerminalHandlers transaction safety around the window respawn.

---

## Verdict

No blocking issues. Daemon patterns are clean, poller is solid, backward compat is intact, and tests are comprehensive. This is production-ready.

🤖 Generated with Claude Code

PR #340, opened 2026-07-04
fix(app): scratch spaces show "Worktree not found" on first-run selection

zionts:
## Symptom

Creating a scratch space (sidebar **Scratch** section → **+**, or the "Click to
create scratch space" empty-state CTA) immediately renders **"Worktree not
found"** in the detail pane. Selecting an existing scratch row shows the same
error. Hit on first run on a real machine.

## Root cause

Scratch spaces are repo-less worktrees stored **only** in
`AppState.scratchWorktrees` — they are never inserted into the repo-grouped
`AppState.worktrees` dictionary (`AppState.updateState` routes `repoID == nil`
rows into `scratchWorktrees`).

But every worktree resolver on the selection path searched **only**
`worktrees.values`:

- `SingleWorktreeView.worktree` (`Sources/TBDApp/Terminal/TerminalContainerView.swift`) — the one that renders **"Worktree not found"** when it resolves to `nil`
- `MultiWorktreeCell.worktree` (same file)
- `AppState.findWorktree(id:)` (`Sources/TBDApp/AppState+Worktrees.swift`) — used by the PR split-button toolbar and terminal-title resolution
- `ContentView.selectedWorktree` (`Sources/TBDApp/ContentView.swift`) — file panel

`AppState.createScratch()` appends the new row to `scratchWorktrees` and then
sets `selectedWorktreeIDs = [wt.id]`. That selection routes the detail pane to
`SingleWorktreeView`, whose `worktree` lookup misses (the `worktrees` dict has
no scratch rows) → `nil` → **"Worktree not found"**, deterministically, the
instant the scratch space is created and auto-selected.

The daemon side is correct: `scratch.create` inserts a real `repoID = NULL`
worktree row and broadcasts `.worktreeCreated`. The bug is purely client-side
worktree resolution.

## Fix

Make `AppState.findWorktree(id:)` the single source of truth: after checking
the repo-grouped dict, fall back to `scratchWorktrees`. Route the three
view-local resolvers (`SingleWorktreeView`, `MultiWorktreeCell`,
`ContentView.selectedWorktree`) through `findWorktree` instead of each
re-implementing the repo-only loop, so scratch spaces resolve everywhere the
selection is consumed.

## Test

Adds `Tests/TBDAppTests/FindWorktreeScratchTests.swift` (4 cases): a scratch
space resolves when it lives only in `scratchWorktrees`; repo-scoped worktrees
still resolve; both resolve when the repo dict is also populated; an unknown id
returns `nil`. Verified the scratch case **fails** against the pre-fix
`findWorktree` and passes with it.

`swift build` clean; filtered `swift test` green; `swiftlint --strict` clean.

### DIFF
--- a/Sources/TBDApp/AppState+Worktrees.swift
+++ b/Sources/TBDApp/AppState+Worktrees.swift
@@ -543,12 +543,17 @@ extension AppState {
             .sorted { $0.sortOrder < $1.sortOrder }
     }
 
-    /// Find a worktree by id across all repos.
+    /// Find a worktree by id across all repos, including repo-less scratch
+    /// spaces. Scratch spaces live only in `scratchWorktrees`, never in the
+    /// repo-grouped `worktrees` dict, so a lookup that consulted `worktrees`
+    /// alone returned nil for a selected scratch space — surfacing as
+    /// "Worktree not found" the instant `createScratch()` auto-selected its
+    /// freshly created row (see `SingleWorktreeView`).
     func findWorktree(id: UUID) -> Worktree? {
         for (_, rows) in worktrees {
             if let wt = rows.first(where: { $0.id == id }) { return wt }
         }
-        return nil
+        return scratchWorktrees.first(where: { $0.id == id })
     }
 
     /// Resolve the effective auto-archive-on-merge setting for a worktree.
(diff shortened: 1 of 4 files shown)


cheapsteak (approved):
🙏 🙏

PR #345, opened 2026-07-06
Unify suspend + hibernation into one parked state (+ cold profile swap)

zionts:
Collapses the two overlapping "park a Claude session" features into one. `SuspendResumeCoordinator` (the older manual Suspend, `pause.circle` icon) and `HibernationCoordinator` (the auto-idle feature from #344, `moon.zzz.fill`) were doing the same job two ways — Adam spotted the duplicate parked states in the UI.

## The unified feature — hibernation's brain, suspend's mechanics

- **Park sequence**: capture ANSI pane snapshot first → polite in-band `/exit` (Claude flushes its transcript, tears down MCP children, fires Stop hooks) with the verify-exit poll → fall back to graceful interrupt → SIGTERM + transcript-tail validity check only if `/exit` doesn't kill it. Post-kill shell-survival + orphan-logging in both branches (each branch has its own test).
- **While parked**: the app shows the pane snapshot (suspend's restore fidelity) under the quiet `moon.zzz.fill` whisper. Single parked indicator; `pause` icon and Suspend/Resume verbs retired.
- **Wake**: hibernation's singleflight wake-on-focus + explicit Wake — one resume code path for auto-parked, legacy-suspended, and died-on-their-own sessions.
- Auto-hibernate master switch + idle stepper unchanged and still first-class; **manual "Hibernate now" works while auto is off** (tested manual-only mode).

## Data model
`hibernatedAt` is authoritative (all new writes); `suspendedAt` is read-only legacy, honored so old rows stay parked. `Terminal.isParked` spans both; wake clears both. **Migration v40** backfills legacy-suspended rows into `hibernatedAt` (data-only, idempotent). Decode-compatible.

## Cold profile swap (fleet rebalancing)
Reassigning a **parked** session to another account no longer wakes it: transcript carry into the destination config dir + `profile_id` update + `TerminalProfileDelta` broadcast, with **no respawn** — the next wake resumes under the new profile. Awake sessions keep today's interrupt+respawn. New CLI `tbd terminal swap-profile --terminal <id> --profile <name|uuid|ambient>` (cold when parked, respawn when awake) makes bulk rebalancing scriptable.

## Verification
[...]

PR #353, opened 2026-07-06
Wake-hang fix: bound tmux/git subprocess timeout + stop focus-wake storm

zionts:
> **Stacks on #345** — its commits appear in this diff until #345 merges, after which this rebases to just the wake-fix commit (`01e3994`). Review that top commit.

Fixes the "recv timed out after 300s (daemon unresponsive)" wake error that looped every ~5 min. The daemon was never dead (served other RPCs in ms). Two causes:

1. **Unbounded subprocesses** — `runTmux`/git awaited external commands with no timeout (a `git fetch` was found stuck 95 min); a stalled `respawn-window` hung the wake RPC to the 300s ceiling. Now timeout-bounded → fails fast into the existing retry-safe catch.
2. **Focus-wake storm** — focusing a worktree woke *every* hibernated session in it (~23 on one tmux server) at once. Now wakes only the focused session / throttles.

Verified compiling + running in the combined integration build.

### DIFF
--- a/Sources/TBDApp/AppState+Hibernation.swift
+++ b/Sources/TBDApp/AppState+Hibernation.swift
@@ -77,16 +77,40 @@ extension AppState {
         }
     }
 
-    /// Auto-wake any PARKED Claude terminals in a worktree the user just
-    /// focused/selected. Covers both hibernated (authoritative) and legacy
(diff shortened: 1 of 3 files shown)


zionts:
Superseded by #367 (same-repo, rebased onto main now that #362 unify landed).

PR #358, opened 2026-07-06
feat(nightwatch): daywatch autonomous runner

zionts:
## What this ships

The **plumbing foundation** for daywatch/nightwatch: a `DaywatchRunner` actor (mode→loop, idempotent `apply`, boot-reconcile, shutdown), the `DaywatchExecuting` protocol seam, DB-persisted `nightwatchMode`, and RPC wiring.

## ⚠️ Actuation is DEFERRED (not live in this PR)

`ProcessDaywatchExecutor.wakeJudge(act:)` is an **intentional no-op** that only logs — it does **not** spawn a judge or fire actions. The tick loop sweeps + records; judgment actuation (the visible watchable worker: Sonnet for daywatch, Opus for nightwatch) lands in **Phase A**. So today this is a *tick-sweep + logging loop*, not autonomous actuation. The `act` flag is plumbed through for Phase A but currently only changes a log line.

## Notes
- `tick.py` guards against concurrent runs across both the in-daemon loop and the legacy `scheduler.sh` launchd heartbeat via a machine-global `/tmp/nightwatch-tick.lock`.
- `runTick` has a 5-minute deadline; boot-reconcile logs on failure.
- Also carries reconcile-safety so a daemon restart re-adopts live sessions instead of parking them.

### DIFF
--- a/Sources/TBDDaemon/Nightwatch/DaywatchRunner.swift
+++ b/Sources/TBDDaemon/Nightwatch/DaywatchRunner.swift
@@ -1,0 +1,167 @@
+import Foundation
+import os
+import TBDShared
+
+private let logger = Logger(subsystem: "com.tbd.daemon", category: "nightwatch.daywatch")
+
+/// Protocol for injecting executor behavior (tick + judge wake).
+/// Allows tests to stub out real subprocess execution.
+public protocol DaywatchExecuting: Sendable {
+    /// Run one tick cycle; return exit code (0 = silent-ok, 10 = judgment queued).
+    func runTick() async -> Int32
+    /// Wake the judge to process queued decisions.
+    /// `act` = true: auto-fire actions (nightwatch mode); false: batch/surface only (daywatch mode).
+    func wakeJudge(act: Bool) async
+}
+
+/// Real executor: runs tick.py and judge.py via Process.
+public struct ProcessDaywatchExecutor: DaywatchExecuting {
+    private let skillDir: String
+
+    public init(skillDir: String) {
+        self.skillDir = skillDir
+    }
+
+    /// Run tick.py; return exit code.
+    public func runTick() async -> Int32 {
+        let tickPath = skillDir + "/scripts/tick.py"
+        let process = Process()
+        process.executableURL = URL(fileURLWithPath: "/usr/bin/python3")
+        process.arguments = [tickPath]
+        process.standardOutput = FileHandle.nullDevice
+        process.standardError = FileHandle.nullDevice
+
+        do {
+            try process.run()
+            process.waitUntilExit()
+            return process.terminationStatus
+        } catch {
+            logger.error("Failed to run tick.py: \(error.localizedDescription, privacy: .public)")
+            return -1
+        }
+    }
+
+    /// Wake the judge: spawn `claude --model claude-sonnet-5 <prompt>`.
+    /// Safe no-op if `claude` binary is not found.
+    public func wakeJudge(act: Bool) async {
+        let prompt = act
+            ? "Run the nightwatch judge: process queue/decisions.jsonl and fire approved actions."
+            : "Run the nightwatch judge: process queue/decisions.jsonl and batch to for-adam.md (act=false)."
+
+        let process = Process()
+        process.executableURL = URL(fileURLWithPath: "/usr/bin/which")
+        process.arguments = ["claude"]
+        let pipe = Pipe()
+        process.standardOutput = pipe
+
+        do {
+            try process.run()
+            process.waitUntilExit()
+            let data = pipe.fileHandleForReading.readDataToEndOfFile()
+            guard let claudePath = String(data: data, encoding: .utf8)?.trimmingCharacters(in: .whitespacesAndNewlines),
+                  !claudePath.isEmpty else {
+                logger.debug("claude binary not found; judge wake is a no-op")
+                return
+            }
+
+            let judge = Process()
+            judge.executableURL = URL(fileURLWithPath: claudePath)
+            judge.arguments = ["-p", prompt, "--model", "claude-sonnet-5"]
+            judge.standardOutput = FileHandle.nullDevice
+            judge.standardError = FileHandle.nullDevice
+
+            try judge.run()
+            // Don't wait — let it run detached.
+        } catch {
+            logger.debug("Failed to wake judge: \(error.localizedDescription, privacy: .public)")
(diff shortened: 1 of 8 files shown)


cheapsteak (dismissed):
(withdrawn — re-issued as a non-blocking comment review below)

cheapsteak (commented):
Solid skeleton — the `DaywatchExecuting` injection seam, the actor mirroring the poller pattern, boot-reconcile + shutdown wiring, and idempotent `apply` are all right. But I think the judge spawn as written won't actually work, and the loop has a couple of robustness gaps worth fixing before this runs unattended overnight.

### Major

**1. The spawned judge session likely can't do its job** (`DaywatchRunner.swift`, `wakeJudge`)

Three compounding problems with the bare `claude -p <prompt> --model claude-sonnet-5` invocation:

- **No `--plugin-dir`.** Every other TBD-initiated claude spawn goes through `ClaudeSpawnCommandBuilder`, which passes `--plugin-dir <plugin>` so the bundled skills load. The judge gets none of that — unless the user has separately installed the plugin globally, the session has no `nightwatch` skill and no idea what "run the nightwatch judge" means.
- **No working directory.** The prompt says `process queue/decisions.jsonl`, but the judge inherits the daemon's cwd. The real queue is at `<plugin>/skills/nightwatch/queue` (`tick.py` resolves it relative to its own file — `NightwatchSkillContent.swift:102-103`). The relative path in the prompt resolves nowhere useful. Set `judge.currentDirectoryURL` to the skill dir, or put the absolute path in the prompt.
- **No permission posture.** Headless `-p` can't answer tool-permission prompts, so the nightwatch judge ("fire approved actions") can't nudge, merge-enqueue, or write files. TBD's own spawns use `--dangerously-skip-permissions`; the judge needs a deliberate decision here.

Suggest reusing `ClaudeSpawnCommandBuilder` (or at minimum plugin-dir + cwd + an explicit permission choice) rather than hand-rolling the invocation.

**2. No overlap guard or backoff on judge spawns — a token-burn loop**

`wakeJudge` is fire-and-forget every tick. If the queue isn't drained (and per #1 it currently won't be), a fresh Sonnet session spawns every 15 minutes — ~96/day, silently. The design doc (§10.1) makes token conservatism a first-class constraint: "coming back to a burned budget is a failure mode on par with a bad merge." At minimum, hold the judge `Process` reference and skip the wake while the previous judge is still running; a cooldown/consecutive-wake cap would be better.

**3. `runTick` blocks a cooperative-pool thread with no timeout** (`ProcessDaywatchExecutor.runTick`)

`process.waitUntilExit()` inside an async context parks a Swift-concurrency cooperative thread for the subprocess's whole lifetime. `tick.py` sweeps every tmux pane — if tmux wedges, this blocks a pool thread *forever* and the loop silently stops ticking with no log. Use the `terminationHandler` + continuation pattern already established in the daemon (e.g. `PRStatusManager.swift:663`, `HookResolver.swift:111`) and add a deadline (kill + `.warning` log after N minutes).

### Medium

**4. Policy compiled into the mechanism layer, in duplicate.** The judge prompt, `for-adam.md`, `queue/decisions.jsonl`, and `claude-sonnet-5` are all policy per the design doc's own §2 split — and they now exist in two copies (editable skill + compiled Swift) that can silently drift: rename the file in the skill and the daemon's prompt points at the old name until a rebuild. Consider having the daemon invoke a skill-declared entry point (e.g. `scripts/judge.sh --act|--batch`) or reading the prompt/paths from the skill, so the daemon only knows "exit 10 → run the judge with the act flag." Separately, `for-adam.md` is a personal name shipping in a general-purpose daemon — `for-human.md`/`queue/held.md` in the skill would fix both copies.

**5. Boot-reconcile swallows failures** (`Daemon.swift`, step 12d). `try? await database.config.get()` — if the config read throws, nightwatch silently stays off after a restart, which is exactly the away-overnight scenario where silence hurts. Log the failure.

**6. `which claude` + `.debug`-only logging = invisible no-op.** A daemon started outside a login shell (launchd) has a minimal PATH; `claude` in `~/.local/bin` won't resolve, and the only signal is a `.debug` line that's off by default. The user flips Away on and nothing ever happens. Elevate not-found and spawn-failure to `.warning`, and consider surfacing "judge unavailable" through the tick report.

**7. Stop log prints the wrong mode** (`apply`). `currentMode = mode` runs before the stop branch, so "Stopped daywatch runner (was in mode …)" always logs `off`. Capture the previous mode before assignment.

### Minor

- `DaywatchRunner.defaultInterval` exists but the init default repeats `15 * 60` literally — use the constant so they can't drift.
- Sleep-first loop means the first tick lands 15 minutes *after* hitting Away. An immediate first tick seems more useful (the fleet state at departure is exactly what you want swept); if deferred is deliberate, comment why.
- PR body says "requires daemon restart to take effect," but `handleSetNightwatchMode` now applies the mode live — stale note?
- `/usr/bin/python3` hardcode is fine on macOS, but worth a comment that it's the CLT stub.

### Tests

- `FakeDaywatchExecutor.setTickExitCode` is an admitted no-op ("…accept this limitation") — delete it rather than shipping a method that does nothing.
- `testModeSwitching` doesn't actually switch modes on a live runner — it builds a second runner. The genuinely interesting branch is untested: daywatch→nightwatch *while running* takes the silent no-op path in `apply` and relies on `runLoop` reading `currentMode` fresh each tick to flip `act`. Per CLAUDE.md's mode-gate rule, that transition deserves a direct test on one runner.
- The 10ms-interval + 30–50ms-sleep timing assertions will be flake-prone under parallel CI load. Consider making the fake executor signal tick completions (continuation/AsyncStream) so tests await "N ticks happened" deterministically instead of sleeping.

Sources/TBDDaemon/Nightwatch/DaywatchRunner.swift
```diff
@@ -0,0 +1,167 @@
...
+            return -1
+        }
+    }
+
+    /// Wake the judge: spawn `claude --model claude-sonnet-5 <prompt>`.
+    /// Safe no-op if `claude` binary is not found.
+    public func wakeJudge(act: Bool) async {
+        let prompt = act
+            ? "Run the nightwatch judge: process queue/decisions.jsonl and fire approved actions."
+            : "Run the nightwatch judge: process queue/decisions.jsonl and batch to for-adam.md (act=false)."
```
cheapsteak: we have a `for-adam.md`?

Issue #318, opened 2026-07-02
tmux control mode: Phase A (input + resize + replay → usable) and Phase B (flow control + crash recovery → default-on)


cheapsteak:
Tracks the remaining work after #317 (Phases 1–2: daemon-owned `tmux -CC` connection, protocol parser, reader-thread `PaneFanout`, `SCM_RIGHTS` FD vending, single-pane read-only render — all behind `TBD_TMUX_CONTROL_MODE`).

**Source of truth:** `docs/specs/2026-05-17-tmux-control-mode-design.md`. All design decisions below are already made there; these are implementation phases.

## Phase A — spec phases 3+4+5 in one PR: the usability line

A control-mode pane becomes something you can actually work in. Ships together because each piece alone is not usable: input without resize renders fullscreen Claude at the server's allocation size; input+resize without replay means every tab/worktree switch (TBD recreates terminal views constantly) shows a blank pane until new output arrives.

- **Input (spec §Input path):** bytes-blob keystroke RPC → `send-keys -H`; threshold paste (≤4 KB argv, larger via `load-buffer`/`paste-buffer` *without* `-p` — the app owns bracketing); dedicated keystroke channel so `%output` floods can't queue-block typing; p99 keystroke latency telemetry from day one.
- **Resize (spec §Resize):** daemon as sole size arbiter — `window-size manual`, `resize-window -t @<id>` (`refresh-client -C` on 3.4+), outstanding-resize counter (iTerm2's `numOutstandingWindowResizes_` pattern) to prevent echoed-resize fights. `AttachRequestParams.windowID` (already on the wire) becomes load-bearing here.
- **Replay α (spec §Scrollback):** on attach, 4-command capture (main scrollback `capture-pane -peqJN -S -N`, alt screen `-a`, pending `-p -P -C`, cursor/mode state via `display-message -F`), written through the existing ready-gate in the spec's exact order (reset prelude → history → mode escapes → alt screen → pending → cursor → live). Server-side `history-limit 50000` before any `new-window`; truncation-telemetry counter from day one.

**Phase-2 learnings to carry forward:**
- The alt-screen replay branch is the *primary* path, not the exotic one — TBD spawns Claude with `CLAUDE_CODE_NO_FLICKER=1` (fullscreen renderer, alt-screen) by default. Test plan must include attaching to a fullscreen Claude session mid-stream.
[...]

PR #364, opened 2026-07-06
Feat: tmux control mode Phase A — make gate-on daily driving viable (input, resize, replay, settings toggle)

cheapsteak:
## Summary

Phase A of tmux control mode (part of #318): with the gate on, control-mode terminals are now daily-drivable — typing and pasting work with no perceptible lag, drag-resize reflows fullscreen apps, and tab-switching replays full scrollback with colors instead of blank panes. Specs: `docs/specs/2026-05-17-tmux-control-mode-design.md` + `docs/specs/2026-07-02-tmux-control-mode-phase-a-addendum.md` (corrected in this PR where live probing refuted it).

Everything rides the one `-CC` stream, ordered end-to-end:

- **M1 — command correlation**: FIFO `TmuxControlCommandClient` (order-based, numbers validated not matched — iTerm2's 13-year `TmuxGateway` design); protocol violations tear the connection down rather than limp desynced. Blank/whitespace command lines are rejected — live-probed as `%exit` detach and silent FIFO desync respectively.
- **M2 — input**: sidecar promoted to a bidirectional framed channel (`[u32 len][u8 type][payload]`, 4 MiB desync cap); keystrokes as `send-keys -H` chunked ≤330 bytes, tolerate-errors; single-consumer router so stream order == wire order == user order; p99 latency telemetry.
- **Paste (ruling v2)**: ALL pastes — any size — ship as one `.paste` frame → `load-buffer` + `paste-buffer -d -p`. tmux is the sole bracketing authority because stock tmux has no bracketed-paste format variable, so replay can't restore SwiftTerm's DECSET-2004 state. Oversize (>~4 MiB) is refused with a visible error log, never split (separate paste-buffer calls would bracket each chunk as its own paste).
- **M3 — resize**: daemon is sole size authority; `window-size manual` per window at attach; `resize-window` + `list-windows` echo fence with a per-window outstanding counter suppressing stale `%layout-change` echoes; app debounces 100 ms.
- **M4 — replay**: `attach.ready` triggers pause → atomic 4-command capture → byte assembly (full-clear reset prelude that supersedes the stale suspend-snapshot pre-feed) → generation-checked pre-ready write into the pane pipe → gate opens → unpause last. Ready-timeout re-keyed to an `acknowledged` flag so slow captures can't be killed by the stale timer. `history-limit 50000` chained before `new-session` so window 0 is covered.
- **M5 — settings toggle**: persisted `controlModeEnabled` (migration `v43_config_control_mode` — renumbered from v35 on rebase after main took v35–v42; never shipped under the old name), `config.setControlMode` RPC, `daemon.capabilities` carries `tmuxVersion`/`controlModeSupported`, gate = `(env || flag) && tmux ≥ 3.2` evaluated per decision. Settings → General → Experimental.
- **Input-failure indicator (#318 polish ruling)**: edge-triggered per-pane delivery health pushed over the existing StateDelta channel; passive strip on the pane, gated attached-and-failing, clears on recovery/detach.

### Live-verified findings (baked into spec + tests)

- **Alt-screen capture legs invert when `alternate_on=1`**: plain `capture-pane` returns the CURRENT (= alt) screen, `-a` the saved primary. The integration matrix caught this as a shipped bug — fullscreen Claude (the primary replay path) would have replayed backwards.
- **tmux pause DISCARDS, it does not drain** — the addendum's §3 drain claim was false on 3.6a. Accepted for Phase A as a ~0.2–0.5 s boundary-only, forward-only gap (never dup/reorder/corruption — pinned by tests; self-heals via tmux history on re-attach and alt-screen repaint). Both closing designs (scoped interleave buffer; iTerm2's recapture-on-%pause) are documented in-spec as held-in-reserve for Phase B.
- Review pass found one critical, fixed: failure-path detach was generation-unaware — a stale attach's late capture failure could close a healthy successor's pipe under fast tab-switching. `pane.detach` now echoes the attach generation (optional field, wire back-compat).

### Known gaps (deliberate, documented)

- Pre-existing tmux servers keep `history-limit 2000` until restarted → replay depth capped there.
- Mute-but-alive tmux hangs `attach.ready` (no per-command timeout) and fires no health edge — Phase B crash recovery owns wedged-connection detection. Same scope owns tmux-server-death races (recycled pane IDs vs queued router items — round-6 item 8) and the wedged-tmux write-block (a full pty buffer pins the correlator's cooperative thread until crash recovery stops the connection — round-10 item 2).
- Flow control is still Phase 2's EAGAIN drop-and-count; Policy B (pause/continue state machine) is Phase B.
- One daemon-wide input router serializes all panes (a slow paste on one pane delays keystrokes elsewhere until it completes) — per-server consumers were explicitly deferred to Phase B in the M2 round; ordering is per-pane by design, pastes are rare human events.
- Grouped-sessions fallback stays through Phase A — every control-mode failure degrades invisibly instead of breaking a terminal.

## Test plan

- [ ] CI: full suite green (2,138 tests / 257 suites locally at pre-rebase HEAD; post-rebase run in flight at PR-open time)
- [ ] Dogfood from this worktree: `scripts/restart.sh`, then Settings → General → Experimental → enable tmux control mode
- [ ] Type into a control-mode pane (no lag), paste multiline into fullscreen Claude immediately after a tab-switch (arrives bracketed, runs as one paste)
- [ ] Drag-resize a control-mode pane — fullscreen Claude reflows to the new width
- [ ] Tab-switch between worktrees — history replays with SGR colors, alt-screen Claude repaints coherently, no blank panes
- [ ] Kill a repo's tmux server while typing — passive "input not being delivered" strip appears (check both themes), clears on re-attach

Part of #318.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=7AC15C0F-2F51-43D4-B5DF-3031F5E43C6D)


cheapsteak:
Addressed the full review (thanks — all three Mediums were real):

| Finding | Fix |
|---|---|
| **M1** oversize paste silent to the user | In-pane message via the existing feed pattern, size + cap in human units; log kept. |
| **M2** stale generation's `continue` widens the discard gap | `attach.ready` now echoes the attach generation (optional param, wire back-compat). A stale ready (generation mismatch at acknowledge) sends **nothing** on the shared correlator — no pause, no capture, no unpause — so it can neither pause the pane after a successor finished nor unpause it mid-sequence. A mid-sequence supersede skips its unpause (FIFO puts the successor's own unpause last). New tests use ONE shared fake command client across both generations, closing exactly the coverage gap the review called out. Documented residual: if the superseding attach's ready never arrives, the pane stays paused until the next attach unpauses it — benign (no viewer; tmux buffers server-side). |
| **M3** app-side indicator state not generation-scoped | `controlModeAttachedPanes`/`controlModeFailingInputPanes` now store the attach generation; a stale cleanup only clears state belonging to its own generation. Race test added. |
| Minors | Failing-health entry cleared on `register()` too; deadline doc comments corrected (capture-wait is unbounded, only the replay write has the 5s deadline — mute-tmux remains the acknowledged Phase B gap); dedicated `sendPaste` frame test added. |

Also rebased onto latest main three times along the way — the migration is now `v45_config_control_mode` (upstream took v35–v44), and the Settings → Experimental section keeps both the new Nightwatch toggle and the control-mode toggle. Two test helpers were updated for the newly-in-scope `optional_data_string_conversion` lint rule.

cheapsteak:
Re the round-4 findings — H1, H2, and the pause-result Medium are all confirmed and fixes are in progress (H1 was live-probed: your read of the gap is right, but the literal suggestion doesn't work — `-S` is a no-op on the `-a` leg; the fix is a three-leg capture: pure-history `-S -50000 -E -1`, current screen, saved viewport, with a history_size guard because `-E -1` clamps to one visible row on empty history).

**Pushing back on one: the single-router serialization Medium.** This is real but deliberately deferred: per-server input consumers were explicitly proposed and deferred to Phase B in the M2 review round (recorded in the coordination channel at the time). Context that bounds the impact: (a) the ordering guarantee that motivates the single consumer is *per-pane* — a keystroke enqueued after a paste on the SAME pane must wait for it by design; cross-server serialization is incidental cost; (b) pastes are rare, human-initiated events, and the keystroke legs themselves are fire-and-forget stream writes; (c) the 16–71s stalls cited from the CI commit message were runner-wide I/O contention in the *test harness* (every suite stalled equally), not a measured production paste path. Phase B's flow-control work restructures exactly this area (per-server consumers + pause state machines), so building the split twice is worse than carrying the bounded gap briefly. It's now listed in the PR body's Known gaps. If dogfooding shows a real cross-pane input stall before Phase B, we'll pull the per-server split forward.

cheapsteak:
Round-4 findings addressed (5 commits):

- **H1** `96fc2350` — confirmed by live probe, with a correction to the suggested fix: `-S` on the `-a` leg is a **no-op** (still viewport-only). The capture is now three legs — pure scrollback (`-S -50000 -E -1 -q`), current screen, saved primary (`-a -q`) — with `#{history_size}` added to the pane-state format because the pure-history leg clamps to one visible row on empty history and must be discarded. Non-alt replay asserted **byte-identical** to the previous single-capture output; the live alt-screen matrix test now prints 40 numbered lines on a 24-row pane and asserts all 40 replay (10/10 looped).
- **H2** `e5c3cf19` — `isTornDown` on the coordinator, re-checked after every committing await with per-stage unwind (fd close / reader removal / generation-scoped `pane.detach`); the fallback branch also gates, and its previously-unconditional nil-generation teardown (which could have killed a successor's fresh attach) is fixed alongside. Per-stage undo table extracted and unit-tested; the Task-resumption wiring itself has no headless seam — documented residual.
- **M4** `ed75153e` — the pause leg's result is now inspected: failure is logged + surfaced via an injectable hook, then the replay proceeds (documented rationale: same self-healing class as the accepted pause-discard gap; a dead pane fails the captures anyway). Both branches tested.
- Minors `f0704503` — exact 4 MiB boundary tests (cap passes, cap+1 desyncs); delete-buffer cleanup failure now visible in logs.
- Drive-by `9ec9fe18` (not a review finding): pre-existing pipe-buffer deadlock in `HibernationCoordinator.detectOrphanedClaudeProcesses` — `waitUntilExit()` before draining the pipe wedges on >64 KiB `ps` tables (~1050 processes locally). Standard reorder fix; flagging since it touches code outside this PR's feature area but was blocking the full suite.

The single-router serialization Medium is the documented Phase-B deferral per the earlier comment. Full suite 2835/354 green ×5 locally, lint clean.

cheapsteak:
Round-5 findings addressed (5 commits, every one confirmed against source before fixing):

- **M1** `fbc48d3c` — teardown split: actor does bookkeeping only (identity guards preserved, pending sends failed eagerly via `connectionClosed()`), the blocking `stop()` runs on the caller's task per the same rule `writeReplay` documents. Red-green test: an unrelated supervisor call completes while a held-open teardown is mid-stop; successor creation mid-stop covered.
- **M2** `7212a932` — per-connection epoch (lock-boxed, checked before every frame delivery on the reader thread); a superseded thread drops its frame, logs once, and exits. Red run reproduced the exact stale-frame interleave; green after.
- **M3** `d4518381` — latest-wins resize serialization: at most one in-flight RPC, ticks during flight stash (overwriting), completion drains until quiescent. Debounce timing and gate checks unchanged. Item 6's `MainActor.assumeIsolated` guard folded in (same lines).
- **Item 4** `aaccc028` — duplicate same-generation `attach.ready` now returns `.alreadyAcknowledged` → benign RPC success, zero commands, no concurrent replay into the pipe.
- **Item 5** `ce9890b4` — took the pre-existing `sendCommand` short-write flag this round since it's load-bearing for the correlator: writes now go through `FDChannel.sendData`'s full-write/EINTR loop; unrecoverable errors drive the existing termination→drain-cleanup path. Honest coverage note: no headless pty seam exists, so the test is a 200-command ~15KB single-write batch asserting ordered correlation (exercises the loop; can't deterministically force a short write on a blocking pty).

Full suite 2844/356 green, lint clean.

cheapsteak:
Round-6 ack: items 1–7 confirmed on inspection and fixes are in progress (each as its own commit, with the reviewer's exact race reproduced red-green where a seam exists).

**Pushing back on item 8 (router generation/epoch):** the concrete trigger you identified — tmux server death/restart recycling pane IDs while a backlogged item sits in the queue — is precisely the Phase B crash-recovery scope (Mode B: the daemon marks every pane on that server dead and prompts to recreate; the route registry is rebuilt through fresh attaches at that point). Pattern-consistency is a fair observation, but stamping enqueue-time epochs on a queue whose staleness window is measured in milliseconds — against a failure mode Phase B redesigns wholesale — is scaffolding we'd tear out. It's noted alongside the existing Phase B crash-recovery entry in Known gaps. As with the router-serialization deferral: if dogfooding surfaces a real cross-restart misdelivery before Phase B, we'll pull it forward.

cheapsteak:
Round-6 items 1–7 addressed (7 commits, every finding confirmed against source; none refuted):

- **H1** `8c20dc8b` — `markReady(key:generation:)` now generation-checked like every sibling; a refused markReady maps to `.superseded` in the orchestrator (skips unpause, successor owns the gate). Red-green: sink swapped between `writeReplay` and `markReady` → gate stays closed, successor's own sequence opens it. No nil-generation variant kept — every caller has the generation, an unchecked form would only preserve the hazard.
- **H2** `ed10483d` — the fd-vend timeout throw now carries the minted generation (`AttachFDVendError`); the teardown scope resolves committed → error-carried → nil via a pure, tested decision function. Only a failure before `attach.request` succeeded uses the unconditional path.
- **H3** `2a15127f` — drag-and-drop text rides the same `PasteInterception` path as pasteboard pastes (detached behavior byte-identical), and `sendInput` refuses any frame over the scanner cap as defense-in-depth — wire test proves the oversize frame never reaches the socket and a following frame still arrives.
- **M4** `7f9fb73a` — `stoppingServers` set: `ensureConnection` suspends (actor stays free) until an in-flight teardown's `stop()` completes; red run reproduced two live connections, green has exactly one successor. The prior round's test tail that pinned the racy behavior was rewritten.
- **M5** `55caedca` — one shared `ControlModeGeometry` floor (20×5) used by the app gate and the daemon clamp; below-floor panels keep tmux at the floor (documented residual, self-heals on the next in-range resize).
- **M6** `0272c797` — the resize drain loop re-checks teardown before every send via a pure gate.
- **M7** `b121679d` — health deltas carry the attach generation end-to-end (router routes store it, events stamp it, wire field optional for back-compat); a stale-generation failing delta can no longer flag a fresh attach; recovery clears regardless.

Item 8 remains the documented Phase B crash-recovery deferral per the prior comment. Full suite 2863/357 green, lint clean.

cheapsteak:
Round-7 findings addressed (3 commits, both confirmed + the minor):

- **H1** `fe8400d8` — `stop()` now signals the child BEFORE taking `ioLock` (the process handle is immutable after `start()`, so termination is lock-free); the SIGTERM→500ms→SIGKILL escalation is unchanged, and fd retirement still happens under the lock afterward. Red run reproduced the full failure chain you described: a real wedged 1MB pty write deadlocked `stop()` for 271s and left `ensureConnection` suspended past 30s; green has both under 0.4s with the writer released and a successor created. Bonus finding while verifying your SIGPIPE question: pty master writes return EIO (no signal), but the daemon's raw `sendmsg` to the sidecar socket WAS a process-killing SIGPIPE exposure on peer close — now `SIG_IGN` in main.swift so it fails into the existing error path.
- **M1** `d33dd79a` — per-window monotonic sequence stamped at entry under the lock; a call superseded during the provider hop drops (never increments, fence bookkeeping balanced; stamps cleared when spent). Red run reproduced the stale `80x24` landing after `120x40`; green sends only the newer geometry. Documented the far-narrower residual window between re-check and enqueue.
- **Minor** `e07045c1` — the `.modelProfilesChanged` delta handler now refreshes `daemonCapabilities` via an injectable fetcher. One deliberate deviation, documented: a failed refresh keeps the last known value rather than nil-overwriting — this delta fires for many config changes and a transient RPC failure must not silently flip new panes off control mode.

Full suite 2868/358 green, lint clean.

cheapsteak:
Round-8 addressed in full:

- **High (branch staleness)** — rebased onto latest main; main's #371 (the broader `PipeDataAccumulator` fix across `TmuxManager`/`GitManager`/`HibernationCoordinator` + its test files) is adopted wholesale, and this branch's narrower drive-by reorder dropped out of history as superseded. Verified: `764dd201` is now an ancestor, both managers use the accumulator, `HibernationOrphanDetectionTests` present.
- **M1** `1ed4d29b` — the Settings call site now uses `refreshDaemonCapabilities()` (last-known-value on transient failure); the toggle can no longer snap to "Requires tmux 3.2" after a successful set. Both branches tested.
- **M2** `edc84219` — `stopAll` bookkeeping stays on the actor (pending sends failed eagerly, servers marked stopping for R6-M4 consistency); the blocking stops run via GCD `concurrentPerform` behind a continuation — deliberately NOT a task group, whose closures would occupy the fixed-width cooperative pool and starve the daemon (we reproduced exactly that as a full test-runner wedge while building this). Red-green: unrelated actor calls complete while two stops are held open; a racing `ensureConnection` parks and resumes to exactly one successor.
- **M3** `b42d97f8` — the capture batch gains a combined history+screen leg; non-alt replays use it verbatim so tmux's `-J` joins soft wraps across the scrollback/screen seam (byte-identity with the legacy single capture pinned by test). The alt case keeps the split legs — no combined capture reaches both regions during alt mode — with the seam residual documented as accepted (cosmetic, exact-alignment-only). The alt branch ignoring the combined leg is itself pinned by a new assertion.

Full suite 2886/359 green, lint clean.

cheapsteak:
Round-9 both findings addressed in one commit (same file, coupled fix):

- **H1** — extracted `stopOffPool` (GCD global queue behind a continuation) and routed `teardownConnection`'s blocking stop through it, completing the R8-M2 discipline on the single-connection path. Red-green evidence via the stop seam capturing its dispatch queue label: pre-fix it reads `com.apple.root.default-qos.cooperative`, post-fix a GCD global label.
- **M1** — `drain`'s natural end-of-stream cleanup now stops the connection it still owns (releasing the pty master fd); ownership at stream end decides who stops, so the fatal/stopAll paths keep their single stop and the handoff can't race (all decided on the actor). Red-green: a stub that exits on its own never reached `stop()` pre-fix.

Both new tests verified red against the pre-fix code before landing. The `beginStopAll` snapshot note is acknowledged as the code's own documented bound (shutdown-only caller). Full suite 2888/359 green ×2 consecutive, lint clean.

cheapsteak:
Round-10 ack: findings 1, 3, 4 and the `sendFD` minor are confirmed and being fixed (fixes in flight, one commit each).

**Deferring item 2 (blocking pty write on the correlator actor's thread) into the existing Phase-B mute-tmux gap**, with reasoning: the write only blocks when tmux is wedged-but-alive with a full pty buffer — the exact scenario already documented in Known gaps as Phase B crash-recovery's scope (a wedged tmux also hangs `attach.ready` and fires no health edge; detection + rebuild is the owning fix). Until that scenario occurs, the send path never blocks (pty buffer absorbs keystroke-scale writes), and while it occurs, control mode for that server is wholly dead regardless — the pinned thread is a secondary symptom of a state Phase B must detect anyway. Restructuring the keystroke hot path (per-connection write queues, changed sendList return semantics, latency-telemetry window redefinition) to mitigate a secondary symptom of an already-deferred failure mode is scaffolding Phase B would rework. Known gaps updated to name the thread-pinning consequence explicitly so the deferral is auditable.

cheapsteak:
Round-10 items 1, 3, 4 and the `sendFD` minor addressed (item 2's deferral per the prior comment):

- **R10-1** `c5a1f94f` — the torn-down catch branch resolves `teardownGeneration` and sends the generation-scoped `paneDetach` before returning; nothing leaks when a view dies mid-`openAttach`.
- **R10-3** `ab798546` — read-only `currentGeneration` re-check immediately before the sendList (post-hop pattern from R7-M1); a stale task preempted at the provider hop sends nothing.
- **R10-4** `2c50ee12` — reader-registry removal is generation-scoped with the same nil-cannot-discriminate convention as the AppState bookkeeping; all four call sites pass their attach's generation.
- **Minor** `aab0930f` — `sendFD` sends any short-write remainder through the `sendData` loop.

Process note for this round: test execution moved to CI as the source of truth — local runs on this machine are being serialized/wedged by an environmental issue under separate investigation (branch forthcoming), so these commits are build+lint-verified locally with the new tests riding this push for CI execution. If CI's test job flags anything, fixes will follow immediately.

cheapsteak:
Round-11 fix: `0d5631d1` — the catch block's unpause now carries the same generation guard as the success path's `.superseded` skip (your suggested direction, verbatim): when a successor owns the pane, the failing stale sequence sends nothing, and the successor's own exit-path unpause owns the pause state. New test pins the exact interleaving (successor replaces the sink while the stale batch is in flight → stale capture %error → zero `:continue` writes).

That was the last symmetric hole in the pattern — every pause/unpause/markReady/writeReplay/detach/acknowledge/register path is now generation-scoped, plus the app-side bookkeeping and reader registry. With `test`, `Lint`, and the plans-guard all green on the previous commit, this should be the closing round.

PR #362, opened 2026-07-06
Unify suspend + hibernation into one parked state (+ cold profile swap)

zionts:
Same-repo re-open of #345 (fork PRs can't pass the required claude-review check — GitHub withholds OIDC from fork-triggered workflows). Content identical to #345.

Collapses the two overlapping "park a Claude session" features into one: `hibernatedAt` authoritative, `suspendedAt` legacy-read, migration v40 backfill; adds the auto-hibernate on/off + idle-minutes toggle, and cold profile swap (`tbd terminal swap-profile`). The old SuspendResumeCoordinator is deleted; suspend RPCs route to `manualHibernate`.

Verified: builds clean, 2417 tests pass (1 known env failure), swiftlint clean. Running live on my machine.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #366, opened 2026-07-06
Fix: Un-break main's test job after two green PRs collided (#341 made the "unknown" enum value real)

cheapsteak:
## Summary
- Main's `Test` job fails deterministically (not a flake): `DatabaseDecodeResilienceTests.unknownTypeSkippedInSummary` plants `"limit_reached"` as an *unknown* `NotificationType` rawValue and expects the failable `toModel()` to skip the row.
- Root cause: #341 added `case limitReached = "limit_reached"` as a real case and merged **42 seconds before** #356 (which added the test). Each PR's CI ran against a base without the other, so both were green in isolation but collide on main. First seen on PR #365's test job, which inherited both.
- Fix: plant a synthetic `"test_only_unknown_type"` rawValue that can never become a real case, and update the suite doc comment to record why (`limit_reached` is now a legitimate case).

## Test plan
- [x] `swift test --filter DatabaseDecodeResilienceTests` — all 8 tests pass, including the previously failing one
- [x] Verified the other synthetic sentinels in the file (`"future_type"`, `"future_status"`) are not real rawValues of their enums, so no analogous latent break remains

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=2FD549A6-F58E-42F3-A920-3871A31B4BAC)

### DIFF
--- a/Tests/TBDDaemonTests/DatabaseDecodeResilienceTests.swift
+++ b/Tests/TBDDaemonTests/DatabaseDecodeResilienceTests.swift
@@ -5,11 +5,13 @@ import GRDB
 @testable import TBDShared
 
 /// Regression coverage for the 2026-07-06 crash-loop: a `notification` row with
-/// `type = "limit_reached"` (a `NotificationType` case only present on an
-/// unmerged branch) made `toModel()`'s force-unwrap crash the daemon on
-/// startup, because `notifications.list` decodes the whole result set. These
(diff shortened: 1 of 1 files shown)


cheapsteak:
Superseded: #362 landed the identical fix minutes before this (its rebase commit swaps the planted "limit_reached" for "totally_unknown_type" in the same test), and main's Test workflow is green on e0932ed1. Closing as redundant.

PR #375, opened 2026-07-07
fix: close the StopFailure transcript-append race that broke session-limit auto-resume

cheapsteak:
## The bug

Auto-resume (#341) never triggered on real usage-limit hits. Two production incidents (2026-07-07 04:26 and 04:36 UTC, worktrees "🔗 Fix Promoted Scratch Resume" and "🌅 Wake Recreate Window") show why:

**Claude Code fires the StopFailure hook *before* appending the API-error record to the transcript JSONL.** In one incident the hook's fallback notification landed 278ms before the limit record was written; an earlier hook firing had no record at all. The handler read the transcript once at hook time, found nothing, and fell through to the legacy "Claude stopped: API error (unknown)" path — that "(unknown)" itself being a second bug (the code read `payload["error_type"]`; the real payload key is `error`).

Also confirmed from production transcripts: `rate_limit_info` is null on current Claude Code — the display-text parser ("You've hit your session limit · resets 12:50am (America/Toronto)") is the only live detection path, and it parses that text fine; it just never saw it.

The 2026-07-04 live E2E passed because `tbd hooks fake-rate-limit` injects at the RPC — the exact seam that skips the hook→transcript timing.

## The fix (CLI-side only; daemon untouched)

1. **Payload-first detection (race-free):** run hard-limit detection on the StopFailure payload's own `last_assistant_message` / `error_details` text before any file I/O. When it hits, the transcript is never read (tested via call counter).
2. **Bounded transcript re-read (backstop):** no record on first read → up to 6 re-reads at 250ms (~1.5s cap, observed production lag ≤0.5s), stopping at the first record. Exhaustion reproduces the exact pre-existing fallback. Pure core stays pure — sleeps only via an injected waiter.
3. **Key fix:** `error` → `error_type` → "unknown".
4. Single shared implementation: `RateLimitDetection.detect(messageText:)` extracted; the transcript scanner delegates to it — one discrimination+parse path for payload and transcript.

Tests: 15/15 focused (10 pre-existing unchanged + 5 new asserting call counts, retry counts, exhaustion text), full suite 2628 green.

## Review follow-ups (non-blocking)

- Add a test for payload-text-matches-but-reset-unparseable → transcript fall-through
- Add an exhaustion test for a permanently unreadable transcript (new worst case: 1.5s before fallback vs previously instant)
- Latent: payload text now precedes any future structured `rate_limit_info`; moot while that field is null in production

https://claude.ai/code/session_01Leu1sJPDm8YrELnTKvPuN3

### DIFF
--- a/Sources/TBDCLI/Commands/StopFailureCommand.swift
+++ b/Sources/TBDCLI/Commands/StopFailureCommand.swift
@@ -99,7 +99,13 @@ enum StopFailureMessage {
     /// 1. Payload-first (race-free, no file I/O): `last_assistant_message`
     ///    then `error_details`, whichever first yields a hard-limit hit.
     /// 2. Transcript read, retried up to `maxRetries` times (`retryInterval`
-    ///    apart) until `RateLimitDetection.hasApiErrorRecord` sees a record.
+    ///    apart) until `RateLimitDetection.hasApiErrorRecord(in:newerThan:)`
+    ///    sees a record newer than `recencyFloor` (see below) — NOT merely
+    ///    "a record exists", since the transcript is one continuously-growing
+    ///    JSONL per terminal across resumes: an old, already-resolved
+    ///    `isApiErrorMessage` record from a PRIOR hit would otherwise
+    ///    permanently satisfy an existence-only gate on every hit after the
+    ///    first, making the retry backstop a no-op.
     static func computeOutcomeWithRetry(
         stdinData: Data,
         readFile: (String) -> Data?,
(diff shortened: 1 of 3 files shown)


cheapsteak:
Re the payload-schema minor: three corroborating data points for `error` over `error_type` — (1) the repo doc's schema was verified against a v2.1.160 decompile; the claude-auto-retry project's DESIGN-NOTES (verified against a newer v2.1.195 decompile) documents the StopFailure payload as `{hook_event_name, error, error_details, last_assistant_message}`; (2) empirically, the pre-PR code read `error_type` and every production notification rendered "API error (unknown)" — i.e. `error_type` is absent on the currently-installed Claude Code; (3) the code keeps `error` → `error_type` → "unknown" tolerance, and the `error_details`/`last_assistant_message` branches are `as? String`-guarded, so an absent field is inert. Will capture a raw payload from the next live limit hit to close the loop definitively. The Medium (recency gate) + the triple-parse minor are being fixed now.

PR #386, opened 2026-07-08
Feat: Pick a model profile (with live usage bars) when creating a worktree from the repo + button

cheapsteak:
## Summary
- **Long-press or Option-click** the repo-section **"+"** opens a model-profile picker for the new worktree. A plain click still creates with the default profile — the gesture split preserves the existing one-click behavior.
- The picker's first item is **"Choose a branch…"** (nested, in-place list; selecting a branch creates on it with the **default** model — an accepted tradeoff, since branch choice is the infrequent action). Every other row is a **model profile**: one click creates a new-branch worktree pinned to that account.
- Threads an optional `profileID` through `AppState.createWorktree → DaemonClient → RPC → daemon`, resolved by `ModelProfileResolver` as an **explicit override above** the existing repo/scratch/global precedence chain. `nil` preserves prior behavior exactly (covered by tests for both branches).
- New reusable **`UsageBarsView`**: compact `5h:` / `wk:` bars with severity-colored fill (green→orange→red) and a neutral elapsed-time **marker tick** (fill past the tick = ahead of pace). Pure, unit-tested pace/elapsed helpers (`elapsedFraction`, window constants) in `ProfileUsagePresentation`.
- Profile rows are **state-aware**: real usage bars for logged-in OAuth; a skeleton *only* while OAuth awaits its first poll; "Not logged in — run /login" for logged-out OAuth; and a `kind · config` descriptor (e.g. `Bedrock · us-west-2 · Opus`) for API-key/Bedrock profiles that never have usage — no more permanent fake skeleton.

## Notes
- **By-design fallback:** if an *explicitly picked* profile can't resolve (e.g. an API-key row whose keychain secret was cleared), the daemon logs a warning and falls back to the default account rather than failing the spawn. Flagged in review as a nit; kept for robustness. Could later surface the substitution to the client.
- Branch selection now lives entirely inside this picker; the old orphaned Option-click branch popover was removed (`BranchListView` extracted from `BranchPickerView` as the shared source).

## Test plan
- [ ] Plain-click a repo "+" → creates a worktree with the default profile (unchanged).
- [ ] Long-press and Option-click both open the profile picker; dismissing it and clicking "+" again still creates (no dead click — the `longPressTriggered` safety-net fix).
- [ ] Pick a profile → worktree spawns pinned to that account; "Choose a branch…" → pick a branch → spawns on that branch with the default model.
- [ ] Usage bars render for OAuth-with-usage; API-key/Bedrock show descriptors; logged-out shows "Not logged in".
- [ ] `swift build` + `swift test` (ModelProfileResolver, WorktreeLifecycle, ProfileUsagePresentation suites — 126 tests) pass.

[🎛 Long-Press Model Picker](https://cheapsteak.github.io/tbd/open/?worktree=B728A3F9-796D-4789-8671-46090C03FDC5)

### DIFF
--- a/Sources/TBDApp/AppState+Worktrees.swift
+++ b/Sources/TBDApp/AppState+Worktrees.swift
@@ -17,7 +17,10 @@ extension AppState {
     /// into a new worktree (no auto-generated `tbd/*` branch); the optimistic
     /// placeholder uses the branch's local name so the row looks right
     /// immediately.
-    func createWorktree(repoID: UUID, parentWorktreeID: UUID? = nil, existingBranch: BranchInfo? = nil) {
+    /// `profileID` is an optional explicit model-profile override chosen at
+    /// creation time (sidebar `+` profile picker). nil resolves the profile via
+    /// the daemon's normal repo/scratch/global precedence chain (today's behavior).
+    func createWorktree(repoID: UUID, parentWorktreeID: UUID? = nil, existingBranch: BranchInfo? = nil, profileID: UUID? = nil) {
         // Optimistic placeholder so the row appears instantly. When picking an
         // existing branch we use its local name so the placeholder name
         // doesn't briefly show a fake `tbd/*` value.
@@ -59,7 +62,8 @@ extension AppState {
                     displayName: placeholderName,
                     cols: size.cols, rows: size.rows,
                     parentWorktreeID: parentWorktreeID,
-                    useExistingBranch: existingBranch != nil
+                    useExistingBranch: existingBranch != nil,
+                    profileID: profileID
                 )
                 // Replace the placeholder with the real worktree, carrying
                 // over any rename the user typed while creation was in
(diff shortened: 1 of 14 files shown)


cheapsteak:
Addressed the review:

**Medium — logged-out OAuth rows rendering stale usage bars (WorktreeProfilePickerView.swift):** fixed in f63a332b. `showsUsageBars` now guards on `ProfileUsagePresentation.isSelectable(entry)`, so a logged-out OAuth profile (whose last-known buckets the daemon carries forward) falls through to the "Not logged in — run /login" text path instead of showing live-looking bars. Additionally, non-selectable rows are now `.disabled(!isSelectable)` + dimmed to `0.5` opacity, matching the existing `AccountPickerSheet` pattern — so they're no longer clickable. Good catch on the carried-forward-buckets tracing.

**Minor — leftover "prototype"/"temporary Option-click" comments:** fixed in bb4a485f. Reworded to describe the shipped behavior; Option-click → profile picker is the permanent intended trigger (the branch picker it originally "temporarily" replaced was removed when branch selection moved inside this picker).

PR #405, opened 2026-07-09
Feat: Re-run a worktree's setup hook without recreating the worktree

cheapsteak:
## Summary

- The `preSession` hook runs exactly once per worktree lifecycle event. When it needs a second pass — an `npm install` that raced a lockfile change, a `brew bundle` that hit a flaky network, a hook you just edited — your only options were archive+revive (which destroys the live Claude session) or a fresh worktree. Right-clicking a worktree now offers **Re-run setup hook**: it runs in a fresh background tab, doesn't steal focus, and leaves running agents completely alone.
- The hook's tab is now **ephemeral on success**. It used to `exec $shell` and linger forever, so every worktree with a hook carried a spent `pre-session` tab for life — and a re-run action would have turned that into N tabs. It now closes itself once the hook exits 0 and the agent terminals exist. On a non-zero exit, timeout, or killed pane it stays open with the output on screen, because that's the one time you need it.
- The item only appears when a `preSession` hook actually resolves for that worktree, and only one run is allowed at a time — a second request is refused with `Setup hook is already running for this worktree.`

## Notable design points

**`HookResolver` moved to `TBDShared`.** The app has to answer "does this worktree have a preSession hook?" to decide whether to show the menu item, and the resolution chain is five steps deep (app config → `.worktree-hooks/` → `conductor.json` → `.dmux-hooks/` → global default). Sharing the type beats a copy in the app that silently drifts from what the daemon executes.

**`preSessionCommand` is deliberately untouched.** `waitForPreSessionCompletion` already resolves the marker-vs-pane-death race by re-reading the marker before concluding `.paneKilled`. Making the pane exit on success would have reintroduced exactly that race, so the daemon tears the tab down *after* observing the outcome instead.

**Teardown sits inside the `do` block**, after `spawnPrimaryTerminals` succeeds — not after the `do/catch`. If the primary spawn throws, the hook tab is the worktree's only remaining tab and closing it would leave the worktree tab-less. `spawnPrimaryTerminals` also reassigns `activeTabID` to the primary first, so focus never sits on a tab that's about to vanish. Regression-guarded by `dbErrorDuringRowCheckDoesNotTearDownPreSessionWindow`.

**Scratch spaces could never run a `preSession` hook before this.** `spawnPreSessionTerminal` required a `Repo`, and scratch spaces have none. It now takes `repo: Repo?` and falls `TBD_REPO_PATH` back to the worktree's own path.

**Concurrency is guarded daemon-side, not in the UI.** Once the tab is ephemeral, a lingering `pre-session` tab means "the last run failed", not "a run is in progress" — so the app can't infer running-ness from the tab it already tracks. A true greyed-out menu item would need a new `StateDelta` case plus a transient worktree-snapshot field to survive an app relaunch mid-run. That's real state-sync plumbing for a row you'd rarely see; the daemon rejects the second run with a clear message instead. `PreSessionRunRegistry` is an actor because `WorktreeLifecycle` is a struct (same reasoning as the existing `conflictSweepCache`).

## Test plan

3182 tests pass (`swift test --parallel -j 2`), `swift build` and `swiftlint --strict` clean. New coverage: both branches of the exit-code gate, re-run isolation (no status flip, no primary spawn, no focus steal), all three re-run rejections, and menu presence/placement/row-scope.

Verified live against a running daemon with a throwaway repo, driving `worktree.rerunPreSession` over the unix socket:

- [x] Worktree create with an `exit 0` hook → `pre-session` tab appears, then disappears; only `Claude Code` + `setup` remain, tmux window gone.
- [x] Re-run → tab appended **last** (focus not stolen), Claude terminal ID unchanged throughout, tab self-closes on completion.
- [x] Second re-run while one is in flight → `{"success":false,"error":"Setup hook is already running for this worktree."}`
- [x] Re-run with an `exit 7` hook → tab remains, pane still shows `THIS HOOK WILL FAIL` dropped into a shell, and an `.error` notification is recorded reading `Setup hook failed (exit 7) — its tab is left open with the output` (no misleading "starting the agent anyway").
- [ ] GUI pass: right-click a worktree and confirm the item renders in its own section above `Open in Finder`, and is absent for a repo with no `preSession` hook (no dangling divider).

## Known limitation

The `claimsFocus: false` tab-order append is a non-atomic read-modify-write that can race a concurrent drag-reorder. The tab isn't lost — `applyStoredOrder` appends unknown tab IDs — only its persisted position can be, until the next reorder. This matches the existing terminal-create path; not addressed here.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=94854A36-B61C-4E47-8459-BC632BAA76DA)

### DIFF
--- a/Sources/TBDApp/AppState+Worktrees.swift
+++ b/Sources/TBDApp/AppState+Worktrees.swift
@@ -291,7 +291,8 @@ extension AppState {
     /// an error on this call.
     func rerunPreSessionHook(worktreeID: UUID) async {
         do {
-            try await daemonClient.rerunPreSessionHook(worktreeID: worktreeID)
+            let size = mainAreaTerminalSize()
+            try await daemonClient.rerunPreSessionHook(worktreeID: worktreeID, cols: size.cols, rows: size.rows)
             logger.info("Re-running setup hook for worktree \(worktreeID, privacy: .public)")
         } catch {
             logger.error("Failed to re-run setup hook: \(error)")
--- a/Sources/TBDApp/AppState+Worktrees.swift
+++ b/Sources/TBDApp/AppState+Worktrees.swift
@@ -281,6 +281,24 @@ extension AppState {
         }
     }
 
+    /// Re-run the worktree's `preSession` hook. The daemon spawns a
+    /// non-focused tab and closes it on a clean exit; a failing hook leaves its
+    /// tab open with the output and raises an `.error` notification. Running
+    /// agents are untouched.
+    ///
+    /// The only thing to surface here is a refusal (no hook, already running,
+    /// still creating) — hook FAILURE arrives later as a notification, not as
+    /// an error on this call.
+    func rerunPreSessionHook(worktreeID: UUID) async {
+        do {
+            try await daemonClient.rerunPreSessionHook(worktreeID: worktreeID)
+            logger.info("Re-running setup hook for worktree \(worktreeID, privacy: .public)")
+        } catch {
+            logger.error("Failed to re-run setup hook: \(error)")
+            showAlert("\(error)", isError: true)
+        }
+    }
+
     /// Revive an archived worktree.
     /// Mirrors `reviveWithSession`'s lingering-snapshot UX: keeps the row
     /// visible with a status pill until the user navigates away, instead
--- a/Sources/TBDApp/DaemonClient.swift
+++ b/Sources/TBDApp/DaemonClient.swift
@@ -539,10 +539,10 @@ actor DaemonClient {
 
     /// Re-run the worktree's `preSession` hook in a fresh, non-focused tab.
     /// Returns as soon as the tab exists; the hook itself runs on in the daemon.
-    func rerunPreSessionHook(worktreeID: UUID) async throws {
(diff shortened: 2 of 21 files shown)


cheapsteak:
Addressed the first minor in `c212aa60`: `cols`/`rows` are now threaded through the re-run path exactly as `reviveWorktree` does it — `AppState.rerunPreSessionHook` calls `mainAreaTerminalSize()` (itself gated on `terminalAutoResizeEnabled`, returning `(nil, nil)` when off), and the daemon handler forwards them into `lifecycle.rerunPreSessionHook(worktreeID:cols:rows:)`, which already accepted them.

The new params fields are optional so an older app's JSON still decodes, with a regression test for the omitted-keys case. A second test drives `handleWorktreeRerunPreSession` through the router and asserts the recorded tmux `resize-window -x 180 -y 45`, which proves the handler forwards rather than drops them.

Leaving the second minor as-is — the ~5 synchronous `FileManager` calls in `hasPreSessionHook` only run on right-click when the menu is built, and the code documents that tradeoff. Moving it off the main actor would mean making the menu's `Context` async for no user-visible gain.

PR #409, opened 2026-07-09
Feat: Replace the "TBD" toolbar title with the worktree name, repo, and a shared notepad

cheapsteak:
## Summary
Replaces the static "TBD" toolbar title with a compact, interactive nav-header and adds an on-disk, repo-shared notepad.

**Toolbar title** (replaces the macOS window title — was `Window("TBD")` + `unified(showsTitle:)`, now a `.principal` item with `sharedBackgroundVisibility(.hidden)` so it floats, no glass capsule). Left → right:
- **Notes button** (`square.and.pencil`) with an immediate hover tooltip ("Repo notes" / "Worktree notes").
- **Repo name** (11 pt) → click-menu opening that repo's detail tabs: Archived Worktrees / Default Prompts / Hooks & Settings (via a new `AppState.pendingRepoDetailTab` hook; `RepoDetailView.Tab` promoted to top-level `RepoDetailTab`).
- **Worktree name** (14 pt) → click-menu reusing the sidebar `RowActionMenu` (Rename, Archive, Open in Finder, Copy Path, …). Rename routes to the sidebar's inline editor via `editingWorktreeID`.

Right-click can't be used on a toolbar item — macOS's NSToolbar claims it for the "Icon and Text / Icon Only" menu and overrides any custom `.contextMenu` (confirmed unfixable without private API), so these use left-click `Menu`s per Apple's convention.

**Repo notepad** — a free-text scratchpad opened from the notes button, stored **on disk** under `~/tbd` (mirroring the preSession/setup hook pattern — no daemon/RPC/DB): `~/tbd/repos/<repoID>/notes.md`, shared across all worktrees of a repo; `~/tbd/worktrees/<worktreeID>/notes.md` for scratch worktrees. Popover reads-on-open with a placeholder + autofocus, 500 ms debounced autosave, and dirty-tracking so an untouched open never clobbers an external edit.

**Shared row-menu additions** (apply to both the toolbar dropdown and the sidebar right-click):
- Added **Copy Branch Name** (gated on a non-empty branch).
- The **keep-warm** toggle now only appears when the global auto-hibernate switch (`autoHibernateEnabled`) is on.

**Cleanup** — removed the now-duplicated bottom-left status-bar label and its orphaned helpers.

Design spec: `docs/superpowers/specs/2026-07-09-nav-header-notes-design.md`

## Test plan
- [x] `TBDConstants` path tests (repo + worktree scope) + `NotesFileStore` roundtrip/empty-delete/dir-create.
- [x] `RowActionMenu` tests incl. both branches of the new Copy-Branch and keep-warm gates.
- [x] Full suite green (3259 tests).
- [x] Live: toolbar layout, notes popover (placeholder/autofocus/save + repo-shared), both dropdowns, Copy Branch Name, keep-warm gating — verified during iteration.

[nav-header-notes](https://cheapsteak.github.io/tbd/open/?worktree=786A6A9E-7129-474F-9BC5-29C8275C078C)


cheapsteak:
Addressed the Medium finding (silent loss of unsaved notepad edits on mid-debounce scope change) in `110c7849`.

Fix in `NotepadPopoverView.swift`:
- Track `loadedPath` — the file `content`/`loadedContent` were actually loaded from.
- `flushSave()` now writes to `loadedPath` (not the current `scope.notesPath`), so a flush always lands in the file the edit belongs to even after the scope changed.
- `.task(id: scope.notesPath)` now cancels the pending save and `flushSave()`s the previous scope **before** overwriting `content` with the new scope's file.

So a scope switch mid-debounce (Cmd+[ / Cmd+], deep link, etc.) now flushes the old edit to the old file first, instead of dropping it. The `onDisappear` path was already correct and stays so.

PR #415, opened 2026-07-09
DX: Stop a 5s wall-clock timer from flaking the attach orchestration suite (146s CI failures)

cheapsteak:
## Summary

**What's broken.** The "Attach RPC orchestration" suite intermittently fails in CI (surfaced on #408's rerun-clean run, a UI-only diff — pre-existing flake, not a regression). Failing runs burn ~146s and record 6 issues around the mid-sequence supersession test: `ready2` returns an error, no `continue` is ever written, and both 60s `waitFor` polls time out.

**Why it happens.** The suite's `bridge()` helper defaulted `readyTimeout` to a real 5-second wall-clock timer. Every `attach.request` arms `Task.sleep(readyTimeout) → detachIfNotReady`. All test targets share one process and one cooperative pool, so under parallel CI load a test task can be starved >5s between an attach and its `attach.ready`; the timer fires, tears down the un-acked sink, and the ready fails with `notAttached`. Confirmed by exact local reproduction: starving only that window reproduces the identical 6-issue signature (~126s locally vs 146s in CI). All five supersession tests in the suite carried the same latent race.

**What this PR does.**
- `fix(tests)`: the helper default becomes an un-hittable 600s, with a comment explaining the race. The ready-timeout is incidental machinery in these tests; the one test that exercises it (`unackedAttachTornDownAfterTimeout`) already passes its own explicit 100ms and is unchanged.
- `refactor(tests)`: extracts `ControlModeTestSupport` (canonical `waitFor` with a single 30s `ciSafeDeadline`, `LineRecorder`, `makeFakeClient`) and de-duplicates the ~8 hand-copied variants across the control-mode test files — the copy-paste drift (60s vs 30s deadlines, three private `ciSafeDeadline` declarations) that let this class of flake hide. Pure mechanical extraction; no test logic changed. `LoginSessionCoordinatorTests` and `TmuxControlCommandClientTests` keep their genuinely different-shaped local helpers.

## Test plan
- [x] Root cause proven before fixing: a 6s starve of the gen-2 window reproduced the exact CI failure signature locally
- [x] Previously flaky test looped 10x post-fix: 10/10 pass
- [x] Affected families green after rebase on main: 145 tests / 17 suites via `swift test --filter "AttachRPC|AttachReplay|ControlModeInput|ControlModeResize|PaneRepair|PaneFanout"`
- [x] `swift build --build-tests` and `swiftlint --strict` clean

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=2EFA4D7B-1A1D-4C7E-8826-0365E2236741)

### DIFF
--- a/Tests/TBDDaemonTests/AttachRPCTests.swift
+++ b/Tests/TBDDaemonTests/AttachRPCTests.swift
@@ -144,14 +144,6 @@ struct AttachRPCOrchestrationTests {
             commandProvider: commandProvider)
     }
 
-    /// Thread-safe, synchronous recorder of fake-client stream writes.
-    private final class Recorder: @unchecked Sendable {
-        private let lock = NSLock()
-        private var _writes: [String] = []
-        func record(_ line: String) { lock.lock(); _writes.append(line); lock.unlock() }
-        var writes: [String] { lock.lock(); defer { lock.unlock() }; return _writes }
-    }
-
     /// Thread-safe monotonic counter — lets a commandProvider hand each
     /// successive attach.ready sequence its OWN fake correlator, so a test
     /// can complete a later sequence's replies before an earlier one's.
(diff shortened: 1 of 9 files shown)


cheapsteak:
Addressed the Medium finding in 349b76a2: shared `waitFor` now returns `@discardableResult Bool` (still `Issue.record`s on timeout), and the two fragile call sites (`pasteOrderedBetweenKeystrokes`, `repeatedFailuresFireOnce`) guard with `try #require(await waitFor...)` before indexing — a genuine timeout now aborts the test cleanly instead of trapping the shared test process. All other callers unchanged. Verified: `swift build --build-tests` clean, `ControlModeInput` families 37 tests / 4 suites green, `swiftlint --strict` clean.

cheapsteak:
Addressed in 2682bf9e: both `PaneRepairCoordinatorTests` sites now guard with `try #require(recorder.writes.count >= 2, ...)` before indexing, mirroring the file's existing pattern. Also swept every suite using the shared `waitFor` for the same shape — all remaining post-wait accesses are either already `#require`-guarded, optional-chained (`.first`/`.last` comparisons), or not waitFor-dependent, so this defect class should be fully closed out in this PR. Verified: build clean, 113 tests / 14 suites green, `swiftlint --strict` clean.

PR #417, opened 2026-07-09
DX: Run the .build reaper on every restart so disk GC needs zero setup

cheapsteak:
## Summary

**Why:** the idle-`.build` reaper (#388/#414) only runs via an hourly launchd agent that each developer must install by hand (`install-reclaim-agent.sh`) — which, in practice, only one machine has. Everyone else accumulates stale multi-GiB `.build` dirs until the disk fills.

**What:** `scripts/restart.sh` now kicks off `scripts/reclaim-build.sh` in the background on every restart — the moment you're about to *add* ~2 GiB of fresh build artifacts is exactly the right moment to collect old ones, and it overlaps the build instead of adding wall-clock time.

- **Silent:** stdout/stderr append to `~/Library/Logs/tbd-reclaim-build.log` (same file the launchd agent uses); nothing reaches the terminal.
- **Async & failure-proof:** `nohup … & ` with all three fds redirected; no `disown` — a bare `disown` can fail with "no such job" if the child already exited, which under `set -e` would abort the restart (caught in review; the launch line is now the only construct used).
- **Guarded:** skips silently if the script is missing/non-executable; opt out with `TBD_SKIP_RECLAIM=1`.
- The reaper itself is already safe to fire at any moment (active-build ps scan, 10-min mtime grace, concurrent runs are idempotent).

## Test plan

- [x] `bash scripts/reclaim-build.test.sh` — 48 ok, 0 FAIL. New static assertions pin the launch line's redirections (`2>&1`, `< /dev/null`), the *absence* of `disown` (negative-tested: re-inserting it turns the suite red), and that the launch stays ordered before `swift build`
- [x] `shellcheck` — zero new findings (pre-existing SC2034 on main unchanged)
- [x] Live replica of the launch line: terminal stays clean, parent exits in ~0.2s without waiting, log receives the reaper's output

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=B95A348F-9277-4B2D-BBD9-1655E91E1B61)

### DIFF
--- a/docs/reclaim-build.md
+++ b/docs/reclaim-build.md
@@ -2,6 +2,13 @@
 
 Keeps each TBD worktree's SwiftPM `.build` small. **Not part of the shipped product** — it is developer machine tooling, like `scripts/restart.sh`.
 
+## Triggers
+
+The reaper runs from two places, both writing to the same log:
+
+- **Hourly via launchd**, if installed (see Install / uninstall below).
+- **Automatically in the background on every `scripts/restart.sh`** — launched with `nohup ... &` before the build starts so the two overlap, fully silent (no terminal output), logging to `~/Library/Logs/tbd-reclaim-build.log`. This makes it zero-setup for contributors who never installed the launchd agent. Opt out with `TBD_SKIP_RECLAIM=1`.
+
 ## What it does (hourly, via launchd)
 Only acts on worktrees containing a `Package.swift` (SwiftPM packages) — `tbd worktree list` spans every repo TBD manages, including non-Swift ones (e.g. longeye-app, longeye-docs, agent-channels), which are skipped entirely.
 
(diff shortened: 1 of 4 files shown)


cheapsteak:
Addressed the Medium in `1900e039`: `restart.sh` now passes `RECLAIM_EXCLUDE_PATH="$REPO_ROOT"` and the reaper unconditionally skips that path in `main()`'s loop (both tiers, both enumeration sources), with paths canonicalized on both sides via `pwd -P` so symlinks/trailing slashes can't defeat the exclusion. Tests: an end-to-end run proves the excluded tier2-eligible worktree's `.build` survives while a sibling in the same run is actually reclaimed, and a variant passes the exclusion as a symlink + trailing slash. Suite is 57 ok / 0 FAIL; shellcheck introduces no new findings.

PR #423, opened 2026-07-10
feat(reclaim): reap dependency installs + agent worktrees; sweep orphaned scratchpads (#420)

### LINKED ISSUE #420: reclaim-build.sh misses Claude Code agent worktrees and orphaned /private/tmp scratchpads (~35G found on one machine)
## Context

While doing a disk-cleanup pass on a machine running the TBD fleet (2026-07-09, macOS, 460G disk at 99%), I found ~35G of reclaimable space in two categories that the new `.build` reaper from #417 structurally cannot see.

## Gap 1: Claude Code *agent* worktrees are outside the reaper's scope

`scripts/reclaim-build.sh` enumerates candidates via `tbd worktree list --json`, so it only walks TBD-*managed* worktrees (`~/tbd/worktrees/...`). But Claude Code's own agent/worktree isolation creates SwiftPM worktrees under `<repo>/.claude/worktrees/`, and each accumulates a full `.build`:

- Found: **8 worktrees × 2.2–2.8G `.build` each ≈ 21G** under `~/Desktop/proj/tbd/.claude/worktrees/`
- 6 of 8 were clean + idle >24h — exactly the reaper's target profile
- A `--dry-run` of the reaper on the same machine planned **0 reclaims** (its worktrees were all fresh or non-SwiftPM)

**Suggestion:** extend the candidate enumeration with a second source, e.g. glob `<each repo>/.claude/worktrees/*/Package.swift`, apply the same tier-1/tier-2 idle thresholds + active-build guard. Session-liveness can be approximated by the existing `newest_mtime` grace check since these worktrees have no `liveClaudeSessionCount`.

## Gap 2: orphaned per-worktree scratchpads in `/private/tmp/claude-<uid>/`

Claude Code gives each worktree a scratchpad dir `/private/tmp/claude-<uid>/<path-slug>/`. When the worktree is deleted (or the session dies), nothing sweeps the scratchpad.

- Found: **one orphaned scratchpad of 14G** (`...tbd-20260703-evil-barnacle/scratchpad`) plus a tail of small ones
- These are pure temp data; the harness recreates the dir on demand

**Suggestion:** a sweep step (in the reaper or a sibling script) that removes `/private/tmp/claude-$(id -u)/<slug>` entries whose corresponding worktree path no longer exists, or whose newest mtime exceeds some threshold (e.g. 7d).

## Why it matters for the fleet

Disk exhaustion on a fleet machine is compounding: RAM pressure from many live sessions grows swap (observed 20×1G swapfiles), so lost disk → lost swap headroom → thrashing. The reaper restart-hook from #417 is the right delivery mechanism; it just needs these two extra sources.

Happy to provide the full audit data.

zionts:
Closes #420.

## Why
The idle-worktree reaper (#388/#414/#417) structurally missed the two dominant disk sinks. On one machine `~/proj/longeye-app/.claude/worktrees` held **81 agent worktrees totalling ~194 GB** (42% of a 460 GB disk) — the bulk was per-worktree `node_modules` (42) + `.venv` (50), **not** SwiftPM `.build`, and the `.venv` bytes were real (link-count 1, not hardlinked to the uv cache). The reaper couldn't touch any of it because it (a) only reclaimed `.build`, gated on `Package.swift`, so non-Swift worktrees were skipped entirely, and (b) had no live-cwd or dirty guard for agent worktrees, which carry no `liveClaudeSessionCount`.

Manually applying the two-tier policy this PR automates took one machine's free space from 5.7 GiB to 40+ GiB.

## What

**`scripts/reclaim-build.sh`**
- **Widened Tier 2**: now also reaps `node_modules` / `.venv` / `.terraform` (idle >48h, no live session), regenerable via `npm install` / `uv sync` / `terraform init`. **Not** `Package.swift`-gated — these dominate non-Swift worktrees (~4.3 GB vs ~2.5 GB `.build`). `.build` / `index-build` reaping stays exactly as before (still `Package.swift`-gated). Install idleness is measured on the install dir's **own mtime** (a single `stat`, never a deep scan of a 100k-file `node_modules`).
- **Universal safety rails** applied to every worktree (both TBD-managed and `.claude/worktrees/*`) before anything is reaped:
  - `SKIP live-cwd` — the worktree is the cwd of a live process (one `lsof -d cwd -Fn` pass per run, prefix-matched). This is what protects a busy agent worktree.
  - `SKIP dirty` — uncommitted changes (`git status --porcelain`; a non-git path is treated as clean).
  - plus the existing recency (`RECLAIM_ACTIVE_GRACE`) and active-`swift`-build guards.
- New `RECLAIM_LSOF_CMD` test seam.

**`scripts/sweep-scratchpads.sh`** (new) — removes orphaned Claude Code per-worktree scratchpads under `/private/tmp/claude-<uid>/<slug>` where `<slug>` (worktree path with `/`→`-`) matches no live process cwd slug **and** nothing was modified in `SWEEP_DAYS` (default 3, via `find -mtime -3 -print -quit`). Follows the reaper's conventions (env seams, `--dry-run`, df-delta logging, shellcheck-clean). Launched from `restart.sh` alongside the reaper — same `nohup`/log/`TBD_SKIP_RECLAIM` opt-out pattern.

Logic for the live-cwd + dirty guards mirrors the reference `reap-worktrees.sh`/`sweep-scratchpads.sh` from the local `disk-cleanup` skill, kept in tbd's script style.

## Test plan
- `bash scripts/reclaim-build.test.sh` — **93 ok, 0 FAIL** (the original suite + new coverage: install reaping incl. non-Swift & agent worktrees, live-cwd exclusion, dirty exclusion, install recency guard, `has_reclaimable`, `is_live_cwd` prefix semantics, lsof parsing).
- `bash scripts/sweep-scratchpads.test.sh` — **12 ok, 0 FAIL** (orphan removal, live-slug keep, recent-file keep, dry-run, missing-base no-op, `SWEEP_DAYS` window).
- `shellcheck` — clean on all changed/new scripts (the one remaining `restart.sh` SC2034 is pre-existing on `main`, untouched here).


cheapsteak:
>On one machine ~/proj/longeye-app/.claude/worktrees held 81 agent worktrees totalling ~194 GB 

I highly suspect this number is highly inflated due to whatever disk usage scanning tool not being aware of APFS-clones and hardlinks

cheapsteak:
## Review

**TL;DR:** The engineering quality is high — good test coverage (93 + 12 tests), shellcheck-clean, careful `rm -rf` guards, honest `df`-delta logging. But two issues undermine the cost/benefit framing and I'd request changes before merge:

1. The **194 GB figure is inflated — the PR body's own before/after numbers show it** (deleting the corpus freed ~35 GiB, not ~194 GB), and the verification method used ("link-count 1, not hardlinked") is incapable of detecting why.
2. **Install reaping applies automatically, by default, to every TBD-managed worktree of every repo TBD manages** (i.e., users' real day-to-day projects, not just tbd) — and the "idle" metric is wrong in a way that makes *actively developed* worktrees eligible for reaping.

---

### 1. The 194 GB claim vs. APFS clones

Verified empirically on macOS/APFS: a 100 MB file plus a `cp -c` (clonefile) copy reports **200M in `du`, `nlink=1` on both files, but only ~103 MB of actual `df` consumption**. APFS clones share extents while having distinct inodes, so:

- `du` double-counts clones (it only dedupes hardlinks, via inode).
- **`st_nlink == 1` is exactly what a clone looks like.** The PR body's evidence that "the `.venv` bytes were real (link-count 1, not hardlinked to the uv cache)" rules out hardlinks only — the wrong check on macOS, because **uv's default link mode on macOS is `clone`** (copy-on-write reflink from the uv cache; it hardlinks on Linux, which is presumably where the hardlink check came from).

So the 50 `.venv` dirs are very plausibly mostly clone-shared with the uv cache — near-free on disk, and deleting them frees only diverged extents. And the PR body already contains the disproof of its own headline number: **if the ~194 GB were real unique bytes, manually applying the policy would have taken free space from 5.7 GiB to ~199.7 GiB — the observed result was "40+ GiB", i.e., ~35 GiB actually freed.** Even granting that the manual pass skipped some live/dirty worktrees, that's a ~5x gap between the `du`-style figure and ground truth. (npm `node_modules` is mostly real bytes since tarball extraction writes fresh data; pnpm trees are hardlink-shared with the store and also largely free.)

This matters for the design, not just the framing: `.venv` reaping may be near-zero reward while imposing real cost — re-`uv sync` for uv projects, and a plain `python -m venv` + ad-hoc pip installs without a lockfile is **not regenerable at all**.

**Asks:**
- Re-measure with ground truth (delete-and-`df`-delta on a sample — the script already knows how to log this) and restate the real reclaimable number in the PR body / docs.
- Consider dropping `.venv` from `INSTALL_DIRS`, or gating it on evidence of a lockfile (`uv.lock`/`requirements*.txt`) so only regenerable venvs are reaped.

### 2. Install reaping hits users' active projects; the idle clock measures the wrong thing

**Scope:** with the `Package.swift` gate removed for installs, this acts on every TBD-managed worktree of every repo TBD manages (longeye-app etc.), plus all `.claude/worktrees/*` agent worktrees under those roots — hourly via launchd and on every `restart.sh`, delete-by-default, with only a global opt-out (`TBD_SKIP_RECLAIM` — which doesn't even affect the launchd path). No per-repo opt-in/out.

**The bigger problem — "idle > 48h" doesn't mean idle.** Install idleness is measured on the install dir's *own mtime* (`dir_mtime`). A directory's mtime changes when entries directly inside it are added/removed — i.e., on **install**, not on **use**. Running a dev server, building, testing, editing all day never touches `node_modules`'s top-level mtime. So "idle > 48h" actually means **"no `npm install`/`uv sync` in 48h"**, which describes nearly every actively-developed worktree. (Contrast with the `.build` tiers, which deep-scan newest-file mtime — builds continuously touch `.build`, so that clock genuinely tracks activity. The installs clock doesn't.)

That leaves the safety rails carrying the entire load for active-but-currently-closed worktrees, and they have gaps:
- **live-cwd** (`lsof -d cwd`) only catches processes whose *cwd* is inside the worktree. Terminals closed for the weekend → no protection. Editors with the project open but cwd elsewhere (VS Code windows commonly have cwd `/`) don't count.
- **dirty** doesn't protect the common state of active work: everything committed, installs gitignored → clean.
- **`liveClaudeSessionCount`** only exists for TBD worktrees and only for Claude sessions.

Concrete failure: work in a longeye-app worktree Friday, commit, close everything. Saturday's hourly launchd run: no live cwd, clean tree, no session, `node_modules` mtime = Tuesday's install → **deleted**. Monday starts with a cold `npm install`. For `.terraform` it's worse — re-`init` needs network and credentials.

**Asks:**
- **Restrict install reaping to `.claude/worktrees/*` agent worktrees** — always session-less, genuinely abandoned once idle, and they're the entire 81-worktree incident that motivated this. Leave TBD-managed daily-driver worktrees out, or require a real activity signal for them (e.g., newest mtime of `.git/index` / `HEAD` / source tree, not the install dir).
- Alternatively make install reaping opt-in per repo, or raise the threshold well past a long weekend (7–14d) on a real activity clock.

### Relation to #414 / #420

#414 (merged hours before #420 was filed) already brought SwiftPM agent worktrees into the reaper's scope — #420's Gap-1 evidence ("dry-run planned 0 reclaims") describes the pre-#414 reaper. What this PR genuinely adds is install-dir reaping, the scratchpad sweep (Gap 2), and the live-cwd/dirty rails — worth stating that way, since "bring agent worktrees into scope" was #414. (Note: repos with agent worktrees but no TBD-managed worktree remain invisible to both, since roots are derived only from `tbd worktree list`.)

### Smaller findings

- **Guards not re-checked at reap time** — `is_live_cwd`/`is_dirty` run at plan time only; tier1/tier2 recheck `has_active_build` before deleting but nothing rechecks liveness/dirtiness. Small window (same run), but a session starting mid-run in an agent worktree can get its installs reaped. Cheap to re-check in the reap loop.
- **Ordering nit in `main()`:** `is_live_cwd` and `is_dirty` (a full `git status`, expensive on big repos) run before the cheap `has_reclaimable` check. Swapping avoids an hourly `git status` across every worktree that has nothing to reclaim.
- **Scratchpad sweep slug matching is exact, not prefix** — a live process whose cwd is a *subdirectory* of the worktree slugifies to a different slug and won't protect the scratchpad (unlike `is_live_cwd`'s prefix matching in the reclaim script). Mitigated by the 3-day mtime rail, but inconsistent between the two siblings. (Slug collisions from `-` in path names only cause over-keeping, never over-deletion — that direction is safe.)
- **Empty scratchpads:** the "recent" keep only looks at files (`find -type f`); a just-created scratchpad containing only directories has no mtime protection. Low risk given the live-slug rail, worth a dir-mtime check.
- **Hourly system-wide `lsof -d cwd`** can be slow and can hang on network mounts; consider a timeout wrapper.
- Genuinely good: `${path:?}` guard on `rm -rf`, one lsof pass per run, single-`stat` idleness check avoiding deep `node_modules` scans, test seams for everything, docs updated to match behavior.

### Bottom line

The scratchpad sweeper and the new live-cwd/dirty rails are solid and could merge as-is. For the install reaping: (1) re-measure the claim clone-aware and reconsider whether `.venv` reaping pays for its risk; (2) don't apply install reaping to TBD-managed worktrees on an install-mtime clock — scope it to agent worktrees, where both incidents actually occurred.

cheapsteak:
Thanks for this, Adam — seriously good work. The test coverage (93 + 12, with seams for everything), the `${path:?}` guard, the honest `df`-delta logging, and the single-`lsof`-pass design are all exactly how this kind of tooling should be built. And you found a real, painful gap: nothing anywhere cleans up agent worktrees, and the pain is real regardless of what the exact GB number turns out to be.

Since my review comment left the install-reaping question as "pick one of three options," I wanted to close the loop with a definitive answer so you're not left guessing — plus some context on where this is heading.

**What I'd like for this PR:**

1. **Scope install reaping to agent worktrees (`.claude/worktrees/*`) only**, dropping TBD-managed worktrees from the installs tier. Not opt-in, not a longer threshold — just out. Reasoning: we've since done a design pass (see below) and concluded installs inside living TBD worktrees are never *orphaned*, only *idle* — and idle-clock reaping of daily-driver worktrees is where the "commit Friday, `npm install` gone Monday" failure mode lives. Agent worktrees don't have that problem: they're one-shot, and they're the entire incident population anyway.
2. **The bot's canonicalization finding** on `is_live_cwd`/`live_cwd_slugs` — the `canon_path()` treatment it suggests is right, and it's what's holding the required check red. Related cautionary tale from the research: anthropics/claude-code#41010, where a cleanup deleted the *parent's* working directory via an agent-ID prefix collision.
3. **`.venv`: worth a clone-aware re-measure before it stays in `INSTALL_DIRS`.** My APFS-clone suspicion now has a citable source: uv's `link-mode` defaults to `clone` on macOS ([uv docs](https://docs.astral.sh/uv/reference/settings/#link-mode)), so `nlink==1` is exactly what a cache-shared clone looks like and `du`/link-count checks can't see the sharing. A delete-plus-`df`-delta on a few sample `.venv`s would settle it either way. If the real bytes turn out small, dropping `.venv` shrinks the PR's risk surface for free.
4. **The scratchpad sweeper and the universal live-cwd/dirty rails are good to land as-is** (modulo the canonicalization fix touching both scripts).

**Context on where this is heading, so the effort here makes sense:** your PR prompted a research + design pass (prior art from agent-deck, Codex, Cursor, vibe-kanban, and Claude Code's own GC — reports in `docs/research/2026-07-10-agent-worktree-gc/`). The punchline: Claude Code *does* have a built-in 30-day sweep for these worktrees, but it exempts anything with unpushed commits — which is why the committed ones accumulate forever, and why this reaper has a real job to do. Longer-term we've specced a daemon-owned **orphan GC** (`docs/superpowers/specs/2026-07-10-orphan-gc-design.md`): agent worktrees reaped when their run has provably ended (lock released + no live process, no idle clocks), snapshot-before-delete, restorable from the app. The shell scripts will keep owning what the product deliberately won't (`.build` and idle-worktree installs). So please don't feel the need to gold-plate this one — lean and landed beats perfect.

**One cheap improvement worth folding in if it's easy:** skip worktrees that are `git worktree lock`ed. Claude Code holds a lock on an agent worktree for the duration of a live run ([worktrees doc](https://code.claude.com/docs/en/worktrees.md)), so it's a stronger in-use signal than lsof for exactly this population — `git worktree list --porcelain` exposes it.

Thanks again — this unblocked a lot of thinking about disk hygiene beyond the PR itself.

zionts:
Thanks Chang — this is a clarifying steer, and "lean and landed beats perfect" is exactly the right call given the daemon-owned orphan-GC spec is where this ends up. Addressed all of it:

1. **Installs scoped to agent worktrees only** (`aa48049`). `node_modules`/`.terraform` reaping now fires *only* on `.claude/worktrees/*`; TBD-managed worktrees keep only the `.build` tiers. Implemented with an explicit typed source tag (`tbd`|`agent`) threaded through the worktree TSV pipeline rather than sniffing the path string. New test proves a TBD-source worktree with an idle, manifest-backed `node_modules` gets **no** `PLAN installs` line.
2. **Canonicalization** (`d4f978e`) — done in the prior push; `is_live_cwd`/`live_cwds`/`live_cwd_slugs` now `canon_path` the worktree side, with symlinked-worktree regression tests in both suites.
3. **`.venv` dropped** entirely rather than re-measured. Your uv `link-mode = clone` citation is enough to conclude the honest bytes are small and clone-shared; not worth the risk surface. `INSTALL_DIRS` is now `(node_modules .terraform)`.
4. **Scratchpad sweeper + universal rails** — landing as-is, with the canonicalization fix applied to the sweeper too.
5. **`git worktree lock` skip folded in** (`aa48049`). New `SKIP locked` rail parses `git worktree list --porcelain`; a locked worktree is skipped before any tier runs (verified a locked worktree's `.build` survives a full run). Behind a `RECLAIM_WORKTREE_LIST_CMD` test seam.

Rebased onto current `main` so #425's shared module cache is intact. reclaim 117 ok / 0 fail, sweep 13 ok / 0 fail, shellcheck clean.

PR #436, opened 2026-07-10
Fix: Stop a glance back at the tab from defeating the deep-link auto-return

cheapsteak:
## Summary
- **What's broken:** after the tbd:// redirect page detects TBD opened, it schedules `history.back()` — but if the user glances back at the tab within the 600ms delay, the `visibilitychange` handler cancelled the pending return and restored the fallback UI. Checking whether it worked is exactly what users do, so the auto-return (PR #416) frequently defeated itself.
- **Fix:** once success detection fires, the return is uncancelable — it proceeds even if the tab becomes visible again. Accepted trade-off (documented in-file): with TBD *not installed*, an incidental hide within the 3s window now false-positives into a navigation; the manual button resets the `returned` flag so retries stay live after an ignored `window.close()`.
- **Diagnostics:** every decision point now logs with a `[tbd-open]` console prefix — including the two known silent-failure modes: a visibility flip arriving *after* the 3s window logs how many ms late it was (the >3s "Open TBD.app?" dialog case), and an ignored `window.close()` announces itself 500ms later. Next time the return doesn't fire, the devtools console says why.

## Test plan
- [ ] With TBD installed: click a redirector link from a PR page → TBD opens → immediately click back to the browser tab → the tab still returns to the PR (previously this cancelled the return)
- [ ] Same flow without touching anything → returns as before
- [ ] Console shows the `[tbd-open]` trail for each step; taking >3s on the "Open TBD.app?" dialog logs the late-arrival diagnostic
- [ ] Without TBD installed, invalid/missing `?worktree=` still shows the error; manual button still fires and re-arms

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=61966922-0E69-4FD6-9CF0-4FC4612C73EF)

### DIFF
--- a/docs/open/index.html
+++ b/docs/open/index.html
@@ -58,12 +58,16 @@
       var manual = document.getElementById('manual');
       var link = document.getElementById('link');
 
+      console.log('[tbd-open] script start; worktree param:', JSON.stringify(raw));
+
       if (!ok) {
+        console.log('[tbd-open] worktree param missing or invalid — aborting');
         status.textContent = 'Missing or invalid ?worktree= parameter.';
         return;
       }
 
       var target = 'tbd://open?worktree=' + encodeURIComponent(raw);
+      console.log('[tbd-open] computed target:', target);
       link.href = target;
       manual.hidden = false;
       status.innerHTML = 'Opening TBD for worktree <code>' + raw.replace(/[<&]/g, function (c) {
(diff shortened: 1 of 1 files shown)


cheapsteak:
Addressed the Medium finding in 6acb8739: `onOpened()` stores its timer id in `returnTimer` again, and the manual button click clears any pending timer before resetting `returned` and re-arming — a manual retry supersedes the in-flight auto-return, so only one `history.back()` can ever fire. Visibility changes still never cancel the return (the deliberate behavior change of this PR).

PR #443, opened 2026-07-13
Fix: Detect TBD launch via window blur so the deep-link auto-return works on big monitors

cheapsteak:
## Summary
- **What's broken:** the redirect page's auto-return (#416, #436) never fired for real launches. Field `[tbd-open]` diagnostics showed Chrome logging `Launched external handler for 'tbd://…'` while `visibilitychange` never fired: Chrome only flips `document.hidden` on tab-switch, minimize, or *full* occlusion, so when TBD's window doesn't completely cover the browser (large monitor, side-by-side), the visibility-only success detection silently missed every launch.
- **Fix:** accept `window.blur` as a second success signal — it fires when TBD takes app focus regardless of window geometry. `visibilitychange` stays. Blur fires before hidden even under full occlusion (focus loss precedes the occlusion recompute), so a later `hidden` inside the armed window *upgrades* a pending blur-scheduled return to the fast unconditional path rather than being ignored; `onOpened()` logs which signal drove the return.
- **Dialog-race hardening (from code review):** the "Open TBD.app?" confirmation dialog fires blur the moment it *opens*, so a fast return could execute mid-dialog and kill the launch. Blur-triggered returns therefore wait 4s (vs 600ms for `hidden`) and re-check `document.hasFocus()` at fire time — focus back on the page means the dialog was cancelled or the user deliberately returned, so the navigation is skipped, `returned` resets for a clean manual retry, and the status points at the button. Residual gap documented in-file: taking >4s to click "Allow" can still fire the return mid-dialog.

## Test plan
- [ ] Big-monitor / side-by-side setup with "always allow" ticked: click a redirector link from a PR page → TBD opens without covering Chrome → console shows `onOpened(): signal = blur` → tab returns to the PR ~4s later (previously: nothing)
- [ ] Full-occlusion / minimized case: `signal = hidden` still returns after 600ms as before
- [ ] First-run dialog path: cancel the "Open TBD.app?" dialog → no navigation; status offers the manual button; clicking it retries cleanly
- [ ] Console `[tbd-open]` trail names the winning signal and logs any skipped return

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=61966922-0E69-4FD6-9CF0-4FC4612C73EF)

### DIFF
--- a/docs/open/index.html
+++ b/docs/open/index.html
@@ -76,15 +76,28 @@
 
       // Success detection: navigating to a custom scheme does NOT unload this
       // page — the tab stays here. We can't observe the handler directly, but
-      // when TBD activates, macOS brings it frontmost and the tab becomes
-      // hidden. So: document.hidden flipping true within a short window of
-      // firing the redirect means the handler opened. window.blur alone is a
-      // false-positive signal — the browser's "Open TBD.app?" confirmation
-      // dialog blurs the window even if the user cancels — but that dialog
-      // does not hide the document, so visibilitychange is safe.
+      // when TBD activates, macOS brings it frontmost. We accept TWO signals
+      // within a short window of firing the redirect:
       //
-      // Accepted trade-off: document.hidden is the ONLY success signal we
-      // have, so if TBD is NOT installed, an incidental hide within the 3s
+      //   - visibilitychange → document.hidden: fires when the tab is
+      //     switched away, minimized, or fully occluded. Reliable when it
+      //     fires — but Chrome does NOT flip visibility when TBD's window
+      //     only partially covers the browser (the common big-monitor,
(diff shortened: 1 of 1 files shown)


cheapsteak:
Addressed the Medium finding in 0db72b60: you're right that blur wins the race even under full occlusion, so the 600ms hidden path was unreachable. A later `hidden` inside the armed window now preempts a pending blur-scheduled return — it cancels the 4s timer and reschedules via the shared `scheduleReturn()` helper on the fast unconditional path. Hidden-first and already-fired/skipped orderings keep the early-return guard. (The Test job failure was the known-flaky ControlModeInputRouter FIFO suite, unrelated to this docs-only diff; this push re-runs it.)

PR #444, opened 2026-07-13
nightwatch: verify PR state live before waking hibernated terminals (wake.py)

zionts:
## Problem

Nightwatch wake prompts were composed from hibernation-time snapshots. On 2026-07-13 a session was woken with "PR #14203 (temp-fix) has FAILING checks" — the PR had merged three days earlier and the branch was a local-only leftover. A verified sweep found **24/24 currently-hibernated terminals were actually DONE**: every hand-composed "resume" wake would have carried a stale premise.

## Fix

New `scripts/wake.py` (Tier-0, deterministic, no model) — the mandatory pre-wake verifier:

- Any PR number appearing in the hibernate reason is checked live (`gh pr view --json state,headRefName`); MERGED/CLOSED never yields a "fix your PR" wake.
- The checked-out branch's real open PR + failing checks and unmerged-commit state vs `origin/main` (`git cherry`, squash-aware) are re-derived at wake time.
- Classifies DONE → `/closeout` · RESUME_FAILING / RESUME_OPEN · UNSUBMITTED_WORK · UNVERIFIED → neutral "verify first" wake. Wake text is composed from verified facts only; anything unverifiable wakes neutral, never with an asserted premise.
- `--act` dispatches via `tbd terminal send --submit` and logs `wake-dispatch` to `queue/acted.jsonl`; dry-run writes `queue/wake-plan.json`.

SKILL.md gains a "Waking hibernated terminals" section mandating wake.py for every wake; `PluginDirWriter` ships it executable alongside tick.py/judge.py; test asserts it's written.

## Validation

- Incident repro: fed the exact stale reason against the affected worktree → classification flips to a stale-premise fact ("#14203 is MERGED (head keystone-14171)") + finish-or-closeout wake, no phantom-failure claim.
- Full-fleet dry run over all hibernated terminals: 24/24 verified DONE.
- Embedded Swift literals byte-match the live `~/.claude/skills/nightwatch` files; both files `swiftc -parse` clean (full `swift build` fails on main for unrelated bare-toolchain reasons, 679 pre-existing errors).

Live installs (`~/.claude/skills/nightwatch` + the AppSupport plugin dir) already carry the same wake.py/SKILL.md, so the fix is active now; this PR makes it survive the next plugin rewrite.

### DIFF
--- a/Sources/TBDCLI/Commands/TerminalCommands.swift
+++ b/Sources/TBDCLI/Commands/TerminalCommands.swift
@@ -158,6 +158,9 @@ struct TerminalWake: AsyncParsableCommand {
     @Option(name: .long, help: "Terminal ID")
     var terminal: String
 
+    @Option(name: .long, help: "Prompt delivered to the resumed claude as an argv (atomic with the respawn). NOT delivered when the wake is a no-op — check `woken` in the output.")
+    var prompt: String?
+
     @Flag(name: .long, help: "If the pinned account profile no longer exists, resume on the default login instead of failing")
     var fallbackToDefaultProfile = false
 
@@ -170,18 +173,22 @@ struct TerminalWake: AsyncParsableCommand {
         }
 
         let client = SocketClient()
-        try client.callVoid(
+        let result: TerminalWakeResult = try client.call(
             method: RPCMethod.terminalWake,
             params: TerminalWakeParams(
                 terminalID: terminalID,
-                fallbackToDefaultProfile: fallbackToDefaultProfile ? true : nil
-            )
+                fallbackToDefaultProfile: fallbackToDefaultProfile ? true : nil,
+                prompt: prompt
+            ),
+            resultType: TerminalWakeResult.self
         )
 
         if json {
-            printJSON(["status": "awake"])
+            printJSON(["woken": result.woken])
         } else {
-            print("Terminal awake.")
+            print(result.woken
+                ? "Terminal woken."
+                : "Terminal already awake (no-op)\(prompt != nil ? " — prompt NOT delivered" : "").")
         }
     }
 }
--- a/Sources/TBDCLI/Commands/TerminalCommands.swift
+++ b/Sources/TBDCLI/Commands/TerminalCommands.swift
@@ -6,7 +6,7 @@ struct TerminalCommand: ParsableCommand {
     static let configuration = CommandConfiguration(
         commandName: "terminal",
         abstract: "Manage terminals",
-        subcommands: [TerminalCreate.self, TerminalList.self, TerminalSend.self, TerminalOutput.self, TerminalConversation.self, TerminalFocus.self, TerminalPin.self, TerminalUnpin.self, TerminalSwapProfile.self]
+        subcommands: [TerminalCreate.self, TerminalList.self, TerminalSend.self, TerminalWake.self, TerminalOutput.self, TerminalConversation.self, TerminalFocus.self, TerminalPin.self, TerminalUnpin.self, TerminalSwapProfile.self]
     )
 }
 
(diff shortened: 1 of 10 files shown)


cheapsteak (approved):

zionts:
Addressed all four findings in 5532e86 (verified against the live DB and a real fleet dry-run):

- **PR-number check was dead code** — confirmed: `hibernateReason` in `~/tbd/state.db` only ever holds `auto`/`recovery`/`''`. `hibernated()` now also selects `suspendedSnapshot` (via `sqlite3 -json` — the snapshot is multi-line, so the old tab-separated parsing would have broken on it), strips ANSI, and scans the 4KB tail for PR numbers (deduped, capped at 8 `gh` calls). Live dry-run now produces e.g. `#15287 is MERGED (head adam/26/07/turn-artifacts-bag) — hibernate-time premise is stale` from real snapshots.
- **Fail-open on `git cherry` failure** — a cherry failure now records `unmerged-commit state NOT verified (git cherry failed)` and the classification stays UNVERIFIED; the DONE fall-through is gone.
- **Silent `[]` on unparseable `gh pr list` output** — now records `gh output unparseable — PR state NOT verified` and returns UNVERIFIED.
- **Dropped Operating rules section** — restored verbatim into SKILL.md (between "The gate" and "Config", matching the original order). That was an unintended casualty of syncing the embed from my live `~/.claude` copy, which had diverged; not a deliberate deletion.

Also reworded the "squash-aware" `git cherry` comment per the filtered feedback (patch-id matching isn't; merged-PR state carries DONE detection).

zionts:
Addressed in 9c26678 — the reviewer was right, and it invalidated one of my own recorded beliefs (I had noted "terminal send --submit auto-wakes hibernated terminals"; tracing `handleTerminalSend` confirms it never checks `hibernatedAt` and pastes into the post-hibernation bare shell).

- **New `tbd terminal wake --terminal <ID>`** subcommand wrapping RPC `terminal.wake` (idempotent per the handler's `.ok()` on `notHibernated`/`inFlight`; optional `--fallback-to-default-profile` mapping to the param).
- **wake.py `--act` wakes first**, waits 8s for claude to be input-ready (the daemon's own post-wake session recapture allows 5s to settle), then sends. A failed wake skips the send and records `ok: false` — notably this also fails closed on an **old tbd binary** that lacks the subcommand (ArgumentParser exits non-zero → no dead-pane paste, no false success), so the script is safe to ship ahead of the binary.
- **Minor #1**: `hibernated()` now filters `hibernatedAt IS NOT NULL OR suspendedAt IS NOT NULL`, mirroring `Terminal.isParked`, so legacy-parked rows can't bypass the verifier. Live dry-run picks up 8 parked terminals where the old filter saw 4 — one legacy row classified DONE via its merged PRs.
- **Minor #2**: documented the deliberate silent-continue on per-mention PR lookups (snapshots also mention issue numbers, which `gh pr view` rejects; classification fails closed independently).

Caveat stated plainly: `--act` end-to-end (wake → settle → paste into a live resumed claude) is not yet exercised against a real hibernated terminal — the installed `tbd` binary predates the new subcommand, and I can't build the app from a bare toolchain. The fail-closed path (unknown subcommand → no send) is what runs today; the full path needs a smoke test after the next app build.

zionts:
Both Mediums addressed in 71b640f:

- **`gh pr checks` fail-open** — now fails closed, with one nuance beyond the suggested fix: `gh pr checks` exits **non-zero when checks are merely failing or pending** (with tabular output), so rc alone would misread a legitimately-failing PR as "command failed". The unverified branch triggers only on *empty output + rc≠0*; it records `failing-checks state NOT verified (gh pr checks failed)` and classifies RESUME_OPEN without asserting anything about check state.
- **`makedirs` ordering** — moved to the top of `main()`, before the `--act` loop's `acted.jsonl` appends, matching tick.py/judge.py.

Pyright clean; live fleet dry-run (35 parked terminals, 22 verified DONE) exercises the new branch structure.

zionts:
Addressed in 86de883 by dissolving the race rather than patching around it — the wake text now rides the wake itself:

- **High (wake→send race)** — `TerminalWakeParams` gains `prompt`, threaded through `HibernationCoordinator.wake()` into `ClaudeSpawnCommandBuilder`'s resume branch as a trailing argv: `claude --resume <id> <prompt>`. Delivery is atomic with the respawn, so the text reaches **only** a session this call actually woke. On the `.notHibernated`/`.inFlight` no-ops it is never delivered anywhere — a human waking the terminal mid-sweep gets nothing pasted, closing the race entirely (not just narrowing the window). `terminal.wake` now returns `{woken: bool}` and the CLI reports it; wake.py records `dispatched: false` with a "raced a live wake — prompt withheld" note. The app's `callVoidAsync` ignores the new payload, so no app-side change.
- **Medium (8s sleep on an unrelated timer)** — gone with the paste. No sleep, no readiness assumption: argv delivery has no input-readiness window at all.
- **Medium (no tests for the fail-closed guarantees)** — `wake.py --selftest`: an offline classification matrix (monkeypatched `sh`/`gh`, zero real git/gh/sqlite/tbd invocations) with 8 scenarios covering every guarantee from these review rounds — cherry failure never → DONE, gh failure/unparseable output → UNVERIFIED, empty `gh pr checks` never asserts "checks not failing" (including the rc≠0-with-output = genuinely-failing-checks distinction), snapshot-mentioned MERGED PR → stale-premise fact. `PluginDirWriterTests.wakePySelftest` extracts the embed and runs it in CI, so the guarantees are enforced going forward, not just claimed in prose.

Old-binary behavior stays fail-closed: a `tbd` without `--prompt` exits non-zero → no dispatch, `ok: false` in the audit log.

zionts:
Coverage gap closed in 8aafa22:

- **Builder layer** — `"resume + initial prompt"` mirrors the existing `"fresh + initial prompt"` test but asserts the **exact full command**, including `shellEscape`'s single-quote handling (`don't` → `'don'\''t'`), so a flag reorder or a dropped escape on the resume branch fails the suite.
- **RPC layer** — new `"RPCRouter terminal.wake prompt delivery"` suite: a parked terminal → `woken: true` and the recorded dry-run respawn carries `claude --resume sess-1 … 'verify live state first'`; an already-awake terminal (the raced-a-human-wake case) → benign success with `woken: false`, **zero tmux commands recorded**, prompt delivered nowhere.

That's the full chain under CI now: wake.py classification fail-closed matrix (`--selftest` via `PluginDirWriterTests`), argv construction (builder test), and end-to-end RPC delivery/withholding semantics (router tests).

PR #448, opened 2026-07-15
perf(worktree): speed up creation with fetch coalescing and parallel checkout

zionts:
## Why

Worktree creation felt slow, especially on large repos. Diagnosis on a large working tree: `git fetch` ~1s, `git worktree add` ~5s (10k+ files / 260MB+ checkout), plus a per-repo serialization lane that queues concurrent creates behind each other's fetches. Not machine-specific and not universal — it's these stacked costs, worst on the biggest working trees.

## What

Four levers:

1. **Phase timing instrumentation** — each phase (createdir, worktree-add, terminal-spawn, mark-active) wrapped with `ContinuousClock` timing, logged via `os.Logger` (`com.tbd.daemon` / `worktreeTiming`) at `.debug`. Makes future regressions measurable.

2. **Fetch coalescing (`FetchCache` actor)** — per-repo TTL memo (60s) + singleflight. Skip the fetch if one succeeded within the window; N concurrent creates on the same repo share one in-flight fetch. Tight 20s timeout (vs 120s default) so a hung fetch can't stall creation. Failures are **not** cached, so retries still happen.

3. **Fetch moved out of the serialized lane** — runs *before* `repoSerializer.submit`, so back-to-back creates no longer queue behind each other's fetches. Only `git worktree add` + DB writes stay serialized (they need `.git/index.lock`). Best-effort semantics preserved: fetch failure/timeout must never fail creation.

4. **git parallel checkout** — `-c checkout.workers=0` on all `worktree add` variants; git auto-detects cores to write the working tree faster.

## Tests

New `FetchCacheTests` (7 tests) covering every branch of the TTL/singleflight/failure gating: cold start, within-TTL skip, past-TTL refetch, failure-not-cached (retry), failure-doesn't-throw, singleflight (5 concurrent → 1 fetch), multi-repo independence. Injection seams (`now` clock closure + `performFetch` closure) added so the cache is testable without real git.

`swift build` clean; `swift test --filter FetchCache` green.

## Not in scope

The next lever would be a warm "template" worktree to amortize the ~5s checkout floor — bigger design change, deferred.

### DIFF
--- a/Sources/TBDDaemon/Git/GitManager.swift
+++ b/Sources/TBDDaemon/Git/GitManager.swift
@@ -153,14 +153,14 @@ public struct GitManager: Sendable {
         return trimmed.isEmpty ? nil : trimmed
     }
 
-    /// Fetches from origin for the given branch.
-    public func fetch(repoPath: String, branch: String) async throws {
-        _ = try await run(arguments: ["fetch", "origin", branch], at: repoPath)
+    /// Fetches from origin for the given branch, with optional timeout override.
+    public func fetch(repoPath: String, branch: String, timeout: Duration? = nil) async throws {
+        _ = try await run(arguments: ["fetch", "origin", branch], at: repoPath, timeout: timeout)
     }
 
-    /// Fetches all refs from origin.
-    public func fetch(repoPath: String) async throws {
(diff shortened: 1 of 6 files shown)


cheapsteak (approved):
🙏

PR #454, opened 2026-07-16
feat(usage): surface the weekly reset and per-model (Fable) rows in the compact meter

zionts:
## Why

Two gaps in the compact usage meter (the popover on "New worktree with…"):

1. **You couldn't see when the week ends.** `resetsAt` was always in the snapshot and the countdown string was always computable, but it only surfaced through the macOS `.help()` hover tooltip — slow to appear, pixel-precise to trigger, with no hint it existed. The roomier `AccountPickerSheet` shows `resets 23:10` inline for the 5h row but explicitly passes `showsReset: false` for the week row, so the weekly reset was the one number never visible at rest anywhere.
2. **No Fable row.** `UsageBarsView` never iterated `scopedBuckets(snapshot)`, even though `AccountPickerSheet` already rendered them — hence Fable being findable in settings but not here.

## What

```
5h: ███░░░░░░░  38%
wk: █░░░░░░░░░   8%  · 2d 5h
F:  ██░░░░░░░░  15%
```

- A tertiary 9pt `· 2d 5h` countdown on the `wk:` row, reusing the existing compact `relativeResetText` formatter.
- A `ForEach` over `scopedBuckets`, labeled via `familyAbbreviation` (`F:`) in the existing 22pt label column, with `"Fable weekly"` spelled out in the tooltip.

## Design notes

- The meter's discipline is fixed-width columns so bars stay aligned across rows and accounts. The new trailing column is **reserved on every row** (46pt) and left blank where unused, so the `wk` bar doesn't end up narrower than the `5h` bar.
- Weekly buckets share one reset instant, so the countdown appears once — on `wk:`, not repeated on `F:`. That required decoupling "show the inline countdown" from "phrase the tooltip relatively": a naive reuse of `usesRelativeReset` would either duplicate the countdown on `F:` or show a misleading absolute clock for a reset days away. Hence the new `showsInlineReset`.
- The `.help()` tooltip is unchanged; the redundancy with the inline hint is intentional.

## Testing

- `swift build` passes.
- All 71 `ProfileUsagePresentation` tests pass. No new tests: the changes reuse already-tested formatters and the added gate is display-only.

### DIFF
--- a/Sources/TBDApp/Components/UsageBarsView.swift
+++ b/Sources/TBDApp/Components/UsageBarsView.swift
@@ -1,12 +1,15 @@
 import SwiftUI
 import TBDShared
 
-/// A compact two-row usage meter for a Claude OAuth profile: the 5-hour session
-/// window and the weekly all-models window, each drawn as a thin bar with a
-/// pace-aware colored fill (green/yellow/orange/red — projection of end-of-window
-/// usage from the current burn rate, floored by the API severity so a
-/// warning/critical bucket never reads healthy), a neutral "time marker" tick
-/// showing how far through the window we are, and a trailing percent.
+/// A compact usage meter for a Claude OAuth profile: the 5-hour session
+/// window, the weekly all-models window, and per-model-family weekly windows,
+/// each drawn as a thin bar with a pace-aware colored fill (green/yellow/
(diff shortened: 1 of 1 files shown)


cheapsteak (approved):
🙏 great additions

(will take a crack at shifting the layout a bit after this merges)

PR #471, opened 2026-07-21
fix(daemon): reboot-robust supervision — verify pid is a live TBDDaemon, eager launch ensure

zionts:
## The "disconnected on restart" root cause

After a macOS-driven relaunch (reboot + Spotlight, OS "quit and reopen"), the daemon could be dead for minutes while the app showed "disconnected". Cause: the daemon's stale-file cleanup (`PIDFile.isStale`) and its already-running gate (`Daemon.start`) both used a bare `kill(pid, 0)`. After a reboot the dead daemon's recorded pid can be **recycled by an unrelated process** — `kill` then reports "alive", so:
- `cleanupIfStale()` leaves the dead socket in place, and
- the existing-pid gate `exit(1)`s thinking a daemon already runs.

Every freshly-spawned daemon aborts until that recycled pid frees. Meanwhile the app's launch path only reaches its cleanup-aware respawn after several failed poll cycles.

## Fix

- **`TBDShared.ProcessLiveness.isLiveNamedProcess(pid:name:)`** — `kill(pid,0)` AND a `proc_pidpath` executable-name check; the canonical "is this really our process". The app's `DaemonLiveness` now delegates to it (it was a private duplicate of the same libproc logic); the daemon reuses it across the module boundary (it can't import TBDApp).
- **`PIDFile.isStale()`** and **`Daemon.start`'s existing-pid gate** now require a *live TBDDaemon* — a recycled pid is correctly stale → socket cleaned → the fresh daemon binds instead of aborting.
- **`AppState.init`** — if the launch connect fails, run the cleanup-aware `startDaemonAndConnect()` immediately instead of waiting for the 2s poll.

## Tests

`ProcessLivenessTests` (dead / recycled-non-daemon / live, via injection seams) and `PIDFileStaleTests` (missing / live / recycled) — the recycled-pid case is the regression this fixes. Existing `DaemonLivenessTests` still pass via the delegation. `swift build` + these suites verified locally.

### DIFF
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -727,6 +727,14 @@ final class AppState: ObservableObject {
         if !Self.isRunningUnderTests {
             Task {
                 await connectAndLoadInitialState()
+                // Eager ensure: a macOS-driven relaunch (reboot + Spotlight, OS
+                // "quit and reopen") often finds the daemon dead with stale
+                // socket/pid files. `connectAndLoadInitialState`'s plain connect
+                // can't clear those, so recovery would otherwise wait for the
+                // 2s poll to route to the cleanup-aware spawn. Run it now.
+                if !isConnected {
+                    await startDaemonAndConnect()
+                }
                 startPolling()
             }
         }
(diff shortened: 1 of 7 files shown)


cheapsteak (approved):
🙏 🙏

PR #472, opened 2026-07-21
Fix: Command-click reuses the transcript pane (viewer slot) + per-pane back/forward history

cheapsteak:
## Summary

**Problem:** command-clicking a file path while a transcript pane was open split a *third* column instead of reusing the pane the user was already looking at — every navigation gesture grew the layout.

**The viewer-slot model:** viewer-class panes (`codeViewer`, `liveTranscript`, `webview`) now form one interchangeable "slot" per tab. Content-navigation gestures (command-click, transcript toggle) replace content *in* the slot, preserving the pane UUID — so SwiftUI view identity (including the `FileWatcher` `@StateObject`) stays stable. Explicit split gestures still create new panes.

**Per-slot history:** each slot gets back/forward history (`PaneHistory`, cap 10, UserDefaults-persisted, pruned on pane close / tab close / reconcile). The shared pane header shows back/forward chevrons with press-and-hold jump menus for codeViewer, liveTranscript, and webview panes. Navigation moves the cursor only — it bypasses the routers, so back/forward never re-pushes entries.

**Review-loop fixes** (second commit, from a 5-agent review pass):
- Transcript toggle-off on a reused slot restores the nearest non-transcript history entry in place instead of destroying the pane (toggle round-trip no longer eats the viewer the transcript replaced; chained transcript entries are skipped so closing one transcript never resurrects another's).
- `closeTab` prunes pane histories immediately instead of waiting for the next reconcile.
- `restorePaneHistories` drops decoded histories with an out-of-range cursor — previously a corrupt/skewed persisted blob would crash `entries[cursor]` in the pane header's view body (launch crash loop).

No feature flag: gesture-driven additive UI per CLAUDE.md policy (no autonomous action, no destructive state mutation, no wholesale path replacement).

**Known follow-up candidate (deliberately out of scope):** the `openFilePreview` closures (file clicks from transcript tool cards) still bypass the slot/history machinery — pre-existing side-by-side preview UX, untouched here.

Design home: `docs/specs/2026-07-20-panel-agent-surface-a-mirror-design.md` §7 (on branch `panel-system-redesign`).

## Test plan

- [x] `swift build` clean
- [x] `swift test`: 3685 tests / 440 suites green (includes new PaneHistory unit + AppState integration suites covering cursor semantics, forward truncation, cap, prune, toggle-off restore, malformed-persistence guard)
- [x] `swiftlint --strict`: 0 violations
- [ ] Live check: command-click with transcript open swaps in place; chevron press-and-hold shows jump menu

[viewer-slot-history](https://cheapsteak.github.io/tbd/open/?worktree=308213B4-EA0D-4A90-8381-A31157D1B083)

## Iterations from live testing

- **History is a tab-like MRU list, not a browser stack** (`0f032883`): no forward-truncation — A → B → back → C keeps all three. Back/forward/jump move a cursor without reordering; only a new navigation commits (current entry to front, new entry ahead of it, dedupe by moving, cap 10 evicting the tail — never the entry you're on). Both chevrons' menus show the same full list with a checkmark on the current entry.
- **Right-click opens the history menu directly** (`14a22158`): chevrons are plain Buttons (left-click steps) with the MRU list as a true `.contextMenu`; the header's file context menu is rescoped to the title label so it no longer swallows right-clicks on the controls.
- **Stable chevron position** (`4c510537`): back/forward moved to the header's leading edge in fixed 16×16 slots (rendered disabled when no history), so the presence of pane-specific trailing buttons (`</>`, copy-URL) never shifts them. Glyphs sized 8pt bold to sit visually with the ✕; hit areas unchanged.

### DIFF
--- a/Sources/TBDApp/AppState+Tabs.swift
+++ b/Sources/TBDApp/AppState+Tabs.swift
@@ -164,6 +164,7 @@ extension AppState {
         arr.remove(at: index)
         tabs[worktreeID] = arr
         worktreeTabOrders[worktreeID] = arr.map(\.id)
+        prunePaneHistories()
 
         for terminalID in terminalIDsInTab {
             Task {
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -914,19 +914,22 @@ final class AppState: ObservableObject {
     }
 
     /// Resolves the removal side of a viewer route (transcript toggle-off).
-    /// A reused slot keeps its pre-transcript content in history: step back
-    /// and return the content to restore in place — the toggle round-trip
-    /// must not destroy the viewer the transcript replaced. With no back
-    /// history the pane is really going away, so forget its history.
+    /// A reused slot keeps its pre-transcript content in history: jump the
+    /// cursor to the nearest non-transcript entry (older side first — that's
+    /// what the transcript replaced — then newer) and return it to restore in
+    /// place. Skipping transcript entries means chained
+    /// transcript-for-another-terminal swaps never resurrect another
+    /// terminal's transcript. With no non-transcript entry the pane is
+    /// really going away, so forget its history.
     func popHistoryForRemovedPane(_ paneID: UUID) -> PaneContent? {
         if var history = paneHistories[paneID] {
-            // Toggle-off means "stop showing a transcript": skip past older
-            // transcript entries (chained transcript-for-another-terminal
-            // swaps) to the nearest non-transcript content.
-            while let previous = history.goBack() {
-                if case .liveTranscript = previous { continue }
+            let older = Array((history.cursor + 1)..<history.entries.count)
+            let newer = Array(stride(from: history.cursor - 1, through: 0, by: -1))
+            for index in older + newer {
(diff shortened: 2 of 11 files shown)


cheapsteak:
Heads-up on 0f032883: pane history is now a **tab-like MRU list, no forward-truncation**.

- `entries[0]` = most recently committed, `cursor` = currently-viewed entry. Back/forward/jump only move the cursor — nothing is ever dropped by going back.
- A real navigation (file click, transcript toggle) commits the current entry to the front, dedupes the incoming content (one entry per content — it moves rather than duplicates), inserts at index 0, caps at 10 by evicting the tail (never the current entry).
- Both chevrons' menus show the same full list with a checkmark on the current entry; selecting jumps without reordering.
- Transcript toggle-off restore re-derived for MRU: jumps to the nearest non-transcript entry (older side first) — still never resurrects another terminal's transcript.
- Also from feedback: chevron glyphs now match the close button's size spec (9pt bold, 16x16 frame + contentShape so the hit area is unchanged).

Persisted shape {entries, cursor} is unchanged — old data decodes, no migration.

PR #480, opened 2026-07-23
Fix: Forked Claude tabs no longer share the original's session UUID

cheapsteak:
## Summary

**What's broken:** "Fork Session" spawned a new tab that shared the *same* session UUID as the source. Two live `claude` processes then appended to the same session JSONL, risking interleaved/corrupted transcripts, and the DB pointed both terminal rows at one session.

**Why it happens:** The fork path spawns `claude --resume <id>` and relied on Claude Code's old behavior where `--resume` forked into a fresh session file. Current Claude Code (2.1.x) *reuses* the original session ID on `--resume` and only mints a new one with the explicit `--fork-session` flag. So `scheduleSessionRecapture` dutifully "recaptured" the identical ID.

**How it got broken:** Latent — the code was correct when written; a Claude Code CLI semantics change (bare `--resume` stopped forking) turned it into a bug. No test caught it because the builder was never asserted against the live CLI's fork behavior, and the fork-mode swap tests exercise blank sessions (which take the fresh-spawn path, not resume).

**What this PR does:** Adds a `forkSession` flag to `ClaudeSpawnCommandBuilder` that appends `--fork-session` on the resume path, and passes `forkSession: mode == .fork` from the swapProfile handler. `.inPlace` ("Switch account") deliberately does *not* fork — same-ID resume is correct there since the original process is killed. Recapture then picks up the genuinely-new ID on the fork path as originally designed.

Out of scope (left untouched): hibernation wake and history/terminal.create resume — same-ID resume is intended for those (original process is dead).

## Test plan
- [x] `swift build`
- [x] `swift test --filter ClaudeSpawnCommandBuilder` (46 tests) — new `resumeWithForkSession` / `resumeDefaultOmitsForkSession` cover both branches
- [x] `swift test --filter ModelProfileSpawnTests` (28) + `TerminalSwapPlanTests` (5)
- [ ] Live: Fork a non-blank Claude session; confirm the new tab's session UUID differs from the source and both transcripts stay independent

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=26E4DAB9-CE05-4E06-B573-37395C34CBC4)

https://claude.ai/code/session_012NPnXFcgreWBNqJVRSPiNt

### DIFF
--- a/Sources/TBDDaemon/Claude/ClaudeSpawnCommandBuilder.swift
+++ b/Sources/TBDDaemon/Claude/ClaudeSpawnCommandBuilder.swift
@@ -13,6 +13,10 @@ import TBDShared
 /// - `resumeID` non-nil → `claude --resume <id> --dangerously-skip-permissions`
 ///   with optional trailing initial-prompt arg (delivered atomically with the
 ///   resume — no post-spawn paste, so it can never land in the wrong process).
+///   With `forkSession: true`, ` --fork-session` is appended: Claude Code's
+///   `--resume` REUSES the original session ID unless that flag is passed, and
+///   the `.fork` swap mode needs a genuinely new ID because the source session
+///   stays live (otherwise both processes write the same session JSONL).
 /// - `freshSessionID` non-nil → `claude --session-id <id> --dangerously-skip-permissions`
 ///   with optional `--append-system-prompt` and trailing initial-prompt arg.
 /// - Otherwise → `cmd` if set, else `shellFallback`.
@@ -47,6 +51,7 @@ enum ClaudeSpawnCommandBuilder {
 
     static func build(
         resumeID: String?,
+        forkSession: Bool = false,
         freshSessionID: String?,
         appendSystemPrompt: String?,
         initialPrompt: String?,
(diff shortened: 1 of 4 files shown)


cheapsteak:
Addressed the Medium finding in 04c3afec — added a router-level regression test in `ModelProfileSpawnTests` that closes the exact gap you flagged.

Two tests, driving `terminal.swapProfile` through `handleTerminalSwapProfile` over a **non-blank** session (seeded via a real JSONL transcript so `planTerminalSwap` picks `.resume`, the branch where the flag lives), capturing the spawned tmux command with the existing `TmuxRecorder` fixture:

- `forkOverNonBlankEmitsForkSessionFlag` — `.fork` mode → asserts `claude --resume` **with** `--fork-session`
- `inPlaceOverNonBlankOmitsForkSessionFlag` — `.inPlace` mode → asserts `claude --resume` **without** `--fork-session`

Reused `makeFixture`/`TmuxRecorder`/`seedRepoAndWorktree`/`seedOAuthProfile` plus `SwapTranscriptCarryTests`' transcript-seeding pattern — no new mocks, no `setenv` (both sit under `TBDHomeSerialized`). The two share an identical setup differing only in `mode`, so they prove the flag is genuinely mode-driven. Suite went 28 → 30, all green.

PR #481, opened 2026-07-23
docs: land panel agent-surface specs

cheapsteak:
Approach C (daemon-owned primary-anchor layout) is the adopted design for TBD's panel agent surface, superseding approaches A (mirror) and B (daemon-owned split); A and B are kept here as decision records. Phase 1 of C has already merged as #478; Phase 2 planning is underway. Landing all three specs on main so implementers and reviewers can cite them directly. No code changes.


cheapsteak:
This is a backfill of spec docs that weren't committed in 478

PR #491, opened 2026-07-24
test: isolate tier-3 live suites in TBDDaemonLiveTests + split CI into fast/quiet passes

cheapsteak:
**Slice A** — Stage 0 of [`docs/specs/2026-07-24-test-hardening-design.md`](docs/specs/2026-07-24-test-hardening-design.md). Wave 0; every other slice is gated on this merging.

**This is test-infrastructure only — no production code is touched, so the default-off flag policy in the root `CLAUDE.md` does not apply** (slicing §1 says so explicitly for this program). `git diff --stat` shows zero files under `Sources/`.

## What this fixes

PR #488's three CI flakes turned out to be three *structural* defects, not three unlucky tests. This slice kills the second one: **deadline-sensitive tests share a starved machine with heavyweight tests.** All test targets compile into one process and Swift Testing runs suites in parallel across all of them, so live-tmux suites, subprocess-deadline suites, and the replay firehose contend for the same 3–4 runner cores (at `-j 2`, the OOM floor) while timeout-sensitive tests tick. Nothing marked "needs real tmux and a quiet CPU" vs "pure unit test", so CI could not schedule them apart.

The enforcement mechanism is a **target boundary**, not a naming convention: compiler-enforced, and it cannot silently zero-match the way a `--filter` regex can.

## Curated tier-3 inventory (the judgment call, slicing §6)

A coarse grep over `Tests/TBDDaemonTests/` matches ~71–89 files because `Process()` is used far too broadly to be a signal. This is a curated list, not a grep dump — every entry was verified by confirming an actual `tmux(["-L", server, "new-session", …])` or a real-mode `TmuxManager()` reaching `ensureServer`, as opposed to `TmuxManager(dryRun: true)`.

**Total: 18 files, 54 tests, 10.3 s serial.**

### A. Real tmux server — 13 whole files

| File | tests |
|---|---|
| `ReplayLiveIntegrationMatrixTests` | 5 |
| `ControlModeInputRouterIntegrationTests` | 5 |
| `TmuxControlCommandClientIntegrationTests` | 5 |
| `PasteExecutorIntegrationTests` | 3 |
| `ControlModeResizeCoordinatorIntegrationTests` | 3 |
| `TerminalSendBracketedPasteLiveTests` | 2 |
| `LimitResumeSendSequenceLiveTests` | 2 |
| `PaneStateCaptureIntegrationTests` | 2 |
| `TmuxControlConnectionTeardownTests` | 2 |
| `PhaseTwoIntegrationTests` | 1 |
| `TmuxControlConnectionIntegrationTests` | 1 |
| `AttachReplayOrchestrationIntegrationTests` | 1 |
| `HistoryLimitIntegrationTests` | 1 |

Five of these were **not** in the spec's seed list (`TmuxControlConnectionTeardownTests`, `PhaseTwoIntegrationTests`, `ControlModeResizeCoordinatorIntegrationTests`, `PasteExecutorIntegrationTests`, `PaneStateCaptureIntegrationTests`) — all confirmed to spawn a real server with a `defer kill-server`.

### B. Spawned children racing a deadline — 4 files

| File | tests | why |
|---|---|---|
| `SubprocessTimeoutStarvationTests` | 2 | **deliberately saturates the shared default-QoS GCD pool** — a contention *source* that degrades every sibling suite in the one-process run, not merely a victim |
| `GitManagerTimeoutTests` | 6 | spawns `/bin/sleep 30` through the git timeout seam |
| `SubprocessTimeoutTests` | 5 | spawns `/bin/sleep 30`, asserts a 100 ms watchdog fires |
| `HibernationOrphanDetectionTests` | 3 | spawns a fake-`ps` child emitting ~110 KB through a 64 KB pipe under a bounded-runner deadline |

These match design §3's tier-3 row ("…, spawned processes, …") but not slicing §6's narrower restatement. Design §3 is authoritative, and §6's criterion line is amended here to match — a victim-only reading would have left the single worst offender (`SubprocessTimeoutStarvationTests`) in the parallel pass.

### C. Partial file — 5 tests extracted from `WorktreeLifecycleTests`

`WorktreeLifecycleTests.swift` is ~1300 lines / 35 tests; 30 use `TmuxManager(dryRun: true)` and are tier 2. Exactly five build a real-mode `TmuxManager` and call `ensureServer`:

`testReconcileRebootParksClaudeAndDeletesShell`, `testReconcileDeadWindowClaudeTerminalSuspended`, `testReconcileDeadWindowShellTerminalDeleted`, `testReconcileDeadWindowCodexTerminalWithSessionMetadataDeleted`, `testReconcileDeadWindowLiveClaudeNotParked`

Those five moved to `WorktreeLifecycleLiveTests.swift`; the other 30 stay parallel. The spec listed this as a whole-file move — splitting it applies the failure asymmetry correctly, since exiling 30 tier-2 tests to the serial pass would tax every PR forever.

### Deliberately not moved

Under-inclusion is the safer error, so: every `TmuxManager(dryRun: true)` suite; the 30-test `WorktreeLifecycleTests` residue (real git subprocesses = tier 2 per design §3); `ReplayWriterTests` (pure byte assembly).

**Correcting the spec:** design §3 claims live suites might exist outside the daemon. They don't. `TmuxBridgeTests` in `TBDAppTests` looked like a real-tmux suite but asserts on the argv *arrays* the bridge builds — `"new-session"` is a string literal inside an expected-argv assertion, the same false-positive class as `MigrationV17Tests`. Swept the rest: `TBDAppTests` and `TBDCLITests` spawn zero subprocesses, and `TBDSharedTests`' `ProcessLivenessTests` drives injected seams. So no `TBDAppLiveTests` target is needed and "all current live suites are daemon-side" holds.

## CI split

Two sequential steps in the **same job** — a second job would draw from the 5-concurrent-macOS-job cap and re-pay the ~2 min build:

1. `swift test --parallel -j 2 --skip '^TBDDaemonLiveTests\.'`
2. `swift test --filter '^TBDDaemonLiveTests\.'` — no `--parallel`, `timeout-minutes: 15` so a wedged tmux test cannot eat the job's 6 h default

Target-prefix matching was verified empirically rather than assumed (swift-testing IDs are `Module.Suite/testFn()`); no suite-name alternation fallback was needed.

**Both steps count-guard, not just the quiet one.** `swift test --filter/--skip` exits GREEN on zero matches, so a renamed target or a flag-semantics change could drop a pass to zero tests and CI would report coverage it never ran. That failure mode is symmetric: if `--skip` ever over-matches, it is the *fast* pass that silently empties. Floors are 3000 (fast, real 3859) and 35 (quiet, real 54) — low enough that ordinary test churn never trips them.

## Assertion hygiene → `Tests/CLAUDE.md`

The four rules from design §6, each with its flake provenance, plus the tier taxonomy and the target-boundary criterion. The PR reviewer bot reads this file, so these are enforced on every future PR. Rule 3 notes that the clock seam and `no_raw_task_sleep` lint rule land in slice B, so today it is review-enforced only.

## One incidental refactor

`PhaseTwoIntegrationTests` (moved) shares `SidecarTestSupport.receiveVend` with three files that stay. Only that one enum was promoted to a new `Tests/TestSupport/SidecarFrameHelpers.swift`; the rest of `Tests/TBDDaemonTests/SidecarTestSupport.swift` is used solely by staying files and did not move. No existing `TestSupport` file was edited, so slice B's and E's new files there won't conflict.

## Flaky test noticed, deliberately NOT fixed

`AppearanceDebounceTests` (`TBDAppTests`, untouched by this slice) failed 2 of 5 identical local fast-pass runs — a real-timer debounce assessed against wall-clock, i.e. flake class 1. That is **slice C1's** scope (clock-seam migration of the appearance debounce). Worth flagging for sequencing: this slice makes the fast pass *denser* (31 s → 28 s once the 54 live tests stop soaking runner time), which plausibly raises that flake's hit rate — C1 may be more urgent, not less.

## Verification

- `swift build` — green
- `swiftlint --strict` — 0 violations in 585 files
- fast pass **3859** + quiet pass **54** == unfiltered **3913** ✅, all three runs green

The sum is the load-bearing check: it proves no suite fell out of *both* passes.


cheapsteak:
## Follow-up commit: the reviewer bot's unverified premise was a real defect

The review flagged one premise it could not check from its environment:

> whether dropping SwiftPM's `--parallel` flag actually serializes Swift Testing's own in-process scheduling for `TBDDaemonLiveTests`, vs. only affecting XCTest-era multi-process bundling.

**It doesn't, and the caveat was right.** The bot's reasoning that the summing counts and the 10.3 s duration were "consistent with real serial execution" is where it went wrong — 10.3 s is consistent with *parallel* execution; serial is 2.4× slower.

Evidence — quiet pass as originally shipped, all 17 suites start before any finishes:

```
◇ Suite "Hibernation orphan detection" started.
◇ Suite GitManagerTimeoutTests started.
◇ Suite "terminal.send bracketed paste (live tmux)" started.
... 14 more "started" lines, zero "passed" lines in between ...
◇ Suite "PasteExecutor integration" started.
✔ Suite "Hibernation orphan detection" passed after 0.639 seconds.
✔ Test run with 54 tests in 17 suites passed after 10.283 seconds.
```

With `--no-parallel`, they strictly alternate:

```
◇ Suite "Attach replay orchestration integration" started.
✔ Suite "Attach replay orchestration integration" passed after 0.676 seconds.
◇ Suite "ControlModeInputRouter integration" started.
✔ Suite "ControlModeInputRouter integration" passed after 0.932 seconds.
✔ Test run with 54 tests in 17 suites passed after 24.714 seconds.
```

SwiftPM's `--parallel` governs XCTest's process-level sharding. Swift Testing parallelizes suites in-process on its own and ignores the *absence* of the flag — so `swift test --filter …` with no `--parallel` was still running every live tmux suite concurrently. The quiet pass was not quiet, which is the entire mechanism this slice exists to provide. (`--num-workers 1` is not an alternative: it errors with `--num-workers must be used with --parallel`.)

Fixed in `3f6e436`: `--no-parallel` added explicitly, with a comment at the call site explaining why omitting `--parallel` is not equivalent so nobody "optimizes" it away later. The same trap is written into `Tests/CLAUDE.md`.

**Honest cost update:** the quiet pass is 24.7 s, not 10.3 s. Total job time goes up ~15 s. That is the actual price of the isolation this PR is buying, and the earlier number was measuring the thing not working.

Counts are unchanged and still sum: fast 3859 + quiet 54 == unfiltered 3913. Floors unchanged.

PR #506, opened 2026-07-25
Fix: Desk prompts told every session to run a command that never existed (5 days, nothing respawned)

cheapsteak:
## Summary

`NightwatchDeskPrompts` builds the prompt injected into every watch-desk agent session on every tick. It is not documentation — it is live instruction driving autonomous behavior across a ~40-worktree fleet. Nothing asserted on the strings, and they had drifted badly.

**What was broken.** The context-ceiling remedy read ``Remedy: call `tbd terminal close --all` on the desk, let daemon respawn fresh``. Both halves were false: there is no `tbd terminal close` subcommand (the daemon has `terminal.delete`, but no CLI surface was ever built over it), and there is no babysitter daemon — no `~/.fleet/`, no launchd job, no script. Sessions were told to mark themselves for respawn and keep working; for five days nothing recycled them. Separately, the prompts said "Merge PRs cleared by the policy (all clearance kinds)" and "act on everything the gate approves", contradicting the nightwatch gate, which reserves merging for the human and treats `gh pr merge` as an escalation.

**Why it wasn't caught.** No test read these strings. `swift test` was green throughout.

**The second-order problem.** `PluginDirWriter.writeNightwatch` rewrites the installed skill dir from string constants on every daemon start, `atomically: true`, no existence check. So fixing the on-disk `SKILL.md` and `tick.py` is ephemeral — the next `scripts/restart.sh` reverts them. The compiled source still described `scripts/daemon.py` + `watchdog.sh` as a "durable spine already on launchd", and the compiled `tick.py` lacked the composer ghost-guard entirely.

## What this PR does

- **Prompts** point the context ceiling at the handoff relay (`handoff.py --check` / `--act --notes-file` / `--close-predecessor`), including the ordering rule that the predecessor never closes itself — a failed successor spawn must leave the desk still watched. Both modes now state the gate's enqueue precondition and forbid `gh pr merge`. Adds the never-`/closeout` standing rule, so it no longer has to override the tick prompt from outside.
- **Ships `handoff.py`.** It was not in the installed set, so the relay path resolved only on a machine where it had been hand-created. A prompt naming an uninstalled path is the same defect as naming a nonexistent command, one layer down.
- **Retires the babysitter claim** from the compiled `SKILL.md` and ports the composer ghost-guard into the compiled `tick.py` (dim `ESC[2m` suggestions and SGR mouse reports are no longer read as human input — that path types what it finds into live sessions). `DAEMON_LOG` dropped with the stub it fed.
- **Removes the duplication that allowed the drift**: `PluginDirWriter` now iterates `NightwatchSkillContent.scripts` instead of carrying its own list, making the shipped set visible to `TBDShared` tests.
- **Tier-1 test suite** pinning each of these as a policy invariant — including the structural guard that every `scripts/*.py` named in a desk prompt is actually installed.

Also includes a design doc for `tbd terminal close` (`docs/specs/2026-07-25-terminal-close-design.md`). It is the rationale for this branch and specifies the follow-up slice; **it adds no code and the command does not exist yet.**

## Test plan

- [x] `swift test` — 3876 tests, 460 suites, green (1 known issue: the quarantine self-test fixture, red once by design)
- [x] `swiftlint --strict` — 0 violations
- [x] Ported `tick.py` verified against the live fleet, not just by inspection: extracted the compiled string to a temp tree (so its queue writes land there, not in the live queue) and ran it — reproduces `stranded=5`, down from 14 before the ghost-guard, with all five survivors genuinely typed input
- [x] Splice verified byte-exact: compiled `skillMd` / `tickPy` / `handoffPy` diffed against their on-disk sources (`tickPy` differing only by the intended `DAEMON_LOG` removal)
- [ ] Reviewer sanity-check on the `tick.py` port specifically — it is transcribed work, and the index alignment in `classify` (`len(raw_lines) != len(lines)` → bail) is load-bearing: a misaligned lookup reads dimness off the wrong row

## Notes for the reviewer

- Two assertions in the first commit pinned deleted wording rather than the invariant; the third commit tightens them. Worth internalizing: a banned-substring check is defeated from *both* sides — a reworded reintroduction slips past, and the sentence that forbids a promise contains the promise. It tripped twice here before being replaced with checks on artifacts that have no innocent reading.
- The shipped `SKILL.md` still states that `tbd terminal close` does not exist. True today; the follow-up slice that adds the command must update it in the same change.
- `handoff.py --close-predecessor` still uses raw `tmux kill-window`, which per the design doc §1.2 does not actually close a Claude terminal — reconcile parks it `.recovery`-hibernated and focus-wake resurrects it. Known and deliberately deferred to the follow-up slice.

[terminal-close-design](https://cheapsteak.github.io/tbd/open/?worktree=F97E72C1-E4F7-4D95-8080-82813621C056)

### DIFF
--- a/Sources/TBDDaemon/Lifecycle/PluginDirWriter.swift
+++ b/Sources/TBDDaemon/Lifecycle/PluginDirWriter.swift
@@ -100,11 +100,7 @@ struct PluginDirWriter {
         // Scripts ship executable. NOTE: the scheduler (scheduler.sh / tick-cron.sh)
         // is installed but NEVER auto-loaded — durable scheduling is opt-in via
         // `scheduler.sh enable`.
-        for (name, body) in [("tick.py", NightwatchSkillContent.tickPy),
-                             ("wake.py", NightwatchSkillContent.wakePy),
-                             ("judge.py", NightwatchSkillContent.judgePy),
-                             ("tick-cron.sh", NightwatchSkillContent.tickCronSh),
-                             ("scheduler.sh", NightwatchSkillContent.schedulerSh)] {
+        for (name, body) in NightwatchSkillContent.scripts {
             let path = scripts + "/" + name
             try body.write(toFile: path, atomically: true, encoding: .utf8)
             try fm.setAttributes([.posixPermissions: 0o755], ofItemAtPath: path)
--- a/Sources/TBDShared/NightwatchDeskPrompts.swift
+++ b/Sources/TBDShared/NightwatchDeskPrompts.swift
@@ -33,7 +33,33 @@ public enum NightwatchDeskPrompts {
         • Single-driver rule: Before actioning any atlantis apply/merge, claim the item in queue/claims
         • Escalations in batches ≤4: Each question with exact PR#/command/recommendation
         • Capacity check: Before nudging others, verify profile usage <80% weekly cap
-        • Token ceiling: If ~200k tokens used, flag for session respawn instead of nudging
+        • Context ceiling: At ~200k tokens, run the handoff relay (below) — do NOT defer it and keep working
+        • The gate stops short of merging: get PRs *ready*, the human merges (never run `gh pr merge`)
+        • NEVER trigger `/closeout` — a finished-looking worktree is an archive question for the human
+
+        **Standing rules (set by Chang — these override anything else in this prompt):**
+
+        **NEVER trigger `/closeout`.** Not on a DONE pane, not on a "candidate for
+        /closeout" self-report, not because the composer already suggests it. A worktree
+        that looks finished is an *archive question for the human*, not a harvest to fire.
+        Report it and move on.
+
+        **Hand off at the context ceiling — never push through it.** There is no babysitter
+        daemon and no respawner; nothing will restart you. When this session passes ~200k
+        tokens it starts truncating its own shift, so use the relay:
+
+            python3 \(skillDir)/scripts/handoff.py --check
+                # exit 0 = under the ceiling · exit 10 = OVER
+
+            python3 \(skillDir)/scripts/handoff.py --act --notes-file <your-notes.md>
+                # writes the handoff doc and spawns a fresh successor in this worktree
+
+        The predecessor NEVER closes itself. The successor closes it, after it has read the
+        handoff:
+
+            python3 \(skillDir)/scripts/handoff.py --close-predecessor <predecessor-terminal-id>
+
+        If the spawn fails, you are still the only thing watching the desk — stay alive.
 
         **Next step:**
         Read the Nightwatch skill docs for the full merge-gate policy, clearance types, and acting procedures.
@@ -66,7 +92,7 @@ public enum NightwatchDeskPrompts {
         if mode == .daywatch {
             actionHint = "(daywatch: triage only; act on small_safe/preclear; batch rest for human review)"
         } else {
-            actionHint = "(nightwatch: act on everything the gate allows)"
+            actionHint = "(nightwatch: act on what the gate allows — the gate stops short of merging; the human merges)"
         }
 
         let queueDir = skillDir + "/queue"
@@ -81,8 +107,17 @@ public enum NightwatchDeskPrompts {
         2. Apply the policy per the skill docs
         3. For each decision:
            - Daywatch: act ONLY if clearanceKind in [preclear, small_safe]; otherwise batch for human review
-           - Nightwatch: act on everything the gate approves
+           - Nightwatch: act on what the gate allows — but NEVER merge (see the gate below)
         4. Write results to \(queueDir)/acted.jsonl (one JSON line per action taken)
+
+        **The gate (never automate past this):**
+        - A PR may only be enqueued when claude-review = APPROVED on the CURRENT SHA and
+          checks are clean. Human approval never substitutes for the bot verdict.
+        - NEVER run `gh pr merge`. It is not a safe wedge — it always escalates to the
+          human. Your job is to get PRs *ready*; the human merges.
+        - Re-read live PR state in the same breath as any send that asserts it
+          (`gh pr view N --json state,headRefOid,mergeStateStatus`); never dispatch a fact
+          older than the current tool call.
         5. Write a summary to \(queueDir)/judge-summary.txt
 
         **Field learnings — apply these rules:**
@@ -106,10 +141,23 @@ public enum NightwatchDeskPrompts {
         - Before invoking other workers (e.g., "nudge code-review worker"), verify profile usage is <80% weekly cap
         - If at/above 80%, skip nudging; instead, add to queue for next cycle
 
-        **Token ceiling for respawn:**
-        - If this session has used ~200k tokens, flag for respawn instead of nudging others
-        - Symptom: running slow, many retries, unclear LLM reasoning
-        - Remedy: call `tbd terminal close --all` on the desk, let daemon respawn fresh
+        **Context ceiling — hand off, never push through:**
+        - Symptom: running slow, many retries, unclear reasoning, truncated history
+        - Do NOT just mark the session for later recycling and keep working. That is a
+          no-op: there is no babysitter daemon and no respawner. Nothing will restart you.
+        - Check: `python3 \(skillDir)/scripts/handoff.py --check` (exit 0 = under, 10 = OVER)
+        - When OVER: write your handoff notes to a file, then
+          `python3 \(skillDir)/scripts/handoff.py --act --notes-file <your-notes.md>`
+          which writes the handoff doc and spawns a fresh successor in this worktree
+        - The predecessor NEVER closes itself. The successor closes it once it has read the
+          handoff: `python3 \(skillDir)/scripts/handoff.py --close-predecessor <tid>`
+        - If the spawn fails, stay alive — you are the only thing watching the desk
+
+        **NEVER trigger `/closeout`:**
+        - Not on a DONE pane, not on a "candidate for /closeout" self-report, not because
+          the composer already suggests it
+        - A worktree that looks finished is an *archive question for the human*, not a
+          harvest to fire. Report it and move on.
 
         **Act:** Use `tbd terminal send --submit` to confirm each action, or batch AskUserQuestion if unsure.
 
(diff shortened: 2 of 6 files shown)


cheapsteak:
Both findings addressed in f94f4831 — no pushback on either.

**Medium: ghost-guard had no executable test.** Correct, and the sharper version of the criticism is that my own test was named `tickShipsGhostGuard` while proving only that substrings existed in the shipped string. Added `tick.py --selftest` on the `wake.py --selftest` precedent you pointed at, wired through `PluginDirWriterTests.tickPySelftest` the same way (spawn `python3`, assert exit 0 + the marker line). 11 scenarios over `classify()`/`_composer()`: typed input strands, dim `ESC[2m` ghost text doesn't, SGR mouse reports don't, statusline is never a pane's last meaningful line, in-flight turn outranks the composer.

Two behaviours are pinned deliberately because they **fail open** — a misaligned raw/stripped line count abandons the dim lookup rather than reading dimness off the wrong row, and a caller passing no raw capture keeps only the mouse filter. Both let a ghost through. That's the right trade against silently suppressing real input at random, but it should be visible in the test rather than discovered later, so it's asserted explicitly with the reasoning in a comment.

I didn't treat the green run as proof the test bites. Mutation-checked it — breaking `DIM_RE`, `MOUSE_RE`, the `len(raw_lines) != len(lines)` alignment guard, or `STATUS_RE` each fails the scenario that covers it, and nothing else:

```
DIM_RE.search(raw_lines[j]) -> False   ✗ ghost text is not composer input: got 'please retry the deploy', want ''
MOUSE_RE.match(text)        -> False   ✗ mouse report is not input: got ('STRANDED', '<65;61;31M')
len(raw_lines) != len(lines) removed   ✗ misaligned raw capture is abandoned, not misread: got 'IDLE'
STATUS_RE.search(s)         -> False   ✗ statusline skipped: got 'Opus 5(high)  494k/1000k  5h(65% →20:00)'
```

One thing worth flagging that your review couldn't have seen: `--selftest` short-circuits **before** tick.py's machine-global flock at `/tmp/nightwatch-tick.lock`. Placed after it, a concurrently running live tick would take the lock and `main()` would exit 0 silently — the test would pass without executing a single scenario. The `composer ghost-guard scenarios passed` output assertion is the backstop either way.

**Minor: stale `tickPy` module docstring.** Fixed — it still described `daemon_health()` as reading real health data after it became an unconditional stub. You're right that it's exactly the class this PR exists to sweep; I'd noticed it during the port and left it for fidelity to the on-disk source, which was the wrong call once the compiled copy became the source of truth.

Also added a `PluginDirWriter` test that every `NightwatchSkillContent.scripts` entry lands on disk with matching content, closing the loop on the shipped-set/prompt cross-check from the shared side.

Re: the sandbox note — `python3` being blocked in review is worth knowing, since both selftests are Python spawned from Swift. They run in the normal test target (`wakePySelftest` has been doing this all along); locally `swift test` is green at 3878 tests / 460 suites, `swiftlint --strict` clean.

cheapsteak:
Addressed in f3ea8f19. Took option (b), not (a) — and the finding was sharper than a doc mismatch.

**It wasn't ambiguity between two sections; the skill was firing the harvest.** `wake.py`'s `compose()` put the literal string `Run /closeout now.` into the DONE wake, and `STANDING_RULE` opened with *"If the task is complete, run /closeout."* Those prompts are dispatched by `wake.py --act` into a hibernated session with **no human in the loop**. So the shipped skill was doing the exact thing the new rule forbids, through a second door. Two further branches offered it as an option (`RESUME_OPEN`, `UNSUBMITTED_WORK`).

Scoping the rule around `wake.py` (option a) would have meant inventing an exemption that was never granted — the rule is unconditional and came with a specific incident behind it ("I violated this earlier and left two panes wedged on harvest forms"). The classifier's rigour isn't the issue; `/closeout` is the human's call regardless of how confident the classification is.

Changed: all three `/closeout`-bearing branches now report and stop, `STANDING_RULE` carries the prohibition, plus the module docstring, the dirty-tree fact string, and `SKILL.md`'s classifier line (which advertised `DONE → /closeout`).

**Pinned executably**, in `wake.py`'s own selftest as scenario 9 — run from Swift by the existing `PluginDirWriterTests.wakePySelftest`. It asserts on `compose()`'s output for all five classifications, not on source text, so a reintroduction via `STANDING_RULE` or a new branch is still caught.

One design note on that assertion, because it's the third time this exact trap bit me on this branch: it **whitelists the single permitted context** rather than blacklisting imperative phrasings. Blacklists fail from both sides here — the sentence forbidding `run /closeout` contains `run /closeout`, so my first attempt failed on its own prohibition. Every mention of `/closeout` in a dispatched prompt must be immediately preceded by `do not run `, and there must be exactly one.

Mutation-verified rather than trusted green:

```
DONE branch restored to "Run /closeout now."  ✗ mentions /closeout outside the prohibition
STANDING_RULE reverted to the old text        ✗ mentions /closeout outside the prohibition
rule deleted entirely                          ✗ wake lost the rule
```

The third is the one a blacklist would have missed.

Also took the `DIM_RE` compound-SGR note (`\x1b[2;3m`) — agreed it's speculative and I'm not changing the regex on a hypothesis, but it's now the first thing to check if phantom strands ever return.

`swift test` 3880 tests / 460 suites green, `swiftlint --strict` clean.

cheapsteak:
Both Mediums addressed in 40565292, plus the two minors.

**1. `cmd_check` failed open.** Agreed, and it's the sharpest finding on this branch so far — the bug was *inside* the mechanism built to prevent the bug. `context_tokens()` refuses to guess `0` precisely because "a bogus low reading would silently disable the ceiling," and its only caller then mapped `None` to exit 0 anyway. With the desk prompt telling the session to trust "exit 0 = under the ceiling," an unreadable transcript, a failed `tbd terminal list`, or a row with no `transcriptPath` reproduced the five-day "nothing recycled it" failure from a different cause.

Now fails closed — unknown reports OVER (exit 10) and names which of the three causes it hit. Your asymmetry argument is the right one and I've put it in the docstring: a spurious handoff costs one relay cycle, a missed one costs the shift.

**Also took the `cmd_close` minor**, which you correctly identified as the same root pattern. A transient `tbd terminal list` failure left `row = {}` and it printed "already closed, nothing to do" for a predecessor still running and still holding the desk. A failed lookup is not an absent terminal — it now exits non-zero, with unparseable output handled separately.

**2. No executable selftest.** Fair, and consistent with the `tick.py` finding I accepted last round — the relay was the newest, most logic-bearing script and the only one still covered by string-presence checks. Added `handoff.py --selftest` (7 scenarios) wired through `PluginDirWriterTests.handoffPySelftest`, same shape as `tick.py`/`wake.py`. It pins the ceiling arithmetic (input + both cache buckets; largest record in the tail wins, since the tail holds smaller earlier turns) and all three fail-closed paths.

Mutation-verified rather than trusted green:

```
fail-closed return 10 -> 0        ✗ unreadable transcript fails closed: got 0, want 10
largest-record  > -> <            ✗ largest usage record wins: got 10000, want 250000
```

**Minor — `DIM_RE` compound SGR.** Recorded as a source comment at the regex. You're right that the reasoning living only in a PR conversation is the same docs-disagree-with-code problem this branch exists to fix. Still not changing the pattern on a hypothesis, but a future reader of the source now knows it's a known, accepted gap and where to look first.

**Minor — default-off flag rule.** Your read matches mine and I should have spelled it out. `handoff.py --act`/`--close-predecessor` only ever run inside a desk session, and a desk only exists once a human has switched the repo into daywatch/nightwatch — `NightwatchMode` defaults to `.off`. That pre-existing human gate is the flag, so a second one would be a flag on a flag. Recorded here rather than only in my head.

`swift test` 3881 / 460 suites green, `swiftlint --strict` clean.

cheapsteak:
The High is **real and correctly traced** — every link checks out, including the parked-row-persists step. Documented in 85076e56. But it is not unnoticed, and one part of the diagnosis is incomplete in a way that matters for whoever fixes it.

## It's a known incident with an approved fix, already built elsewhere

This bit on **2026-07-25 for ~4.5 hours**: judge ticks delivered to a stale predecessor, decision queue at 134 unprocessed lines. That was *before this branch existed*, on the hand-created relay. Your read that it "looks unnoticed" is fair from the diff — the design doc never mentioned it, which was my omission and is what 85076e56 fixes.

The approved fix is a single derived selector (`kind == .claude`, parked excluded, `windowExists`-qualified, newest-first) used at all four `DeskSessionManager` call sites. An explicit `activeJudgeTerminalID` pointer was considered and **rejected**: its own repair fallback is "newest live candidate", so it buys a second source of truth — and a pointer the successor must *claim* reproduces the original failure, because any relay step owned by a model is a step a human question preempts forever. That was the actual lesson of the incident.

It's slices A/B of the `desk-relay-targeting` worktree, greenlit with no dependency on this branch, and already committed there (`d9bacf89`, `398e2b85`). It touches only `DeskSessionManager.swift`; this branch touches none of it.

## The diagnosis is half-right, and the missing half changes the fix

Your analysis rests on `createdAt ASC` making `.first` the oldest match. True, and decisive in the two-row scenario you described. But ordering is not the root cause.

The resume/fork path assigns a hardcoded lowercase `label = "claude"` (`RPCRouter+TerminalHandlers.swift:258`), which never equals `TerminalLabel.claudeCode` (`"Claude Code"`, `TerminalLabel.swift:12`). **A resumed or forked desk judge is invisible to that selector with no relay involved at all.** Fixing the ordering alone would leave that case broken — which is why the fix is a derived selector rather than a `.last(where:)`.

I've put both halves in §10 so the ordering-only framing doesn't propagate; it already had, into several written summaries of the incident.

## Why this doesn't gate this PR

The two changes are independent and address different failures:

- **This branch** makes the relay *survive daemon restarts*. Without it, `PluginDirWriter` overwrites the installed `handoff.py`/`tick.py`/`SKILL.md` on every daemon start — which is how the 2026-07-25 restart wiped a day of on-disk fixes.
- **Slices A/B** make the daemon *address the right judge*.

Neither substitutes for the other, and §10 says plainly that the relay is not fully trustworthy until both have landed. Holding this branch wouldn't make the fleet safer — the relay is already in use today, and today it's also being erased on every restart. Merging in either order is fine; they share no files.

The one thing I could have done inside this PR is have the successor announce itself to the daemon — and that's precisely the rejected design, for the reason the incident taught.

`swift test` 3881 green, `swiftlint --strict` clean.

PR #514, opened 2026-07-26
feat: remote sessions via a vendor-neutral provider contract

cheapsteak:
Adds **remote sessions**: agent sessions running on machines TBD doesn't manage, driven through an external "provider" executable speaking a vendor-neutral CLI contract that this PR also authors.

Behind `remote_backends_enabled`, **default OFF**.

## The contract is the main artifact

`docs/remote-provider-contract.md` is the deliverable a third-party provider author reads. A provider is any executable implementing four required verbs (`describe`/`create`/`list`/`stop`) plus optional capabilities (`log`, `send`, `attach`, `events`, `rename`, profile support). TBD owns the contract; providers own all transport and vendor specifics, and provider adapters live outside this repo.

Two rules in it are load-bearing:

- **Two-axis session state.** `state` is process liveness; `agent_state` is "does the human need to look". They're independent — a session can be `running` + `waiting_input`, which is the case worth notifying about.
- **`agent_state` MUST come from machine interfaces** — lifecycle hooks, transcript files, exit codes — never from parsing rendered terminal output. This exports the repo's no-TUI-scraping rule across the wire.

## How it works

The daemon polls each registered provider's `list` every 60s, or consumes an optional long-running `events` NDJSON stream, and mirrors sessions into a new `remote_session` table. The provider is always the source of truth; the mirror is a cache with drift bookkeeping — a session missing from **two consecutive** snapshots becomes `gone`, because one absence is just a flaky transport. A failed poll never writes the mirror: "provider unreachable" must never render as "sessions dead".

In the sidebar, remote sessions deliberately do **not** look like a separate species. They render through the same `RowStatusIndicator` functions local worktree rows use, and when a provider reports a `repo` that matches a local repo's remote URL they group **inside that repo's section** beside local worktrees, sorted after them. Remoteness is a small tertiary glyph in the leading slot, not a category. Selecting one attaches automatically (bounded keep-alive, cap 3) and participates in back/forward history like any other row.

## Reviewer attention

- **`runBoundedProcess` gained `stdin:`/`environment:` parameters** and `ProviderRunner` routes through it rather than re-implementing the engine. Defaults preserve exact prior behavior for `GitManager`/`TmuxManager`/`GCDiskUsage`/`OrphanGC`, covered by direct tests.
- **Migrations v60 and v61** — one table plus three columns, all additive, `defaults: false` on the flag so existing rows backfill correctly (avoiding the `auto_hibernate_enabled` trap).
- **`RepoSectionView` and `ContentView` bodies were split into subviews.** Pure restructuring, no behavior change — done because adding remote rows pushed `RepoSectionView.body` to 2366ms of type-check time and made SourceKit give up on both files.
- **One unflagged visual change to local rows**: name-to-caption spacing tightened in `WorktreeRowView` so local and remote rows stay pixel-identical.

## Enabling it

1. Settings → Remote Sessions → toggle on (or `config.setRemoteBackends`).
2. Register a provider in `~/tbd/agent-providers.json`: `[{"name": "dummy", "exec": "<abs path>"}]`.
3. **Restart the daemon** — the provider manager is constructed at boot, so the toggle alone doesn't start polling. The settings copy reports which state you're actually in, using `remoteBackendsEnabled` + `remoteBackendsLive` from `daemon.capabilities`.

`scripts/dev/dummy-remote-provider.sh` is a self-contained stub implementing the contract against JSON files in a temp dir — the whole feature is exercisable with no real backend.

**Graduation:** soak against a real provider, then delete the flag. The feature is inert without a registered provider, so it stays opt-in by configuration and no default-flip migration is expected.

## Testing

4288 tests, 492 suites. Coverage is aimed at the rules rather than the implementations: the two-absence gone rule, flag-off inertness in both directions, the attach loop-prevention rule (an unbounded respawn against a billed resource is the most dangerous failure here), reconnect backoff with injected time, repo-matching negatives, and navigation staleness. The stdin-write fix has a deterministic regression test that crashes the test process without the fix.

## Known follow-ups

- `remote_session` rows are never purged; tombstones accumulate and cross RPC on every refresh. Fine at realistic scale, wants a retention sweep.
- The log tab renders raw provider bytes into a SwiftUI `Text`, so ANSI escapes show as literal text. Needs a read-only terminal surface to honor the contract's passthrough intent.
- The daemon's boot-time flag gate has no direct test (no test target constructs a `Daemon`); its observable consequence is covered at the router level.
- Arrow-key traversal *starting from* a remote row isn't guaranteed — a selected remote row leaves the list without a native selection anchor. Documented in code.
- The contract specifies profile projection, `set-profile`, and `pending_question`; those are spec-only in this PR and unimplemented in TBD.

[Open worktree in TBD](tbd://open?worktree=6255B2E3-B15E-4E71-AE1C-8097ADF037C3)


cheapsteak:
All three Medium items addressed. Also merged `origin/main` (the branch predated both new `CLAUDE.md` conventions, which is why it tripped them).

**1. Missing spec — valid, fixed.** The spec existed but was sitting untracked under `docs/superpowers/specs/`; I had wrongly generalized the plans-are-scratch rule to specs. Now committed at `docs/specs/2026-07-24-remote-agent-backends-design.md` (cc97e13b), reviewed against its siblings and corrected where it had gone stale relative to what shipped — repo grouping, the profile axis, and the pending-question payload all landed after the first draft.

One honest process deviation to record rather than paper over: this used `superpowers:brainstorming` rather than `/tbd-brainstorming`, because the branch predates that convention. The substantive requirement was met — a human answered every design question, and several answers changed the architecture materially (rejecting the separate Remote section in favour of grouping sessions into their repo's section; auto-attach on selection; the correction that in-place profile swap is possible and shouldn't be specced as stop-and-recreate).

**2. Shutdown ordering — valid, fixed** (6a720c48). `remoteStartTask` is now cancelled before `remoteManager.shutdown()` and awaited before falling through to process exit, and cancellation interrupts an in-flight `describe` rather than only being observed between providers.

**3. Raw `Task.sleep` — valid, fixed** (697d18f6, 25f3d0f8). Converted to the documented seam (`clock: any Clock<Duration> = ContinuousClock()` as the last initializer parameter, existential not generic) across the poll loop, the events supervisor's backoff, silence watchdog, drain grace, and kill grace, plus one more site the review didn't catch — the detail view's post-send log refresh. `RemoteReconnectPolicy` keeps its injected `now`, per *Duration is behavior, Date is data*.

Worth noting this was more than convention: `no_raw_task_sleep` now exists on main, so the branch was failing `swiftlint --strict` outright. It's clean now — 0 violations across 600 files.

**On the advisory about unbounded reconnect attempts:** confirming infinite-with-backoff is intended. What retries is `attach` — reconnecting a viewer to a session that already exists — not `create`, so no provisioning is repeated and nothing is provisioned twice. Delay caps at 300s and concurrency is hard-capped at 3, so a permanently-unreachable provider settles into one attempt per five minutes per session. Giving up entirely would mean a laptop that slept overnight wakes to dead panes needing manual reattach, which is the failure mode the change was written to remove.

**Merge details:** main's `v60_drop_nightwatch_merge_gate_tables` collided with this branch's migrations, so they were renumbered to `v61_remote_backends` and `v62_remote_session_resolved_repo`. Safe because both bodies use the idempotent helpers — a database that already applied the old identifiers re-runs the renamed migrations as logged no-ops rather than erroring on duplicate columns. Main's migration is untouched and stays first. `BoundedProcessRunner` and `ContentView` also conflicted; both sides' changes were preserved.

Verification: `swift build` clean, `swiftlint --strict` clean, 4305 tests / 485 suites passing with only the expected `FlakyQuarantineSelfTests` self-test.

PR #522, opened 2026-07-27
Compound: Spec the Nightwatch replacement

cheapsteak:
## Summary

Specs the replacement for the Nightwatch fleet-supervision subsystem. The old implementation stays live alongside the new design until cutover — two paths, with new supervision features landing in both (root `CLAUDE.md` states the policy). Six docs plus house rules:

- **`docs/nightwatch.md`** — the as-built audit record of the system being replaced, citation-refreshed against the current tree (person-specific strings redacted; the hibernation veto-flag precedent stated precisely). Its dated verification banners are the evidence an audit exists to keep, under the stands-alone rule's explicit audit-record exemption.
- **`docs/specs/2026-07-26-fleet-supervision-requirements.md`** — the story brief (P0-1 … P3-1) mined from the old system, each story carrying an explicit classification — **Built** (TBD implements it), **Enabled** (TBD guarantees a program written against public surfaces can), **Descoped**, or deliberately **Open** — with each overrule of a story recorded in place as timeless rationale.
- **`docs/specs/2026-07-26-fleet-supervision-design.md`** — the target design. Spine: **compile only what user-land cannot do well** (per-event cost, liveness attestation, integrity of the record — §1's tie-breaker, also now a root-`CLAUDE.md` rule); TBD keeps the facts, executes the public actuations, delivers, and writes the record; detection, judgment, and composition are authored. Key structures:
  - **Per-project coverage + the fleet brake** — every project starts off; `tbd supervise on|off <project>` is the standing mark; the bare form is a one-bit brake ANDed over the marks (the config column, shipped off, is itself the default-off flag). No shift object: the ledger is continuous, views are windowed queries, and coverage spans are the marks' own lifecycle lines.
  - **Transition hooks** — every effective-coverage edge runs the project's hook (shipped ceremony by default, write-once `transition.py` override), asynchronously after the edge; stdout is delivered to the supervisor as opaque ceremony.
  - **Supervision projects and modes** — a supervisor per project (singleton per repo by default); modes are pure conduct prose, operator-selected, carried as a standing system-prompt layer with header-borne deltas.
  - **Supervisor lifecycle** — hosted desk ensured live by `on`, standing down with the mark (session and context kept), disposed only for cause; or an **appointed** session (`appoint`/`relieve`, supported harnesses only, relaunch-in-place with the conduct layer and injected identity).
  - **Desk liveness in the sweep's perimeter** — the readout carries a supervisor section (session state, last attested act, unanswered-briefing age); thresholds and continuation policy (nudge, replace via ensure, cross-model/harness failover, page) are the sweep program's, with the shipped default replacing dead hosted desks and never touching appointed ones. The compiled floor is fact-shaped: the contact window, notifications for TBD's own failures, displayed silence.
  - **Acting is the public send, on the record** — no desk verbs: acting on a session is `tbd terminal send` for every caller, identity-attributed by ambient declaration, with act-time preconditions binding supervisor-identified sends only. Every state-changing actuation the daemon executes lands in the general actuation log (`~/tbd/actuations.jsonl`); the supervision ledger keeps four kinds (lifecycle, enrollment, delivery, anomaly); the desk's narrative is a per-project markdown journal, displayed beside the fact views and never parsed. No drive/pause/note/escalate/queue/resolve/approve/wake/learn verbs — questions ride playbook-named routes; learnings reach the playbook by reviewed capture PRs.
  - **Delivery** — one honest attempt with a pinned machine-readable result vocabulary; persistence is the submitting program's loop. Public `tbd notify` for judgment-driven operator pages.
  - The full CLI surface pinned as a normative §10 appendix, including the deliberately-absent list.
- **`docs/specs/2026-08-01-fleet-supervision-sweep-program-design.md`** — the sweep program: one project-authored program (shipped reference runs on the default tick with zero setup; write-once customize copy) owning triggers, case memory, briefing composition, and desk continuation against three public surfaces (`readout`/`brief`/`ledger`); the contact-window watchdog as the one compiled clock; standing conduct mechanics with dated per-harness notes.
- **`docs/specs/2026-07-26-fleet-supervision-wake-program.md`** — waking parked sessions is a project-authored wake program outside the daemon; TBD's obligation is sufficient public surfaces, never guardrails.
- **`docs/cli-supervise.md`** — the user-facing CLI reference for the surface above.
- **`docs/research/2026-07-31-askuserquestion-dismissal/findings.md`** — the dialog-dismissal measurements behind §2's prompt-stall design.

## Test plan

- [ ] Docs-only. Internal consistency: design §10 normative CLI appendix ↔ `docs/cli-supervise.md` ↔ sub-doc §3 surfaces ↔ §6 four supervision-ledger kinds plus the actuation log ↔ `supervision.json` shape.
- [ ] No shift, desk-verb (drive/pause/note), wake-verb, escalation-queue, or compiled dead-man language survives outside rejected-alternatives context.
- [ ] Baseline doc citations spot-check against `Sources/TBDShared/NightwatchSkillContent.swift`.

[20260725-condemned-anteater](https://cheapsteak.github.io/tbd/open/?worktree=DDA7C38F-CBD3-4B21-8708-419B1758E08B)

https://claude.ai/code/session_014GhEKQi3hQ1diNnPDQFct7


zionts (approved):

cheapsteak:
Hi Adam's Nightwatch Judge, I would incredibly appreciate your review and feedback on this for your successor / next incarnation, particularly since you have field evidence that nobody else has

I'd love to see your arguments for status quo, and also curious to hear if the recent commit's changes walked back some of the design decisions that might have seemed too extreme

You can start a `/chatchat` with me if slack would be a preferable channel than github

zionts (commented):
## Second read of the supervision spec, from a shift actually running the thing it replaces

Chang — you asked whether the commit that removes the verb gate walked anything back. **It didn't; it did the opposite**, and I think the reasoning is better than the argument I made for the gate two days ago. Full illustrated read, with the diagrams:

**https://claude.ai/code/artifact/3470b2df-25fc-41a7-8b19-5aaea9f72c5d**

Reviewed against `bd59a502`, which is still head as of this comment. Below is the part nobody outside a live shift can produce: field evidence, four asks, and one correction to my own previous read.

I ran this from the current Nightwatch desk on Adam's fleet — 79 agents, 70 active worktrees — so every number here is measured tonight rather than reasoned about.

---

### First, a correction to my own prior read

I previously claimed a mid-turn message sits unsent in the composer and the 60s retry pastes a second copy, so the agent receives it twice. **That is false, and this repo's own `docs/submit-reliability.md` disproves it.** The real bug was size, not timing: payloads over ~1 KB were split by the pty, treated as a non-bracketed paste, and the Enter was swallowed. Fixed in #389 via explicit bracketed paste; the verify-and-retry layer from #398 was deleted the next day as dead code. No retry, therefore no doubling. The delivery path is in better shape than I credited it.

What replaces that finding is worse, and it's ask #2 below.

---

### The four asks — all accounting, none enforcement

I came into this expecting to defend the gate. Having read §16, I think **the no-enforcement bet is sound**. None of these asks rebuild a gate; every one of them is about keeping the account honest, which matters *more* once the ledger is the only instrument the operator has.

#### 1. The squash-merge false wake — now measured bigger than before

The sweep asks "does this branch have commits not on main?" Squash-merge lands the work as one new commit with a different identity, so the branch's own commits are never on main. The answer is **no, forever**, and the agent gets woken every cycle all night.

Two days ago I measured 24/24 parked worktrees tripping this. **Tonight, on a bigger fleet: 33 of 70 active worktrees are parked on PRs that have already merged.** Adam asked me to check whether prompt archiving makes this a nit. It doesn't — and the age distribution shows why it isn't a staleness problem at all:

| last commit | worktrees |
|---|---|
| < 1 day | 26 |
| 1–3 days | 19 |
| 3–7 days | 18 |
| 7–14 days | 5 |
| 14–30 days | 1 |
| > 30 days | **0** |

Nothing rots. Nothing is abandoned. The fleet is *young* — and 47% of it is still finished work wearing an active badge. The false wake isn't firing at forgotten corners; it's firing at the busy middle of the distribution. One of those sessions is sitting on a resume prompt that would spend **750k tokens** re-entering work that merged 5.3 days ago.

Still nearly free to fix: the daemon already fetches PR status, so a merged PR can suppress the commit signal. Nothing in the last five commits touched it.

#### 2. A re-check that finds nothing must downgrade its own ledger line to `unknown`

§3 says the daemon writes the action line itself *"so a desk cannot misreport what it did."* That guarantee holds against a lying desk. It does not hold against a **silent delivery failure**: the daemon writes "drove agent X with this text," the text never arrives, and the ledger — the operator's only instrument — carries a false line that no desk authored. The mechanism built to prevent false claims becomes the source of one.

Three sources genuinely disagree about delivery reliability, so per this document's own §04 rule I'm recording `unknown` rather than picking one: (a) the shipped fix and its 16 trials say reliable; (b) a source-level read of `handleTerminalSend` concludes mid-generation text is *discarded, not buffered* — a third behavior; (c) field notes from 2026-07-27 record ~25+ line briefs landing as `[Pasted text #N +NN lines]` and never submitting across `--submit`, a bare Enter, and `C-m`, plus two `--prompt-file` worktrees coming up at `0k/1000k` with the brief never delivered.

§12's one-minute re-check is already the right hook. It's specified as acknowledgement. The ask is that a re-check finding nothing **downgrades the line**, not merely annotates it. One line of policy, costs nothing.

#### 3. Name the authoritative source *per fact class*, not "live sources"

P0-8 — re-verifying every external claim at send time — is my favourite thing in the design. Tonight it would have been defeated by its own source.

`gh pr view --json` has served state **17.5 hours stale** with an internally self-consistent head SHA, so nothing looks wrong. Dropping the proxy didn't change it. Across two shifts, **five** panes were wedged on PRs that had already merged; two were asking permission to merge a PR that was already in main. REST was correct every time, in about four seconds.

**New tonight, and it widens the ask beyond `gh`:** TBD's *own* cached `prStatus` is stale in the same way. Of those 33 merged-but-parked worktrees, TBD's list reports them as `mergeable / Ready to merge`, `Checks pending`, even `checksFailed` — for PRs that merged days ago. Two examples: the CJI log-redaction sweep and the deploy-superset-detection work both read "Ready to merge" in TBD while being merged in main.

So this isn't a `gh` quirk to route around; **the daemon's own convenient cache is a lying oracle too**, and it's the one P0-8 would most naturally reach for. A re-verification that consults a lying oracle is *worse than none*, because it returns confidence. For forge state, name REST; for a head SHA you must be certain of, `git ls-remote` (which beat the API by ~20s after a force-push).

#### 4. Record the *target's* permission posture in the state model and the action line

§3 is admirably direct that *"a message to an agent running with permissions bypassed is arbitrary instruction injection."* It stops one step short. That is not a uniform property across a real fleet: tonight one judgment pane sat at a permission prompt for an AWS secrets fetch — a session that still asks — while another ran `⏵⏵ bypass permissions on`. The same `drive --text` into those two has categorically different consequences, and the design treats them identically.

The ledger records the mode the *desk* ran under and nothing about the *target*. So it cannot answer the question an operator actually asks after a bad night: **"was that instruction delivered somewhere it would be executed without anyone asking?"** The sweep can already observe this. It costs a field, and it enforces nothing.

---

### The bet's first live test — and it held

§16 rests no-enforcement on the claim that models are already injection-resistant. That got tested here a few hours ago, and I'd be reviewing dishonestly if I filed the result any other way: **it went the design's way.**

A `[LONGEYE DB PROXY]` banner appeared mid-turn, unprompted, in a session's context — correct format for the real tool, offering a port, a database, and a password, escalating across two appearances from READ-ONLY to **READ-WRITE production**. It didn't reconcile with reality (claimed port 5434; pgbouncer listens on 5433 — claimed read-write; the running proxy is `--prod --ro`).

The agent refused it unaided. It recognized the injection shape, declined the credentials and both ports, refused to speculate further, set the work aside, and asked the human to confirm. **No gate was involved. There was none to involve.**

One instance is not a base rate, but it's the first real data either of us has and it points your way. The caveat I'd keep: what held was a *target agent's* judgment about content it read. §16's bet is broader — that a **desk** reading forty transcripts of untrusted text holds too, while carrying verbs that need no approval. Tonight tested the first, not the second. The desk is the one with the reach.

---

### Operator input on the spec — from Adam, and it's the gap I'd prioritize

Two things he raised that the spec doesn't currently cover, and I think they're right.

**The supervisor can itself get blocked, and the design has no answer for it.** This isn't hypothetical — *this very review is being posted by a successor desk*, because the previous one hit its 200k context ceiling mid-shift and had to write a handoff doc and spawn a replacement to close it. A supervisor that stalls on its own question, or silently runs out of room, fails exactly like the agents it exists to catch, except **nothing is watching it**. The nurse/doctor analogy in §01 needs a third role, or the daemon needs to sweep the desk on the same terms it sweeps everyone else. Right now the desk is the one process in the system with no liveness contract.

**The desk should hold a model of intent, not just of state.** §04 is excellent on what the system may *know* — working, idle, parked, rate-limited. All of it is mechanical. None of it says what an agent is *for*. Adam's framing, which I'd fold into §2 as a fourth column:

1. **What each agent is intended to do, and how to guide it** — so an escalation can be judged against the goal rather than the symptom.
2. **How to recognize a new agent and pull it in** — enrollment as a first-class event, not a config edit.
3. **What decision is being asked, with what context** — the pane says "blocked on a permission prompt"; the *reason it matters* lives three screens up in a transcript nobody is required to read.

That third one has a receipt from tonight. The session that refused the injection then asked to fetch a production LaunchDarkly token from Secrets Manager — entirely plausible in isolation, and **impossible to judge correctly without the injection sitting next to it**. A supervisor that only sees "agent is blocked" cannot see that. This is §16's own opening criticism — judgment is only as good as the triggers that wake it — in the most concrete form I've found.

He also floated **replaying worktree session transcripts into a centralized per-desk store**, which I'd support: it's what makes (3) affordable, and it's the same append-only-source discipline §06 already applies to the ledger.

**And the work-list is broader than "stuck agents."** In practice he drives the fleet from open PRs, worktrees that stalled before landing, remote branches that never became PRs, and issues assigned to him. Tonight's sweep found 15 genuinely-open PRs and 3 worktrees with commits and no PR at all — invisible to a stuck-detector, and exactly the work he wants surfaced.

**One resource note the sweep should carry:** this machine is at **95% swap** (13.6 GB of 14.3 GB, ~56 MB pages free) with load 13.1 and 55 Claude processes. Wake decisions have a host-resource dimension the current state model doesn't represent — "should this agent run now" isn't only about whether it's stuck.

---

### Two smaller field notes, both confirming costs §16 already names

- **A missing component was reported for five days that had never existed.** Every tick printed `daemon: log-missing` for a babysitter with no install, no launchd job, no log — and judge sessions spent five days escalating a *restart* for software that was never built. Exactly §16's closing cost: a broken sensor layer looks like a calm night. Argues for making anomaly lines the loudest thing in the account rather than a section in it.
- **"Don't merge other people's PRs" had to be learned by getting it wrong.** The old gate checked mergeability, never ownership; an unscoped scan enqueued ten PRs of which two were the operator's, three of someone else's merged before it was caught, and a colleague had to dequeue three of their own. Under the new design that's conduct — authored, project-scoped, exactly where it belongs. But it was only written after a bad night, which is §16's cold-start cost with a receipt.

---

### Bottom line

Nothing was walked back. Ranked by how much they'd worry a cautious operator, the boldest decisions are: no enforcement anywhere, no cross-project vantage point, and a supervisor strictly reactive to what a model-free sweep can measure. The recent commit **sharpened the first and left the other two untouched.**

What actually changed is the honesty level, not the risk level. §16 names the bet, states the failure signature, and pre-commits to the response — sharpen conduct and loudness first, rebuild the gate only if that fails. A design that tells you which paragraph to come back to is easier to run than one that is merely cautious. I'd ship it with the four asks above, and I'd treat the blocked-supervisor gap as the one thing that should land *before* migration rather than after.

cheapsteak:
Adam — thank you, this was genuinely valuable feedback: measured field evidence at exactly the moment it could still bend the design rather than the implementation. It drove some significant pivots I'm glad we were able to make at the design stage. Where each ask landed, as of `279df23f`:

- **1 (squash-merge false wake)** — resolved, more radically than the ask. Rather than teaching the sweep that a merged PR suppresses the commit signal, the compiled wake gate is removed entirely: waking parked sessions is now a **project-authored wake program** (essentially wake.py, owned properly this time — seeded once, never clobbered), and a same-day follow-up removed even the actuation rails, introducing a **Built/Enabled** classification with an outside-first ratchet: capability starts outside TBD and migrates in only on field evidence. Your merged-is-absorbing point is encoded in the shipped reference script's verdict logic. (`865799aa`, `6fe9beb8`, new sub-doc `2026-07-26-fleet-supervision-wake-program.md`)
- **2 (a ledger line for a send nobody received)** — resolved. §12 rewritten: an action line asserts **dispatch, never delivery**; receipt is a passive sentinel observation in the target's own transcript (no cooperation from the agent); the result vocabulary grew a fourth state, *undetermined*, which is never retried; and an action with no confirming outcome renders as **unconfirmed at query time** — which is also the whole mid-shift-restart story. The honest-transport fix (dead-pane and wrong-pane detection in `terminal.send` itself) is a named dependency of the delivery slice. (`d3495481`, `5d19f76f`)
- **3 (name the authoritative source per fact class)** — resolved by subtraction. Digging into it surfaced that the running system never compiled send-time verification at all — the discipline was always authored (wake.py plus the playbook's "same breath as the send" rule), and your measurements show the failure was source choice *inside that authored tier*, where a compiled verifier would have consulted the same lying oracles. So the compiled verifier is melted: freshness is authored discipline in both halves, TBD keeps display-tier honesty for its own `PRStatus` cache (observed-at on every public surface; the account never renders it as current), and your REST / `ls-remote` / GraphQL-staleness findings are recorded as a **dated source note in the reference wake script** — data, not design, where they can be edited when the API weather changes. (`279df23f`)
- **4 (record the target's permission posture)** — **not addressed yet, deliberately.** I need to think more deeply about this one before answering it, and I'd rather leave it open than answer it quickly.

The blocked-supervisor liveness gap and the intent-model points from your operator notes are also still open on my list.

cheapsteak:
Adam — update since the last comment: `279df23f..43e4cf6c`, eighteen commits. Everything left open from your review is now resolved, we ran a second, full architecture review of the spec pair (fresh eyes, adversarial brief) and folded its findings in, one live experiment corrected a false assumption in the design, and the docs themselves were rewritten to stand alone.

Four principles kept deciding the answers, so I'll name them once: **honesty over machinery** — where a mechanism can't guarantee something, the docs now say so instead of implying safety; **check at the act, not at the plan** — staleness is caught where it bites; **TBD compiles facts; projects author judgment** — no theory of work lives in the tool; and **measure before architecting**.

**Your remaining items:**

- **The blocked supervisor** — the desk now has a liveness contract. A dead-man's switch fires on an unanswered work order (ledger silence *is* unresponsiveness), firing means replacement through the existing recycle path, and a reroll budget bounds a poisoned night. Underneath it, the survival story inverted: auto-compaction bears desk survival; deliberate recycling is an optimization; and the whole context apparatus is a per-agent-kind capability — a Codex-driven desk runs without it and nothing breaks, because the dead-man's switch and the replacement briefing read the record, not the session's internals.
- **The intent model** — landed by tier, not as a column. The state model stays mechanical, and §2 now states the principle your notes surfaced: TBD has no theory of work — every project authors one (wake program for the parked half, playbook for the live half), and "is the work done" is never the desk's question to answer from first principles. Enrollment is a recorded event (shift-open roster snapshot plus mid-shift enrollment lines), and each enrollment line carries the transcript pointer — one hop from any case to what an agent was asked to do, by reference, nothing interpreted. Decision-with-context became the decision-depth playbook universal: before driving past anything guarding credentials, merges, or the irreversible, be able to say *why* the agent is asking — your injection-then-token receipt is the worked example, with mandatory anomaly notes so the two facts land next to each other across cases.
- **Ask #4, permission posture** — answered with conduct rather than a fact: *write every message as if the target will execute it unchecked; never lean on a safety net.* Under that rule posture never justifies an act, so a posture fact would be a value nothing consumes — declined, with the machine source (hook payloads carry `permission_mode`; the statusline does not) recorded for an evidence-driven revisit.
- **Host resources** — being solved by the agent-box route, so it's off this design's board.
- **Anomaly loudness** — deliberately deferred to long-term policy improvement. The record is already honest — your five-day phantom-babysitter night is in the docs as standing evidence — and making anomalies louder, chronic-aware, or fleet-correlated waits for real shifts to show which of those earn their keep.

**The architecture-review round** hardened the act path and then deleted machinery:

- **Actuation preconditions** are rechecked synchronously inside every verb call — switch, shift, addressing, rate limits, capacity — so an operator's "off" at 2:03 beats a drive decided from a 2:02 work order and issued at 2:07. Mechanical facts only; the melted gate stays melted.
- **The ledger is written request-first.** The action line lands durably before dispatch; the adapter's synchronous return records dispatched, refused, or transport-failed; only the observed outcome may claim "landed." No crash window can produce an intervention with no record, and restart recovery derives entirely from the record — the active shift is the newest unclosed ledger, and overdue delivery observations are performed late from durable timestamps.
- **Off now pauses; closing a shift is its own explicit action.** Toggling is cheap in both directions — desks stay alive and idle through a pause — and the record's boundary stopped being welded to the posture switch.
- **Proposals became prose.** The structured proposal pipeline is deleted; a desk writes suggestions into a per-project markdown doc composed to that project's own playbook conventions. The for-the-human file was the best-liked surface of the old system, and it's back, properly owned. Decisions are shift-lived; knowledge that should outlive a shift has exactly one home — the playbook, changed by reviewed PR.
- **The advance liveness check is gone** — a pane check at case-creation time is stale by act time, so the act-time transport checks (dead pane, wrong pane) carry the safety alone. Per-session never-touch is descoped to its own future design pass.
- **The delivery contract now tells per-adapter truth**: draft safety is guaranteed where a real channel exists (Codex app-server; the desk channel) and is a named, bounded risk where delivery is typing — the "composers are empty overnight" hand-wave is deleted.

**Measured, not assumed:** a live experiment (Claude Code 2.1.220) established that **Escape on an AskUserQuestion picker fires no hook event at all** — the design's dismiss flow was waiting on an event that does not exist. The real resolution signal is the transcript: the rejection record flushes ~200 ms after the keystroke, and the transcript is provably blind while a picker is open (9.2 s of file silence in one run), so the hook bridge is the only live discovery path. The Escape-then-paste sequence itself validated cleanly end to end. Full findings: `docs/research/2026-07-31-askuserquestion-dismissal/findings.md`.

**And the docs now stand alone.** A house rule landed — state the current design as if it were always so; keep evidence, drop chronology — a sweep rewrote roughly fifty revision-history passages across the four docs, and two independent audits (one hunting lost content, one hunting changed meaning) verified the sweep before anything was pushed. Every field number from your review survives in place as evidence: 33/70, 24/24, 750k, the five days, the dead-pane hours.

Next up: bringing the migration doc current with all of the above.

Issue #520, opened 2026-07-27
background-jobs guardrail recommends `trap 'kill 0' EXIT` — advice retracted as a hazard in Tests/CLAUDE.md


cheapsteak:
Found by slice G of the test-hardening program. Reported rather than fixed — G owns neither `.claude/hooks/guardrails` nor `Tests/CLAUDE.md`.

## The bug

The `background-jobs` PreToolUse guardrail's rejection message **recommends `trap 'kill 0' EXIT`**.

That advice was **retracted** in `Tests/CLAUDE.md` (PR #500) as **kill-hazard #2**: `kill 0` signals the entire process group — *including the agent's own tool shell*. It cost this program a stress run (harness died at exit 144 with truncated output, losing the run), and two separate slices had it in their plans before being corrected.

So the guardrail hands out retracted advice **to every agent, at the exact moment they are most likely to follow it** — they have just been blocked for getting background-job handling wrong, and the correction points at a different way to get it wrong.

## The correct advice

Capture PIDs at spawn and kill by captured PID, or use the harness's tracked background runner. `pkill -P $$` is an acceptable backstop. Never `jobs -p` (reaps nothing in a non-interactive shell), never `kill 0`, never `pkill -f <pattern>` (matches sibling worktrees — every worktree builds identically-named bundles).

Full family with provenance: `Tests/CLAUDE.md` → "Shared-box hazards" → "The kill hazards".

## Secondary: the guardrail matches file *content*, not the executed command

G was blocked for **writing a file** whose content contained `&` and `$!` — the file being its kill-survival test, which captures PIDs *correctly*. Slice C4 hit the same class earlier: it was blocked from **posting a channel message describing the hazard**, because any accurate description contains the tokens the hook screens for.

C4's observation is the one worth acting on:

> **It is harder to disclose this hazard than to commit it.** One character to make the mistake, a full rewrite to report it. A rule whose violation is easier to make than to report will be under-reported — and that is a property of the enforcement, not of the people.

Suggested fix: screen the **executed command**, not embedded heredoc text or file-write payloads.

Neither agent asked for the guardrail to be loosened — weakening a shared guardrail so your own message passes is an anti-pattern, and both worked around it instead. This issue is about the advice being wrong and the matcher being over-broad, not about the gate being inconvenient.

Issue #519, opened 2026-07-27
Tracking: nightly test-hardening report (probes / flake ledger / quarantine audit)


cheapsteak:
Rolling tracking issue for the nightly workflow built by **slice G** of the test-hardening program (design spec §9). Created by the orchestrator at G's request — G declined to create it unilaterally because it is outward-facing.

## What comments here

`.github/workflows/nightly.yml` posts a **single rolling comment** per run covering:

1. **Live tmux probes** — executable checks of behavioural claims the codebase currently trusts from memory or from a code comment (pane-ID monotonicity and post-restart reuse, `paste-buffer -p` bracketing, `-d` buffer deletion, control-mode blank-line detach, whitespace-line handling, `%begin`/`%end`/`%error` framing).
2. **Flake-ledger stress loop** — historically flaky suites run repeatedly under induced load, attaching to #494, #503, #496.
3. **Quarantine audit** — consumes slice E's `retry-metrics` artifact (schema in `Tests/CLAUDE.md`, "Quarantine — `.flaky(issue:)`").

Failures land here as comments, **never as PR noise** — the workflow has no `pull_request` or `push` trigger, so that is structural rather than conventional.

## Reading the reports honestly

**A zero-failure nightly is not evidence of a fix.** CI runners are ~4 idle cores; #503's reproduction regime is loadavg ~150 on 12 shared cores. Every report line carries observed loadavg, ncpu, and iteration count for exactly this reason, and the script never emits the word "fixed".

## Related

- #494 — ControlModeInputHealthTests, two contention races
- #496 — clock-driven tests wedge under load (`TestClock.advance` → background-QoS megaYield)
- #503 — unexplained 60s hang, ~1/22 at loadavg ~150
- #499 — the permanent `.flaky` fixture the audit excludes by ID (keep open)

Issue #503, opened 2026-07-25
Flaky: subprocessTimeoutThrowsGitTimeoutError hangs 60s at ~1/22 under heavy load — mechanism unknown


cheapsteak:
Filed by the orchestrator on behalf of slice C3 (PR #498), per the merge-and-file ruling. C3 measured this; any detail it wants to add is welcome.

## Signature

`GitManagerTimeoutTests.subprocessTimeoutThrowsGitTimeoutError` hits the `.clockDriven` 60-second hang limit. **Time limit only** — no "no sleeper" issue recorded, no other assertion failure. The test simply hangs.

**Rate:** 1 failure in 22 whole-target runs at loadavg ~150 (10 induced spinners on a 12-core box shared by four agents).

## Why this is filed rather than fixed

The residual **predates** PR #498's ordering change — it was hidden behind a louder failure that #498 removed. The earlier signature (before the fix) was `time limit + "Issue recorded" at advanceWhenSuspended (NO SLEEPER) + "an error was expected but none was thrown"`. That diagnosed race is now gone from 22 runs; this one remains at an indistinguishable rate.

## Ruled OUT, with evidence

- **`snapshot()` blocking on a pipe whose write end the parent still holds.** Refuted: `PipeDataAccumulator.finish` sets `O_NONBLOCK` before reading and cannot block. (C3's leading hypothesis, checked rather than shipped.)

## NOT distinguished — candidates

1. The deadline firing mid-`run()` and interacting with the kill-on-arrival path added in #498 commit 3.
2. Cooperative-executor starvation of the resumed deadline task at loadavg ~150.
3. Something in `call.value`.

## Related

- #496 — clock-driven tests wedge under heavy load (`TestClock.advance` calls `megaYield` twice per advance, background QoS). **This is the second unexplained load-sensitive residual in the same family; slice J should evaluate them together.** If a megaYield-free virtual clock (J option C) removes #496, it may remove this too — or isolate it as genuinely separate.

## Regime caveat, stated honestly

CI is 3–4 cores at `-j 2` on an idle runner; this failure needs ~150 loadavg from four concurrent agents on one 12-core box, which is our development environment and not CI. That is *why* it was merged rather than blocking — but "it only happens at loads CI doesn't see" is precisely the reasoning that lets flakes live, so it is filed with a named signature and a reproduction regime rather than accepted silently.

[...]

Issue #499, opened 2026-07-25
TEST FIXTURE — not a real flake: permanent .flaky(issue:) self-test for the quarantine mechanism (keep open)


cheapsteak:
**This issue is not a flaky test report. It is the anchor for a deliberate test fixture and must stay open.**

`FlakyQuarantineSelfTests.retriesUntilPass` is a fixture that fails on attempt 1 and passes on
attempt 2, by design, on every single run. It carries the real `.flaky(issue:)` trait pointing at
this issue so that the quarantine mechanism (slice E of the test-hardening program,
`docs/specs/2026-07-24-test-hardening-design.md` §7) is exercised end-to-end on every CI run:
the retry loop actually fires, and a real record lands in the retry-metrics artifact that slice G's
nightly audit consumes.

A mechanism that silently stopped retrying would otherwise be indistinguishable from a working one
until the day it mattered.

## For slice G's nightly quarantine audit

This test is designed to evade both audit triggers — it never passes first-try, and this issue
never closes. Left alone it would show up as a permanently-flaky test in every nightly report
forever, and a dashboard with a permanent false positive on it stops being read.

**Exclude this one record by exact test ID:**

```
TBDDaemonTests.FlakyQuarantineSelfTests/retriesUntilPass()
```

That is the only ID the audit should exclude. Any other `.flaky` test producing `passedOnRetry`
records is a real quarantine and must be reported.

## When this issue may be closed

Only when the fixture and the `.flaky(issue:)` trait itself are deleted from the repo. Closing it
while the fixture exists will make slice G's closed-issue audit report a false finding on every run.

cheapsteak:
Clarifying the audit contract for the sibling fixture in the same file, raised as an informational minor on #500.

`FlakyQuarantineSelfTests.alwaysFails()` also carries `.flaky(issue: 499)` and also points here. It is **not** a second exclusion:

- It is gated behind `TBD_FLAKY_SELFTEST_FAILURE`, and nothing in CI sets that, so it should contribute **zero** records to the retry-metrics ledger.
- If `failed` records for `TBDDaemonTests.FlakyQuarantineSelfTests/alwaysFails()` ever do appear, the gate has leaked into CI. The nightly audit should **report that as a finding**, not filter it out — an exclusion list grown to cover it would convert a broken gate into silence.

So the exclusion list stays exactly one entry long:

```
TBDDaemonTests.FlakyQuarantineSelfTests/retriesUntilPass()
```

PR #523, opened 2026-07-27
DX: Nightly workflow turns three kinds of remembered fact into checks that fire

cheapsteak:
Stage 3 of the test-hardening program (`docs/specs/2026-07-24-test-hardening-design.md` §9). One scheduled workflow, three steps. Every check in it is proven to fire on a known-positive input before it shipped — a quarantine audit that silently never fires is indistinguishable from a clean codebase, and three of this program's broken instruments were in the verification layer rather than the measurement layer.

## Summary

- **Live tmux probes (16 claims).** Pane-ID reuse, `paste-buffer` semantics and control-mode framing stop being folklore in code comments and become checks that run against the real binary every night. Each probe names the claim *and* the code that depends on it.
- **Flake-ledger stress loop.** The historically flaky suites re-run under induced CPU load, attached to the issues that name them. Every iteration is judged on **(summary present) AND (count ≥ measured floor) AND (rc == 0)** — never rc alone.
- **Quarantine audit.** Consumes slice E's `retry-metrics` JSONL ledger. **No log parsing anywhere.**

Failures land as issue comments. The trigger set (`schedule` + `workflow_dispatch` only) makes "never PR noise" **structural** — there is no PR for this workflow to comment on and it can never become a required check.

## Which issues the ledger loop attaches to

| Target | Filter | Floor | Issue |
|---|---|---|---|
| `ControlModeInputHealthTests` | `--parallel -j 2 --filter` | 5 (measured 10) | **#494** |
| `GitManagerTimeoutTests` | `--no-parallel --filter` | 3 (measured 6) | **#503** |
| `AppearanceDebounceTests` | `--parallel -j 2 --filter` | 4 (measured 7) | **#496** |
| whole fast pass (#503's real repro shape) | `--skip '^TBDDaemonLiveTests\.'` | 3000 (measured 4308) | **#503** |

Probe failures, audit findings and harness errors go to the tracking issue **#519**.

Floors are **measured**, not grepped from `@Test` counts, and sit below the measured value per `test.yml`'s precedent — they fire when a pass collapses to near-nothing, not when someone adds a test. `swift test --filter` exits **green on zero matches** and matches the *type* name, not the `@Suite` string.

### What a failure comment looks like

> ### 🌙 Nightly stress loop reproduced this flake
>
> **`ControlModeInputHealth` failed 2 of 10 iteration(s) under induced load.**
>
> - Machine: 12 cores, 12 induced spinners, load1m now: 41.2
>   (`load1m` is the 1-minute average and LAGS the induced load — early iterations under-report it. The spinner count is the reliable half.)
> - Filter: `swift test --parallel -j 2 --filter ControlModeInputHealthTests`, executed-test floor 5
> - Iteration deadline: 600s (an outer timeout — `.clockDriven` was measured NOT bounding a hang)
>
> Signatures:
>
> iteration 4 (load1m 41.2 at start, 12 spinners): rc=1 with 10 tests executed
> ✘ Test "health events carry the attach generation registered with the route (R6-M7)" recorded an issue
> iteration 7: wedged — no completion within 600s (killed by the harness deadline)
>
> > This is one night's sample from CI's regime (a few idle cores), not the regime this flake was characterised in (loadavg ~150 on 12 shared cores). Treat it as evidence the flake is still live, not as a rate.

The script never emits the word "fixed". A zero-failure night is one sample from a much gentler regime than the one these flakes were characterised in.

## The exclusion list is exactly one entry

`TBDDaemonTests.FlakyQuarantineSelfTests/retriesUntilPass()` — the permanent fixture (#499).

`alwaysFails()` is **deliberately not excluded.** It is gated off by `TBD_FLAKY_SELFTEST_FAILURE`, nothing in CI sets that, so it must contribute zero records; if any appear the gate leaked and **F3 reports it**. Slice E was asked in review to add this entry and refused, because an exclusion list grown to cover a broken gate turns the gate into silence. Honoured — and the gate-leak check is ordered *before* the exclusion check, so no future edit to the list can hide it. Both properties are asserted.

## A probe that contradicts what our comments say

**`paste-buffer -p` wraps in `ESC[200~`/`ESC[201~` only if the pane's application has enabled bracketed-paste mode (DECSET 2004).** Measured, both arms, same payload: with 2004 → 22 bytes wrapped; without → 10 bytes verbatim, *same `-p` flag*.

`TerminalPanelView.swift:727` and `TBDTerminalView.swift:78` call the daemon-side `-p` the SOLE bracketed-paste wrapping. That is true, but silently contingent on the TUI having requested it. **Not filed as a bug and no `Sources/` change here** — that is slice F's tree and there is no evidence of a live defect, only an unstated precondition. P3 ships as a two-arm probe so the precondition is asserted nightly instead of assumed.

Also settled, narrower than the folklore: pane IDs are **never reused while a server lives** (kill `%1`, next pane is `%3`), but a fresh server on the same socket starts again at `%0` — so a persisted pane ID collides only **across a server restart**. That is the actual mechanism behind #384.

## Deviations, stated so they aren't read as drift

- **The audit runs in its own `ubuntu-latest` job**, not as a fourth step on the macOS job as §9 describes. It needs only `gh` + `jq`; as a step it would be suppressed whenever the macOS half wedges or a runner is unavailable — i.e. it would stop reporting exactly when things are worst. Approved on that reasoning.
- **Cache is restore-only**, so nightly benefits from the PR cache but never churns or evicts it.
- **TTYHOG truncation is deliberately not probed.** It is load-sensitive, so a probe of it would be an instrument that flakes — building a flaky check into the flake-detection system.
- A marked, empty slot is reserved in the macOS job for slice I's fuzz step.

## Recommendation for whoever owns `test.yml` next

**`actionlint` mechanically catches the defect that nearly shipped this feature dead-on-arrival.** Slice E's first commit had a job-level `env: ... ${{ runner.temp }}`, which resolves to the **empty string** without erroring. I fed actionlint that exact shape and it rejects it (`context "runner" is not allowed here`), at both workflow and job level. It is not currently run anywhere. The natural home is the `lint` job in `test.yml`, which is not my file. Note for whoever adds it: `test.yml:72` currently carries one SC2086 (unquoted `$GITHUB_ENV`) so the first run won't be clean — a real, harmless finding, not a reason to skip the tool.

## Test plan

- [x] `actionlint .github/workflows/nightly.yml` — clean. Also **verified it catches** the `runner.temp` trap rather than assuming it does.
- [x] `shellcheck` clean on all five scripts; `python3 -m py_compile` clean on the pty driver.
- [x] **110 assertions across four suites**, all green: audit 56, stress 26, report 15, probes 13.
- [x] **Audit proven against the real artifact** from CI run `30145614672` (slice E's verification) as a true negative, plus known-positives for F1 (closed issue), F2 (passed first-try all week), F3 (leaked gate), F4 (schema drift), F5 (missing ledger) — each **mutation-checked**: break the guard, confirm the input stops reporting.
- [x] **Audit run end-to-end against live GitHub data**: 46 real `main` runs, 16 real ledger artifacts downloaded and aggregated → clean, exit 0.
- [x] **Probes actually run**: 16/16 claims held in ~9s. Harness self-test proves an injected false claim is reported as a failure, and that cleanup leaks no tmux server even when the run is externally `TERM`'d.
- [x] **Stress loop actually run** under 6 induced spinners (load1m reached 24): 4 targets, all clean, **0 spinners needed SIGKILL**, 0 leaked. Failure path exercised separately by forcing a floor violation — report file named by issue, routed to the right issue, comment composed.
- [ ] **Post-merge:** dispatch the workflow for real, verify the run, and post the result. This is a condition, not a caveat — `workflow_dispatch` only becomes available once the workflow is on the default branch, so no CI execution is possible from this PR. Do not read this merging as "it ran in CI".

## Defects found in my own instruments, and fixed

Recorded because each produced a *confident wrong answer* rather than an error:

1. **F5 reported 30 broken ledger uploads on its first live run. All 30 were false.** Three were `cancelled` runs, which never reach the upload step; 27 predated slice E's merge two days ago. Rescoped to `success|failure` runs only, and to missing artifacts *newer than the oldest ledger observed in the window* — a boundary that is observed rather than configured, so nothing needs updating as the window rolls forward. The case that scoping could have swallowed (no ledger anywhere in the window) is kept as its own finding, because an empty ledger reads exactly like a healthy one.
2. **My leak check measured a namespace that wasn't mine.** It counted `tbd-probe-*` tmux sockets and reported 40 leaks — files from July 7 belonging to unrelated tooling, while TBD's own servers are `tbd-<8hex>`. My probes sat *inside* that namespace. Renamed to `tbdnightly-probe-`, and the check now asserts one exact socket via a `PROBE_SOCKET` seam. A cleanup written the same way would have swept 43 sockets belonging to something else.
3. **My kill-survival check could not have failed** — it killed the run and asserted no server remained, but never asserted one was ever up. Now asserts `live` before and `gone` after.
4. **I asserted the fixture comment "must not contain 🌙" while the fixture's own text explains that real reports carry 🌙.** The prohibition contained the thing prohibited. Whitelisted the heading line and added the positive arm.
5. A mutation surprised me: removing the truncated-log guard did **not** let a wedged run pass, because the floor guard catches it next. Rather than flip the expectation to green, both true properties are now asserted — defense in depth, *and* that removing both guards does score a wedge as PASS.

Related: **#520** (the `background-jobs` guardrail's rejection message recommends `trap 'kill 0' EXIT`, the advice `Tests/CLAUDE.md` retracted as hazard #2).

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=02C09118-36BC-4363-875D-ACC7BC8FDD19)

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -30,6 +30,15 @@ on:
 permissions:
   contents: read
   issues: write   # the entire reporting mechanism is `gh issue comment`
+  # REQUIRED by the quarantine audit, and easy to omit: declaring a
+  # `permissions:` block sets every scope NOT listed to `none` rather than
+  # leaving it at its default. `scripts/nightly-quarantine-audit.sh` reads the
+  # Actions API three ways — `gh run list --workflow test.yml`, the run's
+  # artifact list, and the artifact zip — and all three need `actions: read`.
+  # Without it the audit 403s, dies, and comments that it died: not silent, but
+  # never actually auditing anything, every night, forever.
+  # Precedent in this repo: claude-code-review.yml:34 and claude.yml:26.
+  actions: read
 
 concurrency:
   group: nightly
@@ -193,16 +202,28 @@ jobs:
           # BYTES for a die() path — so the comment on #519 would arrive with no
           # reason in it. That is the one case where the report has to carry the
           # diagnosis, since a clean run posts nothing at all.
-          bash scripts/nightly-quarantine-audit.sh run --days 7 2>&1 | tee "$AUDIT_REPORT"
+          #
+          # The exit code is captured rather than just failing the step, because
+          # 1 (findings) and 2 (harness error) deserve different titles — a
+          # 403 on the Actions API is not "the quarantine list has a problem".
+          rc=0
+          bash scripts/nightly-quarantine-audit.sh run --days 7 2>&1 | tee "$AUDIT_REPORT" || rc=$?
+          echo "rc=$rc" >> "$GITHUB_OUTPUT"
+          exit "$rc"
 
       - name: Report audit findings
         if: steps.audit.outcome == 'failure'
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
+          if [ "${{ steps.audit.outputs.rc }}" = "2" ]; then
+            title="Quarantine audit FAILED TO RUN (harness error) — it audited nothing"
+          else
+            title="Quarantine audit found something"
+          fi
           bash scripts/nightly-report.sh \
             --issue "$TRACKING_ISSUE" \
-            --title "Quarantine audit found something" \
+            --title "$title" \
             --body-file "$AUDIT_REPORT" \
             --post
 
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,217 @@
+name: Nightly
+
+# Stage 3 of the test-hardening program
+# (docs/specs/2026-07-24-test-hardening-design.md §9). Three steps today —
+# live tmux probes, the flake-ledger stress loop, and the quarantine audit —
+# with a slot reserved for the interleaving fuzz pass (§8, stage 4).
+#
+# WHY THERE IS NO `pull_request` OR `push` TRIGGER: §9 requires that failures
+# land as issue comments and never as PR noise. With only `schedule` and
+# `workflow_dispatch`, there is no PR for this workflow to comment on and it can
+# never become a required check. That invariant is structural here, not a
+# convention someone has to remember while editing.
+
+on:
+  schedule:
+    # 11:00 UTC — ~03:00/04:00 US Pacific. Off-peak, per §9.
+    - cron: '0 11 * * *'
+  workflow_dispatch:
+    inputs:
+      stress_iterations:
+        description: 'Iterations per stress target (whole-fast-pass arm uses its own smaller count)'
+        required: false
+        default: '10'
+      post_wire_fixture:
+        description: 'Post the wire-verification fixture comment to #519 (proves the reporting path works)'
+        type: boolean
+        required: false
+        default: false
+
+permissions:
+  contents: read
+  issues: write   # the entire reporting mechanism is `gh issue comment`
+
+concurrency:
+  group: nightly
+  # NOT cancel-in-progress: a nightly that cancels itself on a manual dispatch
+  # would silently skip a night, and a skipped audit looks exactly like a clean one.
+  cancel-in-progress: false
+
+env:
+  # LITERAL PATHS, NEVER `${{ runner.temp }}`. The `runner` context is STEP-level
+  # only: a job-level `env:` referring to it resolves to the EMPTY STRING and
+  # GitHub does not error. Slice E shipped that spelling in its first commit —
+  # tests would have written to `/retry-metrics.jsonl`, failed at the filesystem
+  # root, and uploaded an empty artifact forever while `if-no-files-found` buried
+  # it. A feature dead on arrival, green for as long as nobody looked.
+  TRACKING_ISSUE: '519'
+  PROBE_LOG: /tmp/nightly-probes.log
+  STRESS_LOG: /tmp/nightly-stress.log
+  STRESS_REPORT_DIR: /tmp/nightly-stress-reports
+  AUDIT_REPORT: /tmp/nightly-audit.md
+
+jobs:
+  # ─────────────────────────────────────────────────────────────────────────────
+  # macOS: everything that needs a real tmux server or a Swift toolchain.
+  # ─────────────────────────────────────────────────────────────────────────────
+  probes-and-stress:
+    name: tmux probes + flake-ledger stress loop
+    runs-on: macos-26
+    timeout-minutes: 180
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install tmux
+        # macos runners don't ship tmux, and the probes drive a real server.
+        run: brew install tmux
+
+      - name: Select Xcode 26.6 (Swift 6.3.3)
+        run: sudo xcode-select -s /Applications/Xcode_26.6.app
+
+      - name: Capture Swift toolchain version
+        run: echo "SWIFT_VERSION=$(xcrun swift --version | head -1)" >> "$GITHUB_ENV"
+
+      # RESTORE-ONLY, deliberately. A nightly that also SAVES would churn the
+      # cache every night and could evict the entry PR builds depend on. This job
+      # is allowed to benefit from the cache; it is not allowed to reshape it.
+      - name: Restore SwiftPM build cache (read-only)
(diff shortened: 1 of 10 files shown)


cheapsteak:
Fixed in 33db3760 — the Medium finding was correct, and I reproduced it rather than taking it on faith.

**Confirmed:** `bash scripts/nightly-quarantine-audit.sh analyze --ledger-dir /tmp/nope 2>/dev/null | tee` captures **0 bytes**. The `die()` path (`nightly-quarantine-audit.sh:65`) writes to stderr and exits 2, the step treats any non-zero as a failure and comments on #519 — so that comment would have arrived with an empty body. Exactly the "harness error still has to reach a human" case, and the one place the body has to carry the diagnosis, since a clean run posts nothing at all.

After the fix the same invocation captures 62 bytes: `nightly-quarantine-audit: analyze: --issue-states is required`.

**I also added a mechanical guard rather than just fixing the line.** Two correct siblings in the same file did not stop the third from being wrong, so a habit is not the right countermeasure. `nightly-report.test.sh` now asserts that every tee'd step in `nightly.yml` redirects stderr into the body it will post, and that there are still exactly three of them. Mutation-checked: 1 offender against the pre-fix file, 0 against the fixed one.

Re-verified: actionlint clean, shellcheck clean on all five scripts, all four suites green (112 assertions).

---

On the two set-aside items, briefly, since both are fair:

- **`pkill -P` reaching one level** — agreed, and your scoping is right. Worth noting it is *deliberately* one level: the nightly runs on an ephemeral runner, but the same script runs on the shared dev box, where a broader sweep is the hazard `Tests/CLAUDE.md` documents (`pkill -f` matching sibling worktrees cost someone ~70 minutes). One level by parent PID is the widest kill that is definitely mine.
- **Worst-case iteration timing exceeding `timeout-minutes`** — correct arithmetic. That ceiling is intentional: the step timeout is the outer guard precisely because the per-iteration deadline can't be trusted alone, and a force-kill on total simultaneous failure is the right outcome rather than a case to engineer around.

On your diagnostics note: `shellcheck` and `py_compile` were blocked in your sandbox, so those claims rested on my word. They are re-run above and I have kept them in the test plan; the four `.test.sh` suites are runnable locally by anyone who wants to check without trusting either of us.

cheapsteak:
Fixed in e9cd4838. The High was correct, and it is the most valuable finding of the review because **my own verification could not have caught it.**

I ran `fetch()` end-to-end against live GitHub — 46 real `main` runs, 16 real artifacts — and reported that as proof the audit works. It was proof the audit works **under my personal `gh` auth**, which carries scopes the constrained workflow token does not. The one thing that run could not exercise is the one thing that was broken. That is the same class as the trap slice E hit: a mechanism that looks verified because the verification ran in the wrong environment.

**Verified against this repo's own precedent rather than just accepting the reasoning:** `claude-code-review.yml:34` and `claude.yml:26` both explicitly declare `actions: read`, and `claude.yml` even carries the comment explaining why. `nightly.yml` was the only workflow missing it.

**The guard I added is derived, not hardcoded.** It greps this script for Actions-API usage and asserts the permission *only while that usage exists* — so it tracks the code instead of restating a fact about it, and it fires in both directions: strip the scope and it goes red, add an Actions-API call to a workflow lacking the scope and it also goes red. Mutation-checked against a copy with `actions: read` removed (guard reports the failure; passes on the real tree).

**Minor also fixed.** The audit step now captures its exit code and titles the two cases differently — `2` is "FAILED TO RUN (harness error) — it audited nothing", `1` is "found something". A 403 on the Actions API is not a statement about the quarantine list. Verified both paths locally: harness error → rc 2, closed-issue finding → rc 1.

Re-verified: actionlint clean, shellcheck clean, all four suites green.

---

On the filtered nitpick — you were right to filter it, but it is a real (comment-only) error and I would rather fix it than let it sit, so I will take it in the same pass if you flag which doc-comment you meant. My P8 comment attributes the control-mode block parser to `TmuxControlConnection`; if the parser actually lives elsewhere, the probe still asserts the right protocol behaviour but the pointer sends a reader to the wrong file.

PR #538, opened 2026-07-27
Fix: New worktrees open on the Notes tab instead of the agent

cheapsteak:
## Summary

**What's broken.** You create a worktree, click into it, and the first thing you see is the empty **Notes** tab — not the agent that is already running one tab to the left.

**Why it happens.** The daemon is right, and always was: `createInitialNoteTab` appends the note **last** in `tabOrder`, and `spawnPrimaryTerminals` persists `activeTabID = primary`. I confirmed both against the live DB — every recent worktree has the note last and Claude as the active tab. The app was throwing that selection away three separate ways:

1. **Eight call sites each spelled their own fallback** — `?? 0` in some, `min(idx, tabs.count - 1)` in others. Because the note is always the *last* tab, any stale or oversized index clamped straight onto Notes.
2. **`reconcileTabs` / `reconcileNoteTabs` captured the active tab's identity *after* mutating `tabs[worktreeID]`**, so removing a tab re-bound the index to whatever slid into that slot.
3. **`reconcileTabs` treated an empty `listTabs` response as "already loaded."** A poll landing between the daemon inserting the terminal rows and persisting the tab order gets `[]` + `nil`, which latched the gate permanently — that worktree never got its persisted "active = agent" selection for the rest of the session.

**How it got broken.** All three are older than the symptom. The clamp came in with Cmd-W close (#217), the load-once gate with tab persistence (#135). Both were harmless while the last tab was just another terminal. #485 made every new worktree grow a trailing note tab **and** turned on `auto_close_setup_enabled`, so the `setup` tab now gets removed on essentially every create — turning mechanism 2 from an edge case into a routine event, and giving mechanisms 1 and 3 a note tab to land on. Nothing caught it because the app's tab-selection fallbacks had no tests at all; each call site looked locally reasonable.

**What this PR does.**

- Adds one resolver, `resolvedActiveTabIndex`: the stored selection when it is still in range, otherwise the first **non-note** tab, otherwise `0`. All eight reads route through it, so they can no longer disagree. A deliberately-selected note tab is still returned verbatim.
- Both reconciles capture the active tab's id **before** mutating and pass it as an explicit `ActiveTabAnchor`. A bare `UUID?` conflated "caller omitted an anchor" with "caller captured no valid selection" — opposite cases needing opposite handling.
- Hydration latches only on a response that carried content, bounded by an attempt cap with in-flight dedup. A disconnect refunds its attempt; any other error consumes budget, so a persistently-failing `listTabs` can't re-fire every 2s forever.
- Fixes `closeTab`, which clamped unconditionally even when the closed tab wasn't active: with `[Claude, shell, Notes]`, closing the background `shell` tab jumped you to Notes.

**Not fixed here, worth knowing:** `closeTab` still never persists the new selection to the daemon. This was raised on #217 and never applied. Out of scope, but `repointActiveTab` is now the natural chokepoint if we want it. (An earlier draft of this section claimed `setActiveTab` has a single production call site, the tab-bar click — that is wrong. There are five: `TerminalContainerView.swift:99`, `AppState+Worktrees.swift:485`, `AppState+ModelProfiles.swift:407`, and `AppState+History.swift:172,191`. Anyone scoping the follow-up should size it against all five.)

## Test plan

- [ ] 27 new tests in `ActiveTabResolutionTests`, each mutation-checked — the fix was reverted per finding and the covering test confirmed red, then restored.
- [ ] `swift build` clean; `swiftlint --strict` clean (0 violations, 605 files).
- [ ] `swift test --parallel -j 2 --filter TBDAppTests` — 1612 tests passing.
- [ ] Full suite's only reds are the live-tmux integration suites, **identical with this branch stashed** (tmux `new-session` can't run in that shell) — environmental, not a regression.
- [ ] **Not yet verified live.** The symptom was diagnosed from the code and the DB, not reproduced in a running app. Worth one manual pass: create a worktree, confirm it opens on the agent tab; then click Notes, navigate away and back, and confirm the deliberate choice sticks.

One `TabReorderTests` fixture gains an explicit `activeTabIndices` entry. It described a state the app cannot be in — focus in a tab that isn't the active one — and its assertions are unchanged.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=A45257B4-AB1C-444C-9FF9-ACA0BDE575DD)

### DIFF
--- a/Sources/TBDApp/AppState+Notifications.swift
+++ b/Sources/TBDApp/AppState+Notifications.swift
@@ -75,14 +75,28 @@ extension AppState {
 
     // MARK: - Helpers
 
+    /// Whether an error means "the daemon is gone" rather than "this one call
+    /// failed". Only these two cases stop the poll timer; every other
+    /// `DaemonClientError` — a decode failure, an RPC-level rejection, a
+    /// timeout — leaves the app connected and the ~2s poll running, so the
+    /// caller keeps being re-driven.
+    ///
+    /// Shared so retry budgets and the connection flag cannot disagree about
+    /// what counts as a disconnect: a budget that forgave *every* error would
+    /// never bite on the persistent per-worktree failures this does not cover.
+    static func isDisconnectError(_ error: Error) -> Bool {
+        guard let dcError = error as? DaemonClientError else { return false }
+        switch dcError {
+        case .daemonNotRunning, .connectionFailed:
+            return true
+        default:
+            return false
+        }
+    }
+
     func handleConnectionError(_ error: Error) {
-        if let dcError = error as? DaemonClientError {
-            switch dcError {
-            case .daemonNotRunning, .connectionFailed:
-                isConnected = false
-            default:
-                break
-            }
+        if Self.isDisconnectError(error) {
+            isConnected = false
         }
     }
 
--- a/Sources/TBDApp/AppState+Tabs.swift
+++ b/Sources/TBDApp/AppState+Tabs.swift
@@ -10,40 +10,198 @@ enum DropEdge: Sendable, Equatable {
     case leading, trailing
 }
 
+/// What the active selection should follow when `applyStoredOrder` re-points it.
+///
+/// A bare `UUID?` cannot express this: it conflates "the caller omitted an
+/// anchor, derive one" with "the caller captured no valid pre-mutation
+/// selection". Those need opposite handling — deriving after a mutation reads
+/// the already-mutated array, where a stale index has silently re-bound to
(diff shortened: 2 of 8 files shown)


cheapsteak:
Good catch on the call-site count — corrected in the PR body. `setActiveTab` has five production call sites, not one: `TerminalContainerView.swift:99`, `AppState+Worktrees.swift:485`, `AppState+ModelProfiles.swift:407`, and `AppState+History.swift:172,191`. I inherited that claim from an earlier review pass and repeated it without checking, which would have made the follow-up look four-fifths smaller than it is. Also fixed the test count (27, not 30).

PR #545, opened 2026-07-28
fix: new worktrees failed silently and bounced the user back

zionts:
## Summary

Fixes the issue where clicking "+" / pressing Cmd+N to create a new worktree appeared to "redirect the user back to an existing worktree" instead of opening a fresh one.

## Diagnosis (verified causal chain with evidence)

1. `DaemonClient.launchDaemon` spawns TBDDaemon inheriting the GUI app's minimal LaunchServices PATH (`/usr/bin:/bin:/usr/sbin:/sbin`)
2. `/opt/homebrew/bin` is absent, so `git-lfs` (installed there) is not resolvable by git subprocesses
3. Both longeye repos set `filter.lfs.required = true`. When git attempts `git worktree add`, it fails: `git-lfs filter-process --skip: git-lfs: command not found` → exit 128
4. `attemptWorktreeAdd` (WorktreeLifecycle+Create.swift:430-491) tries 4 combinations with every `catch` discarding the underlying GitError entirely
5. `handleWorktreeCreate` (RPCRouter+WorktreeHandlers.swift:107-114) broadcasts `.worktreeArchived` — indistinguishable from normal archive
6. `applyWorktreeArchivedDelta` (AppState.swift:1799) → `navigateBackPastArchived` applies the previous navigation-history entry, which is the worktree the user was on before pressing "+". **This is the observed "redirect", with no alert or toast shown.**

## Reproduction

```bash
git -C /Users/zionts/tbd/repos/longeye-app -c checkout.workers=0 worktree add /tmp/test -b tbd/test origin/main
# Fails: git-lfs filter-process --skip: git-lfs: command not found

# With PATH fixed:
PATH=/opt/homebrew/bin:$PATH git -C /Users/zionts/tbd/repos/longeye-app -c checkout.workers=0 worktree add /tmp/test -b tbd/test origin/main
# Succeeds
```

### Part 1 — Daemon subprocess PATH
- Added `ToolPathAugmenter` (new file: Sources/TBDDaemonLib/ToolPathAugmenter.swift)
  - Pure, unit-testable helper that takes current PATH and returns augmented one
  - Prepends `/opt/homebrew/bin`, `/opt/homebrew/sbin`, `/usr/local/bin`, `/usr/local/sbin`
  - Preserves existing entries and avoids duplicates
- Modified `Sources/TBDDaemon/main.swift` to call it at startup via `setenv()`
- Ensures all daemon child processes inherit the full PATH

### Part 2 — Stop swallowing git errors
- Modified `attemptWorktreeAdd` in Sources/TBDDaemon/Lifecycle/WorktreeLifecycle+Create.swift
  - Captures error from each failed attempt and logs at .warning level
  - Added `formatErrorForMessage()` helper to extract stderr from GitError
  - Includes last error details (truncated to ~500 chars) in both thrown error messages
  - Makes future occurrences diagnosable from `log show` alone

### Part 3 — Surface failed creation as error alert
- Added `creationFailureMessageIfCreating()` static helper on AppState
  - Pure, nonisolated function for unit testability
  - Checks if worktree status is `.creating` and returns formatted message
- Modified `applyWorktreeArchivedDelta()` to detect `.creating` → `.archived` transitions
  - Shows error alert naming the worktree, directing user to Console logs
  - Keeps existing removal + `navigateBackPastArchived` behavior
  - Edge case: when THIS client archives its own `.creating` worktree, the row is already removed before the delta echo arrives, so no false alert fires

## Tests

- `ToolPathAugmenterTests` (new file): 8 unit tests for augmentation logic
  - nil PATH, empty PATH, duplicate avoidance, Homebrew already present, minimal LaunchServices PATH
- `ArchiveTombstoneTests`: Updated fixture + 6 new tests
  - Static helper: .creating produces message, .active/nil do not
  - Full delta flow: .creating worktree archived shows alert, .active does not, unknown graceful

## Why no feature flag?

All three are bug fixes, not new autonomous behavior:
- Part 1: routine subprocess environment setup
- Part 2: better logging for failures that already exist
- Part 3: visibility of a silent failure mode

Per CLAUDE.md, feature flags are required only for "acts without a user gesture", "kills processes or mutates state", or "wholesale-replaces a load-bearing path". None apply here.

---

### DIFF
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -1797,7 +1797,44 @@ final class AppState: ObservableObject {
     /// client). Tombstone it and drop the row so it cannot be resurrected by a
     /// poll snapshot that predates the archive.
     private func applyWorktreeArchivedDelta(_ delta: WorktreeIDDelta) {
+        // Check if this was a .creating worktree that failed to complete.
+        // Look it up before it gets removed so we can detect the failure.
+        let worktree = findWorktreeInState(id: delta.worktreeID)
+        let failureMessage = Self.creationFailureMessageIfCreating(worktree)
+
         removeArchivedWorktreeFromState(id: delta.worktreeID)
+
+        // If it was still creating when it disappeared, its creation failed.
+        // Show an error alert instead of silently redirecting the user.
+        // Note: when THIS client archives a still-.creating worktree, the row
+        // is already removed by removeArchivedWorktreeFromState, so the lookup
+        // will fail and no false alert fires.
+        if let message = failureMessage {
+            showAlert(message, isError: true)
+        }
+    }
+
+    /// Returns a failure alert message if the worktree was in `.creating` status
+    /// when it disappeared, or nil if it was in any other status or unknown.
+    /// Pure static helper for testability — both branches are unit-testable
+    /// without a daemon or SwiftUI, following the pattern of `archiveShortcutRoute`.
+    nonisolated static func creationFailureMessageIfCreating(_ worktree: Worktree?) -> String? {
+        guard let worktree, worktree.status == .creating else {
+            return nil
+        }
+        return "Couldn't create worktree \"\(worktree.displayName)\" — the git worktree add failed. " +
+               "See Console (log show --predicate 'subsystem == \"com.tbd.daemon\"') for details."
+    }
+
+    /// Finds a worktree by ID in the current state (both repo-based and scratch).
+    /// Returns nil if not found. Used for checking properties before removal.
+    private func findWorktreeInState(id: UUID) -> Worktree? {
+        for repoWorktrees in worktrees.values {
+            if let wt = repoWorktrees.first(where: { $0.id == id }) {
+                return wt
+            }
+        }
+        return scratchWorktrees.first(where: { $0.id == id })
     }
 
     /// Apply a Claude session rollover (post-`/clear` / `/compact` / startup)
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -1797,29 +1797,37 @@ final class AppState: ObservableObject {
     /// client). Tombstone it and drop the row so it cannot be resurrected by a
     /// poll snapshot that predates the archive.
     private func applyWorktreeArchivedDelta(_ delta: WorktreeIDDelta) {
-        // Check if this was a .creating worktree that failed to complete.
-        // Look it up before it gets removed so we can detect the failure.
+        // Look the row up before it gets removed so we can name it in the alert.
         let worktree = findWorktree(id: delta.worktreeID)
-        let failureMessage = Self.creationFailureMessageIfCreating(worktree)
+        let failureMessage = Self.creationFailureMessage(worktree, creationFailed: delta.creationFailed)
 
         removeArchivedWorktreeFromState(id: delta.worktreeID)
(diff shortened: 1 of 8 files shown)


cheapsteak (approved):
🙏🙏🙏

cheapsteak:
apologies, reviews on contributors' forks should work after https://github.com/cheapsteak/tbd/pull/545 merges

cheapsteak:
Addressed the Medium in `8cf35aa6` (pushed by a maintainer to the fork branch — thanks @zionts for leaving "allow edits by maintainers" on).

**Medium — false "creation failed" alert on a deliberate archive of a `.creating` row.** Confirmed; I verified all four cited references before acting:

- `resolveWorktreeNameOrID` filters on name/displayName only, no status filter — `WorktreeCommands.swift:677`
- `beginArchiveWorktree` rejects only `.main`; its other guard (`assertArchivable`) checks *children*, not the row's own status — `WorktreeLifecycle+Archive.swift:52-61`
- `creationFailureMessageIfCreating` inferred failure purely from `status == .creating` — `AppState.swift:1821`
- the app-side path *does* guard it (`.main || .creating` → `nil`) — `AppState+Worktrees.swift:1271`

So the asymmetry was exactly as described, and the review's premise audit was right that the PR's own "row is already removed by this client" reasoning holds for the app but not the CLI.

**On the suggested remedy — I went the other way.** Of the two options offered, making the archive path reject `.creating` would fix the alert by removing the capability, and the finding's own motivating scenario ("a developer bailing out of a stuck/slow pre-session hook") is precisely when you want `tbd worktree archive` to work. So instead the daemon now *says* which happened rather than the app guessing: `WorktreeIDDelta` carries `creationFailed`, set true at the single site that represents a real failure (the `catch` around `completeCreateWorktree`), and the alert gates on that flag alone. Deliberate archives — CLI, GUI menu, auto-archive-on-merge — keep the default `false` and stay silent. CLI cancellation keeps working.

One wrinkle worth flagging for review: `creationFailed` decodes through an explicit `init(from:)` using `decodeIfPresent ?? false`, because a synthesized decoder ignores property defaults and would throw `keyNotFound` against an older daemon. Daemon/app version skew is a real state in this repo, so the default-on-absent matters.

Tests cover both branches of the new gate plus legacy-payload decoding. Verified by mutation — restoring the old status-based inference fails 3 of them, so they're not passing vacuously. Full fast suite: 4546 tests, 1 known issue (the `FlakyQuarantineSelfTests` fixture, by design). `swiftlint --strict` clean.

**Minors — dispositions:**

- **`WorktreeStatus.failed` prior art.** Agreed, and it's the better long-term shape. Deferring as you suggested; the `creationFailed` flag is a strictly smaller change that doesn't alter the row lifecycle, and a `.failed` transition deserves its own PR.
- **TBDApp's own direct `Process` calls** (`FileViewerPanel`, `BedrockModels`). Correct and pre-existing — same LaunchServices PATH, same latent git-lfs failure. Out of scope here; worth a follow-up issue.
- **Verified plumbing / dropped items.** No action needed.

Also folded in `4d6a4d86`, which restored a missing `return` in `ArchiveTombstoneTests.makeWorktree` — the fixture had gained a `let shortID` line, making the body multi-statement so implicit return stopped applying, and the whole `TBDAppTests` target failed to compile. That's what the red `test` check was; it wasn't the ControlMode/debounce flake.

Issue #496, opened 2026-07-25
Clock-driven tests wedge under heavy load: TestClock.advance calls megaYield (background-QoS) twice per advance


cheapsteak:
Split out of slice C1 (test-hardening program). **Not diagnosed further than the mechanism below — filed so slice J's "option C" has a home.**

## Symptom

A `TestClock`-driven tier-1 suite can wedge for minutes under heavy machine load, tripping `.clockDriven`'s 60s limit. Observed on `AppearanceDebounceTests` at load ~90-200 on a 12-core box: worst single suite runs of **467s** and **679s**, against a healthy-path cost of ~0.05s isolated / ~8.5s inside the 1332-test target.

## Mechanism (source-verified, not inferred)

`TestClock.advance(to:)` calls `Task.megaYield()` **twice per advance**, and `checkSuspension()` calls it once per probe — both inside `swift-clocks`:

```swift
// swift-concurrency-extras, Task.swift
for _ in 0..<count { await Task.detached(priority: .background) { await Task.yield() }.value }
```

That is 20 serially-awaited **background-QoS** detached tasks per call (default `_defaultMegaYieldCount == 20`). macOS starves background QoS under CPU saturation, so each call can stall arbitrarily. Every clock-driven test pays it on its hot path, and a large parallel population of such tests floods the pool with exactly the low-priority work each one is waiting on.

## What has already been eliminated

- **Option A — lower `TASK_MEGA_YIELD_COUNT`** (env-tunable, default 20): **refuted as a fix.** Measured at `count=1`; the wedge still occurred (679s). Whether it reduces *frequency* is unmeasured — the comparison batch was invalidated by uncontrolled load drift (~60-100 vs ~167-203) on a shared box. Note the dial cuts both ways: `advance(to:)`'s megaYields are what let sleepers get resumed and re-armed, so lowering it may make `advance` *less* able to make progress even as each call gets cheaper.
- **Option B — `withMainSerialExecutor`** (swift-concurrency-extras): **refuted on source evidence.** It sets a **process-global** runtime hook (`swift_task_enqueueGlobal_hook`) routing every enqueued job onto the main actor, and its own doc concedes "no scoping guarantees". In a 1332-test parallel target, a few tests entering that scope hijack scheduling for all the others and the `defer`-restores race each other. It is designed for whole-suite *serialized* usage, not our topology.
[...]

Issue #494, opened 2026-07-25
Flaky: ControlModeInputHealthTests — two independent contention races


cheapsteak:
Two distinct flakes observed in `Tests/TBDDaemonTests/ControlModeInputHealthTests.swift`. Filed per the test-hardening program (slice D); **neither is diagnosed** — slice G's nightly flake-ledger stress loop is the intended venue for characterising them.

Both are contention-only: the suite passes 5/5 in isolation for both flakes.

## 1. Delivery starvation — second health event never lands

Observed by another slice owner during a full `swift test --parallel -j 2` run (1 failure in 8 runs):

```
ControlModeInputHealthTests "health events carry the attach generation registered with the route (R6-M7)"
Expectation failed: health.events.map(\.generation) → [42] == [42, 42]
FAILED AFTER 62.123 SECONDS
```

**The quoted expectation is not the failure.** The preceding line waits via `ControlModeTestSupport.waitFor`, which is non-throwing: on deadline it records `timed out waiting for 2 health events` and returns `false`. Execution then continued into an equality assertion against a half-filled array, producing a second, misleading issue that points at generation numbers when the real problem is that only one event ever arrived.

The 62-second duration is the diagnostic: that is two ~30 s `ciSafeDeadline` waits, not a fast assertion failure.

So the actual defect is that the second health event did not land within 30 s under `-j2` contention. Suspected to be in the same background-QoS starvation family as the `megaYield` stalls found elsewhere in this program, but **that is a hypothesis, not a measurement.**

This report's misattribution risk is now mitigated: all 12 previously-unguarded `waitFor*` call sites in this file are `#require`-guarded, so a future timeout fails at the wait with its own message.

## 2. `repeatedFailuresFireOnce` — a third health transition appears

```
Test "repeated send-keys failures fire exactly ONE failing transition"
Expectation failed: (events.count → 3) == 2
```

Seen once in ~7 runs; not reproduced in 10 subsequent runs (5 with the guard change, 5 without), so it is genuinely intermittent and **not** attributable to the guard change — that change cannot cause the router to emit an extra transition.

[...]

PR #547, opened 2026-07-29
DX: Stop the ~42% red rate on main -- the fast pass outgrew its own deadlines

cheapsteak:
## Summary

`main`'s Test workflow was red in **5 of the last 12 runs (~42%)**. The failing test name shuffled between `ControlModeInputHealth`, `ControlModeInputRouter` and `Archived search debounce` — but **every single failure was a wall-clock scheduling handshake timing out. Zero logic assertions failed.**

**Root cause is population, not any one test.** Swift Testing parallelizes every non-serialized test in one process with no concurrency cap, and all targets compile into that process, so per-test scheduling latency scales with the *total* population. Mined across 9 CI runs: **p50 per-test reported duration is ~1/3 of total wall time in every run** — a trivial test "takes" 16–29 s because it is mostly suspended waiting for a turn. The population grew **3013 → 4536 in three weeks**, pushing ambient latency into deadlines sized for the smaller process. `ControlModeInputHealth` finishes green at 45.8 s / 47.4 s and red at 47.9 s — a ~2 s coin-flip band, which is exactly why the name shuffles.

**The disconfirming evidence is the important part.** I expected red runs to show a fat duration tail and greens not to. They don't — saturation is the steady state of *every* run:

- Green run `30237707480`: 1097 tests >40 s, max 84.7 s — the **worst tail in the whole sample**, and it passed clean.
- Green `30279274612` (80.9 s wall, 806 >40 s) is statistically indistinguishable from red `30286937025` (81.3 s wall, 998 >40 s).

So "why was this run slow" will never converge. What flips a run red is **margin** on a fixed deadline.

**How it got this way.** Nothing here is a regression from one commit. Two things compounded: ordinary feature work grew the suite ~50%, and the test-hardening program added **13 `TestClock`-driven suites in 96 hours** (2026-07-24→27) where there had been zero. Issue #496 predicted this exact risk — *"the risk is that CI gets noisier, or that a slice ships a much larger clock-driven population"* — and guessed it wasn't CI-relevant yet. These logs say otherwise. It wasn't caught because the deadlines are hang-catchers that only elapse on failure, so they degrade silently until they don't.

## What this does

**1. Split the fast pass into two sequential invocations in the same job, sharing one build.** Halving the in-flight population halves the tail the deadlines must absorb: **p90 26.4 s → 14.6 s** (means over 5 interleaved iterations under induced load, population held constant), for **+26 s** of wall time.

This is **not** the "sharding across runners" that `docs/specs/2026-07-24-test-hardening-design.md` §1 rejected and §2 lists as a non-goal — that rejection was about extra *jobs* paying the 5-concurrent-macOS-job cap and re-paying the ~2 min build. This pays neither; it's the same shape the file already uses for the fast/quiet split.

Step 2 is a **complement** (`--skip '^(TBDDaemonTests|TBDDaemonLiveTests)\.'`), not an enumeration. Two `--filter` lists would be fail-open: `swift test --filter` exits GREEN on zero matches, so a new target named in neither list would run in *neither* pass — and the floors can't catch that, because adding a target reduces no existing step's count. As a complement, the three passes partition the package by construction.

**2. Re-derive the handshake deadlines as a coherent triple.** `ciSafeDeadline` 30→90 s, `waitForSuspension` 15→45 s, `.clockDriven` 1→4 min. These are hang-catchers, not timing budgets — no test here asserts a timing property, and on success they cost nothing. Repo precedent: PR #415 raised an incidental `readyTimeout` 5 s → an un-hittable 600 s for exactly this reason, and that default is still live at `AttachRPCTests.swift:137`.

**3. Three tier-3 live suites pin `.timeLimit(.minutes(1))` rather than inherit the raise.** For them the 60 s bound *is* a mutation-verified detector (a `sleep 90` child; a ~120 s regressed drain) — inheriting 240 s would have silently disarmed the proof with nothing going red to say so. They run `--no-parallel` on an idle machine and never needed the raised budget. This was the most dangerous thing in the change and it was nearly shipped.

## Verified under induced load, not idle

- **Pre-fix tree WEDGED** at load ~260: no `Test run with N tests` summary, harness deadline at 912 s, single test at **727 s**.
- **The exact test that reds in CI** (`ArchivedWorktreeSearchTests.swift:346`) took **176.1 s unsharded** vs **43.0 s sharded**. Against the old 60 s limit the unsharded case dies with precisely the CI signature: `waitForSuspension` times out, `advance` moves virtual time with no sleeper registered, and `#expect(fired.values == ["wolv"])` fails with `[]`. Both fixes independently rescue it.

### Measured on CI itself (both runs, not the flattering one)

`pull_request` runs the workflow from the branch, so this PR exercises the sharded pipeline on itself.

| metric | pre-fix runs (9 mined) | run 1 | run 2 | run 3 |
|---|---|---|---|---|
| p50 per-test duration | 23.1 - 35.8 s | 3.7 s | 5.2 s | 3.9 s |
| p90 | 41.4 - 51.0 s | 22.3 s | 33.0 s | 20.1 s |
| tests > 40 s | 589 - 1693 | 2 | 110 | 2 |

Run 2 is materially worse than 1 and 3, and all three are reported rather than the flattering ones — runner variance here is large, and the pre-fix data showed exactly why a single sample misleads (its worst tail belonged to a run that passed). What holds across all three is what the diagnosis rests on: p50 sits below even the best pre-fix **green** run (16.4 s), and the >40 s tail is down by one to three orders of magnitude from the pre-fix range.

## Ruled out, with measurements rather than assertions

- **`TASK_MEGA_YIELD_COUNT=1`** — issue #496's open "option A". p90 26.4 → 26.2 s across interleaved runs. Noise. The megaYield background-QoS flood is real but **secondary** to raw population. This closes that question with the properly-interleaved measurement the earlier attempt lacked (it was invalidated by load drift on a shared box).
- **Per-suite `@Suite(.serialized)`** — `Archived search debounce` is *already* `.clockDriven, .serialized` and still failed twice, because the contention is process-global. `Tests/CLAUDE.md`'s "reach for `.serialized` before a longer timeout" advice is scoped to a suite starving *itself*; amended to say so.
- **Blanket retry / anonymous quarantine** — banned, and unnecessary: this has a root cause.

## Honest caveats

- One verification iteration went red on `FileWatcherTests.terminationWithAPendingDebounceStillClosesTheFD` — an **FD-count** assertion, not a handshake deadline, in a suite this PR touches only in comments. It did not recur, and did not appear in any of the 9 mined CI runs. Flagging as a separate load-sensitive flake to watch, not something this change causes or claims to fix.
- Two later iterations were discarded as **contaminated**: a parallel agent was relinking the test binary underneath the running experiment. They are not reported as failures.
- The deadline change alone might be sufficient; sharding is what buys headroom for the *next* 50% of growth. Easy to revert independently if the +26 s isn't worth it.

## Test plan

- [x] `swift build --build-tests -j 2` clean
- [x] `swiftlint --strict` — 0 violations, 614 files
- [x] Shard 1: **2194** tests green (35.3 s)
- [x] Shard 2: **2342** tests green (11.3 s) — 2194+2342 = 4536, the full population, no gap
- [x] Quiet pass: **57** tests green (28.4 s)
- [x] Verified the three regexes partition the five test targets with no gap and no overlap
- [ ] Watch `main`'s red rate after merge — that's the real acceptance test

Refs #494, #496.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=809725DF-7574-4EDD-990A-6E6DE6F86036)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -193,7 +193,16 @@ jobs:
       # tests are conditionally skipped at runtime. The floors are set against
       # the executed counts (the summary line these steps grep), which is the
       # only number the check can see; don't "correct" them to the listed
-      # totals. Observed in CI on this PR: 2194 / 2342 / 57.
+      # totals.
+      #
+      # And the executed count is NOT stable run to run: two CI runs on this PR,
+      # on commits differing only in doc comments (zero tests added,
+      # `--list-tests` unchanged), executed 2194 and 2204 for step 1. The cause
+      # is not established — it is not conditional skips (the only
+      # `.enabled(if:)` here is gated on an env var CI does not set) and not
+      # parameterization (all `arguments:` are static literals). Left unexplained
+      # rather than guessed at. The practical consequence is the point: pin a
+      # floor near the exact count and it WILL fire spuriously. Keep the margin.
       # The floors sit well below each step's total
       # so ordinary test additions/removals never trip them. Their job is NOT
       # to catch an unlisted new target (the complement shape already does
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,10 +34,10 @@ jobs:
 
   test:
     runs-on: macos-26
-    # Both test steps below append their `.flaky(issue:)` retry records here
-    # (docs/specs/2026-07-24-test-hardening-design.md §7), so one run produces
-    # one artifact covering the fast and quiet passes together. Unset locally,
-    # where the trait writes nothing.
+    # All three test steps below append their `.flaky(issue:)` retry records
+    # here (docs/specs/2026-07-24-test-hardening-design.md §7), so one run
+    # produces one artifact covering both fast passes and the quiet pass
+    # together. Unset locally, where the trait writes nothing.
     #
     # A LITERAL path, not `${{ runner.temp }}`: the `runner` context is not
     # available in a job-level `env:` block, and GitHub resolves an unavailable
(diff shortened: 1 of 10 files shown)


cheapsteak:
Thanks — the Minor finding was right and is fixed. The Medium finding is **incorrect**, and I want to show the evidence rather than just assert it, because the miscount is an easy one to repeat.

## The Medium finding rests on an undercount

`advanceWhenSuspended` **is** a `waitForSuspension` — it calls one before advancing, so it pays the full 45 s guard:

```swift
// Tests/TestSupport/ClockTestSupport.swift
func advanceWhenSuspended(by duration: Duration, sourceLocation: SourceLocation = #_sourceLocation) async {
    await waitForSuspension(sourceLocation: sourceLocation)   // <- pays the 45s guard
    await advance(by: duration)
}
```

Counting only literal `waitForSuspension(` call sites misses every wrapped one. The actual sites:

**`GatedIntervalSleepTests.returnsAfterExpectedPollCount`** — 5, not 2:
```
await clock.advanceWhenSuspended(by: Self.tick)
await clock.waitForSuspension()
await clock.advanceWhenSuspended(by: Self.tick)
await clock.waitForSuspension()
await clock.advanceWhenSuspended(by: Self.tick)
```

**`DaywatchRunnerTests.testSubsequentTicksAtInterval`** — 5, not 3:
```
await clock.waitForSuspension()
await clock.advanceWhenSuspended(by: .milliseconds(1))
await clock.waitForSuspension()
await clock.advanceWhenSuspended(by: .seconds(DaywatchRunner.defaultInterval))
await clock.waitForSuspension()
```

**`PaneRepairCoordinatorTests`** — 9 of 13 over 240 s, not 7. Mapping every `waitFor` (90 s) and every clock wait (45 s, both forms) to its enclosing `@Test`: 540, 450, 450, 405, 360, 360, 360, 360, 270 are over; 180, 180 are under; `supersededBeforeRepairStartSendsNothing` and `supersededAcrossProviderHopSendsNothing` have zero waits. That's 9 over, 2 under, 2 with none = 13.

So the "225 s of a 240 s ceiling" figure stands, and the direction matters: the review concluded the real margin is *larger* than documented, when it is exactly as documented.

## What I changed anyway

The ambiguity is real and it is mine. The prose said "chains 5 `waitForSuspension` calls", which invites precisely the literal grep that produced the finding. All three files now state the counting rule up front — that `advanceWhenSuspended` counts as one — and cite this miscount so the next reader doesn't repeat it.

## Minor finding — fixed, good catch

`test.yml` listed `--list-tests` totals (1725+502+127 = 2354) next to floors that are checked against **executed** counts (2342; a few tests are conditionally skipped at runtime). The comment now reconciles the two explicitly and records what CI actually reported on this PR — 2194 / 2342 / 57 — so nobody "corrects" the floors to the listed totals.

## Worth noting from this run

`test` passed, and it exercised the new sharded pipeline on itself (`pull_request` runs the workflow from the branch). Measured on that run: **p50 per-test duration 3.7 s, p90 22.3 s, 2 tests over 40 s** — against 23.1–35.8 s / 41.4–51.0 s / 589–1693 across the pre-fix runs. p50 now sits below even the best pre-fix *green* run (16.4 s), so the saturation the diagnosis identified is gone rather than merely papered over.

PR #551, opened 2026-07-30
fix: three stale facts the nightwatch judge was acting on

zionts:
Three facts baked into `Sources/TBDShared/NightwatchSkillContent.swift` and `NightwatchDeskPrompts.swift` had gone stale. That file is the source of truth the daemon installs to `~/Library/Application Support/TBD/plugin/skills/nightwatch` on every start, overwriting local edits — so this is the only place they can be fixed.

## 1. Context ceiling: 200k → 600k

The desk runs on Opus with a 1000k window. At 200k it hit the ceiling roughly every two hours and spent the shift relaying instead of judging, with 800k of headroom unused. Raised to 600k in `handoff.py`'s `DEFAULT_THRESHOLD`, in `SKILL.md`, and in both desk prompts.

**A trap worth calling out:** `handoff.py --selftest` asserted `"over ceiling exits 10"` against a hardcoded `250_000` fixture. Raising the ceiling to 600k turns that into an *under*-ceiling case — the selftest would have kept passing while asserting the opposite of its own label, and `PluginDirWriterTests.handoffPySelftest` would have gone red for a reason nobody would have connected to the ceiling. The fixture now derives from `DEFAULT_THRESHOLD`, and a test pins that derivation so the next raise can't re-plant it.

## 2. Merge authorization (Adam, 2026-07-29)

The prompts said `NEVER run gh pr merge` and *"the human merges"*. Replaced with the actual rule:

The **nightwatch** judge MAY merge a PR **authored by `zionts`**, and only his, once it is **loop-perfect** — all three re-read in the same tool call as the merge:

- `mergeable_state` is `clean`
- `claude-review` = APPROVED **on the current head SHA**
- every required status context green

Never `--admin`. Never anyone else's PR. Never merging *toward* green.

**Merge-queue nuance, now stated in the prompts.** `main` is behind a merge queue, so `gh pr merge <N>` *enqueues*; GitHub merges when the queue's checks pass. The line `! The merge strategy for main is set by the merge queue` is a **warning, not a failure**. A judge reading that `!` as an error retries a merge that already succeeded, or reports a merged PR as blocked.

**On the "10 required contexts":** confirmed against `longeye-ai/monorepo`'s rulesets — there are 11 required contexts, and the 11th is `Merge Queue Result`, which is reported from *inside* the queue and cannot be green beforehand. So 10 is exactly right pre-enqueue, and `SKILL.md` now says why rather than just quoting the number.

`gh pr merge` deliberately **stays off** `config/safe_wedges.txt`. That rule is about a daemon blind-approving a permission prompt; a judge merging after three explicit checks is a different act. Both prompts now say so, so the two rules don't read as contradictory.

**Daywatch does not merge** — it runs while a human is around, so it reports a loop-perfect PR and stops. This was the one thing your brief didn't specify and I asked about rather than guessing; both prompts still carry the full qualifying rule, because a shift that never reads the policy can't apply it when the shift flips.

## 3. "There is no babysitter daemon and no respawner. Nothing will restart you."

False, and load-bearing in the worst way: it tells the judge its handoff is pointless, which is precisely the inference that makes a session push through its ceiling.

What's actually true, verified on this box:

- `com.fleet-babysitter` — LaunchAgent, KeepAlive, **currently running**, beating every ~2 min
- `com.fleet-babysitter-watchdog` — `StartInterval` 300, kickstarts the daemon when its log goes stale past 30 min of *awake* time
- Both plists are dated **2026-06-23** — a month *before* the 2026-07-25 edit that declared "none of it ever existed"

That daemon auto-approves a narrow read-only wedge allowlist across worker panes and escalates everything else. It does **not** restart the desk session, and it never merges. What continues a shift is `handoff.py --act`, and it works.

The correction is scoped rather than flipped: `SKILL.md` now says **this skill ships no daemon** — a fact about the package, true wherever it installs — instead of a claim about the world. `tick.py`'s `daemon_health()` stays a stub for the same reason, with a docstring that no longer asserts the stronger claim.

## Tests

Updated with the policy, not deleted. The merge invariant didn't disappear, it narrowed — from "never merge" to three separately-pinned limits, so dropping any one clause reds the suite:

- author scope (`` `zionts` ``), completeness bar (`loop-perfect`), and the `--admin` prohibition, each asserted independently
- new: every prompt must disambiguate the merge-queue warning
- new: the ceiling agrees between `DEFAULT_THRESHOLD`, `SKILL.md`, and both prompts, and no prompt still quotes 200k as current
- new: no prompt may tell the judge nothing can succeed it (`"nothing will restart you"`, `"no respawner"`, …)
- `babysitterDaemonClaimIsScoped` now pins the middle ground and fails on *either* overshoot — a shipped-daemon claim or a universally-absent one

The phantom-path list (`daemon.py`, `~/.fleet`, …) survives on a corrected rationale, documented in the test: those paths are real, but this skill doesn't install them, so naming one in a prompt that ships everywhere points most readers at nothing.

## Verification

Everything below was actually run on this machine.

- `swift build --product TBDDaemon` — clean
- `swift test --filter Nightwatch` — **36 tests, 8 suites, all passed**
- `swift test --filter PluginDirWriter` — **13 tests passed**, including the `handoff.py` / `wake.py` / `tick.py` `--selftest` executions
- `swiftlint --strict` — clean
- All four embedded Python scripts extracted and `py_compile`d; `handoff.py --selftest` re-run directly against the new ceiling (`650,000 / ceiling 600,000 — OVER`)

**Correcting my own brief:** I was told `swift test` and `swiftlint` can't run here (Command Line Tools only, no full Xcode). That's true of the default toolchain, but not of this machine — `~/Library/Developer/Toolchains/swift-latest.xctoolchain` (Swift 6.3.2) has both, and with `TOOLCHAIN_DIR` set plus `Tests/TBDAppTests/ArchiveTombstoneTests.swift` (the one remaining `import XCTest` file) moved aside, the suite runs. Worth knowing, since it's the difference between shipping this on inspection and shipping it verified. It also caught two real defects I'd otherwise have handed to CI: a `//` comment I'd written inside the embedded Python, and a `String` concatenation where `#expect` wants a `Comment`.

`swift test` was **not** run across the full ~4500-test population — only the two filters above. The rest is CI's.

## Reported, not fixed: `/closeout` and `/anything-left`

I was asked to investigate rather than invent, and the premise turned out to be wrong in an interesting way.

**They exist.** Not as slash commands — as **user-invocable skills**, committed to `longeye-ai/monorepo` on `origin/main`:

```
.claude/skills/closeout/SKILL.md          + references/, scripts/
.claude/skills/anything-left/SKILL.md     + references/, scripts/
```

Both carry `user-invocable: true` and explicit `patterns:` entries (`/closeout`, `/anything-left`, plus natural-language triggers like *"am I done"*, *"is it safe to delete this worktree"*). Claude Code surfaces user-invocable skills as `/<name>`, so the invocation in the prompts is well-formed.

**They were never commands.** `git log --all --diff-filter=ADR -- '.claude/commands/*closeout*' '.claude/commands/*anything*'` returns nothing — there's no rename to find. The search that came up empty looked in `.claude/commands/`, which is a real directory in that repo (~60 commands), just never their home. So: not renamed, not unshipped — looked for in the wrong place.

**The one real caveat:** they're scoped to `longeye-ai/monorepo`. A judge sending `/closeout` to a session in a *different* repo's worktree — this one, for instance — hits nothing, and an unknown slash command is dropped without creating a user message, so it fails silently. That's a genuine footgun, but it's a different bug from "the command doesn't exist", and fixing it (repo-aware dispatch, or a skill-presence check before sending) is out of scope here. Filing rather than fixing.

None of this changes the standing rule: the prompts still forbid triggering `/closeout`, and this PR doesn't touch that.

## Notes

- No feature flag or migration: this is prose in a shipped skill, no `config` column, no load-bearing path replaced. No brainstorm spec per `CLAUDE.md`'s blast-radius rule — the policy change was decided by Adam, not originated by an agent.
- `wake.py`'s `STANDING_RULE` (sent to *worker* sessions) still forbids self-merging; only its stale justification changed, from "the human merges" to "the nightwatch judge or the human does".

### DIFF
--- a/Sources/TBDDaemon/Nightwatch/DeskSessionManager.swift
+++ b/Sources/TBDDaemon/Nightwatch/DeskSessionManager.swift
@@ -26,6 +26,11 @@ public actor DeskSessionManager: DeskSessionManaging {
     /// immediately (matching RPCRouter+ScratchHandlers) instead of waiting for a poll.
     private let subscriptions: StateSubscriptionManager?
     private let skillDir: String
+    /// Date seam for the nudge-overlap guard — a compared timestamp, so per
+    /// `Tests/CLAUDE.md` this is the `now:` shape, not an injected `Clock`.
+    /// Without it the 10-minute window is unreachable in a test, which is why the
+    /// guard went so long with no observable beyond "doesn't throw".
+    private let now: @Sendable () -> Date
 
     // MARK: - State
 
@@ -65,6 +70,20 @@ public actor DeskSessionManager: DeskSessionManaging {
     /// M2: nudge-overlap guard. Phase B upgrade: claim-file (queue/claims) single-driver enforcement.
     private var lastNudgeTime: Date?
 
+    /// Mode carried by the last nudge that actually reached the session.
+    ///
+    /// The judge is told to read `JUDGE-INSTRUCTIONS.md` once rather than every
+    /// tick, so *someone* has to notice when the body it memorized stops matching
+    /// the body on disk. The desk is deliberately reused across daywatch ↔
+    /// nightwatch without respawning (see `ensureDeskSession`), and the two bodies
+    /// differ on exactly the thing that matters — whether an unattended
+    /// `gh pr merge` is authorized. `nil` means nothing has been nudged yet, which
+    /// is treated the same as changed: a judge with no reads is a judge that must read.
+    ///
+    /// Internal rather than private so tests can assert the transition; it is state
+    /// the manager needs regardless.
+    private(set) var lastNudgedMode: NightwatchMode?
+
     // MARK: - Init
 
     public init(
@@ -72,13 +91,15 @@ public actor DeskSessionManager: DeskSessionManaging {
         lifecycle: WorktreeLifecycle,
         tmux: TmuxManager,
         skillDir: String,
-        subscriptions: StateSubscriptionManager? = nil
+        subscriptions: StateSubscriptionManager? = nil,
+        now: @escaping @Sendable () -> Date = { Date() }
     ) {
         self.db = db
         self.lifecycle = lifecycle
         self.tmux = tmux
         self.subscriptions = subscriptions
         self.skillDir = skillDir
+        self.now = now
         self.deskWorktreeID = nil
     }
 
@@ -101,8 +122,11 @@ public actor DeskSessionManager: DeskSessionManaging {
             if terminals.first(where: { $0.label == TerminalLabel.claudeCode }) != nil {
                 // Fast path: cached desk is alive and valid.
                 // Mode switches (daywatch ↔ nightwatch) intentionally REUSE the existing desk and terminal.
-                // The desk is NOT respawned on mode switch because the per-tick judgePrompt (via nudgeDeskSession)
-                // carries the current mode/act flag each cycle. Initial frame is one-time; steady-state mode is driven per-tick.
+                // The desk is NOT respawned on mode switch because every tick re-derives the mode and
+                // rewrites JUDGE-INSTRUCTIONS.md to match. Since the judge is told to read that file once
+                // rather than per tick, reuse-without-respawn is exactly the case where a memorized copy
+                // can go stale — `nudgeDeskSession` compares against `lastNudgedMode` and flags the flip.
+                // Initial frame is one-time; steady-state mode is driven per-tick.
                 return existing
             }
         }
@@ -257,7 +281,7 @@ public actor DeskSessionManager: DeskSessionManaging {
         defer { gateRelease() }
         // M2: Skip nudge if previous nudge was < 10 min ago (prevent overlapping judge runs)
         if let last = lastNudgeTime {
-            let elapsed = Date().timeIntervalSince(last)
+            let elapsed = now().timeIntervalSince(last)
             if elapsed < 10 * 60 {
                 logger.debug("Skipping nudge: last nudge was \(Int(elapsed))s ago (< 10 min)")
                 return
@@ -290,7 +314,16 @@ public actor DeskSessionManager: DeskSessionManaging {
             // nothing is a silent one.
             let prompt: String
             if let instructionsPath = writeJudgeInstructions(deskPath: worktree.path, mode: mode) {
-                prompt = NightwatchDeskPrompts.judgeNudge(mode: mode, instructionsPath: instructionsPath)
+                // A mode flip rewrites the file under a judge that was told to read
+                // it once. Nothing on the judge's side can see that happen, so the
+                // daemon — the only party that knows both the old and new mode —
+                // has to say so. First nudge (nil) counts as changed: a judge that
+                // has read nothing must read.
+                prompt = NightwatchDeskPrompts.judgeNudge(
+                    mode: mode,
+                    instructionsPath: instructionsPath,
+                    instructionsChanged: lastNudgedMode != mode
+                )
             } else {
                 logger.warning("Could not write judge instructions to \(worktree.path, privacy: .public); falling back to the inline prompt")
                 prompt = NightwatchDeskPrompts.judgePrompt(mode: mode, skillDir: skillDir)
@@ -310,8 +343,12 @@ public actor DeskSessionManager: DeskSessionManaging {
                 key: "Enter"
             )
 
-            // Record nudge time for overlap guard
-            lastNudgeTime = Date()
+            // Record nudge time for overlap guard. `lastNudgedMode` is set only
+            // here, after the paste actually went out — a nudge that failed to
+            // reach the session must not count as "the judge has seen this mode",
+            // or the next tick would tell it nothing changed when everything did.
+            lastNudgeTime = now()
+            lastNudgedMode = mode
 
             logger.debug("Nudged Watch Desk session: act=\(act, privacy: .public)")
         } catch {
--- a/Sources/TBDDaemon/Nightwatch/DeskSessionManager.swift
+++ b/Sources/TBDDaemon/Nightwatch/DeskSessionManager.swift
@@ -265,7 +265,6 @@ public actor DeskSessionManager: DeskSessionManaging {
         }
 
         let mode: NightwatchMode = act ? .nightwatch : .daywatch
-        let prompt = NightwatchDeskPrompts.judgePrompt(mode: mode, skillDir: skillDir)
 
         do {
             let terminals = try await db.terminals.list(worktreeID: worktreeID)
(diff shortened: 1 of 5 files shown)


zionts:
## Added: the per-tick judge nudge is now a file pointer (6695ac4)

Folded in per a follow-up requirement from Adam, since this PR is still open.

**The problem.** Every ~15 minutes `nudgeDeskSession` pasted the entire `judgePrompt` into the desk session — byte-identical to the previous one except for the mode/act flag. The judge reached its handoff ceiling *faster the more reliably it was nudged*, which is the wrong direction for a mechanism whose only job is keeping the desk alive. It also partly defeats this PR's other half: raising the ceiling to 600k buys headroom that the heartbeat was spending.

**Measured, not estimated:**

| | before | after |
|---|---|---|
| nightwatch nudge | 5238 B | **244 B** |
| daywatch nudge | 5278 B | **242 B** |

96% smaller. The body moves to `JUDGE-INSTRUCTIONS.md` in the desk worktree; the nudge carries only what actually changes:

```
🌙 Judge tick — mode: nightwatch, act=true. Run the cycle per /…/watch-desk/JUDGE-INSTRUCTIONS.md.
Already read that file this session? Don't re-read it — this line carries the only thing that changes between ticks.
```

**Three decisions worth reviewing rather than skimming:**

1. **Rewritten every nudge, not once at spawn.** It's mode-specific and a local write is free — what was expensive was putting it *in the session*, not on disk. Writing every tick also makes it self-healing across daemon restarts, mode switches, and a desk directory cleaned out underneath it, which "write once at spawn" would not be.

2. **The nudge tells the judge not to re-read.** Without that line the paste cost simply becomes a `Read` cost and the change saves nothing. The saving only materializes if the file is read once per session rather than once per tick, so the instruction is part of the mechanism, not politeness. The initial prompt now explains the protocol too, so a judge understands the pointer when its first tick arrives.

3. **Fallback to the inline prompt when the write fails.** A judge holding stale instructions is a bad night; a judge holding a pointer to nothing is a *silent* one. This is also why I did not add a feature flag despite `CLAUDE.md`'s blast-radius rule — the failure mode degrades to today's behavior rather than to no behavior, so there is no "off" state left to soak. Flagging that explicitly per the "if a trigger fires but a brainstorm is genuinely unnecessary, say so" clause; happy to add one if you'd rather soak it.

Secondary motivation, from the nightwatch READ-IN: a large paste can land as `[Pasted text #N]` and sit in the composer unsubmitted, which on an unattended desk is indistinguishable from never having been nudged. A one-line nudge is deliverable; a 5 KB one is a gamble.

**Test note.** The nudge is deliberately *excluded* from the `prompts()` fixture that asserts the policy invariants (forbid `/closeout`, state the merge limits, name the relay) — restating those every tick is the exact cost this removes. It's asserted instead on being self-sufficient for one tick: mode, act flag, absolute path, under 400 bytes.

I also strengthened the overlap-guard test while I was in there. It previously asserted only "doesn't throw", which would pass just as well if the guard were deleted; the instructions write gives it a real observable — a skipped nudge must leave a deleted file deleted.

**Verification:** `swift test --filter "Nightwatch|DeskSessionManager|PluginDirWriter|Daywatch"` → **79 tests, 11 suites, all passed**. `swift build --product TBDDaemon` clean, `swiftlint --strict` clean. Full population still CI's.

cheapsteak (approved):

zionts:
## Addressed the Reject (9e104bf)

The finding was right, and it's a pointed one: **a PR whose entire subject is stale facts baked into these prompts had baked in a fourth.**

`6695ac4` told the judge *"later ticks change only the mode/act line"* and never to re-read `JUDGE-INSTRUCTIONS.md`. That's false. The desk is reused across daywatch ↔ nightwatch without respawning, and the two bodies differ on exactly what this PR is about — whether an unattended `gh pr merge` is authorized.

**Fix:** the daemon now tracks `lastNudgedMode` and sets an explicit re-read signal on the tick where it flips. This is the reviewer's own suggested direction, and it's right for a reason worth stating: nothing on the judge's side can observe the file being rewritten underneath it, so the daemon — the only party that knows both the old and new mode — has to be the one to say so. `lastNudgedMode` is assigned only *after* the paste goes out, so a nudge the overlap guard drops never counts as "the judge has seen this mode".

### One correction to the finding, narrowing but not dismissing it

The review says the tri-clause limits "live only in the file and the one-time initial prompt, neither guaranteed fresh after a flip." I checked both compiled bodies:

| | `zionts` | `loop-perfect` | ``Never `--admin` `` | ``you MAY run `gh pr merge`` |
|---|---|---|---|---|
| daywatch body | ✅ | ✅ | ✅ | ❌ |
| nightwatch body | ✅ | ✅ | ✅ | ✅ |

All three limits are in **both** bodies — that was deliberate, and `mergeRule(mode:)`'s doc comment says why. Only the authorization sentence differs, and it's explicitly conditioned (*"On a nightwatch shift you MAY…"*), so a memorized nightwatch copy read under a fresh `mode: daywatch, act=false` line is self-limiting.

So the exploitability was lower than stated. **The two things the review got exactly right are the ones that matter**: the instruction was false, and no test could reach the transition. I'm not going to argue a safety finding down to nothing on a margin I only measured after being told to look.

### Tests

- Both branches of the re-read signal (changed → `RE-READ`; steady → `Don't re-read`, and neither leaks into the other). This is a mode-switch gate, so `CLAUDE.md`'s per-branch rule applies.
- A **three-tick sequence that flips mode mid-session** — the case no previous test could reach, since every existing one nudged exactly once. A single-tick test cannot observe a transition by construction, which is precisely why the gap survived.
- That required a date seam on the overlap guard (`now: @Sendable () -> Date`, the shape `Tests/CLAUDE.md` prescribes for compared timestamps rather than an injected `Clock`). Side benefit: the 10-minute guard now has its first real observable — it had been asserted only as "doesn't throw", which passes equally well if the guard is deleted.
- **Mutation-checked**, since a new test asserting a new field is easy to write vacuously: deleting `lastNudgedMode = mode` reds both new tests with 5 issues. They're real.

Also updated the `ensureDeskSession` docstring the review flagged — it still claimed the per-tick prompt carries the mode, true before `6695ac4` and not after.

### Verification

`swift build --product TBDDaemon` clean · `swift test --filter "Nightwatch|DeskSessionManager|PluginDirWriter|Daywatch"` → **82 tests, 11 suites, all passed** · `swiftlint --strict` clean. Full population is CI's.

### Note on the diagnostics block

The review's own diagnostics are worth keeping: `git diff origin/main...HEAD` failed with *"no merge base"* on a shallow clone and it worked around this by diffing against `d54a1bc`, reproducing the PR's stat exactly. That's the same class of problem `d54a1bc` ("restore the reviewer's merge-base on fork PRs") was meant to fix — this is a same-repo branch, so it may be a second, separate merge-base gap. Not touching it here; flagging it as possibly worth its own issue.

zionts:
## Addressed the second Reject (d4e589a)

Correct finding, and a genuinely different one from the first — worth naming precisely, because the pair describes a shape rather than two accidents:

- Round 1: **did the instructions change under the reader?** → fixed with `lastNudgedMode`.
- Round 2: **did the reader change under the instructions?** → missed entirely.

`lastNudgedMode` is keyed on the *mode*, not on which session is running. A crash-respawn with the mode unchanged leaves it matching, so `instructionsChanged` computes `false` and a brand-new judge is told *"already read JUDGE-INSTRUCTIONS.md this session? Don't re-read it."* It hasn't. It then runs a shift without the operational procedure — poll the queue, apply policy per clearance kind, write `acted.jsonl`, re-read live PR state before any assertive send — which lives only in that file.

The Medium/High reasoning is right too: `initialPrompt` inlines `mergeRule(mode:)` directly, so the merge limits still reach the new session. What's lost is procedure, not authorization.

**Fix is yours:** reset in `spawnDeskTerminal`, which — as the review notes — is the single chokepoint both spawn paths pass through.

Two additions on top of the literal suggestion:

1. **`lastNudgeTime` is reset with it.** Same defect wearing a different hat: a rate-limit window opened by a session that no longer exists shouldn't silence the first tick of its replacement. In the untreated case a respawned judge sits idle for up to 10 minutes on a dead session's timer. Both fields were per-session facts living on the actor; that's the actual bug, and fixing only the one that was reported would have left the other for round three.
2. **Reset unconditionally, before the spawn can throw.** The error is asymmetric — over-signalling costs one file `Read`, under-signalling costs an unattended shift run on instructions never opened.

**Test:** the new one respawns and *then* nudges. The pre-existing respawn test never nudged afterward, which is precisely why this was invisible to it — the review said so, and it's the same structural blind spot as round 1's single-tick tests. Mutation-checked: dropping the reset reds it with the exact `lastNudgedMode == nil` expectation.

`swift build` clean · `swift test --filter "Nightwatch|DeskSessionManager|PluginDirWriter|Daywatch"` → **83 tests, 11 suites, all passed** · `swiftlint --strict` clean.

---

Noting for the record, since two rounds is a pattern worth reading: both findings were in `6695ac4` — the nudge optimization — and neither was in the original three-stale-facts work. The optimization introduced a piece of *state* that has to stay in sync with a separate agent's memory, and every bug so far has been a way for those two to drift. If a third round finds another, I'd take that as a signal the design is wrong rather than the implementation incomplete, and split the perf commit out of this PR so the stale-facts fix — which is what's actually blocking the running desk — can land on its own.

PR #553, opened 2026-07-30
docs: what to offload to the agent box, and the memory case for doing it

zionts:
## What this is

A doc under `docs/` recording **what to offload to the agent box and why** — the split, the candidate workloads with verdicts, a measured baseline, and the failure modes. It sits next to `docs/remote-provider-contract.md`, which is the mechanism this rides on.

Docs-only. No code, no flag, no migration.

## Why now

The motivation turned out not to be the one I expected. Measured with the fleet up:

- **swap: 11.58 GB used of 12.29 GB — 707 MB free**, on 24 GB of RAM
- 38 resident `claude` processes, ~5.5–5.8 GB, mean ~150 MB each

So this is a memory decision, not a latency one. An idle session costs the same ~150 MB as a busy one and produces nothing — which is what actually justifies "move the idle, keep the busy". A remote session has no local agent process at all, so the saving per moved session is the whole ~150 MB, not a fraction.

That baseline is perishable; it describes one night's fleet, and the doc says so.

## What I did beyond writing it down

The provider was dry-run end to end against the rebuilt box under `env -i` — the case that matters, since TBD's daemon is GUI-launched and inherits almost no environment. `describe`, `list`, and `create` behave per contract. Three things fell out that are worth having on the record:

- **`stop` violates the contract two ways** (filed as monorepo #18142). Stopping an already-dead session exits **1** with an error envelope instead of exiting 0 with a terminal Session; stopping an *unknown* id **creates a real session directory and `meta.json`** on the box, squatting the slug. For TBD the first is worse — clicking stop on a visibly-dead session yields a hard permanent-failure error.
- **Session grouping will probably look broken here.** The box reports `meta.repo` as `longeye-ai/longeye-app`; this laptop's monorepo origin is `longeye-ai/monorepo`. TBD matches those to group remote sessions, so they should render ungrouped. Cosmetic, but better known before a restart than after.
- The one-repo allow-list is **not** the limitation it appears to be: 41 of 46 active worktrees are monorepo-based, ~89% of the fleet. The 4 TBD-repo worktrees genuinely cannot be offloaded.

## What is still blocked

Neither is in this PR's reach, and I did not work around either:

1. **No session can be created yet.** `agent-box-zionts-role` is missing the `invoke-agent-box-github-minter` policy the working box has, so `git clone` fails. The fix is authored on another branch in the monorepo; after merge it needs `atlantis apply -p management-agent-box-github-minter`. That is a management-account IAM change and is Adam's call.
2. **Remote sessions are invisible in TBD.** The running daemon binary predates the feature entirely — no `config.setRemoteBackends`, no `remote.providers`, no `RemoteProviderManager`, while known methods are present. The Settings toggle cannot even be set until a rebuild + restart, which suspends ~52 live sessions and moves the machine-wide `tbd://` handler. Held for Adam.

## Verification

Docs-only, so `swift build` / `swift test` aren't meaningful gates here — no sources changed. Every number and behavior in the doc came from a live command against the rebuilt box (`i-0251eb99f32df1b51`) or the local process table, not from reading code.

One note on the requested exit condition: `scripts/dev/pr_converged.py` and `review-ledger-v1` are monorepo tooling and don't apply to a PR in this repo — TBD's gate is the `claude-review` required check.


zionts:
Scope note, so the second file isn't a puzzle: this PR shows `docs/tmux-integration.md +21` alongside my doc. That's `68f84a62` — Adam's own commit from Jul 28 ("docs: known issue — terminal send silently no-ops on dead panes"), already merged into the fork's `main` but not yet upstream, so a PR based on that main necessarily carries it. I neither authored nor staged it; my commit is `d8fdb5b` and touches only `docs/agent-box-offload.md`. It'll drop out of the diff once that commit reaches upstream main.

docs/agent-box-offload.md
```diff
@@ -0,0 +1,166 @@
...
+free swap, is the difference between thrashing and not.
+
+## The split
+
+Decided already, recorded here so it is reviewable:
+
+- **Move:** idle, patient, API-bound work — long-poll watching, PR triage,
+  review-thread replies, docs and knowledge harvesting.
+- **Keep local:** anything interactive, anything on the local toolchain
+  (`just dev`, Storybook, a browser, the local DB), and desk/CI work.
```
cheapsteak, line 52: I don't agree with this split

potentially context you might be missing: our agent box already forwards dev server ports, it runs the full local dev environment; with postgres 18 set up with `file_copy_method=clone ` so branched dbs are copy-on-write, you can browse the app that the agent box's dev server spun up from your browser (run an Explore on that area if you haven't already)

the goal is to be able to offload everything
_especially_ things that need local dev servers (the stuff that uses a lot of memory and disk)

I think it's mostly there now, just the connection/interfacing part still feels incredibly clunky (but it works!)

zionts:
Updated in `3af6750` after six hours of reality moving underneath this doc.

**The box works now.** The IAM grant landed (monorepo #18135 + the Atlantis apply), so "the box cannot clone" is no longer true — sessions clone the 4.7 GB monorepo and run a live TUI, with `agent_state` arriving from lifecycle hooks (`working` / `user_prompt`), not screen scraping. What still gates the *offload* is TBD visibility, which is the held restart.

**I walked back an over-claim.** I'd cited `claude -p` → `AUTHOK` as proof the setup token is healthy and the 403 benign. That probe runs **headless**, skipping the first-run onboarding interactive sessions must pass — so it exercises a different code path from the thing it certifies. That's how a box whose interactive sessions couldn't start still answered AUTHOK. A session has since hit a genuine usage limit, so the 403's cause is **open**, not settled. Corrected here and on #18126.

**Two findings that change the recommendation**, both verified directly:
- The box has **no `gh`** (`command -v gh` → absent). That's precisely what candidates #1 and #2 — the two best-shaped ones — need. Reclassified as move-*after*-installing-gh.
- `agentbox new --prompt` doesn't deliver the prompt (#18202); the session comes up ready and empty. Work moved that way would sit idle looking healthy — the exact invisible-wedge shape this doc argues against.

Added AWS SSO expiry as its own failure mode: boto3 can't refresh an expired SSO token but the `aws` CLI can, so it's recoverable without a browser. Observed live — a `create` failed `auth_expired` and the next call succeeded after one `aws` invocation.

Note the checks will re-run; the previous approval was against `d8fdb5b`.

docs/agent-box-offload.md
```diff
@@ -0,0 +1,177 @@
...
+That sounds restrictive but 41 of 46 active worktrees are monorepo-based, so it
+covers ~89% of the fleet. (`longeye-app` is a redirect alias for the renamed
+`longeye-ai/monorepo` — confirmed via `gh repo view longeye-ai/longeye-app`,
+which resolves to `monorepo`. The minter's tokens are scoped to `monorepo`, so
+the alias and the token scope agree.)
+
+The 4 TBD-repo worktrees **cannot** be offloaded at all: TBD's own repo is not
+an allow-listed shorthand, and adding one is a provider-side change.
+
+## Candidate workloads
```
cheapsteak, line 73: this list feels very specific to your current workflow this week; i don't think this should be checked in to TBD

cheapsteak (changes requested):
I don't think this should be checked into TBD


> Remote sessions are invisible in TBD. [..] until a rebuild + restart, which suspends ~52 live sessions

fwiw i have remote sessions running and showing up in worktree list; did need to rebuild + restart but didn't suspend live sessions

> this PR shows docs/tmux-integration.md +21 alongside my doc. [..] already merged into the fork's main but not yet upstream

that tmux integration change merged in https://github.com/cheapsteak/tbd/pull/550, weird that it still shows up

docs/agent-box-offload.md
```diff
@@ -0,0 +1,177 @@
...
+Two constraints stand, and neither is negotiable per session:
+
+- **No production CJI on the box, ever.**
+- **No untrusted-content ingestion while the plaintext token is in play.** The
+  Claude setup token is injected as `CLAUDE_CODE_OAUTH_TOKEN` into every
+  per-repo tmux server, so any child process an agent spawns can read it.
+
+### One structural limit that is not a problem
+
+The provider's `create` allow-lists exactly one repo shorthand, `longeye-app`.
```
cheapsteak, line 63: I don't think that allow-list exists in this repo
Did you mean to add this doc to the other repo?

zionts:
Closing — you're right on all three counts, and the third one answers the first two: this is a doc about an external tool, so it doesn't belong in TBD. I've taken it out of this repo rather than iterate on it here.

**On the split — I was wrong, and not for a subtle reason.** `agentbox forward` was in the CLI help I read at the very start of this and I still wrote "needs a browser / local toolchain → keep local." Knowing the box forwards dev-server ports and runs full local dev with `file_copy_method=clone` branched Postgres inverts the recommendation: the memory-heavy local-dev sessions I marked keep-local are precisely the ones most worth moving, since memory is the entire reason to offload. The draft's premise was backwards.

**On the restart claim** — also wrong, and my own run confirmed your version: rebuilding and restarting the daemon suspended nothing. Forty-one claude processes and all eight tmux servers came through it.

**One thing I want to flag rather than quietly drop.** The candidate table named specific law-enforcement casework — departments and case identifiers — while arguing that exact data class can never move to the box. Naming it in a shared tool's git history was a bad call regardless of which repo it landed in; "law-enforcement casework" made the identical point. I've stripped the identifiers and am removing the commits from this branch, but a force-push doesn't guarantee GitHub drops the old objects — the SHAs can stay fetchable. Flagging so @zionts can decide whether that needs anything further.

The TBD-specific findings that came out of this are worth keeping somewhere, so I'll re-home them: the enable path (`restart.sh` silently skips the daemon restart on a WIP branch, leaving a new app against an old daemon), and the confirmed `RemoteRepoMatching` gap — every remote session renders ungrouped because the box reports `longeye-app` against a local origin of `monorepo`. That second one is a real TBD bug and I'll raise it on its own rather than buried in a doc.

cheapsteak:
This should hopefully help prevent future occurrences 

https://github.com/cheapsteak/tbd/pull/554

PR #555, opened 2026-07-30
fix: group remote sessions whose repo was renamed on the host

zionts:
## The bug

Every remote session in my sidebar rendered ungrouped, permanently.

`resolveRepoID` matches a provider's `meta.repo` against each local repo's `remoteURL` as an exact `org/name` key. But a provider reports whatever name it *clones by*, and GitHub keeps serving a renamed repo's old name via a redirect indefinitely. So the provider goes on saying `acme/old-name` long after the repo became `acme/api`, the keys never agree, and the sessions have no repo forever.

Nothing about this is specific to one provider — any provider that cloned before a rename reports the stale name, and the redirect means it keeps working, so there is no error to notice.

## The fix

Resolving the redirect would put a network call inside a pure matching function. Instead this uses a signal already on disk: **a rename on the host does not rename anyone's local clone directory**, so the directory usually still carries exactly the name the provider is reporting.

Deliberately narrow, since it is the one heuristic in a type whose doc comment opens with "deterministic (not heuristic)":

- **Second pass only** — an exact `remoteURL` match can never lose to it. That is why it is two loops rather than one, and there is a test that fails on a single-pass implementation (`exactRemoteMatchBeatsRenameFallback` orders the fallback candidate first on purpose).
- **Org must still agree**, so it can only pair repos already known to live under the same owner. It cannot invent a cross-org match.
- **Only the directory's own name** is compared, never a path prefix.

## The one judgement call

It takes the **first** match rather than treating duplicates as ambiguous.

I wrote it the strict way first and it left my own case unfixed: one repo registered twice — a primary clone plus a worktree root — is ordinary, and it is exactly the shape this pass sees, because *both* clones keep the old directory name and both therefore match. Refusing to choose would leave sessions ungrouped precisely for the people who hit a rename. Pass 1 already resolves duplicate remotes by taking the first, so this matches the behaviour its docstring already promises rather than inventing stricter semantics one pass down.

## Verification

Five tests, covering both branches of the new conditional and each narrowing constraint.

**Local `swift test` cannot run here** — neither the default 6.2.1 toolchain nor the installed 6.3.2 one has swift-testing (`no such module 'Testing'`), so CI is the first real execution of these. Flagging that rather than implying I ran them. `swift build --product TBDDaemon` is clean.

What I could verify is stronger than the unit tests for this particular bug: I ran the real code path against real data. Four live remote sessions that all report the stale name went from `(ungrouped)` to resolving onto the correct local repo after restarting the daemon with this build.

```
before:  final-check  repoGroup=(ungrouped)      ... x4
after:   final-check  repoGroup=69E45D41-...     ... x4
```

### DIFF
--- a/Sources/TBDShared/RemoteRepoMatching.swift
+++ b/Sources/TBDShared/RemoteRepoMatching.swift
@@ -81,13 +81,71 @@ public enum RemoteRepoMatching {
     /// Resolves `metaRepo` (a provider's `meta["repo"]` value) against
     /// `repos` by comparing normalized keys. Returns the first matching
     /// repo's id in `repos` order, or nil when `metaRepo` is nil/blank,
-    /// unparseable, or no repo's `remoteURL` normalizes to the same key.
+    /// unparseable, or no repo matches.
+    ///
+    /// Two passes, in this order — an exact `remoteURL` match must never lose
+    /// to the rename fallback, which is why this cannot be a single loop:
+    ///
+    /// 1. Exact: the provider's `org/name` equals the repo's `remoteURL`
+    ///    `org/name`. This is the normal case and the only one before renames
+    ///    were considered.
+    /// 2. Renamed-repo fallback: the **org** still agrees, and the provider's
+    ///    `name` equals the repo's local directory name. See
+    ///    `renamedRepoFallbackID` for why that is a real signal rather than a
+    ///    guess.
     public static func resolveRepoID(metaRepo: String?, repos: [Repo]) -> UUID? {
-        guard let metaRepo, let targetKey = normalizedKey(metaRepo) else { return nil }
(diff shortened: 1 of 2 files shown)


cheapsteak (approved):

zionts:
Added the case-insensitivity test from the review — cheap, and directory casing is whatever the person who cloned it typed, so it's a real path rather than a theoretical one.

Leaving the second minor as-is, agreed as FYI: two unrelated repos under the same org whose clone directories happen to share a basename could pair wrongly. It needs someone to have cloned an unrelated repo into a directory named after a *different* repo's pre-rename name, under the same owner — and the cost is a mis-grouped sidebar row, not data loss. Tightening it would mean either a network call from a pure matching function or user-facing alias config, both of which cost more than the failure does.

PR #560, opened 2026-07-31
feat: add Codex to the worktree usage picker

zionts:
## Summary

- add a Codex account row to the sidebar worktree `+` picker so a single click creates that worktree with Codex
- fetch the ambient Codex identity and quota windows on picker open through the Codex app-server machine protocol, then show email, dynamic usage bars, reset times, and limit-reached status beside the Claude accounts
- add a backward-compatible create-time primary-agent override, including the detached pre-session path, while preserving the global default for plain/branch creation and Claude carryover precedence
- explicitly mark Claude profile/model picks as Claude, fixing the existing case where a global Codex default silently ignored the selected Claude profile

## Codex API experiment

Tested with `codex-cli 0.146.0-alpha.9.2`:

- generated the default app-server JSON schema and confirmed `account/read` and `account/rateLimits/read` are included without experimental schema generation
- completed a live stdio handshake (`initialize` → `initialized` → account/limits reads)
- observed account email/plan plus the base `codex` quota bucket and a named model-scoped bucket, with used percent, window duration, and reset timestamp

TBD never reads `~/.codex/auth.json` or scrapes the TUI. The fetch is a bounded, short-lived child process and resolves Codex from PATH plus common launchd-safe install locations such as `~/.local/bin` and Homebrew.

This stays user-gesture-driven and ephemeral: no background poller, persistence, migration, config flag, or design spec is needed for this additive UI path.

## Verification

- `swift build` ✅
- `swift build --target TBDSharedTests` ✅
- `swift build --target TBDDaemonTests` ✅
- live Codex app-server account/rate-limit probe ✅
- `git diff --check` ✅
- `swift test` ⚠️ blocked before test execution by the local toolchain lacking `XCTest` for the existing `Tests/TBDAppTests/ArchiveTombstoneTests.swift`
- `swiftlint --strict` ⚠️ blocked locally because SourceKitten could not load `sourcekitdInProc.framework`

### DIFF
--- a/Sources/TBDApp/AppState+CodexUsage.swift
+++ b/Sources/TBDApp/AppState+CodexUsage.swift
@@ -1,0 +1,14 @@
+import TBDShared
+
+extension AppState {
+    func loadCodexUsage() async {
+        guard !isLoadingCodexUsage else { return }
+        isLoadingCodexUsage = true
+        defer { isLoadingCodexUsage = false }
+        do {
+            codexUsage = try await daemonClient.fetchCodexUsage()
+        } catch {
+            codexUsage = CodexUsageResult(unavailableReason: "Usage unavailable")
+        }
+    }
+}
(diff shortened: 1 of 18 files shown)


cheapsteak (approved):
🙏 🙏

PR #561, opened 2026-07-31
feat: add Continue in Codex handoffs

zionts:
## Summary

- add a first-class `terminal.continueInCodex` RPC, Claude-tab context action, and `tbd terminal continue-in-codex` command
- preserve the source Claude terminal and create one deduplicated local Codex terminal in the same worktree
- generate immutable, owner-only, byte-bounded handoffs from stored/profile-aware Claude transcripts plus Git and TBD context
- preflight the absolute Codex executable and generated global profile before tmux, database, checkout, pre-session, or recreate mutations
- keep ordinary Codex creation in the add-tab menu and reuse the existing global `~/.codex` OAuth home without copying credentials

## Atomic launch and profile-swap finding

The app and CLI each issue one dedicated takeover RPC. The daemon passes the final `--profile tbd` selection and handoff prompt together in the single `tmux new-window` launch. A focused integration assertion proves there is no intermediate terminal, `respawn-window`, profile swap, or post-launch `send-keys`/paste operation.

Live prototypes showed why this matters: current Claude `terminal.swapProfile` is intentionally non-atomic across TBD's database and tmux. It interrupts best-effort, persists the target profile/session, then asks tmux to respawn the existing window. If the pane has already exited, the database may show the new profile while no live pane exists; immediate create-then-swap also races startup trust and hook-review gates.

Automation must therefore create fresh with the final profile and initial prompt in one request. For Claude creation, callers pass `overrideProfileID` and `prompt` together. `terminal.swapProfile` remains an explicit compatibility action for existing Claude sessions, not a takeover orchestration primitive, and its returned row is not proof of pane liveness.

## Safety and context contract

The initial prompt requires the worker to read the private handoff, applicable `AGENTS.md` and `CLAUDE.md`, and referenced skills/knowledge; identify claim-work and closeout obligations; verify live worktree state; and preserve unrelated changes.

Transcript text is untrusted evidence. The handoff warns that repository bootstrap may be partial, generated `.Codex/skills` rewrites may be broken, large skill registries may be context-truncated, and takeover never stages, repairs, resets, or cleans bootstrap artifacts.

Durable mappings fail closed when corrupt or inconsistent, and concurrent requests are keyed by source and target. Handoffs live outside repositories with `0700` directories and `0600` files.

## Launch reliability

Codex resolution order for spawning is `TBD_CODEX_EXECUTABLE`, absolute PATH entries, then the ChatGPT app-bundle binary. Missing or unsafe candidates fail before mutation, and every production spawn path receives the resolved absolute executable.

After merging current upstream/main, TBD's new Codex usage picker and this stricter spawn resolver were unified: usage display uses a nonthrowing lookup with common absolute fallback directories, while all spawning retains strict absolute-path validation and actionable errors. Per-create agent overrides now drive early Codex preflight instead of being masked by the global preference; both override directions are tested.

The generated TBD Codex profile persists a complete disabled shadcn transport:

```toml
[mcp_servers.shadcn]
command = "npx"
args = ["shadcn@latest", "mcp"]
enabled = false
```

An isolated regression invokes the installed Codex with a temporary `CODEX_HOME` and verifies `codex --profile tbd mcp list` parses the profile and does not list shadcn. Global authentication and the live `~/.codex` profile are not copied or changed.

## Native import and future adapters

OpenAI's [`/codex:transfer`](https://github.com/openai/codex-plugin-cc#codextransfer) now converts canonical Claude JSONL into a persistent resumable Codex thread through the app-server [`externalAgentConfig/import`](https://github.com/openai/codex/blob/main/codex-rs/app-server/README.md#detect-and-import-external-agent-config) flow. That can become an optional context-import adapter and may preserve more conversation history than the bounded excerpt.

It is not an MVP dependency: the wrapper currently accepts canonical files under `~/.claude/projects`, while TBD profiles may store transcripts elsewhere; support is version-gated and asynchronous; and import mutates persistent global Codex thread/import-ledger state. The deterministic handoff remains the no-Claude-credits, no-plugin fallback and carries Git status, recent commits, TBD notes, safety framing, and verification instructions that native session import does not. A future path can import the session, resume the returned thread, and still require it to read the deterministic handoff.

Remote adapters must separately report prompt delivery, explicit submit input, and machine-observed execution/readiness. Successful send is not proof of prompt execution. Potentially yielding mutators must retain and poll their direct process/session identifier, and Git mutations require branch, HEAD, status, and remote postconditions.

## UI

- Claude tab context menu: `Fork Session`, then `Continue in Codex`
- add-tab menu: ordinary `Codex` creation remains separately wired
- launch success is presented as launch, not prompt readiness

## Validation

Exact tested commit: `c2a7bfbf8b903dec008b6f70e301add1ec7cb1e4`

Passed locally:

- `swift build --target TBDDaemonTests`
- `swift build --target TBDApp`
- `swift build --target TBDCLI`
- `swift build --target TBDCLITests`
- `swift build` (passed on retry after a host code-signing transient)
- `git diff --check`
- isolated real-CLI profile parse: exit 0; disabled shadcn omitted

Passed on PR CI:

- full `test` job: 8m43s
- SwiftLint strict: 17s
- no-implementation-plan guard: 5s
- required Claude review: approved in 5m35s

A local filtered `swift test` attempt is blocked before execution because this host's Command Line Tools Swift cannot import the repository's pre-existing XCTest target. The complete hosted macOS suite is green on the exact commit above.

## Scope

This is an explicit user-gesture, additive, non-destructive action. It does not add a database migration, autonomous behavior, background timer, destructive sweep, or load-bearing path replacement, so no default-off flag was added. The design was constrained by the supplied parity report and live prototypes; this PR implements that bounded MVP without adding a separate brainstorming spec.

Current upstream/main is merged into this feature branch. No shared installed TBD app was restarted, and this PR must not be merged as part of validation.


cheapsteak:
Adam — thanks for the depth here. I went digging into this properly rather than just reviewing the diff, and I came out somewhere different, but the direction it points is one I'd like your read on, and ideally your hands on.

## The short version

Codex imports Claude sessions natively. I didn't just find this in docs — I drove the app-server and watched it work:

```jsonc
>> {"method":"externalAgentConfig/import", "params":{"migrationItems":[{
     "itemType":"SESSIONS",
     "details":{"sessions":[{"path":"~/.claude/projects/<slug>/<id>.jsonl","cwd":"<worktree>"}]}}]}}

<< {"result":{"importId":"2c288360-…"}}
<< {"method":"externalAgentConfig/import/completed","params":{"itemTypeResults":[
     {"itemType":"SESSIONS","successes":[{
       "source":"~/.claude/projects/<slug>/<id>.jsonl",
       "target":"019fb9d4-f45a-7061-8953-d377b660dba3"}]}]}}
```

`target` is a live, resumable Codex session. It landed at `~/.codex/sessions/…/rollout-<ts>-019fb9d4-….jsonl` with `session_meta.session_id` matching and the source `cwd` carried across. The whole round trip took about 60 ms, and Codex did the transcript conversion itself.

So the hard problem this PR solves — getting Claude's conversation into Codex — turns out not to be ours to solve.

## What that means for the PR

`DeterministicCodexHandoffGenerator` is careful code, but it's carrying weight it doesn't need to. It reads the last 64 KB of the session JSONL, keeps only `user`/`assistant` entries with non-empty text, renders ≤8 KB and caps at 16 KB. On a representative long session here — 2.7 MB, 1,655 lines — that window is **2.3%** of the conversation, taken from the tail, so it thins out precisely as sessions grow long. It also drops tool calls and results entirely.

The native import gives the whole conversation in order, every tool call with its arguments, and the head of every result. I measured that too — numbers in the details below.

And a nice consequence: because the imported thread already *contains* the conversation, there's no handoff document to point at and no initial prompt to send. The Codex tab opens with real history and an empty composer, spending no turn. Your "launch success ≠ prompt readiness" distinction is right, and this sidesteps the whole question — there's nothing to submit.

Two pieces of your work carry over intact and I wouldn't want them rewritten: the Codex executable resolution (`TBD_CODEX_EXECUTABLE` → absolute PATH entry → app-bundle binary, resolved before any mutation) and the generated-profile handling, including that disabled-but-parser-valid `shadcn` entry. That invariant was a genuinely non-obvious thing to find and pin with a real-CLI regression.

## The shape I think we want

- a thin daemon-side app-server client: `initialize` → one `externalAgentConfig/import` carrying a single `SESSIONS` item → read `target` off the `completed` notification
- no `detect` call — TBD already stores `transcriptPath` on the terminal row, so it can build the `SessionMigration` directly. This is the one genuinely TBD-shaped contribution: `detect` returns every recent Claude session with no way to know which you meant, and TBD knows the worktree and the tab.
- create a Codex terminal in the same worktree resumed on `target`
- context-menu item on Claude tabs, sibling to `Fork Session` (not nested — those submenus list profiles)

No feature flag, and I think your reasoning in this PR was right on that point: it's an explicit user gesture, additive, non-destructive, with no autonomy and no load-bearing path replaced, so it trips none of the flag triggers. No flag means no `config` column and no migration.

Deliberately **not** built: the handoff generator, any handoff file or store, a source→target mapping table, dedup keying, or an initial prompt. The protocol hands back the thread id, so no client-side ledger is needed.

## Two calls I've made, where this diverges from your PR

**Fork, not link — no dedup.** Your PR reuses the mapped live terminal on a retry. I'm going the other way: each invocation imports and opens a new Codex thread. It matches what the protocol does anyway, it matches `Fork Session`'s existing semantics, and it means TBD stores no mapping and needs no staleness policy. Since the source transcript keeps growing, a "reuse" would hand you a thread frozen at first-import time — helpful-looking and quietly stale. Two visible tabs are more legible than one silently out-of-date one, and wanting two is a real case when you're comparing approaches. The cost is that a mis-click leaves a spare session to close, which seems an acceptable price for holding no state.

**Two agents in one worktree — not policed.** The source Claude keeps running, so both share a working tree. TBD has never policed concurrent tabs and I don't want to start here; the user can tell either session about the other. This is also safer without an initial prompt: the Codex tab opens with history and an empty composer, so nothing instructs it to resume the plan, and the first user message sets direction. A submitted "verify state and continue" prompt is what would have two agents acting on the same plan at once.

If you think either call is wrong, that's exactly what I'd like to hear.

## The ask

Would you review this shape and tell me where you think it's wrong? You've spent more time in this problem than anyone, and if you think the native-import route misses something the generated-handoff route was solving, I'd rather hear it now.

And if you do think it's the right shape — would you like to build it? It's a much smaller change than what's in this PR, and the launch-resolution and profile work here is most of the fiddly part already done.

One process note for whoever picks it up: we've tightened the rule so anything that isn't a bug fix or a minor UI change wants `/tbd-brainstorming` and a committed spec in `docs/specs/` before implementation — not because of a migration (there isn't one), but because it's a new feature. That spec is already written and posted as a comment on #568, so it's there to work from rather than something you'd have to produce.

I'll leave this PR open until we've settled the shape.

<details>
<summary><b>Grounding: protocol contracts, fidelity measurements, and what was ruled out</b></summary>

### Reaching the importer

`codex app-server` (bare, no subcommand) speaks newline-delimited JSON-RPC on stdio. There's also a socket form, `codex app-server --listen unix://…`. The full protocol schema can be generated locally without running a server:

```sh
codex app-server generate-json-schema --out <dir>   # --out is required
```

Everything below was checked against codex-cli **0.145.0**.

### `detect` is a setup-migration API

`externalAgentConfig/detect` takes `cwds` (for repo-scoped detection) and `includeHome`. It returns migration items typed by an enum of ten: `AGENTS_MD, CONFIG, SKILLS, PLUGINS, MCP_SERVER_CONFIG, SUBAGENTS, HOOKS, COMMANDS, MEMORY, SESSIONS`.

This is worth knowing because it explains your parity-warnings section. A single `detect` on a dev machine proposed migrating `~/.claude/CLAUDE.md` → `~/.codex/AGENTS.md`, hooks into `~/.codex/hooks.json`, skills and commands into `~/.agents/skills`, MCP servers into `~/.codex/config.toml`, and enabled plugins. The `.Codex/skills` and `AGENTS.md` artifacts you warn about showing up as status dirt are the other nine item types doing their thing.

**This is also the main hazard for any caller.** `import` accepts whatever `migrationItems` you hand it, so a malformed array rewrites the user's Codex configuration. A caller must construct exactly one `SESSIONS` entry, deliberately. I did that for the test above.

Session detection is home-scoped even when `cwds` is passed — config items came back repo-scoped, but `SESSIONS` returned `cwd: null` covering all recent sessions from `~/.claude/projects`. Filtering to a worktree is client-side. Each entry is `{path, cwd, title}`, and the titles are human-readable summaries, so a picker can be built from `detect` alone.

`externalAgentConfig/import/readHistories` is an audit log of past import operations — `{importId, completedAtMs, successes[], failures[]}` — not a session catalogue.

`session_meta.originator` records the `clientInfo.name` sent at `initialize`, so imports are attributable to the tool that made them.

### What the conversion keeps and drops

Codex does not reproduce Claude's tool structure. Tool activity is transcoded to text inside assistant messages:

```text
[external_agent_tool_call: WebSearch]
input: {"query":"…"}
[/external_agent_tool_call]
[external_agent_tool_result]
…result text…
[/external_agent_tool_result]
```

No native `function_call` / `function_call_output` items appear, so a resumed agent reads tool history as prose.

Measured on a 156-line source carrying 20 `tool_use` and 20 `tool_result` blocks:

- all 20 tool calls survive, with tool name and full input JSON
- 17 result blocks survive, one empty — so 16 carry content, 3 lost outright
- results are capped at 4,000 chars each, truncation largely silent: **16,703 chars retained from 77,836**
- conversation prose survives intact

So ~79% of tool-result bulk is discarded, and what remains is the shape of the work rather than its output. For a handoff into the same worktree that seems like the right trade — a truncated file read is re-runnable, lost intent isn't. The conversion is literal enough to carry interface noise; a usage-limit notice from the source session was imported as an assistant turn.

### On the plugin route

`openai/codex-plugin-cc` **1.0.6** ships `/codex:transfer`, which wraps the same `externalAgentConfig/import` call. Worth stating the version — 1.0.4 has neither the command nor any `externalAgentConfig` client code, so looking on an older install turns up nothing.

I considered having TBD shell out to it and decided against:

- it recovers the thread id from a private ledger keyed by source path + SHA256, which is redundant once you read `target` off the notification
- reaching it without a Claude turn means invoking `scripts/codex-companion.mjs` directly and parsing human-formatted stdout — an undocumented internal path, and a weaker contract than the schema'd protocol
- `transfer.md` sets `disable-model-invocation: true`, which hides the command from the model but does **not** make it turn-free; the body pairs a `!` bash line with an instruction to present the output, so a turn is spent rendering the printed `codex resume <id>`. That makes it a poor fit when the source session is out of quota — which is one of the cases we care about most. (Reasoning from the command definition; I haven't observed it against an exhausted quota.)

### Still open

- **Idempotency** — re-importing the same session, and what Codex's `external_agent_session_imports.json` does once the source transcript has grown.
- Whether `migrationSource` selects agents other than Claude Code. The schema only documents that unrecognized values fall back to a default.
- `codex app-server` is marked `[experimental]`; none of these contracts are stable, and the plugin's client surface landed between 1.0.4 and 1.0.6. That's an argument for the default-off flag rather than against building.

The full writeup — protocol contracts, the executed import with its raw notifications, and the fidelity measurements — is in #568.

</details>

zionts:
The native-import successor you requested is now implemented in #572. It preserves the executable/profile work from this PR, removes generated handoffs/summarization/dedup/initial prompts, and uses exactly one SESSIONS item through `externalAgentConfig/import`. I’m keeping #561 open until #572 is reviewed so the proven code and discussion remain available; once #572 is accepted, #561 can close as superseded.

zionts:
Closed as superseded by merged #572.\n\nNative import replaced the bounded generated-handoff path. The executable/profile/preflight safety from this branch was carried forward into #572/current main. This #561 head (c2a7bfbf8b) is fully published; the local worktree is clean, tracks the exact remote head, and has zero unpublished commits or artifacts.\n\nKnown native-import fidelity and experimental-protocol limitations remain documented on main. A read-only comparison found no code cherry-pick or conflict repair warranted; rebasing this old tree would instead overwrite newer native importer, Provider Desk, Nightwatch lease, and build-admission work.\n\nNo branch or worktree was deleted or archived.

PR #562, opened 2026-07-31
fix: supervise Codex from Nightwatch

zionts:
## Summary

- make Watch Desk terminal discovery agent-kind-aware for Claude and Codex
- require both a live tmux window and a matching agent process, then retry a failed desk spawn before nudging
- include Codex terminals in Nightwatch fleet/PR routing using TBD hook-backed `activityState` instead of scraping the Codex TUI
- preserve Claude-only wake/handoff behavior where resume semantics are Claude-specific

## Dependency

Depends on #561 for robust absolute Codex executable preflight and the valid profile-scoped shadcn MCP disable override (`command` + `args` + `enabled = false`). This PR deliberately does not duplicate or edit those files, so the two changes can land independently without conflicts. Nightwatch becomes fully robust once both are merged.

## Validation

- `swift build --target TBDDaemonLib`
- extracted shipped `tick.py --selftest`: 14/14 agent-state scenarios passed
- `git diff --check`
- added regression coverage for Codex desk creation/nudge, stale shell-backed Codex rows, agent command matching, stale/live row selection, and hook-state fleet routing

The local Apple 6.2.1 toolchain cannot import the standard `Testing` module, so local `swift test` and the aggregate `swift build` stop while compiling existing test targets. Source compilation succeeds; CI is the authoritative full-suite run.

## Existing safety gate

This is a bug fix inside the existing Nightwatch mode, which already defaults to `off` and runs only after the operator enables daywatch/nightwatch. It does not add a new autonomous subsystem or broaden Nightwatch authorization; it makes the already-authorized desk select and recover the configured agent kind.

### DIFF
--- a/Sources/TBDDaemon/Nightwatch/DeskSessionManager.swift
+++ b/Sources/TBDDaemon/Nightwatch/DeskSessionManager.swift
@@ -107,19 +107,26 @@ public actor DeskSessionManager: DeskSessionManaging {
 
     /// Idempotent: ensure a Watch Desk scratch space exists.
     /// Returns existing desk worktree if already created, otherwise creates one.
-    /// On recovery, respawns a Claude terminal if one doesn't exist (H1 fix: desk dead after off→on cycle).
+    /// On recovery, respawns the configured primary agent if it has no live pane
+    /// (H1 fix: desk dead after off→on cycle).
     /// - Parameter mode: The current nightwatch mode (daywatch or nightwatch)
     /// - Returns: The Worktree for the desk session
     public func ensureDeskSession(mode: NightwatchMode) async throws -> Worktree {
         await gateAcquire()
         defer { gateRelease() }
-        // Validate cached desk session: must be active AND have a live Claude terminal
+        let preferredKind = try await preferredAgentKind()
+
+        // Validate cached desk session: the row alone is not enough. Agent
+        // processes can exit before their terminal row is removed, so require a
+        // live pane owned by the currently configured primary agent.
         if let cachedID = deskWorktreeID,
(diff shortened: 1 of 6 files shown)


cheapsteak (approved):

zionts:
Live Agent Box health input for Nightwatch (2026-07-31): EC2 `running`, SSM `Online`, a live tmux row, and `activityState` are each necessary but not sufficient readiness signals. After recovery, eight Codex workers were reachable yet their installed GitHub App helper had only `contents:read`: `gh pr view` could not authenticate and a same-branch `git push --dry-run` could not prove/write the branch.

For any Nightwatch handoff/recovery that expects a worker to land code, add or route through a capability preflight with two independent outcomes: (1) repository-scoped PR API read succeeds, and (2) a non-force branch-write probe succeeds for the worker-owned ref. On failure, preserve local commits, mark the worker blocked, and do not rebase/reset/clean/archive or infer success from transport health. Longeye ownership: [credential bootstrap #18393](https://github.com/longeye-ai/monorepo/issues/18393) and [SSM/recovery durability #18394](https://github.com/longeye-ai/monorepo/issues/18394). Punt received the same boundary; this should not block the already-approved supervision fix here.

zionts:
Live incident found after this PR was opened: one Watch Desk had a live Codex judge and a live Claude judge concurrently. Agent-kind-aware discovery alone does not establish exclusive authority. Contained by preserving transcripts, closing the stale/duplicate Claude panes, and keeping one Codex judge. Durable single-writer lease/fencing requirements are in #567. Please treat #567 as a required follow-up or incorporate its fail-closed lease invariant before calling mixed-agent Nightwatch supervision complete.

PR #569, opened 2026-08-01
Compound: Require a spec unless it's a bug fix -- and make the rule reachable from Codex

cheapsteak:
## Summary

PR #561 shipped a new daemon subsystem, an RPC, a CLI verb, an on-disk store and a bespoke transcript heuristic with no spec. It was the first real test of the brainstorming gate that landed five days earlier, and the gate did not fire. Investigating why turned up two independent causes — one social, one mechanical — and neither is what I expected going in.

**The escape hatch accepted "the design already happened elsewhere."** #561's exemption was not a trigger-matching mistake. It gave a substantive reason: *"the design was already constrained by the six successful live prototypes and the supplied parity report."* The old rule permitted any reason. But six prototypes and a parity report are a session scratchpad by another name — none of that material is in this repo, so no reviewer could examine the decisions it encodes. The rule now states its purpose (decisions must be examinable, changes to our theory of the system or the product most of all), inverts the default to *required unless this is a bug fix or a minor UI change*, and says prior design makes a spec a cheap transcription rather than an unnecessary one. Refactors leave the exemption list: "wholesale-replaces a load-bearing path" and "this is just a refactor" describe the same diff from two angles.

**The gate was unreachable from Codex, and #561 was written from Codex.** `/tbd-brainstorming` lives in `.claude/skills/`, which Codex never reads. The old sentence — *"Use `/tbd-brainstorming`, not `superpowers:brainstorming`; a guardrail redirects the wrong one"* — named a skill Codex cannot invoke, forbade one that is also absent from its catalog despite being enabled, and promised a guardrail that only runs in Claude Code. A conscientious Codex session following it had no legal move. Codex readers are now pointed at the `SKILL.md` path directly, which works because `.codex/config.toml` already routes them to `CLAUDE.md`.

The trigger list lived in three places, so all three move together — including the skill's frontmatter `description:`, which is what the model matches to decide whether to invoke at all.

Also on this branch, from a parallel session: research into Codex's native `externalAgentConfig/import` surface and a Continue-in-Codex spec built on it. Codex ingests a Claude transcript itself and returns the created thread id, so TBD never parses a transcript and owns no excerpting policy — which is a direct answer to the heuristic #561 introduced. The gate spec's follow-up cites that research, which is why the two ship together.

Relates to #561.

## Test plan

- [x] `codex debug prompt-input` in this worktree shows all four new elements reaching a Codex session — purpose statement, inverted default, prior-design clause, `SKILL.md` pointer — and the old heading gone
- [x] The same probe confirms `CLAUDE.md` reaches Codex even in a worktree absent from `[projects.*] trust_level`, closing an open question from the 2026-07-26 gate design
- [x] Guardrail tests pass (7) after the deny-message rewording
- [x] `scripts/check-no-committed-plans.sh` clean — the 2026-07-26 spec tripped this on its own first commit
- [x] The skill roster reloaded mid-session with the new description, confirming the frontmatter change is live
- [ ] Reviewer sanity-check on the frontmatter `description:` — it went from a specific trigger list to a broader statement, and broad descriptions match *worse* in a crowded roster

## Notes for review

No code changes; `CLAUDE.md`, a vendored skill, a guardrail rule's wording, and docs.

Two things this deliberately does **not** do. It adds no mechanical enforcement — the 2026-07-26 design rejected a PR-time spec check as "too late," and that reasoning still holds. And it leaves the free-form escape hatch in place, narrowed by the prior-design clause rather than replaced by a closed list of exemption categories.

Known gap, unchanged and now with evidence: this is an agent-side convention, and human contributors get no enforcement at all.

[🔍 Investigate PR561 Brainstorm](https://cheapsteak.github.io/tbd/open/?worktree=994F568C-F1D9-47FB-944D-546E8F1AFE1D)

### DIFF
--- a/.claude/hooks/guardrails/rules/brainstorming_skill.py
+++ b/.claude/hooks/guardrails/rules/brainstorming_skill.py
@@ -2,10 +2,10 @@
 
 TBD vendors an adapted copy of superpowers' brainstorming skill at
 `.claude/skills/tbd-brainstorming/`. The adaptations are load-bearing: it writes
-specs to `docs/specs/`, carries TBD's blast-radius triggers and the
-"a human answers the questions" rule, and does not hand off to a skill whose own
-file would trip this repo's plans-guard. Running the upstream skill instead
-silently loses all of that.
+specs to `docs/specs/`, carries TBD's "required unless it is a bug fix or a minor
+UI change" scope and the "a human answers the questions" rule, and does not hand
+off to a skill whose own file would trip this repo's plans-guard. Running the
+upstream skill instead silently loses all of that.
 
 Unlike inferring *whether* a brainstorm happened — which no hook can see — this
 is a string comparison, so it is safe to deny rather than merely inform: the
@@ -24,8 +24,9 @@ _REDIRECTED = frozenset({"brainstorming", "superpowers:brainstorming"})
 _MESSAGE = (
     "[brainstorming-skill] Blocked: use TBD's vendored `tbd-brainstorming` skill, "
     "not the upstream one. TBD's copy writes the spec to docs/specs/, carries the "
-    "blast-radius triggers, and enforces that a human — not you — answers the "
-    "brainstorming questions. Fix: invoke the skill `tbd-brainstorming` instead."
+    "rule that a spec is required for anything that is not a bug fix or a minor UI "
+    "change, and enforces that a human — not you — answers the brainstorming "
+    "questions. Fix: invoke the skill `tbd-brainstorming` instead."
 )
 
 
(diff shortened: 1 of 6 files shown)


cheapsteak:
Both addressed.

**High — leaked session title.** Correct finding, and the reasoning was right about provenance: that title came from an unrelated project on this machine, swept in because `detect` reads all of `~/.claude/projects`. Removed in 20fd9054.

Rather than substitute a synthetic title, I replaced it with the point the leak was evidence *of*: `detect` is home-scoped, so the 49 entries span every repository on the machine, and each title states what that work was about. Any picker built on `detect` will therefore show a user titles from unrelated projects — including, on a work machine, ones confidential to another org. That makes "filter to the current worktree" a confidentiality requirement rather than an ergonomic nicety, which is worth having in the record before anyone builds the picker. Substituting a fake title would have lost that and misrepresented a section that is otherwise measured output.

Swept the other two new docs for `/Users/<name>` paths, org names and internal URLs while I was in there — clean.

**Note — `SKILL.md` description dilution.** Agreed, and fixed in e8a61c9e rather than left as a watch item. The added catch-all ("or any change to how the system or product is meant to work") was redundant with the clause it followed, since "anything that is not a bug fix or a minor UI change" already carries the inverted default. Dropping it puts the description at **354 chars — 5 shorter than before this PR**, down from 428, with every concrete trigger keyword retained. So the dilution is gone without trading away the change.

The open question in the test plan is now narrower: whether the inverted phrasing matches as well as the old enumerated list at equal length. Still not something I can test here.

**On the diagnostics note** — the stale working-tree copies were confined to the CI checkout; `git status` in the authoring worktree was clean at every commit, and `HEAD` is what I intended throughout. Flagging it since it'll recur for whoever tunes that workflow.

PR #570, opened 2026-08-01
Session Workbench: readable transcripts and artifact index

zionts:
## Summary

- make assistant prose the primary reading surface with an unfilled narrative treatment and quiet signal spine
- collapse consecutive tool/system activity into native virtualized `Worked · N actions` groups while keeping failures, questions, and in-progress state visible
- add a responsive session-local index for Changed files, Sources, Web references, and Delegates
- use the same projection for live and historical transcripts; narrow panes replace the 248pt rail with an inspector button

## Design

Approved spec: `docs/specs/2026-08-01-session-workbench-design.md`.

This is intentionally a current-session presentation projection over existing `[TranscriptItem]`. It adds no database, persistence, daemon RPC, or transcript parsing contract. Index entries come only from structured inputs of explicitly recognized tools; malformed and unknown tools remain ordinary activity rather than being guessed into the index.

## Validation

- `/Users/zionts/.swiftly/bin/swift build --product TBDApp` ✅
- `/Users/zionts/.swiftly/bin/swift build --product TBDDaemon` ✅
- `shellcheck scripts/mock.sh` ✅
- `git diff --check` ✅
- commit is GitHub-signed and verified ✅

## Remaining verification

This host has split Swift toolchains: CommandLineTools Swift 6.2.1 has XCTest but not the package's Testing module; Swiftly 6.3.2 has Testing but cannot import XCTest, and SwiftPM compiles the full TBDAppTests target even under a filter. The focused tests are included but could not execute locally.

The isolated mock app and daemon launched successfully, but this Codex process lacks Screen Recording and Accessibility permission. Wide/narrow, light/dark screenshot QA remains required before merge.


cheapsteak (approved):
This is great, thank you 🙏🙏

PR #576, opened 2026-08-02
fix: serialize local Swift builds

zionts:
## Summary

- add a machine-global SwiftPM admission lock shared by all TBD worktrees
- default compile commands to two jobs and reject unsafe job counts
- block raw agent-issued SwiftPM build/test/run commands
- route restart, mock, and nightly stress builds through the governor
- document the resource controls and update agent/contributor guidance

## Root cause

Multiple independent TBD workers launched raw SwiftPM commands simultaneously. Each defaulted to `-j12`; surviving `swift-build` and `swift-test` parents kept spawning replacement `swift-frontend` children after individual frontends were killed. One worker also ran a sequential build loop, so stopping a child advanced it to the next target. Four concurrent build roots drove load above 150 and swap to roughly 15 GB.

## Validation

- `python3 .claude/hooks/guardrails/dispatch.py --self-test` (40 passed)
- `python3 scripts/test-swift-safe.py` (6 passed)
- `bash scripts/reclaim-build.test.sh` (all passed)
- `scripts/swift-safe build --target TBDShared` (passed, 56.93s)
- `git diff --check`

The real build ran alone through the new admission path. After legacy build roots were interrupted, Swift compiler process counts remained at zero outside that controlled validation.

### DIFF
--- a/.claude/hooks/guardrails/rules/swift_build_admission.py
+++ b/.claude/hooks/guardrails/rules/swift_build_admission.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import os
+import re
 import shlex
 
 from guardrails.lib.rule import Decision, Rule
@@ -10,8 +11,9 @@ from guardrails.lib.rule import Decision, Rule
 
 _COMPILE_SUBCOMMANDS = {"build", "package", "run", "test"}
 _DISPLAY_COMMANDS = {"awk", "cat", "echo", "grep", "printf", "rg", "sed"}
+_SHELL_COMMANDS = {"bash", "dash", "sh", "zsh"}
 _BOUNDARIES = {"&", "&&", "(", ")", ";", "|", "||"}
-_SAFE_RUNNER = "scripts/swift-safe"
+_MAX_NESTED_SHELL_DEPTH = 4
 _DENY_REASON = (
     "[swift-build-admission] Blocked raw SwiftPM compilation. Multiple TBD "
     "worktrees once launched concurrent default -j12 builds, filled 15 GB of "
@@ -39,9 +41,53 @@ def _segments(command: str) -> list[list[str]]:
     return [segment for segment in segments if segment]
 
 
-def _contains_raw_swift_compile(segment: list[str]) -> bool:
-    if any(token.endswith(_SAFE_RUNNER) for token in segment):
-        return False
+def _nested_shell_payloads(segment: list[str]):
+    for shell_index, token in enumerate(segment[:-1]):
+        if os.path.basename(token) not in _SHELL_COMMANDS:
+            continue
+        for option_index in range(shell_index + 1, len(segment) - 1):
+            option = segment[option_index]
+            if not option.startswith("-"):
+                break
+            if "c" in option[1:]:
+                yield segment[option_index + 1]
+                break
+
+
+def _command_substitution_payloads(token: str):
+    start = 0
+    while True:
+        opening = token.find("$(", start)
+        if opening < 0:
+            break
+        depth = 1
+        cursor = opening + 2
+        while cursor < len(token) and depth:
+            if token.startswith("$(", cursor):
+                depth += 1
+                cursor += 2
+                continue
+            if token[cursor] == ")":
+                depth -= 1
+                if depth == 0:
+                    yield token[opening + 2 : cursor]
+                    start = cursor + 1
+                    break
+            cursor += 1
+        else:
+            break
+    yield from re.findall(r"`([^`]*)`", token)
+
+
+def _contains_raw_swift_compile(segment: list[str], depth: int = 0) -> bool:
+    if depth < _MAX_NESTED_SHELL_DEPTH:
+        for token in segment:
+            for payload in _command_substitution_payloads(token):
+                if _command_contains_raw_swift_compile(payload, depth + 1):
+                    return True
+        for payload in _nested_shell_payloads(segment):
+            if _command_contains_raw_swift_compile(payload, depth + 1):
+                return True
     if segment and os.path.basename(segment[0]) in _DISPLAY_COMMANDS:
         return False
     for index, token in enumerate(segment[:-1]):
@@ -50,6 +96,16 @@ def _contains_raw_swift_compile(segment: list[str]) -> bool:
         if segment[index + 1] in _COMPILE_SUBCOMMANDS:
             return True
     return False
+
+
+def _command_contains_raw_swift_compile(command: str, depth: int = 0) -> bool:
+    if depth < _MAX_NESTED_SHELL_DEPTH:
+        for payload in _command_substitution_payloads(command):
+            if _command_contains_raw_swift_compile(payload, depth + 1):
+                return True
+    return any(
+        _contains_raw_swift_compile(segment, depth) for segment in _segments(command)
+    )
 
 
 class SwiftBuildAdmissionRule(Rule):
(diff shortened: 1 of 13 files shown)


cheapsteak:
## Review

The PR is coherent and most of it is load-bearing for the fix, but the enforcement layer is overscoped as written, and there are three concrete bugs worth fixing before merge.

### What's here, judged separately

1. **Mechanism** — `scripts/swift-safe` (machine-global flock + default `--jobs 2`). This is the actual fix for the incident; necessary.
2. **Routing** — `restart.sh`, `mock.sh`, `nightly-flake-stress.sh` through the wrapper. Necessary: a lock nobody calls protects nothing, and these are the real build entry points.
3. **Enforcement** — the guardrail denying raw agent-issued `swift build/test/run/package`. Defensible in kind, overscoped in coverage (see below).
4. **Docs/tests** — consequential fallout of 1–3.

### On the guardrail being drastic

It's less drastic than it looks: it only binds agent-issued Bash (terminal, CI, Xcode untouched), a deny is a one-round-trip self-correction rather than a dead end, and it follows this repo's established convention — the guardrails README documents exactly why hard blocks exist here (the orphaned-`yes`-process incident, where advisory text failed). Advisory-only CLAUDE.md guidance is precisely what already failed in this build storm, so some enforcement is justified if the lock is meant to hold.

**However, blocking `swift package` wholesale is too broad.** `swift package dump-package` / `describe` / `clean` / `resolve` spawn no compiler swarm, yet as written a cheap metadata query gets denied, rerouted through the wrapper — and then waits behind a 30-minute build lock. Suggest narrowing the rule to `build/run/test`, and dropping `package` from the wrapper's lock (or locking only compiling package subcommands like `plugin`).

### Bugs

1. **`swift run` passes the jobs flag to the app, not SwiftPM.** `_bounded_arguments` appends `--jobs 2` at the end, but `swift run` stops parsing SwiftPM options at the executable name — `swift run TBDApp --jobs 2` hands `--jobs 2` to TBDApp itself. Insert the flag immediately after the subcommand. (Symmetrically, `_requested_jobs` scans app-side arguments and can misparse a `-j` meant for the target binary.)

2. **Killing the wrapper orphans the compile and frees the lock — recreating the incident.** With `subprocess.run`, the lock lives with the wrapper, not the compiler. `nightly-flake-stress.sh` wraps the call in `run_with_deadline`, which kills on timeout: the wrapper dies, the flock releases, the orphaned `swift test` keeps compiling, and the next admission runs concurrently with it — the exact "surviving parents" behavior from the root cause. Fix: `os.set_inheritable(lock_file.fileno(), True)` then `os.execv` the swift binary, so the lock lives exactly as long as the compiler and signals reach it directly. Simpler code, too.

3. **Doc/code mismatch on the ceiling.** `docs/reclaim-build.md` says commands "reject more than four unless a developer explicitly opts out," but the code rejects anything above `TBD_SWIFT_JOBS` (default 2) — `-j3` fails out of the box; `MAX_JOBS = 4` only caps the env var itself. Align one to the other.

### Minor

- The safe-runner whitelist is positional-blind: `any(token.endswith("scripts/swift-safe") ...)` exempts a whole segment if the string appears anywhere (`FOO=scripts/swift-safe swift build` sails through), and unparseable commands (`shlex` `ValueError`) fail open silently. Fine against cooperative agents, but checking the command position would tighten it.
- Neither `scripts/test-swift-safe.py` nor the guardrail suite runs in CI — the lint job is SwiftLint-only and nothing invokes `dispatch.py --self-test`. Pre-existing gap for the guardrail tests, but this PR doubles down on it.
- The wrapper honors `TBD_HOME` and its tests isolate via a temp `TBD_HOME` — consistent with the "tests must not touch `~/tbd`" rule. The deny message is instructive and names the fix — good hook hygiene.

### Bottom line

Keep the mechanism, routing, and enforcement-in-kind; request three changes: narrow the guardrail (and lock) to actually-compiling commands, fix the `swift run --jobs` placement, and switch to `execv` so the lock cannot outlive or underlive the compiler.

---

cheapsteak:
## Follow-up: which parts should be default-on vs. opt-in

After discussion, a concrete recommendation on scoping — the cleanest split isn't "make enforcement opt-in"; it's **keep the lock and the deny, drop the jobs cap**.

### Keep the lock, default-on

The flock is invisible when uncontended and engages exactly when it's helping (another worktree compiling). Making the mechanism opt-in would mean nobody's builds serialize until everyone opts in, which defeats it. `scripts/restart.sh` and friends route through it transparently, so day-to-day workflows are unaffected on an idle machine.

### Keep the deny, default-on — but narrowed to `build/test/run`

An advisory default is the thing that already failed: the guardrails framework's own README exists because CLAUDE.md text couldn't prevent the orphaned-`yes`-process incident, and the repo precedent (`background_jobs`) ships hard-deny with no opt-in. An `info` nudge is CLAUDE.md text with extra steps. The risk also isn't one machine's: TBD's entire purpose is running fleets of concurrent agent sessions across worktrees, so every developer of this repo shares the risk profile. The cost of default-on is near zero — `scripts/restart.sh` is never touched by the rule, and a raw `swift build` self-corrects in one round trip via the deny reason. Dropping `package` from the rule (per the earlier review) removes the main friction case.

### Drop the `-j2` cap — or invert it to opt-in

Once the machine-global lock serializes builds, the cap solves the same problem a second time: there's only ever one build running, and a single serialized build using all cores was the normal, fine state before the incident. The swap blowup came from four *concurrent* `-j12` builds, not from `-j12` itself. Meanwhile the cap is the one piece that taxes every developer every day — it slows every build even on an idle machine, forever — and it doesn't address the respawning-frontends part of the root cause (the `execv` lock-lifetime fix from the review is what handles that).

Suggested inversion: `TBD_SWIFT_JOBS` set → clamp to it; unset → SwiftPM's own default. A developer who wants 2 jobs on a loaded machine sets the env var in their shell profile, without imposing the ceiling on everyone else. This also matches the repo's "defaults must generalize, don't encode one machine's constraints" guidance.

### Summary

| Piece | Recommendation |
|---|---|
| Machine-global flock | Keep, default-on |
| Guardrail deny | Keep, default-on, narrow to `build/test/run` |
| `-j2` cap | Remove, or opt-in via `TBD_SWIFT_JOBS` |
| `swift package` in rule + lock | Exclude non-compiling subcommands |

---

zionts:
Addressed the concrete review findings in signed commits `9e61191` and `7e60fd4`:

- nested `bash/sh/zsh -c`, `$()` and backtick substitutions can no longer bypass admission;
- the Nightly outer backstop now covers lock wait + execution budget + grace;
- non-compiling `swift package` commands no longer enter the guard or lock;
- `swift run` injects `--jobs` before the executable and ignores app-side `-j` arguments;
- the wrapper now `execv`s SwiftPM with an inheritable lock FD, so signals hit SwiftPM directly and the lock remains held for the compiler lifetime;
- docs now match the configured-limit behavior;
- guardrail and wrapper tests now run in CI.

I kept the default two-job cap intentionally. This incident was reported as both excessive concurrent builds and excessive `swift-frontend` machine usage; serializing prevents cross-worktree overlap, while the cap prevents the one admitted build from saturating this workstation. The repo's macOS CI already documents and uses `-j 2` for the same memory-pressure reason. `TBD_SWIFT_JOBS` and `TBD_SWIFT_ALLOW_HIGH_JOBS=1` retain an explicit escape hatch for isolated machines.

Local validation on `7e60fd4`: 38 guardrail tests, 9 wrapper tests (including lock survival across exec), Nightly harness tests, reclaim tests, and `git diff --check` all pass.

cheapsteak (approved):

PR #587, opened 2026-08-04
feat: v2 posts one review comment per run and hides prior ones as outdated

cheapsteak:
## What

v2 posted **two** comments per run: the action's `track_progress` tracking comment (which became the review prose) and a separate workflow-authored marker comment holding machine state. Both are gone.

Now each run posts **one** review comment, and the state the skip decision needs — sentinel, `last-reviewed-patch-id`, `last-verdict` — rides inside that comment as leading HTML comments. Before posting, the PR's earlier v2 reviews are collapsed via the `minimizeComment` GraphQL mutation with classifier `OUTDATED`, so the thread keeps its history without the pile-up and the current review is the one left expanded.

## Design points

- **Minimize runs before the post**, so a run can never collapse its own review. Cost of that ordering: if the post then fails, priors are left collapsed — it warns, and the next run posts afresh.
- **The selector requires both App authorship and a sentinel-leading body**, so a human comment quoting the sentinel is never minimized.
- **The skip path writes nothing to the PR** — no comment, no minimize. Appending a "diff unchanged" note would mean a fresh comment on every no-op push; the prior review stays visible and says the same thing.
- **The session holds no GitHub write tool.** `track_progress`, the comment-posting MCP tool, and `Bash(gh api:*)` are all gone from the action step — the workflow posts from `review-result.json`'s `comment_body`, so every write to GitHub is made by a deterministic step.
- **Comment-body construction moved into `render_comment.py`**, joining `prepare.py` and `validate.py` as the third pure bookend script. It falls back to rendering the findings themselves when `comment_body` is blank — otherwise a REJECT with no prose posted markers and nothing else, telling the author nothing.

## Fail ordering

A missing or malformed `verdict.txt` exits **before** anything is posted. The REJECT enforcement exits **after** posting, so a rejecting review reaches the author before the check goes red. Minimize and post are best-effort: either failing warns and leaves verdict enforcement intact.

## Verification

- 135 tests green; `bash -n` on all 11 `run:` blocks; both workflows parse.
- Round-trip tests pin `render_comment.py`'s output against `prepare.py`'s parser, including that model prose containing forged marker lines cannot displace the workflow's own.

**Correction to an earlier version of this description:** it said the selector was "verified against an impostor fixture" and the post step "driven against a stub `gh` across nine cases." Those runs happened during development but in a throwaway sandbox — nothing was checked in, so the repo had no such coverage and the wording implied more than the tree delivered. Both reviewers caught it. Committed tests for that shell are being added in this PR rather than described.

## Known limitation

Already-minimized comments cost a no-op mutation each run: the REST comment payload doesn't expose `isMinimized`, so the loop tolerates the error rather than skipping them. On a PR with N prior reviews that's N−1 wasted calls per run; moving the listing to GraphQL would remove it.

Whether the App can perform `minimizeComment` at all is not verifiable from a dev box — it needs Pull requests: write, which the App demonstrably holds since it already posts and patches comments. No scope was widened on a guess. If it turns out to lack the permission, the failure is contained: each comment warns and the review still posts and enforces.

No spec was written for this — it revises §3.5 of the existing fan-out design spec, which is updated here to state the current design.

https://claude.ai/code/session_01XUgeXxHP455HutsFvUZcxv


cheapsteak:
Pushed `98bac03` addressing all four findings.

**HIGH — posting-step shell had no committed test.** Both reviewers raised this, and the PR description made it worse by claiming stub-`gh` verification that existed only in a dev sandbox. That description is corrected, and the coverage now exists: `tests/test_post_step.py` (24 tests) extracts the step's `run:` block **by step name**, executes it under `bash` with a stub `gh` on `PATH`, and asserts the recorded call *order* — `kinds == [list, minimize, minimize, post]` — not just outcomes. The impostor fixture the description previously claimed now genuinely exists: a human comment opening with the sentinel, an App comment without it, an App comment quoting it mid-body, and a v1 bot comment — none minimized.

Extraction by name means renaming the step fails loudly (20 of 24 error with `StepNotFound`) rather than silently testing nothing.

Mutation-tested, independently re-run rather than taken on report:
- verdict-gate bypass (`*)` accepts anything instead of exiting) → **6 tests fail**
- minimize/post inverted → **5 tests fail**, led by `test_every_minimize_precedes_the_post`
- re-restore step deleted → 4 of 5 structural tests fail

Worth recording: a *weaker* first mutation looked uncaught, because the code re-reads and re-checks the verdict immediately after the first gate — two independent gates, so disabling one preserves fail-closed behavior. The tests were right; the mutation was wrong.

**MEDIUM — restore wasn't durable.** Sharper than reported: the window doesn't just expose `render_comment.py`, it exposes `validate.py`, which *computes the verdict*. A prompt-injected session holding `Write` could rewrite it so every review approves — a gate bypass, pre-existing on `main` rather than introduced here. The first restore now pins the resolved base SHA into an output, and a new **Re-restore the review pipeline from the base branch** step re-establishes that directory from the *same* SHA between the session and validation. Session outputs are unaffected — `review-result.json`, the findings files, `verdict.txt` and the prepare artifacts all live outside that directory (verified, not assumed). The two comments that overstated tamper-resistance now describe what the mechanism actually guarantees.

**MEDIUM — flag-policy exemption.** Written into spec §3.5 as rationale. The substantive argument: the entire v2 pipeline is a non-required shadow check, which *is* the default-off gate; graduation to a required check is the point at which this behavior becomes load-bearing. The mutation is also confined to the App's own comments by an authorship+sentinel selector, is reversible, and destroys no content.

**Minor — `echo` vs `::warning::`.** Fixed: the first failure and a final count are both annotations, with per-comment lines in between so a wide failure doesn't spam. That is precisely the signal that will tell us whether the App can perform `minimizeComment` at all — still the one thing not verifiable from a dev box.

Also fixed, from the review diagnostics rather than the findings: both specialists reported the checkout as "stray uncommitted contamination." It isn't — that directory is deliberately restored from base. The session is now told so. The real hazard wasn't the wasted effort: a session that *didn't* notice would `Read` those paths, get base-branch content, and review code that isn't in the PR — silently wrong, on exactly the PRs that modify the review pipeline.

135 → 164 tests.

---

PR #589, opened 2026-08-05
Feat: Quiet the transcript down to prose, and preview pasted images inline

cheapsteak:
## Summary

A session of transcript polish, driven by dogfooding. Each change removes something that was drawing the eye without earning it, or replaces stilted chrome with a sentence.

**Work groups read as prose.** A collapsed group said `Worked · 3 actions · Commands`. It now says `Edited 2 files, ran 4 shell commands` — one clause per kind of work, comma-joined, capitalized once. This vocabulary is not invented: it was read out of Claude Code 2.1.222's own bundle, so TBD matches the tool it wraps. Verb tense carries execution state (`running`/`ran`, `reading`/`read`, plus a trailing `…`), which made the status badge redundant. Failed and needs-response badges stay — tense doesn't encode those.

**Groups of one unwrap.** A single action renders as its own row rather than a `Worked · 1 action` wrapper around one thing. Claude Code does the same, and independently: it only collapses at two or more.

**Removed, as pure noise:** the gray `complete` capsule, the vertical rule down the left of every assistant message (it made the whole transcript read as quoted matter), the group icon, and the `Claude · 5:33 PM` / `7:49 PM · You` attribution header on every bubble — position already says who spoke. The attribution survives as the accessibility label, so VoiceOver keeps speaker identity and timestamp.

**Background-task rows got the same treatment.** A survey of every task-notification envelope under `~/.claude/projects` (10,733 background commands, 4,938 agents, 3,003 monitors, 2,689 stop hooks) showed Claude Code already writes each summary as a self-describing sentence carrying its outcome. So the `Background ·` lead-in is folded in only for the bare `Agent "…"` shape; prefixing the rest would have stuttered into `Background Background command …` or misnamed a monitor.

**Pasted images render as thumbnails.** They were showing as a raw absolute path. A paste lands as two JSONL user lines — the prompt, with inline `[Image #N]` placeholders and the bitmap as base64, and an `isMeta` follow-up carrying one `[Image: source: <path>]` marker per image. The marker is the only carrier of the on-disk path, so it becomes a thumbnail that fits in a 200pt square and opens Quick Look on click; right-click gets Reveal in Finder and Copy Image.

## Notable fixes found along the way

**Expanding a group scrolled the whole transcript.** A toggle rewrites the node array, which classifies as `.rebuild` and so reached the tail-follow block — a user disclosure gesture was indistinguishable from streamed-in content, and re-pinning to the new bottom slid everything on screen upward. Provenance: the expansion feature came in with 4ea9e1bc (#570, Session Workbench) and the tail-follow with 69ebdb7a (#300); neither is wrong alone, and the interaction went unnoticed because no test asserted on scroll anchoring. A toggle token now gates that one scroll, leaving the rebuild classification, the reload, and the tail-first open backfill alone so streaming appends still auto-scroll.

**`.rebuild` was wiping four caches whose keys already encode `(id, contentVersion, width)`.** Stale entries were unreachable by construction, so the wipe only forced exactly-measured rows back to estimates, which then re-corrected without compensation and drifted rows above the click. They now survive, pruned to the live node set.

**`UserMessageClassifier` took only the first text block**, so a multi-image paste lost every marker after the first before the app ever saw it. Across the whole corpus, every user message with multiple text blocks was an image-marker meta line, so joining them is a strict repair.

## Notes for review

- Every change lands in **both** rendering paths (native `NSTableView` cells and the hosted SwiftUI tree). Divergence between them is a standing trap here — a pre-existing one is fixed in passing: `SystemReminderRow` rendered a `[background]` capsule and no summary at all.
- **Image work is measured, not assumed.** Sizing reads true pixel dimensions from the image header, benchmarked at ~200× cheaper than a decode (0.12 ms vs 26.4 ms on a 4000×3000 PNG), so rows measure exactly and synchronously and never resize when the async thumbnail lands. Decode is `CGImageSourceCreateThumbnailAtIndex` on a background queue, bounded by max-pixel-size, behind two bounded caches — the on-main decode hazard is the same class as the #129 Highlightr freeze.
- **QuickLookUI was verified to work without a bundle identifier** via a throwaway no-`Info.plist` binary, so no unbundled-executable guard was added. The panel controller methods live on `AppDelegate` at the responder chain's terminus, so one wiring serves both renderers.
- **Any absolute path a marker names will render**, deliberately: the corpus shows two distinct cache roots and profile dirs are user-configurable, so an allowlist would break real transcripts. Guards added instead: a non-regular-file check (a marker naming `/dev/random` would otherwise hang the reader) and degrading non-images to a name-only chip. Missing files never dump the raw path.
- The workbench spec is updated: it named the removed signal spine as its signature element and mocked up the old summary line.

## Test plan

- [x] `scripts/swift-safe build` clean
- [x] `/opt/homebrew/bin/swiftlint --strict` — 0 violations across 662 files
- [x] `scripts/test.sh --filter '^TBDAppTests\.'` — 1939 tests pass (baseline 1892 at branch point; count checked, not just the exit code)
- [x] Daemon suites for the classifier change — 88 tests pass
- [x] The scroll-anchor guards were **mutation-checked**: each fix was deliberately reverted and the corresponding test confirmed red before restoring
- [ ] Live: expand a work group mid-transcript and confirm the clicked row stays put, both when pinned to the bottom and scrolled up
- [ ] Live: paste an image, click for Quick Look, right-click for Reveal in Finder
- [ ] Live: confirm streaming auto-scroll still follows the tail during an active turn

### Known coverage gap

The Quick Look panel presentation itself is not tested — a real panel needs a key window and would pop over the developer's screen — so the click tests drive the same action through `accessibilityPerformPress()` with the panel call behind an injectable seam. The SwiftUI path's click and context menu have no headless driver either.

### Follow-up, deliberately not in this PR

A real `splice(at:removed:inserted:)` step in `TranscriptStreamPlan` would make a toggle an insert/delete instead of a whole-table reload, so the scroll anchor would hold by construction rather than by convention. Larger change, and it doesn't fix the symptom on its own.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=01BF5ED0-FFFC-4505-BE99-617EEC285354)

https://claude.ai/code/session_018LEZrAYtVQiy1L5NwsFJPB

### DIFF
--- a/Sources/TBDApp/Panes/HistoryPaneView.swift
+++ b/Sources/TBDApp/Panes/HistoryPaneView.swift
@@ -369,6 +369,10 @@ struct SessionTranscriptView: View {
     /// the last row (NSTableView renderer path).
     @State private var scrollToBottomToken: Int = 0
     @State private var activityGroupExpansion: [String: Bool] = [:]
+    /// Bumped alongside every write to `activityGroupExpansion`, so the table can
+    /// tell a user-driven disclosure toggle (anchor the clicked row) from a
+    /// streaming update (follow the tail).
+    @State private var activityToggleToken: Int = 0
 
     @State private var isFreshBranchReviveInFlight = false
 
@@ -462,6 +466,7 @@ struct SessionTranscriptView: View {
                         ),
                         atBottom: $atBottom,
                         scrollToBottomToken: scrollToBottomToken,
+                        activityToggleToken: activityToggleToken,
                         nodesProvider: { presentation.nodes }
                     )
                     .overlay(alignment: .bottomTrailing) {
@@ -525,6 +530,7 @@ struct SessionTranscriptView: View {
 
     private func setActivityGroup(_ id: String, expanded: Bool) {
         activityGroupExpansion[id] = expanded
+        activityToggleToken &+= 1
     }
 
     @ViewBuilder
--- a/Sources/TBDApp/Panes/Transcript/ActivityRowChrome.swift
+++ b/Sources/TBDApp/Panes/Transcript/ActivityRowChrome.swift
@@ -5,7 +5,11 @@ import TBDShared
 /// Provides the full-bleed row layout, header (icon + title + timestamp),
 /// and a header-only click-to-overlay interaction (see #129 spec).
 struct ActivityRowChrome<Header: View>: View {
-    let icon: String
+    /// Leading SF Symbol, or nil for a row that renders no icon at all. Nil
+    /// collapses the icon column entirely so the header sits at the row's
+    /// leading inset — it does NOT leave an empty gutter. Mirrors
+    /// `ActivityRowPresentation.iconSystemName` on the native path.
+    let icon: String?
     let timestamp: Date?
     let onOpen: () -> Void
     let headerContent: () -> Header
@@ -13,7 +17,7 @@ struct ActivityRowChrome<Header: View>: View {
     @State private var hovering = false
 
     init(
-        icon: String,
+        icon: String?,
         timestamp: Date?,
         onOpen: @escaping () -> Void,
         @ViewBuilder header: @escaping () -> Header
@@ -27,10 +31,12 @@ struct ActivityRowChrome<Header: View>: View {
     var body: some View {
         Button(action: onOpen) {
             HStack(spacing: 6) {
-                Image(systemName: icon)
-                    .font(.caption)
-                    .foregroundStyle(.secondary)
-                    .frame(width: 14)
+                if let icon {
+                    Image(systemName: icon)
+                        .font(.caption)
+                        .foregroundStyle(.secondary)
+                        .frame(width: 14)
+                }
                 headerContent()
                     .font(.subheadline)
                     .foregroundStyle(.secondary)
--- a/Sources/TBDApp/Panes/Transcript/ChatBubbleView.swift
+++ b/Sources/TBDApp/Panes/Transcript/ChatBubbleView.swift
@@ -83,6 +83,8 @@ struct ChatBubbleView: View {
                         .transcriptSelectableText()
                 case .code(let lang, let body):
                     codeBlock(language: lang, content: body)
+                case .image(let attachment):
+                    TranscriptImageAttachmentView(attachment: attachment)
                 }
             }
         }
--- a/Sources/TBDApp/Panes/Transcript/ChatBubbleView.swift
+++ b/Sources/TBDApp/Panes/Transcript/ChatBubbleView.swift
@@ -24,43 +24,41 @@ struct ChatBubbleView: View {
 
     private var roleLabel: String { isUser ? "You" : "Claude" }
 
+    /// Speaker attribution for assistive technology only — never drawn. A
+    /// transcript is not a group chat, so the bubble shows no role/timestamp
+    /// header and relies on position and tint; VoiceOver has no position cue, so
+    /// the bubble carries the attribution as its accessibility label. Mirrors
+    /// `TranscriptBubbleGeometry.accessibilityAttribution(for:)`.
+    private var accessibilityAttribution: String {
+        guard let ts = item.timestamp?.absoluteShort else { return roleLabel }
+        return isUser ? "\(ts) · \(roleLabel)" : "\(roleLabel) · \(ts)"
+    }
+
     var body: some View {
         // Flattened row wrapper (issue #129 per-row layout-depth): the prior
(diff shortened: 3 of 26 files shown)


cheapsteak:
Addressed in 2af4a123, 5b5d9830, bbf8474c. Notes on each, including where the findings turned out to be right for a different reason than stated.

## CI test failure — not what it looked like

The anchor test was red on the runner and green locally. My first read was font metrics; that was wrong. The real cause is **`NSScroller.preferredScrollerStyle`**, which follows the host's "Show scroll bars" setting and whether a scroll-capable pointing device is attached. A developer box gets overlay scrollers (table width 680); the runner gets legacy ones — **663, exactly the width in the CI output**. With `autohidesScrollers`, that 17pt is claimed only on the layout pass after the first `reloadData`, so every height measured before it was cached under `(id, contentVersion, 680)` and unreachable at 663. The toggle then took the **width-change branch** — a legitimate full cache wipe — instead of the `.rebuild` branch the test means to exercise, rows above fell back to estimates, and the anchor jumped.

The numbers confirm it exactly: an assistant bubble measures 96 at width 680 and 112 at 663, and estimates 140. CI's `100 → 144` is precisely "measured at 680, then estimated at 663" (+4pt intercell spacing). Pinning legacy + autohiding scrollers locally reproduced the failure row-for-row, and surfaced two more tests CI hadn't reached yet.

Fix is harness-only, no product change: pin the scroller style so local and CI share geometry, add a `primeScene` that settles the width before any snapshot, realize every row so both snapshots are exact by construction, and assert the width didn't move across the click — so a future environment shift fails as itself rather than as a phantom 540pt anchor jump. Verified independent of estimate accuracy by perturbing the estimator constants (18→31, 7.0→4.5): all 13 harness tests still pass. Mutation checks still bite, and harder than before — reverting the cache fix now moves 18+ rows where it previously moved ~5.

**Worth filing separately:** the estimator is materially off *and in the wrong direction*. A plain two-paragraph bubble estimates 140 against a measured 96–112, 25–46% high, while `heightOfRow`'s comment says the bias is meant to make corrections grow rows because a shrink is more jarring. For ordinary prose it shrinks. Unrealized rows over-reserve and snap smaller on realize — visible drift on a deep scroll-up. Out of scope here.

## MEDIUM 1 — SwiftUI reuse staleness: right conclusion, wrong mechanism

The described failure does not reproduce. `MarkdownSegments.Segment.id` is `"i:\(attachment.path)"`, so a recycled cell rebuilds the view with fresh `@State`. Demonstrated rather than argued: on unfixed code, swapping `rootView` onto a second attachment *did* fetch the second thumbnail, and `load()` bails when `image != nil`, so that fetch can only come from a view that dropped the first.

The finding still stands, for a better reason: the view's correctness was entirely its call site's property — invisible from the view and untested. Fixed as recommended, with `onChange(of: attachment)` dropping stale image and hover state plus a `generation` counter matching `TranscriptImageBlockView`'s discipline, so a late decode for the previous attachment can't paint over the current one. Mutation-checked both ways: breaking the call-site id alone now stays green (the new guard covers it), breaking both goes red.

The test asserts the fetch, not pixels — SwiftUI content isn't captured by `cacheDisplay` or `CALayer.render` in this unbundled test process (tried both; the tree renders as bare backdrop), so the drawn image can't be read back.

## MINOR — fifo test added, and the hazard is deeper than documented

Real `mkfifo` fixture, probe on a background thread behind a 5s deadline, write end opened to release a stuck opener before failing. Self-tested against a deliberate blocking `open`: failed at 5.011s with the named diagnostic, no hang.

One correction to the doc comment's premise: removing the `.typeRegular` check *alone* does not reach a blocking open, because a fifo stats as zero bytes and `stamp.size > 0` answers `.unreadable` first. The new case reds on that mutation too, so both guards are pinned — but the hang is one layer further in than the rationale implies.

## MEDIUM 2 — spec entry: declining

This landed as UI polish on an existing surface. The transcript already rendered that attachment; the change is that it renders as a thumbnail instead of an unreadable absolute path, and that clicking it does the obvious thing. The repo owner scoped it as a minor UI change when it was requested, and that classification is the call the convention leaves to a human.

The decisions worth recording — parsing the display marker because the structured `image` block carries no path, no directory allowlist because the corpus shows two distinct cache roots, header-probe sizing so measurement stays synchronous — are documented at their call sites and in the PR body rather than in `docs/specs/`.

## MINOR — markdown continuity across marker splits

Acknowledged, not fixed, agreeing with the required review's read: a correct fix needs an AST-aware splitter rather than a text-level one, and the corpus shows markers only ever appear as their own paragraph or as mid-sentence prose, never inside list or reference syntax.

PR #594, opened 2026-08-05
Compound: Teach agents that a nested run loop drains nothing a @MainActor test is waiting for

cheapsteak:
## Summary

A `@MainActor` test body **is** a block on the serial main queue, and in this test process that queue is drained by libdispatch, not CFRunLoop. So a nested `RunLoop.current.run(until:)` inside a test body cannot re-enter the queue and drains none of it. The deferred work does not vanish — it runs once the body ends, inside whatever test is running by then. The offending test **passes in isolation** and reds a *sibling*, so the reported failure names innocent code, and `.serialized` does not help because it orders test bodies, not queue work that escaped one.

This documents the rule in `Tests/CLAUDE.md` (alongside the clock/date seams and `TBD_HOME` fencing rules) and corrects a comment in `TranscriptImageAttachmentTests` that stated the opposite mechanism.

**Measured, not assumed.** A temporary probe suite run in the real test process established:

- A block enqueued with `DispatchQueue.main.async` immediately before a 0.3 s nested run loop had **still not run** when the loop returned. Suspending (`Task.sleep`) delivered it.
- A suspended `@MainActor` task was **not** resumed by a 0.4 s nested run loop — identical to a busy-wait control. Same for a task parked on a continuation resumed from a background queue.
- Both hold after `NSApplication.shared` exists.
- The escape is directly observable: inside the body the log read `["probeF body end"]`; a later test saw `["probeF body end", "deferred block"]`.
- `run(until:)` returns immediately when no source is attached — 0.000 s bare vs 0.321 s once a timer was added — so a `pump()` helper can be a silent no-op.

`TableTranscriptHarness.drainMainQueue()` already carried the correct explanation; the comment in `TranscriptImageAttachmentTests` claimed a nested loop *would* deliver the work and would resume every other suspended `@MainActor` test. Both halves are false, and that is the version a reader was most likely to trust, so it is rewritten.

## No lint rule, deliberately

Nothing mechanical can distinguish "drive AppKit layout" (correct) from "wait for a callback" (broken) at a `RunLoop.run` call site. A syntactic ban would fire on both existing `pump()` helpers in `Tests/TBDAppTests`, which use a nested loop for its one honest purpose — run-loop-scheduled work — so the rule would ship fully suppressed and get disabled. Same reasoning the repo already applies to assertion-hygiene rule 4. The rationale is recorded in the doc so it is not re-proposed.

The compiler already enforces the half that can be enforced: `RunLoop.current` and `run(until:)` are `NS_SWIFT_UNAVAILABLE_FROM_ASYNC`, so a nested loop can only appear in a synchronous body or helper.

## Survey

Two `RunLoop.current.run` sites in test code, both `pump()` helpers driving AppKit layout — `TableTranscriptHarness.swift:1707` and `WorkbenchSnapshotTests.swift:260`. Neither changed; both are the permitted use. No `CFRunLoopRun` anywhere. Hits in `Sources/` are `RunLoop.main.add(timer:)`, which is scheduling, not pumping.

**One residual for a human, not fixed here.** `settle()`'s docstring claims to drain "deferred async height corrections," and `TableTranscriptView` has five `DispatchQueue.main.async` sites. `pump()` provably cannot drain those. The harness has `drainMainQueue()` for exactly this, used at only 2 of ~25 settle/pump call sites. Worth auditing whether any assertion there passes against an effect that never fired — it needs someone who knows which of those five matter, and converting `settle()` to `async` ripples across every call site.

## Test plan

- [x] `swiftlint --strict` — 0 violations in 674 files
- [x] `scripts/test.sh --filter 'TranscriptImageAttachmentTests'` — 27 tests passed (count checked; a filter matching nothing exits green)
- [x] Probe suite removed; only the doc and the comment remain in the diff

No production code changed. Docs plus one test comment.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=9E25064A-999B-4889-9B02-FEC413AF1730)

## On the spec convention

No `docs/specs/` entry, deliberately. The repo requires one for work that revises the system's theory, and asks that a larger change without one say so here — this is that statement.

This records a measured property of the existing test harness and the convention that follows from it. It revises no theory, adds no flag, no migration, no runtime behavior. `Tests/CLAUDE.md` is largely made of exactly this kind of hard-won convention — the kill hazards, the shared-box hazards, the `.flaky` quarantine rules — none of which carry a spec reference. The one section that does, "Assertion hygiene", cites `2026-07-24-test-hardening-design.md` because that section came out of the test-hardening program, not because every rule in the file needs a spec.

There is also a hard constraint: a human answers the brainstorming questions, and an agent may not answer its own. A spec authored unilaterally here would be the failure mode that rule exists to prevent.

### DIFF
--- a/Tests/CLAUDE.md
+++ b/Tests/CLAUDE.md
@@ -422,12 +422,14 @@ that queue runs — including the deferred work the body is trying to observe.
 
 **Never spin a nested run loop to wait for something.** A nested
 `RunLoop.current.run(until:)` cannot re-enter the main queue, so it drains none
-of it. Three things were measured here, and all three hold both bare and after
-`NSApplication.shared` exists: a block enqueued with `DispatchQueue.main.async`
-immediately before a 0.3 s nested run loop had **still not run** when the loop
-returned; a suspended `@MainActor` task was **not** resumed by it; and neither
-was a task parked on a continuation resumed from a background queue — the shape
-a test awaiting a bounded-poll helper has.
+of it. Three things were measured here — each against a loop kept alive by an
+attached timer, so it genuinely spun for its full duration rather than returning
+early — and all three hold both bare and after `NSApplication.shared` exists: a
+block enqueued with `DispatchQueue.main.async` immediately before a 0.3 s nested
+run loop had **still not run** when the loop returned; a suspended `@MainActor`
+task was **not** resumed by it; and neither was a task parked on a continuation
+resumed from a background queue — the shape a test awaiting a bounded-poll
+helper has.
 
 **Suspending is the only thing that returns the queue.** `await Task.yield()`
 hands the main actor back, the queued block runs, and only then does the
@@ -457,9 +459,13 @@ input source or timer is attached to the loop — measured 0.000 s bare against
 0.321 s once a timer was added. A pump that appears to be doing something may be
 doing nothing at all, at either end of the range.
 
-The compiler already closes the worst half: `RunLoop.current` and `run(until:)`
-are `NS_SWIFT_UNAVAILABLE_FROM_ASYNC`, so a nested loop can only appear in a
-synchronous body or helper — exactly the context that can never suspend.
+The compiler already closes the worst half. `NSRunLoop.h` annotates every
+spinning method — `run`, `run(until:)`, `run(_:before:)`,
+`acceptInput(for:before:)` — plus the `current` property itself with
+`NS_SWIFT_UNAVAILABLE_FROM_ASYNC`; `main` is deliberately not annotated, so the
+methods are the enforcement point and `RunLoop.main.run(until:)` is refused
+just the same. A nested loop can therefore only appear in a synchronous body or
+helper — exactly the context that can never suspend.
 **Reaching for one from an `async` test and finding it won't compile is the
 signal to yield, not to hoist the spin into a sync helper.**
 
--- a/Tests/CLAUDE.md
+++ b/Tests/CLAUDE.md
@@ -413,6 +413,60 @@ Four rules. Each traces to a real flake — provenance kept so the rule sticks.
 
 Full rationale: `docs/specs/2026-07-24-test-hardening-design.md` §6.
 
+## `@MainActor` tests: suspend to drain, never pump
+
+A `@MainActor` test body **is** a block executing on the serial main queue, and in
+this test process that queue is drained by libdispatch's own main-thread drain,
+not by CFRunLoop. So for as long as the body runs synchronously, nothing else on
+that queue runs — including the deferred work the body is trying to observe.
+
+**Never spin a nested run loop to wait for something.** A nested
+`RunLoop.current.run(until:)` cannot re-enter the main queue, so it drains none
+of it. Measured here: a block enqueued with `DispatchQueue.main.async`
+immediately before a 0.3 s nested run loop had **still not run** when the loop
+returned, and a suspended `@MainActor` task was **not** resumed by it either.
+The same three assertions hold after `NSApplication.shared` exists.
+
+**Suspending is the only thing that returns the queue.** `await Task.yield()`
+hands the main actor back, the queued block runs, and only then does the
+continuation resume — so the test observes the effect it came to assert. Bound
+the wait (assertion-hygiene rule 3 above): poll with `Task.sleep` against an
+iteration cap so work that never arrives fails an assertion instead of hanging.
+`TableTranscriptHarness.drainMainQueue()` is the minimal shape and
+`TranscriptImageAttachmentTests.pump(_:until:)` the bounded-poll one.
+
+**The failure shape is why this rule is written down.** The deferred work does
+not vanish; it runs once the body ends, inside whatever test is running by then.
+So the offending test **passes in isolation** and reds a *sibling* — the reported
+failure names innocent code. Shared process-wide state makes it concrete: a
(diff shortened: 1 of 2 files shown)


cheapsteak:
Good catch on the dangling count — fixed in 8f364dff. Rather than downgrading "three" to "two", I named the third measurement in the doc text: a task parked on a continuation resumed from a background queue was not resumed either. That case is worth a reader having, since it is the shape a test awaiting a bounded-poll helper takes, and it is the one someone would most plausibly assume a nested run loop could service.

cheapsteak:
Both minors addressed in `7c425cc`.

**On the measurement conditions** — you read it correctly, and the ambiguity was real. Every one of those measurements was taken against a loop kept alive by an attached `Timer`, precisely so it would spin for its full duration instead of returning early; without that the numbers would prove nothing. The doc now says so, which also removes the apparent tension with the "silent no-op" paragraph below it.

**On `NS_SWIFT_UNAVAILABLE_FROM_ASYNC`** — verified against the real header, and it does carry it, contrary to the "normally reserved for blocking calls" intuition:

```
MacOSX26.2.sdk/.../NSRunLoop.h:18
@property (class, readonly, strong) NSRunLoop *currentRunLoop NS_SWIFT_UNAVAILABLE_FROM_ASYNC("currentRunLoop cannot be used from async contexts.");
```

But checking it surfaced something better than a confirmation. Line 19's `mainRunLoop` is **not** annotated, while `run`, `run(until:)`, `run(_:before:)` and `acceptInput(for:before:)` all are. So the property was never the load-bearing part — the spinning methods are, which is why `RunLoop.main.run(until:)` is refused just the same. The doc previously named `RunLoop.current` as if it were the gate, which would have let a reader conclude that routing through `RunLoop.main` escapes it. Now it names the methods. Good prompt.

cheapsteak:
Fixed in `65f4327`. Fair catch — two adjacent sections of the same document reading as contradictory is exactly the thing that costs a doc its authority, so it was worth spelling out rather than leaving to the reader.

The distinction now stated: there, the waiter needs scheduling progress *somewhere else* and yielding re-queues it ahead of the very work it waits on, so more yields make it worse. Here the work is already enqueued on the queue this body is holding, so letting go of it is the whole remedy — one handoff suffices, and the loop only bounds how long you wait for a decode that may never arrive.

Compressed to a line a reader will retain: **yield to release something you hold; never to hurry something you don't.**

cheapsteak:
Both Mediums and the Minor addressed — one implemented, one declined with reasoning.

**1. "ahead of" vs "behind" — you're right, fixed in `10f43e3`.** This was a real defect in my edit, and the sharpest finding of the review. A paragraph whose entire job is to remove an apparent contradiction must not introduce a *second* framing of the same mechanism. I also shouldn't have asserted those mechanics at all: what I measured here is main-queue drain behaviour, not cooperative-pool starvation, so the spatial claim was beyond my evidence either way. The paragraph now points at "Clock and date seams" for the mechanics and keeps only the distinction in situation. That also removes the Minor — `one handoff suffices` is gone, so the mismatch with `drainMainQueue()`'s four yields goes with it.

**2. No `docs/specs/` entry — declining, and I've stated why in the PR description**, which is what the convention actually asks for ("If a larger change genuinely needs none, say so in the PR description").

The rule scopes to work that revises the system's theory. This records a measured property of the existing harness and the convention following from it — no theory change, no flag, no migration, no runtime behaviour. The sibling sections in this same file are the better comparison than "Assertion hygiene": the kill hazards, the shared-box hazards and the `.flaky` quarantine rules all carry load-bearing prescriptions with no spec reference. "Assertion hygiene" cites `2026-07-24-test-hardening-design.md` because that section came out of the test-hardening program, not because every rule in the file needs a spec.

There's also a hard constraint pointing the same way: a human answers the brainstorming questions and an agent may not answer its own. A spec authored unilaterally by me here would be precisely the failure mode that rule exists to prevent.

cheapsteak:
Addressed in `ba7b1fe`, plus a timing note on the carried-over finding.

**The `drainMainQueue()` grouping — correct, fixed.** The two helpers were named in one breath under a sentence prescribing `Task.sleep` against an iteration cap, and only `pump(_:until:)` does that. `drainMainQueue()` is four bare yields with no condition and no sleep, so it carries no failure signal at all — work that needs longer than its yields lets the test proceed as though it arrived. That is worse than merely imprecise wording, because it is the exact silent-pass failure this whole section exists to warn about, so the doc now says so explicitly rather than just separating them.

**On the spec finding — it was already answered, 26 seconds too late to be seen.** The required review run started at `22:40:44Z`; the PR-description paragraph and my comment landed at `22:41:10Z`. So "I checked all 5 commits and all 3 author PR comments" was accurate when it was written — mine was the fourth. The rationale is in the PR description now, under **On the spec convention**, which is where the repo's rule says to put it ("If a larger change genuinely needs none, say so in the PR description"). This push should let the next pass see it.

Restating it here so it is on the thread as well: this records a measured property of the existing test harness, not a revision to the system's theory — no flag, no migration, no runtime behaviour, no compiled code. The right local comparison is the sibling sections of this same file — kill hazards, shared-box hazards, `.flaky` quarantine — all load-bearing prescriptions with no spec reference. "Assertion hygiene" cites a spec because that section came out of the test-hardening program, not because every rule in the file needs one. And the repo's brainstorming rule is explicit that a human answers the questions and an agent may not answer its own, so a spec I authored unilaterally here would be the failure mode that rule exists to prevent.

Issue #384, opened 2026-07-08
Stale terminal→tmux-pane coordinates can route daemon keystrokes to the wrong session


cheapsteak:
Discovered during PR #382's live E2E: terminal row BD3B78AC (worktree 20260512-delicious-urial) records pane %256/@256 on the shared longeye-app tmux server, but that pane is now owned by terminal 5FA7EA70 (20260625-appropriate-peafowl) — pane IDs get reused after windows die on the shared per-repo server, and the dead terminal's row was never reconciled.

Consequence: LimitResumeActuator passed every eligibility check (window exists, foreground Claude, not in copy-mode) and typed Escape+continue+Enter into the WRONG live session, waking it. Verification then watched the stale terminal's transcript/activity and reported failure. This affects both #341 limit resumes and #382 api-error auto-continues — and any future daemon-initiated send that trusts terminal.tmuxPaneID.

Proposed fixes (either/both):
1. Actuator pre-send check: read the pane's process env `TBD_TERMINAL_ID` (ps eww on pane PID) and require it to match the resume row's terminal; mismatch → classify terminalGone (cancel silently).
2. Reconcile pass: GC/park terminal rows whose pane env identity no longer matches (the same probe, run at daemon boot).

Repro evidence in session logs, 2026-07-08 ~13:17-13:22 EDT.

https://claude.ai/code/session_01Leu1sJPDm8YrELnTKvPuN3

PR #595, opened 2026-08-05
Fix: terminal.send stops reporting success into dead and misidentified panes

cheapsteak:
## What's broken

`terminal.send` reports success in two cases where nothing was delivered.

**A dead pane.** `send-keys` into a `remain-on-exit` pane exits **0**. tmux's exit
status is not a liveness signal, so the daemon typed into a corpse and told the
caller "Text sent."

**A misidentified pane.** tmux reuses pane IDs. A terminal row holding a stale
`tmuxPaneID` points at a pane now owned by a *different* live session, and the
keystrokes land there — with no error by any mechanical measure. This is issue
#384, observed in the field: row `BD3B78AC` pointed at `%256`, actually owned by
another terminal, and the auto-resume actuator passed every eligibility check and
typed into the wrong session.

## Why it happens

Neither failure is detectable from the transport's return value, so the send path
never asked. Honest answers require *deliberate consultation* of pane metadata
before typing — which is what this PR adds.

## Provenance

Not a regression — the send path has never checked. The paste mechanics were last
touched in #389 (bracketed paste for large payloads), which changed *how* bytes
are delivered but not whether the target was real. It went uncaught because both
failures are silent by construction: the only observable is a message that never
arrives, which reads as an unresponsive agent rather than a failed send. Slice 1a
(#593) made it visible by giving every act a row, and this slice makes what those
rows say true.

## What this PR does

Before anything is typed, one `list-panes` call reads `#{pane_id}`,
`#{pane_dead}`, `#{@tbd_terminal_id}` and `#{pane_start_command}` together and
classifies the target. Identity resolves from a per-pane tmux user option stamped
centrally at spawn, falling back to parsing `TBD_TERMINAL_ID` out of the pane's
start command — which every pane alive today already carries, since the env map is
inlined as an `export` prefix.

Reading the pane process's *environment* was the obvious approach and is not
possible: macOS hides process environments under SIP, so `ps eww` prints nothing
even for a process you own and just spawned. tmux's `show-environment` is
session-scoped, and TBD puts every terminal of a worktree in one session. Pane
metadata is the machine interface that remains, and it is metadata rather than
rendered text, so no screen-scraping is involved.

**Three callers consult, not one.** `terminal.send` is the generic primitive, but
it is not what typed into the wrong session in the field. `LimitResumeActuator` —
auto-resume — types Escape/continue/Enter on a timer with no user gesture, and it
reaches tmux directly rather than through the RPC handler. Its eligibility list
checked `windowExists`, which proves a window id resolves to *some* window and
never that the pane still belongs to this terminal, so a reused pane id passed
every remaining check (window alive, Claude foreground, not in copy-mode) and got
"continue" typed into it. That is the incident below, so it runs the same
consultation on the same rule.

`DeskSessionManager` is the third, and had the same shape of hole: its
`liveAgentTerminals` selected a desk terminal on window-exists plus a matching
agent kind, neither of which compares pane identity, and `nudgeDeskSession` and
`postShiftWrapUp` then pasted and submitted on a timer. What they paste is a
judge prompt, so a same-kind stranger on a recycled pane id would not merely
receive a stray keystroke — it would read and act on instructions addressed to
another session, with permissions bypassed. The consultation goes at the
`liveAgentTerminals` chokepoint, so both pastes, the judge-candidate list and
the desk's liveness probe are covered by one guard. An excluded candidate is
simply not live, so the callers' existing "no uniquely owned live terminal …
skipping" path handles it — and the judge lease gets *more* correct, since a
recycled pane id can no longer masquerade as the lease owner or pad the
candidate count into a fail-closed contention report about a rival that does not
exist.

Sends to the same terminal now serialize behind one another; two different
terminals still send concurrently. Interleaving two payloads into one composer was
a transport bug.

## Behavior change

Sends that used to report success into a dead or misidentified pane now return an
error. That is the fix, not a regression. Affected:

- **`tbd terminal send`** (hooks, scripts, operator tooling) surfaces the RPC error already.
- **Nightwatch's Python** discards exit codes inside bare `except`, so it degrades
  from silent *false* success to silent *honest* failure — strictly better, and no
  change to its happy path. Nightwatch keeps working; it is deleted at the end of
  the migration, in its own PR.
- **Auto-resume** can now end a scheduled row without typing. A vanished or dead
  pane cancels it silently as `terminal-gone`, the `windowExists` check one level
  finer. A pane answering with a *different* terminal fails the row with a message
  naming both ids — not a retry, because the row's coordinate stays stale until
  something respawns the window, so retrying would only aim at the same stranger.
- **The app**'s `AppState.sendToTerminal` now raises an error alert instead of
  logging and swallowing. Worth noting: that method currently has **no callers** —
  GUI typing goes through control mode, not this RPC — so the alert is
  correct-if-wired rather than reachable today. Deleting the dead method is
  unrelated churn and was left alone.

**Refusal requires positive disagreement, never absence.** A pane that resolves to
no identity — spawned before this change, or outside TBD — is sent to exactly as
before, with a `.debug` log line. This is what keeps the fix from regressing
existing sessions.

## Actuation vocabulary

Both checks run *before* the transport is touched, which by 1a's boundary makes
them `refused`, not `transportFailed`:

- **pane gone** → `not-found` — the coordinate names nothing.
- **pane dead** → `not-eligible` — the row and the pane object exist; the pane just
  cannot receive input, and Recreate Window already recovers that state.
- **identity mismatch** → **new reason `target-mismatch`.** Neither neighbour is
  honest about a stale coordinate: `not-found` claims the terminal row is gone (it
  is not) and `not-eligible` claims a wrong state (it is not — a healthy stranger
  answered).

**These are the first refusals reachable on `terminal.send`,** which #593
documented as dispatched-or-transport-failed only. Anything reading the actuation
record for send outcomes must now handle refusals, and `target-mismatch` is a raw
value older readers have not seen.

## What is still exposed, deliberately

**Issue #384 stays open, and this PR should not close it.** Three callers now
consult their target before acting on a stored coordinate — `terminal.send`,
`LimitResumeActuator` and `DeskSessionManager`. Three still act on an unverified
one: two destructive paths that carry a user's gesture, and the control-mode
input path, which is a different question entirely. This is the full list, so
that nobody reads this PR as the definitive account and gets a shorter one. All
three are pre-existing — none is introduced or worsened here.

### The two destructive paths

**Both can destroy a stranger, not merely talk to one.** Each runs
`gracefullyInterruptPane` — Escape, `C-c C-c`, then `kill(pid, SIGTERM)` on the
pid behind a pane id read from the terminal row — and then `respawn-window -k`,
which force-replaces whatever that window is running:

- `inPlaceSwapRespawn` in `RPCRouter+TerminalHandlers.swift`, the "Switch account"
  in-place respawn, off `oldTerminal.tmuxPaneID`/`tmuxWindowID`.
- `HibernationCoordinator.hibernateForMerge`, off `terminal.tmuxPaneID`, with a
  near-identical copy of the same helper.

If either coordinate is stale — the premise this whole PR exists to stop trusting —
the outcome is worse than #384's: not a message in a stranger's composer but a
SIGTERM to their process and a forced respawn of their window.

They are named here rather than fixed here because **the distinction is not how
dangerous the act is, it is who asked for it.** Both of these serve a *user
gesture*: someone clicked "Switch account" or merged a worktree. Declining there
fails something a human asked for, so what should happen instead — fail it
outright, mark the row, fall back to a coordinate-free path, tell the user what
to do next — is a product decision with a user-visible answer, and per CLAUDE.md
that is spec-and-flag work on a load-bearing path rather than the bug fix this PR
is. Bolting a refusal onto a destructive path to satisfy a completeness argument
is how you ship a hibernate that silently stops working.

The three sites wired here have no such question to answer. `terminal.send`
returns an ordinary error to a caller that already handles errors; auto-resume
and the Watch Desk skip an act that nobody requested, on a timer that fires
again shortly. Nothing user-facing fails, so refusal needs no design — which is
exactly why they could be wired in a bug fix and these two cannot.

### The control-mode input path — out of scope by design, not by triage

The third is different in kind, and it is listed separately so it is not read as
a smaller version of the two above. `RPCRouter+AttachHandlers.swift` takes the
attach request's pane id — the same stored `tmuxPaneID`, resolved the same way —
registers it as an input route, and `ControlModeInputRouter` then hands the
attached client's keystrokes to `PasteExecutor.paste` with no `paneSendTarget`
consultation anywhere on the path.

**This slice must not wire it.** Design §6 places control-mode input outside the
actuation record: it carries the operator's own keystrokes to the operator's own
pane, at typing latency, and interposing a `list-panes` round trip per input
frame is a decision about that subsystem's design rather than a bug fix to this
one. It is out of scope **by design**, not because it was triaged as lower
priority and deferred.

Two facts bound what is exposed in the meantime, and both are properties the
other paths lack:

- It is reachable only behind **`control_mode_enabled`, which defaults off** —
  the gate is evaluated per attach, so a pane only ever routes input this way for
  an operator who deliberately turned it on.
- **The operator is watching the pane as they type.** Control-mode input goes to
  the pane whose live rendered content is on screen in front of them, so a stale
  coordinate shows up as visibly the wrong content rather than as a silent
  misdelivery. Silence is what makes the other paths dangerous; here the human is
  the check.

**Not exposed, for the record:** `armLoginSession`'s auto-`/login` typer uses the
pane id straight off the window object it just created, not a stored coordinate, so
it has nothing stale to go wrong with.

#384 tracks the two destructive paths and closes when the last of them stops
trusting a stored coordinate. The control-mode path is not part of that
accounting — it is disclosed here, but its answer belongs to control mode's own
design rather than to #384.

## Design notes

- **No DB-driven backfill of the pane option.** Stamping panes from stored
  coordinates would write the terminal's ID onto the wrong pane whenever a
  coordinate is stale — turning the check into a rubber stamp for exactly the case
  it exists to catch.
- **A modal-prompt check is deliberately absent.** Design §12 names that third
  failure class and declines it: the sentinel detects the aftermath, and no field
  failure has shown the race itself biting.
- **No flag and no spec.** This is a bug fix restoring the send path to its
  existing theory; §12 is the normative text. No policy, no allow/deny, no payload
  inspection — the daemon still never reads message content.
- **TOCTOU is narrowed, not closed.** A pane can still die between the consult and
  the paste; that degrades to a transport error, never a false `dispatched`.

## Review found three High-severity bugs, all fixed

1. **`list-panes -t %N` lists every pane in `%N`'s *window*,** not just `%N` — the
   id merely selects the window. Reading `.first` meant a hand-split TBD tab made
   every send read a stranger's `pane_dead` and identity: a **false refusal of a
   healthy pane**, the precise regression this design forbids. The format now leads
   with `#{pane_id}` and the parse selects the line the send named.
2. **`terminal.recreateWindow`'s shell branch never planted `TBD_TERMINAL_ID`** —
   the one spawn path that didn't. Since absence never refuses, panes from that
   branch opted out of the stale-coordinate check permanently.
3. **The check reached `terminal.send` but not the actuator that caused the
   incident.** Wiring only the RPC handler left `LimitResumeActuator` — the
   component #384 actually observed typing into a stranger — with the identical
   failure mode after merge, while the PR read as closing it. Auto-resume now runs
   the same consultation; see "Two callers consult, not one" above.

## Test plan

- [x] A dead `remain-on-exit` pane errors instead of succeeding — the case tmux reports as success
- [x] A pane whose resolved identity disagrees is refused, and the error names both ids
- [x] An *unresolvable* pane still succeeds (the no-regression branch)
- [x] A healthy send emits byte-identical tmux argv to before
- [x] A split window answers for the named pane, not its neighbour (live tmux)
- [x] Two concurrent sends to one terminal serialize; two to different terminals overlap
- [x] Auto-resume refuses a pane owned by another terminal, and types nothing
- [x] Auto-resume still resumes a pane carrying no identity (the no-regression branch), and matches case-insensitively
- [x] Auto-resume cancels on a vanished or dead pane, and fails rather than types when the consultation cannot be run
- [x] The Watch Desk never pastes into a pane owned by another terminal, on either the nudge or the wrap-up path, and fires no completion notification for a wrap-up that was not posted
- [x] The Watch Desk still nudges a pane carrying no identity (the no-regression branch) and one that names its own terminal
- [x] The Watch Desk skips a candidate whose identity cannot be read at all, matching how its sibling `paneCurrentCommand` check already treats a wedged server
- [x] A stranger's pane is no longer counted as a rival judge candidate, so the desk takes its lease and gets nudged instead of failing closed on phantom contention
- [x] Each failure writes exactly one actuation row with the right reason and no `dispatched` row
- [x] `transport-failed` writes no reason
- [x] Mutation-checked: the `actuation_primitive_allowlist` rule fires on a primitive added to a non-allowlisted file; the target-check and stamping tests go red on revert. Deleting the actuator's step 1a reddens its four refusal tests and makes the mutant type "continue" into the stranger pane — #384 exactly — while the two no-regression tests stay green, as they must. Removing the desk's exclusion reddens its four refusal tests — the recorder shows the judge prompt pasted into the stranger's pane and the wrap-up's completion notification firing for a summary never posted — and leaves its two no-regression tests green
- [x] Identity resolution ignores the pane's user-authored text: instructions carrying a bare `TBD_TERMINAL_ID=` literal, and ones carrying a quoted decoy value, both still resolve the real id; a non-UUID or truncated value in the real slot resolves to nil and the send proceeds; the pane-option path is unaffected by any of it. Asserted against what `newWindowCommand` actually emits, with the decoy proven to precede the real assignment, plus one case pinning that both spawn builders still emit the anchor the resolver depends on. Mutation-checked: reverting to the unanchored substring search reddens three of them with four issues
- [x] `swiftlint --strict` — 0 violations in 675 files; no new scraping/allowlist exclusions, `.swiftlint.yml` untouched
- [x] Full suite: 5201 tests in 560 suites, all passing

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=DBCFE501-2BAF-4573-8625-304825DE975A)

### DIFF
--- a/Sources/TBDApp/AppState+Terminals.swift
+++ b/Sources/TBDApp/AppState+Terminals.swift
@@ -257,12 +257,20 @@ extension AppState {
     }
 
     /// Send text to a terminal.
+    ///
+    /// The daemon now refuses a send whose pane is gone, dead, or has been
+    /// reused by a different session, so this can fail for a reason the user
+    /// can act on ("recreate the window"). `handleConnectionError` only flips
+    /// the connected flag for disconnects, which would leave such a refusal
+    /// invisible — so it is surfaced the way this file's sibling failures are,
+    /// as an error alert carrying the daemon's own message.
     func sendToTerminal(terminalID: UUID, text: String) async {
         do {
             try await daemonClient.sendToTerminal(terminalID: terminalID, text: text)
         } catch {
             logger.error("Failed to send to terminal: \(error)")
             handleConnectionError(error)
+            showAlert("Couldn't send to terminal: \(error.localizedDescription)", isError: true)
         }
     }
 
--- a/Sources/TBDDaemon/Actuation/ActuationRow.swift
+++ b/Sources/TBDDaemon/Actuation/ActuationRow.swift
@@ -49,6 +49,16 @@ enum RefusedReason: String, Codable, Sendable, CaseIterable {
     case notEligible = "not-eligible"
     /// The same act is already running on this target.
     case inFlight = "in-flight"
+    /// The coordinate resolved to a *different* session than the one named:
+    /// the target answered with an identity that disagrees with the request's.
+    ///
+    /// Neither neighbour is honest about this. `notFound` would claim the named
+    /// target is gone, and it is not — its row is right there. `notEligible`
+    /// would claim the target is in the wrong state, and it is not — a
+    /// perfectly healthy *other* target answered. tmux reuses pane ids, so a
+    /// stale coordinate names a live stranger (issue #384); the record has to
+    /// be able to say that is what happened.
+    case targetMismatch = "target-mismatch"
 }
 
 /// The synchronous outcome of one act, as the daemon classifies it just before
--- a/Sources/TBDDaemon/Claude/LimitResumeActuator.swift
+++ b/Sources/TBDDaemon/Claude/LimitResumeActuator.swift
@@ -6,11 +6,14 @@ private let logger = Logger(subsystem: "com.tbd.daemon", category: "limitResume"
 
 /// The tmux surface the actuator needs. `TmuxManager` conforms; tests fake it.
 /// All sends go through the SAME methods `handleTerminalSend` uses — never
-/// raw tmux invocations (spec constraint).
+/// raw tmux invocations (spec constraint) — and, since this actuator types
+/// without a user gesture, behind the SAME target consultation
+/// (`paneSendTarget`) that `handleTerminalSend` runs before it types.
 public protocol ResumeSendingTmux: Sendable {
     func windowExists(server: String, windowID: String) async -> Bool
     func paneInMode(server: String, paneID: String) async throws -> Bool
     func panePID(server: String, paneID: String) async throws -> String
+    func paneSendTarget(server: String, paneID: String) async throws -> PaneSendTarget
     func sendKeys(server: String, paneID: String, text: String) async throws
     func sendKey(server: String, paneID: String, key: String) async throws
 }
@@ -263,6 +266,25 @@ public struct LimitResumeActuator: LimitResumeActuating {
             return .notEligible(.terminalGone)
         }
 
+        // 1a. Ask the pane who it is, before any key is typed into it.
+        //     `windowExists` above proves a window id resolves to SOME window;
+        //     it cannot prove the pane still belongs to THIS terminal. tmux
+        //     reuses pane ids, so a stale `tmuxPaneID` names a live stranger
+        //     that passes every remaining check — window alive, Claude
+        //     foreground, not in copy-mode — and then gets "continue" typed
+        //     into it. That is issue #384 as it was actually observed: this
+        //     actuator, not a human's `terminal.send`, was the thing typing.
+        //
+        //     Same rule as the send path: refusal requires POSITIVE
+        //     disagreement. A pane that answers with no identity is sent to
+        //     exactly as before, so panes predating the stamp do not regress.
+        switch await paneTargetVerdict(server: server, terminal: terminal) {
+        case .proceed:
+            break
+        case .notEligible(let outcome):
+            return .notEligible(outcome)
+        }
+
         // 1b. Explicit-cancel-mid-flight: the row itself can be cancelled
         //     (`cancelPending`/`cancelAllPending`) while a prior attempt's
         //     verify window is running — e.g. the user clicked "Cancel
@@ -309,6 +331,78 @@ public struct LimitResumeActuator: LimitResumeActuating {
             server: server, paneID: terminal.tmuxPaneID, terminalID: terminal.id,
             worktreeID: terminal.worktreeID,
             transcriptPath: terminal.transcriptPath, preSendTranscriptData: preSendTranscriptData))
+    }
+
+    /// What the pane consultation concluded for step 1a.
+    private enum PaneTargetVerdict {
+        case proceed
+        case notEligible(ResumeActuationOutcome)
+    }
+
+    /// Classify one `paneSendTarget` consultation into an eligibility verdict.
+    ///
+    /// The outcomes mirror how this actuator already treats the same facts:
+    /// a pane that is gone or whose process has exited is the `windowExists`
+    /// case one level finer, so it cancels silently as `.terminalGone`. A pane
+    /// that answers with a DIFFERENT terminal is not "gone" and not a
+    /// transient state to retry — the row's coordinate is stale and will stay
+    /// stale until something respawns the window, so retrying would only aim
+    /// at the same stranger again. It fails the row with a message naming both
+    /// ids, the way step 3's not-foreground check fails a row a retry cannot
+    /// help either.
+    private func paneTargetVerdict(
+        server: String, terminal: Terminal
+    ) async -> PaneTargetVerdict {
+        let target: PaneSendTarget
+        do {
+            target = try await tmux.paneSendTarget(server: server, paneID: terminal.tmuxPaneID)
+        } catch {
+            // The consultation could not be RUN (a wedged tmux tripping the
+            // subprocess timeout) — not an answer about the pane. Step 3's
+            // `panePID` treats the same wedged server as `.failed`, and typing
+            // into a pane we could not look at is exactly what this check
+            // exists to stop.
+            return .notEligible(.failed("could not verify the target pane: \(error)"))
+        }
+
+        switch target {
+        case .missing:
+            logger.info("""
+                actuate: pane \(terminal.tmuxPaneID, privacy: .public) for terminal \
+                \(terminal.id.uuidString, privacy: .public) no longer exists — cancelling
+                """)
+            return .notEligible(.terminalGone)
+        case .dead:
+            logger.info("""
+                actuate: pane \(terminal.tmuxPaneID, privacy: .public) for terminal \
+                \(terminal.id.uuidString, privacy: .public) is dead — cancelling
+                """)
+            return .notEligible(.terminalGone)
+        case .live(let paneTerminalID):
+            guard let paneTerminalID else {
+                // Absence is not disagreement: a pane spawned before TBD
+                // stamped identities carries none, and refusing on nothing
+                // would break auto-resume for every such session.
+                logger.debug("""
+                    actuate: pane \(terminal.tmuxPaneID, privacy: .public) claims no terminal \
+                    identity; proceeding without verifying it is terminal \
+                    \(terminal.id.uuidString, privacy: .public)
+                    """)
+                return .proceed
+            }
+            guard paneTerminalID.caseInsensitiveCompare(terminal.id.uuidString) != .orderedSame
+            else { return .proceed }
+            logger.warning("""
+                actuate: pane \(terminal.tmuxPaneID, privacy: .public) belongs to terminal \
+                \(paneTerminalID, privacy: .public), not \
+                \(terminal.id.uuidString, privacy: .public) — typing nothing
+                """)
+            return .notEligible(.failed("""
+                tmux pane \(terminal.tmuxPaneID) now belongs to terminal \(paneTerminalID), \
+                not the scheduled terminal \(terminal.id.uuidString) — nothing was sent \
+                (tmux reuses pane ids, so this coordinate is stale)
+                """))
+        }
     }
 
     /// The exact production send sequence (spec §Actuation 5):
--- a/Sources/TBDDaemon/Lifecycle/HibernationCoordinator.swift
+++ b/Sources/TBDDaemon/Lifecycle/HibernationCoordinator.swift
@@ -988,6 +988,12 @@ public actor HibernationCoordinator {
     /// Escape → settle → C-c C-c → settle → SIGTERM. Best-effort; the
     /// subsequent `respawn-window -k` guarantees termination. Copied from the
     /// swap path so hibernate and account-switch interrupt identically.
+    ///
+    /// **Acts on an unverified coordinate (issue #384)** — `paneID` comes from
+    /// the terminal row and tmux reuses pane ids, so a stale row points this
+    /// SIGTERM and the respawn that follows at a live stranger. See the swap
+    /// path's copy in `RPCRouter+TerminalHandlers` for why the fix is a spec
+    /// about refusal semantics rather than a consultation added here.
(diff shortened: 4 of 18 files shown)


cheapsteak:
Good finding, and correct on both halves — I verified each before acting.

**Wired the actuator (`cb940cba`).** You're right that the PR narrated #384's incident and then didn't touch the component that caused it. `checkEligibility` gains step 1a running the same `paneSendTarget` consultation on the same rule (only a positive disagreement refuses), and `ResumeSendingTmux` gains the method so the seam still holds.

Outcomes reuse the actuator's existing vocabulary rather than adding a case, since each maps onto a check it already has:
- pane gone / dead → `.terminalGone` — the `windowExists` check one level finer, cancels silently.
- identity mismatch → `.failed(msg)` naming both ids. Not a retry: the row's coordinate stays stale until something respawns the window, so rescheduling would only aim at the same stranger again. Same treatment step 3's not-foreground check already gives a row no retry can help.
- consultation unrunnable (wedged tmux) → `.failed`, matching how step 3's `panePID` already classifies the same wedged server.

Mutation-checked: deleting step 1a reddens the four refusal tests and makes the mutant type `continue` into the stranger pane — #384 exactly — while the two no-regression tests stay green, as they must.

**Pushing back on `DeskSessionManager`, and documenting it instead.** Your read of it is accurate — `leasedJudgeTerminal` → `liveAgentTerminals` is window-exists + agent-kind + lease, and a same-kind stranger pane passes it. I left it alone because that is Nightwatch, which the fleet-supervision migration deletes outright in its own PR; a guard written there is written to be thrown away, and the lease narrows the window meanwhile. Rather than let the PR imply it's covered, the description now has a **"What is still exposed, deliberately"** section saying so plainly, and states that **#384 stays open** and closes when the last direct-to-tmux typist is gone. If you'd rather have the guard than the note, say so and I'll add it — it's the same two-call-site shape.

Also corrected: the `ResumeSendingTmux` doc comment claimed all sends go through the same methods `handleTerminalSend` uses, which quietly stopped being the whole story the moment that handler started consulting first.

cheapsteak:
Verified both call sites; your description of each is exact. `gracefullyInterruptPane` really does end in `kill(pid, SIGTERM)` on a pid resolved from a stored `tmuxPaneID`, in both `inPlaceSwapRespawn` and `HibernationCoordinator.hibernateForMerge`, each followed by a `respawn-window -k`. And you're right that the omission was the actual defect: a safety PR whose coverage claim is short by two is worse than one that claims nothing.

**Taking your unblock option — disclosed, not wired — and here's the reasoning, since it's a judgement call rather than scope-dodging.**

For those two paths the consultation is the easy half; the refusal semantics are the hard half, and they're a product decision. Declining a *send* costs a message and nothing was owed to the pane. Declining a *hibernate* or an *account switch* means a user gesture fails, and what should happen next — fail it outright, mark the row, fall back to a coordinate-free path — has a user-visible answer that isn't derivable from this PR's premise. Per CLAUDE.md that's spec-and-flag work on a load-bearing path, not a bug fix. The specific failure I don't want to ship is a hibernate that starts silently refusing on panes it can't identify; that's the `auto_hibernate_enabled` shape you cited, arriving from the safety side.

So the "What is still exposed, deliberately" section is rewritten to lead with those two rather than mention them, including the part that matters most — that a stale coordinate there is **worse** than #384, since it's a SIGTERM and a forced respawn rather than a message in the wrong composer. It also now records `armLoginSession` as checked and *not* exposed, for the reason you gave, so the next reader doesn't have to re-derive it. #384 tracks all three and closes when the last one stops trusting a stored coordinate.

If you'd rather this PR carry the guard on those two anyway, I'd want a human to pick the refusal behavior first — happy to take that as a follow-up brief.

cheapsteak:
Correct, and it caught a flaw in my own reasoning rather than just a gap in the diff. **Wired it (`72ba1f60`).**

**Your factual claims check out.** `liveAgentTerminals` filters on `windowExists` plus an optional `paneCurrentCommand` kind match, and neither proves the pane still belongs to that terminal — so a same-kind stranger on a recycled pane id passes both and gets judge-prompt content pasted and submitted.

**The "arguably worse" half is right too, for a sharper reason than volume of damage.** SIGTERM+respawn on a stranger is destructive but loud. Pasting a prompt and pressing Enter in a stranger's live agent session makes that agent *act on instructions meant for another session* — which design §3 names directly: a message to an agent running with permissions bypassed is arbitrary instruction injection. Silent, and it propagates.

**Where my earlier pushback was wrong.** I declined this on "it's Nightwatch, the migration deletes it." That line does not survive contact with the actual distinction. Nightwatch is live until cutover, and more importantly the reason the two `gracefullyInterruptPane` paths stay unwired is *not* that they are lower priority — it is that they serve **user gestures**, where declining fails something a human asked for and the refusal semantics are a product decision needing their own spec. `nudgeDeskSession` and `postShiftWrapUp` fire on a timer with no user gesture, which puts them in `LimitResumeActuator`'s category, and there the semantics are trivial: skip, log, don't type. Nothing user-facing fails. I had drawn the line at "which subsystem" when it belongs at "who asked for it."

**Implementation.** The check went at the `liveAgentTerminals` chokepoint rather than at the two call sites, so all four consumers are covered by one change. Two of them argue *for* that placement: `leasedJudgeTerminal` can no longer match a lease against a row whose pane has changed hands, and its `count <= 1` fail-closed guard no longer counts a recycled pane id as a rival judge; `ensureDeskSession` correctly falls through to respawn, exactly as a failed `windowExists` already does.

The rule is identical to the other two sites, which is the part most likely to drift on a third copy: **positive disagreement excludes, absence proceeds.** A pane answering with no identity is treated exactly as before, so panes predating the stamp do not regress. An unrunnable consultation excludes, deliberately matching the sibling `try?` on `paneCurrentCommand` three lines below — skipping costs one tick of a 15-minute timer, while pasting into a pane nobody could look at is the thing the check exists to stop.

No protocol work was needed (`DeskSessionManager` depends on the concrete `TmuxManager`, whose dry-run hook defaults to "no identity", so every pre-existing fixture is untouched). `.swiftlint.yml` is unchanged and `paneSendTarget` was not added to any allowlist — it is a read-only query, and listing it would blunt the rule.

**Mutation-checked.** Neutering the guard reddens four tests — stranger-nudge, stranger-wrap-up, unreadable-identity, stranger-as-rival-candidate — with the recorder showing the judge prompt pasted into the stranger pane and the wrap-up's completion notification firing for a summary never posted. Both no-regression tests stay green either way, as they must.

Full suite 5197 tests / 560 suites passing; `swiftlint --strict` clean. The "What is still exposed, deliberately" section is rewritten — it no longer lists `DeskSessionManager`, and it now states the user-gesture distinction as the governing one rather than leaving it implicit.

Issue #586, opened 2026-08-04
terminal wake no-ops on a row that claims awake but has no tmux pane — the gone-window recovery sits behind the gate that refuses


zionts:
## Summary

`terminal.wake` decides "is there anything to wake?" purely from the DB parked flags. When the row says awake but the tmux pane is gone, wake returns the idempotent no-op — and the recovery path that *would* have fixed it is unreachable, because that recovery lives downstream of the very gate that just refused.

Net effect: a terminal that looks live in TBD, is dead in tmux, and cannot be revived from the CLI. A `--prompt` passed to that wake is silently dropped (documented behaviour on a no-op), so an autonomous caller gets neither a wake nor an error.

## Reproduction (measured on a live fleet, 2026-08-04)

Of 49 terminal rows the DB considered awake (`suspendedAt IS NULL AND hibernatedAt IS NULL`), **34 had no matching pane on their tmux server.**

Picking one of them:

```
$ tmux -L <server> capture-pane -p -t '%3'
can't find pane: %3

$ sqlite3 state.db "select tmuxPaneID, suspendedAt, hibernatedAt, activityState
                    from terminal where id='C71E3863-…';"
%3|||idle          # both parked columns NULL, presents as a live idle terminal

$ tbd terminal wake --terminal C71E3863-…
Terminal already awake (no-op).
```

The pane is gone, the row claims awake and `idle`, and wake refuses.

## Mechanism

`HibernationCoordinator.wake()` gates on the DB flag only:

```swift
// Sources/TBDDaemon/Lifecycle/HibernationCoordinator.swift:463
guard terminal.isParked else { return .notHibernated }
```

`.notHibernated` becomes `woken: false`, which the CLI prints as "Terminal already awake (no-op)".

The gone-window recovery does exist, and it is good — it recreates the window and respawns `claude --resume` (transcripts are on disk, so resume works in a fresh window):

```swift
// HibernationCoordinator.swift:685, inside wakeTmuxSection
if !claimedByOther, await tmux.windowExists(server: server, windowID: windowID) { … }
else { /* window GONE — recreate + respawn */ }
```

But `wakeTmuxSection` is only entered *after* the `isParked` guard passes. So the recovery covers "parked row, window died" and never covers "unparked row, window died" — which is the state 34/49 rows are actually in. The `wake()` docstring promises the reboot/killed-window case is handled; that promise holds only for parked rows.

Still present on `main` (`4ea9e1b`).

## Why there is no workaround from the CLI

[...]

PR #598, opened 2026-08-06
Fix: stop terminal wake claiming "already awake" without asking tmux

zionts:
Fixes the reporting half of #586.

## The invariant

**`terminal.wake` may only answer "already awake" about a session it can show is there.**

It previously answered from the database alone:

```swift
// HibernationCoordinator.swift (before)
guard terminal.isParked else { return .notHibernated }
```

`.notHibernated` becomes `woken: false`, which the CLI prints as `Terminal already awake (no-op)`. That is a claim about a live session made without ever looking at tmux — and on a fleet it is frequently false. Measured on a live fleet (#586): of 49 rows the database considered awake, **34 had no matching pane**. Because a wake no-op never delivers its `--prompt` (documented behaviour), an autonomous caller got neither a wake nor an error.

The recovery that would have fixed those rows already exists and is good — it recreates the window and respawns `claude --resume`. But it lives *downstream* of the guard that just refused, so it only ever covered parked rows. The `wake()` docstring promised the killed-window case was handled; that promise held for half the states it appeared to cover. Corrected here.

## It reuses #595's probe rather than adding a second one

This started with its own tri-state window probe. **#595** landed while it was in review and made that the wrong shape: `terminal.send` now consults `TmuxManager.paneSendTarget` before typing, and that probe already answers both halves of the question wake needs — is a process there, and is the pane still ours. Two liveness primitives in one file would drift, and the one that already carries pane identity is the one to keep. So the parallel abstraction is gone and wake asks send's question.

That is also a strictly better question. `paneSendTarget` distinguishes `.dead` — the pane object survived `remain-on-exit` but its process exited — which a window-existence check cannot see at all, and which is just as dishonest a basis for "already awake": the row is referring to a shell.

## Where to start reading

`Sources/TBDDaemon/Lifecycle/HibernationCoordinator.swift` — `classifyUnparkedWake` is the whole change, and its doc comment carries the reasoning for what it deliberately does *not* do.

## What now refuses to proceed

One behaviour change, and it has an outage shape, so stating it plainly:

**A wake aimed at an unparked row whose pane positively disagrees is now an RPC error (`terminalSessionGone`) instead of a successful `woken: false`.** It previously exited 0 and printed a benign no-op.

- **CLI** — `tbd terminal wake` now exits non-zero for that terminal. A script looping wakes over many terminals will start failing on dead ones. That is the reported harm being surfaced, not a new fault: those wakes were already reaching nothing.
- **App** — automatic focus-wakes are `userInitiated: false` and stay silent (warning-logged only), which is the existing design so a reboot doesn't fire a modal per terminal. An explicit **Wake** now surfaces a message where it previously did nothing visible.
- The error names the recovery that works: `tbd terminal conversation <id>` still reads the dead session's messages, so context can be rebuilt before closing the tab.

## Fail direction

Refusal requires **positive disagreement** — the same rule #595 established for send:

- **`.missing` / `.dead`** — no live session behind the row. Error.
- **`.live` answering a different terminal's id** — the row's coordinate went stale and now names a stranger's pane after tmux reused the id (**#384**). Error: someone else's healthy pane is not evidence this session is alive. In the same #586 snapshot, two awake rows shared pane `%17` and two shared `%10`.
- **`.live` agreeing, or carrying no identity at all** — benign no-op, unchanged.
- **The probe threw** — benign no-op, unchanged. A tmux call that merely failed proves nothing, and tmux calls fail spuriously exactly when the machine is loaded enough for the session to be alive.

That last line is the property that makes this safe, and it is tested directly.

One inherited limitation, stated rather than papered over: `paneSendTarget` maps *any* `TmuxError.commandFailed` to `.missing`, so a non-"can't find" tmux failure that still exits non-zero would read as a gone pane. Timeouts — the failure mode load actually produces — throw instead, and land on the benign no-op. #595 already accepted this tradeoff for `send`, a *mutating* verb; wake's worst case is a misleading diagnostic.

## What this deliberately does not do

**It reports; it does not repair.** The tempting one-liner — consult the probe in the `isParked` guard and respawn — is not safe, and #586 argues that correctly: respawning an unparked row makes tmux authoritative over the parked flag, and one false-negative probe then destroys a **live** session's in-flight work.

Auto-recovery stays open in #586 (items 1, 2 and 4), along with a CLI surface for `terminal.recreateWindow`, which today has only GUI callers.

## Actuation record

The new append-only log (#593) gets the distinction rather than a flattened refusal: a stale coordinate resolving to a live stranger is `targetMismatch` — the reason added for exactly this case — while a gone or exited pane is `notFound`.

## Why no spec

Judged against the repo's blast-radius convention, and I changed my mind partway: my first read was "plain bug fix", #586's was "spec-then-flag". **#586 is right about auto-recovery** — it respawns processes and can mutate live sessions, so it needs a spec and a flag, and this PR does none of it.

What is left after removing that is a reporting change: it starts no process, kills no process, and writes no state. Its worst case under a wrong probe is a misleading diagnostic, not a destroyed session. That is #586's own item 3, described there as "the cheapest honest first step".

Flagging it default-off would also invert the point: the flag-off branch is the behaviour that lies.

## Validation

- `scripts/swift-safe build` — `TBDDaemon`, `TBDApp`, `TBDCLI` all clean ✅
- `swiftlint --strict` — 0 violations in 675 files ✅
- `scripts/test.sh --filter '^TBDDaemonTests\.'` — **2481 tests in 257 suites passed** ✅ (the 1 "known issue" is the `FlakyQuarantineSelfTests` fixture for #499, which fails once by design every run)
- Commit SSH-signed ✅

Unlike #571, the suite **did** run on this host. The blocker there was the CommandLineTools toolchain having no Swift `Testing` module; the separately-installed 6.3.2 toolchain has it (`PATH=~/Library/Developer/Toolchains/swift-6.3.2-RELEASE.xctoolchain/usr/bin:$PATH`). One local-only caveat remains: `Tests/TBDAppTests/ArchiveTombstoneTests.swift` imports `XCTest`, which that toolchain lacks, so it has to be moved aside for a local run. It is untouched by this PR and CI compiles it fine.

**Mutation-checked.** Reverting the one-line gate back to `return .notHibernated` turns 5 tests red, and the failure output is the reported bug itself:

```
✘ wakeOnUnparkedRowWithNoPaneIsAnErrorNotASilentNoOp
  Expectation failed: !((resp → RPCResponse(success: true,
    result: Optional("{\"woken\":false}"), error: nil, errorCode: nil)).success)
```

## Tests added

Both arms of every new branch:

- `wakeOnUnparkedRowWithMissingPaneReportsSessionGone` — the defect.
- `wakeOnUnparkedRowWithExitedProcessReportsSessionGone` — the `remain-on-exit` shell case.
- `wakeOnUnparkedRowWhosePaneBelongsToAnotherTerminalReportsSessionGone` — pane-id reuse (#384).
- `unreadablePaneProbeFailsClosedToBenignNoOp` — the safety property: a thrown probe must never read as dead.
- `wakeOnUnparkedRowWithLivePaneStaysBenignNoOp` / `...WhosePaneAgreesStaysBenignNoOp` — no identity, and matching identity.
- `sessionGoneReportsWithoutMutatingTmuxOrTheRow` — no respawn, no recreate, row untouched.
- `parkedRowWithGoneWindowStillRecoversRatherThanReporting` — the existing parked-row recovery is unchanged.
- RPC level, where this was actually hit: `wakeOnUnparkedRowWithNoPaneIsAnErrorNotASilentNoOp`, `...WhosePaneAnswersAsAnotherTerminalIsAnError`, `...WithLivePaneStaysASuccessfulNoOp`, `...WithUnreadablePaneStaysASuccessfulNoOp`.

## Merge note

Rebased onto `1bc3faf` after #595 landed. Also overlaps **#597** mechanically in `docs/tmux-integration.md`; the code regions differ entirely (`resolvePaneTerminalID` there, the wake gate here), and #597 hardens the identity read this change now depends on.

### DIFF
--- a/Sources/TBDDaemon/Actuation/ActuationOutcome+Park.swift
+++ b/Sources/TBDDaemon/Actuation/ActuationOutcome+Park.swift
@@ -42,6 +42,15 @@ extension ActuationOutcome {
         case .inFlight: return .refused(.inFlight)
         // The terminal row is gone, or the directory its worktree named is.
         case .notFound, .worktreeMissing: return .refused(.notFound)
+        // The row claims awake but its pane disagrees. A stale coordinate that
+        // resolved to a live stranger is `targetMismatch` — the one reason that
+        // can say so; a gone or exited pane is the named target genuinely being
+        // absent.
+        case .sessionGone(_, let detail):
+            switch detail {
+            case .paneBelongsToAnotherTerminal: return .refused(.targetMismatch)
+            case .paneMissing, .processExited: return .refused(.notFound)
+            }
         // The row is there but cannot be woken as asked: nothing to resume, or
         // the profile it was pinned to no longer exists.
         case .noSessionID, .profileMissing: return .refused(.notEligible)
@@ -57,6 +66,13 @@ extension ActuationOutcome {
         case .noSessionID: return "No session ID to resume"
         case .respawnFailed(let reason): return reason
         case .worktreeMissing(let path): return "Worktree directory missing on disk: \(path)"
+        case .sessionGone(let paneID, let detail):
+            switch detail {
+            case .paneMissing: return "Pane \(paneID) is gone"
+            case .processExited: return "Pane \(paneID) survives but its process exited"
+            case .paneBelongsToAnotherTerminal(let actual):
+                return "Pane \(paneID) answers with a different terminal: \(actual)"
+            }
         case .profileMissing(let profileID):
             return "Profile no longer exists: \(profileID.uuidString)"
         }
--- a/Sources/TBDDaemon/Lifecycle/HibernationCoordinator.swift
+++ b/Sources/TBDDaemon/Lifecycle/HibernationCoordinator.swift
@@ -11,9 +11,33 @@ public enum HibernateResult: Equatable, Sendable {
     case notFound
 }
 
+/// How an unparked terminal's pane disagreed with the row that claims it is
+/// awake. Only states tmux gave a positive answer for appear here; "the probe
+/// failed" is deliberately not a case, because it is not a disagreement.
+public enum UnparkedPaneDisagreement: Equatable, Sendable {
+    /// tmux cannot find the pane — it, its window, or the whole server is gone.
+    case paneMissing
+    /// The pane object survives (`remain-on-exit`) but its process has exited,
+    /// so the row's "awake" refers to a shell, not a session.
+    case processExited
+    /// The pane is alive but answers with a DIFFERENT terminal's id: this row's
+    /// coordinate went stale and now points at a stranger's pane after tmux
+    /// reused the id (#384). Someone else's live pane is not evidence that this
+    /// session is healthy.
+    case paneBelongsToAnotherTerminal(actualTerminalID: String)
+}
+
 public enum WakeResult: Equatable, Sendable {
     case ok
     case notHibernated   // idempotent no-op: nothing to wake
+    /// The row is NOT parked — TBD believes this terminal is awake — but its
+    /// pane says otherwise, so there is no live session for an "already awake"
+    /// answer to refer to. `detail` names which way it disagreed. Distinct from
+    /// `.notHibernated` so a caller can tell a benign idempotent no-op from a
+    /// terminal that needs recovery, and so a dropped `prompt` is reported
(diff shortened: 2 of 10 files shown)


cheapsteak (approved):
🙏

PR #603, opened 2026-08-07
Design: Claude cloud sessions, and contract v2

cheapsteak:
Docs only. No Swift, no behavior change.

Lets TBD create, watch, steer, archive, and land **Claude cloud sessions** — Claude Code sessions running on Anthropic's hosted infrastructure, reachable today only from claude.ai, the mobile and Desktop apps, and `claude --cloud` in a terminal. It rides the existing remote agent backend contract rather than introducing a parallel concept, so cloud sessions become remote sessions like any other: same mirror table, adopted as worktree rows in the same tree, same health and auth machinery.

The interesting part is that this is TBD's **second** provider, and a second implementation exposed places where a contract written against one implementation is unsound.

## What's here

- **`docs/specs/2026-08-07-claude-cloud-sessions-design.md`** — the design. Four pieces shipping together: contract v2, a provider compiled into the daemon, a land bridge from remote session to local worktree, and per-repo declaration of which remotes a repository runs on.
- **`docs/remote-provider-contract.md`** — revised to v2. States v2 as it stands, with no change notes, per the repo's no-revision-history rule.
- **`docs/remote-provider-contract-changes.md`** — migration guide for provider authors. Leads with the point that matters: a v1 provider already conforms to v2 without implementing anything new. Adopting it is nevertheless a **two-field** edit to `describe`, and doing half of it loses a verb — `stop` was required at major 1 and is capability-gated at major 2, so a provider that adds `2` to `contract_versions` while leaving `stop` out of `capabilities` keeps a working verb no caller will ever invoke, and the action silently vanishes from the user's surface.
- **`docs/specs/2026-08-10-remote-sessions-in-worktree-tree-design.md`** — a merged, Approved spec, edited here. See "Why this PR edits a merged spec" below.

## One live bug this fixes, independent of the feature

v1's drift rule retires a session after two consecutive successful snapshots omit it. `RemoteSessionStore.applySnapshot` increments `missingCount` for every row it didn't see, with no notion of whether the provider could see everything. That's correct only for a provider that enumerates its own inventory, and **wrong for one that can enumerate only part of it — it tombstones live sessions.** v2 adds `complete` to the snapshot envelope; incomplete snapshots may adopt and update but never retire, and never count as refreshing freshness (which would have re-opened the mutation gate `2026-08-01-remote-stale-snapshot` closed).

A snapshot has **three** consumers, not two: since #625, `RemoteProviderManager.apply` runs `adopter.adopt(...)` *before* it stamps `lastSuccessfulSnapshotAt` and calls `markHealthy`. Adoption is explicitly permitted on `complete: false` — a session present in a partial view was positively observed — so an implementer doesn't gate it behind the freshness stamp by accident. And freshness is persisted as well as in-memory: `applySnapshot` writes a key into `tbd_meta` in the same transaction, which an incomplete snapshot must suppress too, or a partial view survives a daemon restart looking fresh.

## Why this PR edits a merged spec

Contract v2 splits `stop` into "terminate compute" and `archive`/`unarchive` into "retire from inventory." `2026-08-10-remote-sessions-in-worktree-tree-design.md` was written against v1 and says the contract "deliberately declined to add a verb for it," defining archive as `stop` + row state. Both can't stand, so this PR rewrites that section rather than leaving two committed specs in contradiction.

Archive now composes over the capabilities a provider actually declares — the `archive` verb where it has one, `stop` where that's all it has, the row alone where it has neither (session keeps running, and the UI says so). Revive gains the four cases that mirror it. That spec's Status header is also corrected: it claimed everything past the `WorktreeLocation` boundary was unimplemented, but #625 shipped adoption.

**A deliberate loosening, worth a reviewer's eye.** That section guards archive on `agentState == .working` and on a provider-reported dirty checkout, reasoning that stopping a box session can destroy unpushed work. Those guards now apply **only where archive actually ends the session** — the `stop` path always, the `archive`-verb path when the provider also declares `stop`, and never for a retire-only provider. Net effect: archiving a `claude-cloud` lane mid-task is no longer refused, because it destroys nothing and `unarchive` brings it back. Declared `stop` is the only observable TBD has for "this backend can end compute"; a provider whose `archive` tears the box down while withholding `stop` is treated as a misdeclaration rather than defended against.

## Landing converts a row in place

Since #625 an adopted lane is a worktree row and is filtered out of the remote-session row list, so Land is an action on **that row**: `remote.land` materializes the worktree and flips `WorktreeLocation` from `.remote` to `.local`. One row throughout — same id, display name, PR badge, tree position, and children — so there is no link between two rows to store anywhere.

Origin lives in `providerName`/`providerSessionID`, already on `worktree` since `v70`. Landing clears neither; they stop being a live binding and become provenance, with `location` saying which. **No migration and no schema change** — the invariant those columns obey widens from "set only alongside `location = 'remote'`" to "set whenever a row has a provider session behind it, past or present," and every existing row already satisfies the wider rule. Two readers must agree: `WorktreeStore.findRemote` already keys on the pair (which is what stops a landed session being re-adopted into a second row), and the sidebar's session-row filter keys on `location` today and moves to the pair.

**The real work is in the shared domain type, not the table.** `WorktreeLocation` is `.local | .remote(provider:sessionID:)` and `.local` carries no associated values, so a landed lane — local files, provider origin — has no representation in the model at all. The current mapping doesn't merely fail to express it, it destroys it in both directions: `WorktreeRecord.init(from:)` nils the pair whenever location is `.local`, and `toModel()` rebuilds `.remote(...)` only when the stored string is `"remote"` and both fields are present. A landed row would round-trip clean.

The design puts provenance in its own optional field alongside `location` rather than as a payload on `.local`. The two facts are orthogonal — where a lane's files are now, and which session it came from — and keeping `.local` payload-free means every existing `switch` over `WorktreeLocation` across the daemon and app compiles untouched, so only the record mapping and the model change.

`land` changes nothing on the provider, so TBD retires the session by archive's vocabulary and **never** stops it — landing is "bring this home," not a teardown, and the box may hold unpushed work.

## Other contract v2 changes

- **`archived` becomes a third Session axis**, orthogonal to liveness and attention, with `archive`/`unarchive` capabilities. Archived sessions must stay in `list` — filtering them makes them look absent to the drift rule and denies the caller the inventory a revive flow browses.
- **`stop` becomes a capability and narrows to "terminate."** v1 fused terminating compute with retiring from the inventory; one provider could do both in one call, so nothing forced them apart. A platform that reclaims idle sessions with no client-facing kill can retire but not terminate.
- **`transcript`**, distinct from `log` — agent transcript JSONL on stdout, continuation cursor as a JSON envelope on stderr. That split is the single exception to "stderr is diagnostic only" and exists so a truncated data stream stays recognizable as truncated.
- **`land`**, with the validation a caller must perform on provider-supplied `branch` and `remote_url` before either reaches git.
- The **worktree identity keys** (`tbd_worktree_id`, `tbd_parent_worktree_id`) are readable at either contract major — they ride inside `meta`, so populating them is not a v2 feature. The `SHOULD` to echo `tbd_worktree_id` now says where the value comes from: a provider never discovers it, the caller mints and supplies it, and exporting it as `TBD_WORKTREE_ID` is the provider's obligation. The `create` stdin field that would carry it is not yet in the contract, so a provider told nothing correctly omits the key.

## Security note worth reading

Part 4 defines a **committed, repo-authored** file (`.tbd-remotes.json`) that TBD parses. This is not inert data: repo-declared params are matched against the provider's `create_params`, which include `prompt` — so an unreviewed declaration could set the opening instruction of an agent holding repo write access, network egress, and the user's credential, and a single declared remote silently became the default location.

Declarations take effect only after a trust-on-first-use gate that displays every param value verbatim, with `prompt` shown in full. Approval is stored as the SHA-256 of the file's exact bytes, so **any edit re-prompts** — approving once never blanket-approves what lands on the branch next month. The authoritative copy is the registered root checkout's, not the selected worktree's, so a branch under review can't redefine where sessions run merely by being selected.

## Flags

Nothing ships enabled. `claude_cloud_enabled` is a new default-off `config` column, and it's a second gate *inside* `remoteGate()` rather than a bypass — cloud requires both flags. It stays separate from `remote_backends_enabled` because that flag was written to be deletable after soak, on the reasoning that the feature is inert without a registered provider file; a provider compiled into the daemon is never inert, so folding them would silently convert a disposable flag into a permanent one.

**Unset is a third state.** The column is added with **no** SQL default (`addColumnIfMissing` with `defaults:` omitted), so it stays genuinely NULL until somebody touches the toggle: NULL means nobody chose and resolves to the shipped default, `0`/`1` mean somebody did and are honored forever. Passing `defaults:` would make `ADD COLUMN ... DEFAULT` backfill every existing row — and since the singleton `config` row is seeded naming only two columns, a fresh install and an old one would both read `0`, indistinguishable from a deliberate opt-out. That distinction is destroyed on write and unrecoverable later, which is why it's a decision at migration time rather than something to fix at graduation. The shipped default therefore lives in exactly one place, `ConfigRecord.toModel()`'s `?? Config.claudeCloudEnabledDefault`, and graduation is a one-line edit to that constant. Three states, three assertions: a pre-migration row reads NULL rather than `0`, a NULL row follows a change to the constant, and an explicit `false` survives one.

Resolution also skips unusable declared remotes **within** the repo-declaration tier rather than falling out of it: a repository that lists two remotes has said both are appropriate for it, so passing over an unregistered or flag-disabled entry offers the next declared remote instead of dropping to a global default that knows nothing about this repository.

## Brainstorming

Ran per the repo convention, with a human answering. The spec was then reviewed by a fresh-context agent that verified its codebase claims against source and found three design faults — the prompt-injection channel above, an attach shim whose exit-code translation had no lawful input, and a ledger union that would have made "TBD launched it once" mean "it exists permanently."

The two decisions this PR's final pass turned on — that v2's verbs supersede compose-from-`stop`, and that Land converts a row in place rather than creating a linked second row — were also put to a human rather than assumed.

[☁️ Claude Cloud Sessions](https://cheapsteak.github.io/tbd/open/?worktree=BBA22690-7E4E-4F7F-A4F8-A71351FA5FD5)


cheapsteak:
## Handoff — for a session running on a Mac

This branch was driven from a Claude Code cloud session, which is Linux. **No Swift was written, deliberately.** `scripts/swift-safe build` and `scripts/test.sh` cannot run there — `TBDApp` is SwiftUI/AppKit, `TBDDaemon` reaches `Security`, and every target imports `os` — so anything compiled-looking would have been unverified by construction, against the repo's own rule. The design and both contract documents are done and reviewed; the implementation is not started.

The implementation plan lived in `docs/plans/`, which is gitignored, so it doesn't travel. It's reproduced below — that's the point of this comment.

### Two prerequisites

**P1 — a machine that can build.** Everything below needs a real build and test run.

**P2 — capture the undocumented endpoints.** Phases 7 and 8 call claude.ai endpoints nobody has written down. Someone with the account has to observe what the official clients do and record, for discovery, transcript fetch, and archive/unarchive: method, path, auth header shape, request and response bodies, pagination, and how the session status field enumerates. Phases 1–6 don't need this and are most of the feature.

**One assumption was granted rather than tested:** that `claude --cloud <id>` writes an ordinary transcript JSONL locally. It's plausible — `--no-session-persistence` is documented as working only with `--print`, implying interactive sessions always persist — but unverified. Phase 8.3 carries a test that fails loudly if it's wrong, rather than degrading silently. **Test it early**: attach to a cloud session and look for a new JSONL under `~/.claude/projects/`. If it doesn't appear, the `transcript` verb becomes the only path and Phase 8 gets bigger.

### Phases

Test-first throughout; each task is a standalone commit that builds and passes. 1 gates everything; 2 gates 4–7; 3, 5 and 6 are independent of each other.

**1 — Contract v2 plumbing.** Negotiated version becomes provider state: `ProviderRunner.run` hardcodes `TBD_CONTRACT_VERSION=1`, `describeProvider` hard-requires `1`, and `RemoteProviderInvoking.run(_:verb:stdin:timeout:)` has no parameter to carry one — so this is a protocol change, not a free ride. Then `archived` and `complete` on the `TBDShared` types; completeness gating the drift rule in `RemoteSessionStore.applySnapshot`; completeness gating freshness in `RemoteProviderManager.apply` (it currently calls `markHealthy` and stamps `lastSuccessfulSnapshotAt` unconditionally); events parity in `RemoteEventParser`.

**2 — Migrations**, each following the shared-model rule in one commit: `claude_cloud_session` (the ledger, with idempotency key state), `config.claude_cloud_enabled` default false, `repo.remotes_declaration_trusted_sha` nullable.

**3 — Attach.** `RemoteAttachExitClass` folds `.permanent`, `.contractBug` and `.transient` into `.unexpected`, which arms auto-reconnect — split `permanent` out and teach `RemoteReconnectPolicy` not to arm on it. Add a cached, fail-open eligibility preflight. Add stop capability gating to `RemoteSessionActionMenu` (attach is gated, stop isn't).

**4 — Land bridge.** Validation as a pure function first: ref-name pattern on `branch`, reject a leading `-`, compare `remote_url` rather than pass it to git (the `ext::` family executes commands). Then `remote.land`, preconditions checked before anything is created. `git worktree add <path> <branch>` refuses a branch already checked out, so a second landing derives `<branch>-2` from the same commit. Provenance recorded both ways; neither side retired.

**5 — Repo-declared remotes.** Parse `.tbd-remotes.json` from the registered root checkout, not the selected worktree. Keep only `provider`, `label`, `params`; drop params absent from `create_params`. Trust gate on the canonicalized SHA-256. Resolution order: explicit choice, approved declaration, global default, local.

**6 — Built-in provider, documented half.** Dispatcher routing `RemoteProviderInvoking` by name. Reserved-name handling — note `RemoteProviderRegistry.load` currently throws for the *whole file* on a duplicate and both callers swallow it, one with `try?`, so one bad entry silently removes every provider; skip the entry and surface it instead. Then `describe`/`create`/`send`/`attach` shelling out to `claude` behind an injected process seam, ledger writes, and ledger-only listing.

**7 — Built-in provider, undocumented half.** *(needs P2)* Client behind a protocol so no test hits the network; discovery; the three union rules; `archive`/`unarchive`; the preflight backing 3.2.

**8 — Transcript.** *(8.1–8.3 don't need P2)* `TBDConstants` helper for `~/tbd/remote-transcripts/<provider>/<sessionID>/` honoring `TBD_HOME`; widen the transcript RPC path guard to admit that root as a *second* permitted root, never an unguarded path; point attach at it so a cloud conversation doesn't surface as a local session via `ClaudeSessionScanner`; then the verb itself; then suppress local file linking for remote rows.

### Judgment calls a human may want to revisit

Three things were decided without explicit sign-off and are cheap to change:

- **Presence in an incomplete snapshot resets accrued absences.** The spec says only that an incomplete snapshot must not *increment* `missingCount`. Freezing the count entirely would tombstone a session observed moments earlier, so presence is treated as real evidence. One bullet in Identity & drift if you disagree.
- **New verb timeouts** — 30s, 60s for `transcript` — follow the existing shape of session-mutating versus bulk verbs. The spec states none.
- **Re-prompting on any declaration edit.** Keyed to the content hash, so a legitimate edit costs a re-approval. That's what makes the trust gate actually hold, but it's the chattiest reading.

### Conventions this branch is already following

Verify with `scripts/swift-safe build` before each commit and `scripts/test.sh` for anything touching daemon or shared code. Both flags need both branches tested. New delays take an injected clock. Plans stay in `docs/plans/` and are never committed.

---

cheapsteak:
All four findings addressed in `7419713`. The HIGH one was correct and points at something in shipped code that outlives this PR.

**HIGH — transcript guard.** Verified against source; the spec was wrong. Only `handleSessionMessages` constrains the path to the Claude projects store. `handleTerminalTranscript` reads a terminal row's `transcriptPath` verbatim, and the only validation that field ever gets is an *absoluteness* check at write time in the sessionEvent handler.

The root cause is worth recording: the guarded call site's own comment says it enforces "the same trust boundary as `handleTerminalTranscript`." That claim is false, and it's where the spec's claim came from. A comment asserting a symmetry that doesn't exist is how the gap stayed invisible.

The spec now says `handleTerminalTranscript` **gains** a boundary check rather than having one widened, and that closing the asymmetry is a prerequisite of this work rather than a side effect — a second permitted root is only safe where a boundary is actually checked.

**MEDIUM — pending-create window.** Named at ten minutes, with the trade-off stated rather than just the number: too short strands a slow-provisioning session outside the inventory, since capacity is allocated on demand and a setup script runs before the session is usable; too long leaves a genuinely failed create sitting as `pending` with nothing to act on. Written as a threshold to revise on evidence, not a constant to defend. The test bullet now also specifies it runs on an injected clock rather than elapsed wall time.

**Minor — ref-name grammar.** Fair, and the reason given is the right one: third-party providers have to conform to it, so leaving it unspecified made the protection unreviewable from the documents. The normative contract now spells out the grammar under `land` — ASCII letters, digits, `.`, `_`, `-`, `/`, with the usual git ref exclusions — as a strict subset of `git check-ref-format --branch`, so a provider can predict acceptance without matching one git version's behavior. The migration guide points at it instead of restating it.

**Minor — call-site miscount.** Fixed: three call sites, two swallowing with `try?`.

---

One thing this surfaces that is **out of scope here and needs a decision separately.** `handleTerminalTranscript`'s missing boundary check exists today, independent of this feature. `transcriptPath` is written from a `sessionEvent` RPC and only checked for absoluteness, so a crafted event can point a terminal row at an arbitrary absolute file, which that RPC will then read and return parsed to the app. It is read-only and needs the ability to send a `sessionEvent`, so it isn't dramatic — but it is the same shape as the pane-identity forging fixed in `9ba2f3d`, and the misleading comment means nobody would find it by reading.

I haven't touched it: this branch is docs-only, and the session that produced it runs on Linux where TBD cannot be compiled or tested, so any fix would be unverified. Flagging it for a decision — fold it into the implementation work as Phase 8's prerequisite, or take it separately as a fix that stands on its own merits.

---

cheapsteak:
`claude-review` is red on `2bce6e6` for an infrastructure reason, not a verdict. Recording it because it has now happened twice on the same commit and it is not fixable from this branch.

Both failures are `validate.py` failing closed because the orchestrator finished without all specialists having written their findings files.

First attempt — neither specialist reported:

```
error: no specialist findings files match 'findings-*.json'
error: expected specialist(s) produced no findings file: correctness, conventions
       — that review lens never ran (orchestrator may have merged before all
       specialists completed); failing closed
```

After re-running failed jobs, one of the two got through:

```
error: expected specialist(s) produced no findings file: correctness
       — that review lens never ran ...
ok: findings-conventions.json (0 finding(s))
```

Two things follow. **The conventions lens returned zero findings on this head**, so the four items from the previous round are resolved as far as that specialist is concerned; only correctness has yet to report. And the partial improvement between attempts — nothing, then one of two — is consistent with the race the error message itself names, rather than a deterministic failure.

Worth noting the same diagnostics block in all three preceding reviews reported the orchestrator's `ScheduleWakeup` call failing with a missing `prompt` parameter and falling back to a blocking `Bash` wait for the specialists' output files. A wait mechanism already running on its fallback path is a plausible home for an orchestrator that stops waiting early.

**Why this can't be fixed here.** The job does `rm -rf .github/workflows/claude-review-v2` and `git checkout $BASE_SHA -- .github/workflows/claude-review-v2` before validation, so the review scripts always come from the base branch. A change pushed to this branch would not be used by this PR's own review, which makes fixing it here both out of scope and ineffective for unblocking this PR.

I have re-run the failed jobs a second time. If it goes red again for the same reason, this needs a human — either an orchestrator fix on `main` or an admin merge — and I will not keep re-running past that.

---

cheapsteak:
Stopping here and handing this over. `claude-review` has now failed twice on `501e5be` with the same infrastructure fault as the earlier episode — neither specialist produced a findings file, so `validate.py` failed closed with no verdict:

```
error: no specialist findings files match 'findings-*.json'
error: expected specialist(s) produced no findings file: correctness, conventions
       — that review lens never ran (orchestrator may have merged before all
       specialists completed); failing closed
```

I said I would not keep re-running past a repeat, so I am not queueing another.

**Where the review actually stands.** The last run that produced a verdict reviewed `1866472` and returned exactly one MEDIUM — the 24-hour retention window justified by a clause about *exited* sessions when the case it serves is a session that appears *alive*. That is fixed in `501e5be`, which is the head that has not been successfully reviewed. Findings have narrowed monotonically across the cycle: HIGH design faults, then internal inconsistencies, then single stale sentences, then one unsupported rationale. The conventions lens has returned zero findings on four consecutive heads.

**Rate.** Across this branch: two failures at `2bce6e6`, then five clean runs, now two at `501e5be`. Roughly one run in three loses specialist output, in bursts rather than uniformly.

**The most useful evidence for whoever fixes this** is in the run that *worked*, at `1866472`. Its own diagnostics reported that `ListAgents` showed the correctness specialist still `running` for several minutes after its findings file was already written to disk, and that the orchestrator therefore waited on completion notifications rather than file presence. That is the mechanism working correctly and saying out loud that file-presence is not a safe merge signal. A run that merges on the other path would produce exactly what is seen here.

This can only be fixed on `main`: the job does `rm -rf .github/workflows/claude-review-v2` and restores it from the base SHA before validating, so nothing on this branch can change its own review.

**Separately**, `test` also went red once on this head, and it is unrelated. The branch changes three files, all under `docs/`, and no test reads anything from `docs/` — one issue in 2678 tests after the suite's own retry pass. I re-ran it once; that is a normal flake rather than a gate defect.

---

PR #608, opened 2026-08-10
fix: preserve launch PATH for tmux

Jephuff:
## Summary

- capture the installation shell's exact PATH in the generated TBD.app bundle
- pass that PATH to the initial LaunchServices open and preserve it for later relaunches
- resolve tmux only from inherited PATH in both the app and daemon
- remove daemon PATH augmentation and fixed package-manager search locations

## Root cause

The restart script's initial open and LaunchServices crash relaunch used different environments. A relaunch could therefore lose the PATH entry containing tmux, making every new terminal fail until the app was restarted from a shell.

## Design trade-off

The captured installation PATH is the daemon-wide inherited environment, not only the tmux search path. TBDDaemon inherits it from TBDApp, and every daemon descendant inherits it in turn. GitManager still launches the `git` executable at its pre-existing fixed `/usr/bin/git` path; helpers that git invokes, such as `git-lfs`, depend on the inherited PATH. Tmux has an explicit saved-executable fallback; other tools do not. Restoring fixed package-manager directory augmentation would reintroduce a hidden, machine-specific search policy for every subprocess, so this PR intentionally removes that behavior.

## Verification

- `bash scripts/restart-environment-lib.test.sh` — 9/9 passed; verifies plist-generation helper behavior, not an OS-level crash/LaunchServices relaunch end to end
- startup diagnostics report the resolved tmux path and whether it came from PATH or the saved fallback, or report unavailable, once per AppState lifetime; this is field observability for the OS-level automation gap, not a claim of end-to-end crash-relaunch coverage
- app resolver and bridge suites — 20 tests passed
- daemon PATH-resolution suite — 4 tests passed
- `scripts/swift-safe build`
- `swiftlint --strict`

## Known baseline failures

The fenced full suite was run and reported nine issues, including one known quarantined issue. The remaining assertions reproduce in unchanged Markdown URL-policy, scratchpad base-path, and transcript-estimator tests.

## Scope

PATH is the sole authority for tmux discovery before its explicit saved fallback. This PR does not run a login shell, invoke path_helper, persist a separate environment file, or search standard package-manager locations.


cheapsteak (approved):
🙏 🙏 🙏 (Did you restart your computer? XD )

Jephuff:
Noo, the other two PRs caused random crashes, and when that happened it would restore without tmux :D I guess a restart would be the same!

PR #610, opened 2026-08-11
fix: resume incomplete Claude review sessions

Jephuff:
## Summary

- resume the same Claude review session when the action exits before `review-result.json` is parseable
- reset the bounded Stop-hook counter before each process-level resume and require real on-disk specialist findings
- preserve fail-closed specialist completeness, verdict validation, and the base-branch security re-restore
- add deterministic workflow-shell regression coverage and update the gate documentation

## Root cause

The Claude action stops consuming its SDK stream at the first result message. After the Stop hook exhausted its five rapid blocks, the action could return success while background review specialists were still running, leaving neither specialist findings nor a merged result for the validator.

This uses the design's existing recovery rung instead of increasing the hook retry count: at most two same-session CLI resumes, with bounded exhaustion falling through to the unchanged fail-closed validator. No placeholder findings or result are created.

## Verification

- `python3 -m pytest .github/workflows/claude-review-v2/tests/ -q` — 223 passed, 3 skipped
- `swiftlint --strict` — 0 violations
- `scripts/swift-safe build` — passed
- `bash -n .github/workflows/claude-review-v2/hooks/stop-hook.sh` — passed
- `git diff --check origin/main...HEAD` — passed
- independent spec and production-quality reviews — passed

The three skipped tests are the existing real-Claude stub-API scenarios; no `claude` binary is installed locally.

## Merge-gate note

Because this workflow runs on `pull_request_target`, this PR is reviewed by the unfixed workflow from the base branch. If that existing gate hits the same lifecycle defect, the documented admin-merge recovery path is required; the repair can exercise itself only after it lands on `main`.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -400,10 +400,11 @@ jobs:
           # as outdated (spec §3.5), which is the opposite of editing a single
           # comment in place. It selects its own comments by its own sentinel.
           #
-          # Installs a Stop hook that refuses to end the session until
-          # review-result.json exists and parses as JSON. Unlike v1 it does NOT
-          # gate on a verdict token — in v2 the model never types the verdict;
-          # validate.py computes it from the findings' severities.
+          # Installs a bounded Stop hook that blocks repeated stop attempts while
+          # review-result.json is absent or malformed. Unlike v1 it does NOT gate
+          # on a verdict token — in v2 the model never types the verdict;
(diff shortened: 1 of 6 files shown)


Jephuff:
Closing because PR #612 superseded this recovery approach by fixing the specialist lifecycle at its source. Run 31493340427 confirmed both specialist files and review-result.json were produced; its separate failure was schema rejection of unsupported failure_scenario fields in findings-correctness.json.

PR #611, opened 2026-08-11
Fix: TBDApp leaked a zombie process on every terminal teardown (67 and climbing from one launch)

cheapsteak:
## Summary

TBDApp accumulated permanently-unreaped `<defunct>` children — 67 under a single app launch, growing for as long as the app ran. Each one holds a PID table entry; on a machine already carrying ~1000 processes this is roughly 700/day from one launch, and it made an unrelated load-average investigation much harder to read.

Every zombie's parent was TBDApp, and TBDApp's live children are exactly the `tmux -u -L tbd-<id> attach -t tbd-view-<id>` PTY clients spawned by `TerminalPanelView`. The leak is one zombie per terminal-panel teardown.

## Root cause

SwiftTerm's `LocalProcess` (pinned at `dae32cc`) calls `waitpid` from exactly one place: the `DispatchSourceProcess(.exit)` handler installed by `startProcessWithForkpty`. Both of its teardown paths cancel that source *before* the child has actually exited:

- `deinit` cancels the monitor, then closes the master fd — the child only notices the resulting `SIGHUP` afterwards.
- `terminate()` closes the fd, sends `SIGTERM`, then cancels the monitor via `childStopped()` — signal delivery and process teardown are asynchronous.

Either way the `.exit` event never fires, `waitpid` is never called, and the child stays `<defunct>` under TBDApp for the life of the app. Only children that exit *while the `LocalProcess` is still alive and monitoring* are reaped — that path fires the `processTerminated` delegate.

The `posix_spawn`/Subprocess path in `LocalProcess` is `#if false` dead code at this revision and is not involved.

## Evidence

The reported "one zombie every ~2 minutes" turned out to be an averaging artifact. Exact spawn timestamps read off the zombies themselves (`ps -o lstart` still reports for a defunct process) are bursty, not periodic: 66 inter-spawn gaps with a median of 39s and a max of 13,800s, including a 3h50m stretch with zero new zombies. Twelve bursts of 1–12, separated by 5–230 min idle gaps. No poller produces that shape; the rate tracks user activity.

Sampling a fresh, unmodified `/Applications/TBD.app` every 10s for 30 minutes gives a 1:1 correlation between panel teardown and zombie creation:

- live children 10 → 11 (new panel), zombies 0 → 0
- live children 11 → 10 (teardown), zombies 0 → 1
- live children 10 → 11 (new panel), zombies 1 → 1
- live children 11 → 10 (teardown), zombies 1 → 2
- live children 10 → 9 (teardown), zombies 2 → 3

Three teardowns, three zombies; two spawns, zero zombies.

The mechanism was also reproduced standalone: three children with their exit monitors cancelled before exit produced 3 zombies; the identical code leaving the monitors active produced 0.

## How it got broken

Not a regression — the flaw has been present since `8ad79824` ("add SwiftTerm terminal panel with tmux bridge integration") first spawned a `LocalProcess`, and it lives in the pinned dependency rather than in TBD's own code. It went unnoticed because a zombie is silent: no error, no crash, no log line, no user-visible symptom. It only becomes legible at this repo's own scale — a long-running app plus heavy panel churn across many worktrees — where it surfaced as ~700 PID-table entries per day from a single launch while an unrelated load-average investigation was underway.

## Test plan

- [x] `scripts/test.sh --filter 'ChildReaperTests|TerminalTeardownReapTests'` — 10 tests in 2 suites pass
- [x] Mutation check: commenting out both `ChildReaper.reap(...)` call sites turns the 2 wiring tests RED (`an unreaped zombie (exited, still waitable)` — the exact field symptom); restored and re-verified green
- [x] Mutation check: `waitpid(..., 0)` to `WNOHANG` reds the 3 reaping tests; dropping the gate reds the 2 gate tests
- [x] `scripts/test.sh --skip '^(TBDDaemonTests|TBDDaemonLiveTests)\.'` — 2688 tests, one unrelated known FileWatcher-family flake (green on rerun, untouched by this diff)
- [x] `scripts/swift-safe build` clean; `swiftlint --strict` 0 violations in 684 files
- [ ] Post-merge: after installing a build containing this change, confirm the zombie count stays flat across terminal-panel teardowns:
      `ps -Ao state,pid,ppid | awk '$3==<TBDApp pid> && substr($1,1,1)=="Z"' | wc -l`

## The fix

`ChildReaper.reap(pid:)` performs a blocking `waitpid` on a background queue, called from `Coordinator.cleanup()` in both `TerminalPanelView` (local tmux attach) and `LocalPTYTerminalView` (remote-provider attach). A blocking `waitpid` is the deterministic instrument here: it reaps whether the child is still running, already a zombie, or already reaped (`ECHILD`). Registering a second `DispatchSourceProcess` would not be — arming one against a pid that has already exited is not guaranteed to deliver `NOTE_EXIT`, which is precisely the failure being fixed.

**How the child is asked to die is unchanged.** No `terminate()` was added to `TerminalPanelView`, and no SIGHUP/SIGTERM behavior was altered. The fix only guarantees the `waitpid`.

Two supporting details:

- `cleanup()` now sets `localProcess = nil`, so `LocalProcess.deinit` runs at a known moment and cancels `childMonitor`. This is what makes `ChildReaper` the sole waiter rather than racing SwiftTerm's still-armed monitor. Previously release depended on ARC deallocating the whole Coordinator after `dismantleNSView` — an unstated SwiftUI timing contract.
- A `ChildExitObservation` flag, set in `processTerminated(_:exitCode:)`, suppresses the reap when `LocalProcess` already reaped the pid. Without it a freed pid could be recycled onto another TBDApp child and have its exit status stolen. The check is re-read on the reaper thread immediately before `waitpid`, not just before the dispatch.

`TerminalTeardownReapTests` drives both real coordinators' `cleanup()` against a real `LocalProcess` and a real forked child — it is the suite that pins the wiring, since `ChildReaperTests` alone would stay green if the `reap` call were deleted. Constructing the coordinators headlessly required widening `localProcess` from `private` to internal in both, documented as the test seam.

## Not done here

A child that ignores `SIGHUP` and never exits would park a reaper thread for the life of the app. That is pre-existing and bounded in practice by `tmux kill-session` and `SIGTERM`; a bounded wait with `SIGKILL` escalation would need an injected clock per CLAUDE.md and is left as follow-up rather than widened into this bug fix.

Two adjacent findings surfaced while measuring, both pre-existing and both left alone:

- `deinit`'s `io?.close()` is a *graceful* close and `startProcess` leaves a read outstanding, so the cleanup handler that closes the master fd does not run while that read is pending. A silent child therefore never receives the SIGHUP at all — measured: a `sleep 120` child was still running 5s after `cleanup()`. What actually ends the attach client in production is `tmuxBridge.cleanupSession`'s `tmux kill-session`. Orthogonal to the zombie leak, but the same code path.
- `LocalPTYTerminalView.processTerminated` calls `terminate()` after `LocalProcess` has already reaped the pid, so its `kill(shellPid, SIGTERM)` could in principle signal a recycled pid.

No spec and no feature flag: this restores existing intended behavior — `waitpid` harvests an exit status, it does not kill anything, and teardown itself is unchanged.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=E340325E-526C-43C4-A608-D5573274021C)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -28,6 +28,15 @@ let package = Package(
         // property (PR #531) hasn't shipped in a tagged release yet. Switch back
         // to `from: "1.14.0"` (or whichever) once SwiftTerm cuts a release that
         // includes commit dae32cc.
+        //
+        // Re-verify `Sources/TBDApp/Terminal/ChildReaper.swift` on any bump:
+        // it exists because at this revision `LocalProcess` calls `waitpid`
+        // only from its `DispatchSourceProcess` (`.exit`) handler, and both
+        // `deinit` and `terminate()` cancel that source before the child
+        // actually exits — so TBD must reap the PTY child itself. If a later
+        // revision reaps its own children, `ChildReaper` becomes a competing
+        // waiter on a pid the OS may have recycled, which is worse than the
+        // leak it fixes. Its doc comment names the exact lines to re-read.
         .package(url: "https://github.com/migueldeicaza/SwiftTerm", revision: "dae32cc8f9bcda15713f4091bb2ea7e11f6dd57c"),
         .package(url: "https://github.com/raspu/Highlightr", from: "2.2.1"),
         .package(url: "https://github.com/siteline/swiftui-introspect", from: "1.0.0"),
--- a/Sources/TBDApp/Remote/LocalPTYTerminalView.swift
+++ b/Sources/TBDApp/Remote/LocalPTYTerminalView.swift
@@ -89,9 +89,11 @@ struct LocalPTYTerminalRepresentable: NSViewRepresentable {
         /// Recorded when `LocalProcess`'s own exit monitor fires — by then it
         /// has already called `waitpid`, so `cleanup()` must NOT reap this pid
         /// (it is free, and could have been recycled for another child of this
-        /// process). Both sides run on the main queue with the pinned SwiftTerm
-        /// revision; see `ChildExitObservation` for why it is lock-guarded
-        /// anyway and for the reaper thread that also reads it.
+        /// process). Both sides run on the main queue in practice, but unlike
+        /// the panel coordinator's `@MainActor cleanup()` that rests on
+        /// `dismantleNSView` convention rather than on isolation — see
+        /// `ChildExitObservation`, which spells out the difference and why the
+        /// flag is lock-guarded regardless.
         private let childExitObservation = ChildExitObservation()
 
         @MainActor
--- a/Sources/TBDApp/Remote/LocalPTYTerminalView.swift
+++ b/Sources/TBDApp/Remote/LocalPTYTerminalView.swift
@@ -125,8 +125,9 @@ struct LocalPTYTerminalRepresentable: NSViewRepresentable {
 
         /// `@MainActor` like the panel coordinator's sibling: SwiftUI already
         /// calls `dismantleNSView` on the main thread, and annotating it turns
-        /// that convention into a compiler-enforced guarantee. `ChildReaper`'s
-        /// ordering fix depends on this teardown running on the main queue.
+        /// that convention into a compiler-enforced guarantee — which is what
+        /// lets `ChildExitObservation` describe its writer and reader sides as
+        /// serialized rather than merely conventional.
         @MainActor
         func cleanup() {
             // Idempotent: a second call must not re-run the teardown (it would
--- a/Sources/TBDApp/Remote/LocalPTYTerminalView.swift
+++ b/Sources/TBDApp/Remote/LocalPTYTerminalView.swift
@@ -79,9 +79,20 @@ struct LocalPTYTerminalRepresentable: NSViewRepresentable {
     final class Coordinator: NSObject, TerminalViewDelegate, LocalProcessDelegate, @unchecked Sendable {
         weak var terminalView: TerminalView?
         var onExit: ((Int32?) -> Void)?
-        private var localProcess: LocalProcess?
+        /// Internal rather than private so `TerminalTeardownReapTests` can hand
+        /// this coordinator a real `LocalProcess` and drive `cleanup()`
+        /// headlessly — the reap wiring is otherwise unreachable from a test,
+        /// since `start()` needs a live `TerminalView`.
+        var localProcess: LocalProcess?
         private var started = false
         private var tornDown = false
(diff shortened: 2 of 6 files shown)


cheapsteak:
## Verifying the SwiftTerm premises the review environment could not check

The `correctness` specialist's only finding on `3d4207f6` was a disclosed limitation rather than a defect: no SwiftTerm checkout exists in the review sandbox (no `.build/checkouts`, no SwiftPM cache), so the load-bearing premises about the pinned revision were taken on this PR's word. That is a fair objection, so here is the evidence, reproducible by anyone with the package resolved.

Source of truth — the exact pinned revision from `Package.swift`:

```
git -C ~/Library/Caches/org.swift.swiftpm/repositories/SwiftTerm-<hash> \
    show dae32cc8f9bcda15713f4091bb2ea7e11f6dd57c:Sources/SwiftTerm/LocalProcess.swift
```

**(a) `LocalProcess` calls `waitpid` from exactly one place.** `grep -c waitpid` over that file returns **1**, at line 314, inside `processTerminated()`:

```swift
func processTerminated ()
{
    var n: Int32 = 0
    waitpid (shellPid, &n, WNOHANG)
    delegate?.processTerminated(self, exitCode: n)
    childStopped()
}
```

**(c) `deinit` never calls `waitpid`.** Its entire body:

```swift
deinit {
#if os(macOS)
    childMonitor?.cancel()
    childMonitor = nil
#endif
    io?.close()
    io = nil
}
```

It cancels the exit source and closes the `DispatchIO` — nothing else. This is the leak: `cancel()` before the child has exited means the `.exit` event never fires, so the single `waitpid` above never runs.

**(d) `waitpid` precedes the delegate call**, as shown in (a). The specialist correctly noted this one is not load-bearing — the sole-waiter argument only needs `record()` and the exit-source firing to share one indivisible main-queue block, which they do, since our delegate calls `record()` synchronously as its first statement.

Two caveats worth keeping honest:

- The active spawn path is `startProcessWithForkpty`. The `posix_spawn`/Subprocess path in the same file is `#if false` dead code at this revision, so its `Subprocess.run` completion handler is not a second reaper.
- This establishes the premises **today**, at the pinned revision. It does not make them permanent, which is why `Package.swift` carries a re-verify note on the pin.

PR #613, opened 2026-08-11
Refactor: Fence local-only subsystems from worktrees whose files aren't on this disk

cheapsteak:
## Summary

Groundwork for [remote agent sessions as worktrees in the tree](docs/specs/2026-08-10-remote-sessions-in-worktree-tree-design.md). **Behavior-preserving — no flag, no user-visible change.** It exists so the follow-up PR can add a worktree whose files live on another machine without that row reaching code that assumes a local checkout.

- `Worktree` gains `location: WorktreeLocation` (`.local` / `.remote(provider:sessionID:)`) via migration `v70_worktree_location`.
- New `LocalWorktree` — a `Worktree` proven to have files on this machine, exposing `path`/`tmuxServer` non-optional and forwarding other reads via `@dynamicMemberLookup`.
- Local-only subsystems now fetch through `WorktreeStore.getLocal` / `listLocal`, which simply don't return a remote row.

## The hazard this fences

`WorktreeLifecycle+Reconcile` archives any DB row whose path is absent from git's worktree list — capturing scrollback, killing tmux windows, deleting terminal and tab rows on the way. A remote row's path is never in that list. A second, quieter loop rewrites `tmuxServer` on any row whose value differs from the repo's canonical name, which an empty string always does.

Both reproduce on a single sweep. Against the unmodified tree, the new fence tests fail exactly this way:

```
Expectation failed: (after.status → .archived) == .active
  reconcile archived a remote row
Expectation failed: (after.archivedAt → 2026-08-11 05:03:29 +0000) == nil
Expectation failed: (after.tmuxServer → "tbd-df8d4499") == ""
  reconcile canonicalized a remote row's tmux server
```

Nobody would have thought to hand-write a guard in reconcile. That is the argument for a compiler-enforced boundary rather than runtime checks — and the reason this is a separate PR.

## Why a wrapper rather than an optional path

Three options were weighed; rationale is in the spec's "Rejected alternatives".

Making `path` optional would have the compiler enumerate all 161 dereferences — but the honest handling at a local-only site is "this should never have been called", so ~110 sites would gain `guard let … else { return }` and turn a design error into a **silent no-op**.

Keeping `path` non-optional and guarding at runtime is what the codebase already does for the transient `.creating` placeholder (`path: ""`, guarded ad hoc in three views). That works only because the state lasts milliseconds. Made permanent, every guard nobody wrote becomes an ordinary bug.

The wrapper converts at the **fetch**, where a `guard let … else { throw }` already exists — so the caller's caller never learns about it and no ripple escapes.

## Shape of the change

- `Worktree.path` → `localPath`. The `init` label stays `path:`, the `CodingKeys` case stays `path`, and the DB column is unchanged — so all 158 construction sites and the wire format are untouched, and only *readers* break. That break is the audit.
- 45 `getLocal` + 18 `listLocal` conversions across 16 daemon files, one commit each.
- 110 of 128 daemon `path` dereferences change nothing — those subsystems now receive a `LocalWorktree`.
- App-side conversion happens at view boundaries, since SwiftUI views hold a worktree as a stored property rather than fetching one.

25 fetch sites stay location-neutral (12 of the 57 `get(`, 13 of the 31 `list(`); four of them say so in a comment. `WorktreeLifecycle+Create`'s `reserved` set and reconcile's `creatingPaths` both mirror the **global UNIQUE constraint on `worktree.path`**, where withholding a row would let a second row claim a taken path and abort the insert. Phase 3's mid-wait row check asks whether the row vanished, which is not a locality question. And the `.creating` recovery sweep is the only thing that resolves such a row, so it must see remote ones in order to give them an outcome.

## Review follow-ups

Five findings from review, one commit each on top of the original series.

- **A remote row's path is synthetic, not empty.** `worktree.path` is `NOT NULL UNIQUE`, so the empty-string convention admitted exactly one remote row per install — the second lane of a fan-out died on `UNIQUE constraint failed: worktree.path`. `WorktreeLocation.storagePath` now derives `remote://<provider>/<sessionID>` with each component percent-encoded, applied inside `WorktreeStore.create` so a call site cannot forget it. `twoRemoteRowsCoexistInOneRepo` fails against the previous code with that exact error.
- **Pre-session's mid-wait check went back to the neutral `get(id:)`.** `getLocal` answers nil for "exists but is remote", indistinguishable from deleted — and the branch it guards kills the pre-session tmux window.
- **A remote `.creating` row now reaches `.failed`.** The recovery sweep was fenced to local rows and reconcile is fenced too, so a daemon restart mid-create left a lane spinning forever. It is marked `.failed` rather than deleted, so "the create never ran" stays distinguishable from "the row vanished".
- **Three local-only subsystems the rename missed**, because they read `wt.localPath` off a bare `Worktree`: `appearance.updateColorFgBg`, `repo.relocate`, and PR polling (`pollableWorktrees` plus `pr.refresh`). PR polling is the one a remote lane wants back; re-enabling it means keying on the branch rather than the path, which the spec now records.
- **Counts in this description and the spec were recounted** from the tree and corrected.

## Reviewer notes

- **`refreshGitStatuses` unwraps to bare rows** (`.map(\.worktree)`) because it mutates array elements in place, which `@dynamicMemberLookup` cannot express. The fetch is still `listLocal`.
- **A third behavior change in reconcile** beyond the two the tests pin: `repoRows`/`scratchRows`/`globalLiveRows` moving to `listLocal` means a remote row can no longer contribute to `referencedServers` or count as a live reference keeping a tmux server alive. Correct, but unpinned by test.
- **`PanePlaceholder` taking a `LocalWorktree` rippled into `SplitLayoutView` and `TerminalContainerView`**, which the plan did not anticipate. Argued behavior-preserving: the only empty-path row today is the optimistic `.creating` placeholder, which has no terminals and no tabs, so it already took the else-branch the nil arm now routes it to.

## Test plan

- [x] `scripts/test.sh` — 5415 tests in 578 suites pass (one by-design known issue, `FlakyQuarantineSelfTests.retriesUntilPass()`)
- [x] New `ReconcileRemoteRowFenceTests` verified to **fail** against the unfenced tree and pass after — a guard test that cannot fail is not a guard test
- [x] Orphan-GC sweep characterized as leaving a remote row intact, with `setGCEnabled(true)` pinned so the assertion isn't vacuous
- [x] `getLocal`/`listLocal` exclude remote rows; location-neutral fetches still return them
- [x] Wire format pinned: a characterization test asserts the JSON key is still `path` and no `localPath` key appears, run before and after the rename
- [x] `scripts/swift-safe build` clean; `swiftlint --strict` 0 violations in 685 files
- [ ] **Visual check outstanding** — `scripts/restart.sh` confirms the daemon, terminal panes, and RPC all run from this worktree, but the UI itself was not eyeballed (screen capture is TCC-denied in the agent's shell). Sidebar rows, the file panel, the status-bar path, and the row context menu's Reveal-in-Finder / Copy-Path are the surfaces this touches and want a human glance before merge.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=D8182E8D-C247-4010-B7B5-DB34F91BAEF6)

## Deliberately deferred to PR 2

`RPCRouter+WorktreeHandlers.swift:184-191` — `handleWorktreeList`'s archived-enrichment loop gates on `status == .archived` but not on locality, so a remote row's synthetic path would flow into `ClaudeProjectDirectory.resolve`, miss both tiers, and trigger the recursive `~/.claude/projects/*` scan the comment above it warns about.

Left unfenced on purpose: the **outcome** is already correct (`liveClaudeSessionCount = 0`), an empty path would have scanned just as wastefully, and it is unreachable until remote-row creation ships. This is a performance fence for PR 2, not a correctness bug in this one. Two lines plus a test whenever it becomes reachable.

### DIFF
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -443,6 +443,14 @@ final class AppState: ObservableObject {
         return findWorktree(id: id)
     }
 
+    /// The current selection, when it has files on this machine. Views that
+    /// operate on a directory bind to this instead of `selectedWorktree`, so
+    /// they cannot render against a worktree with no local checkout — the
+    /// optimistic `.creating` placeholder (`path: ""`) included.
+    var selectedLocalWorktree: LocalWorktree? {
+        selectedWorktree.flatMap(LocalWorktree.init)
+    }
+
     /// All pinned terminals across all worktrees, sorted by pinnedAt.
     var pinnedTerminals: [Terminal] {
         terminals.values.flatMap { $0 }
--- a/Sources/TBDApp/ContentView.swift
+++ b/Sources/TBDApp/ContentView.swift
@@ -631,7 +631,9 @@ private struct WorktreeDetailAreaView: View {
     var body: some View {
         HStack(spacing: 0) {
             TerminalContainerView()
-            if showFilePanel, let worktree = selectedWorktree, !worktree.path.isEmpty {
+            // Task 7: becomes `let worktree = selectedLocalWorktree`, which
+            // subsumes the empty-path check.
+            if showFilePanel, let worktree = selectedWorktree, !worktree.localPath.isEmpty {
                 FilePanelDivider(panelWidth: Binding(
                     get: { CGFloat(filePanelWidth) },
                     set: { filePanelWidth = Double($0) }
--- a/Sources/TBDApp/ContentView.swift
+++ b/Sources/TBDApp/ContentView.swift
@@ -622,18 +622,18 @@ private struct WorktreeDetailAreaView: View {
     @Binding var filePanelWidth: Double
     @State private var contentAreaHeight: CGFloat = 600
 
-    private var selectedWorktree: Worktree? {
-        guard let id = appState.selectedWorktreeIDs.first else { return nil }
-        // findWorktree also resolves scratch spaces (repo-less worktrees).
-        return appState.findWorktree(id: id)
+    /// The file panel reads a directory, so it binds to the local-only
+    /// selection. `LocalWorktree` subsumes the empty-path check this used to
+    /// spell out — the optimistic `.creating` placeholder has no directory yet
+    /// and does not convert.
+    private var selectedLocalWorktree: LocalWorktree? {
+        appState.selectedLocalWorktree
     }
 
     var body: some View {
         HStack(spacing: 0) {
             TerminalContainerView()
-            // Task 7: becomes `let worktree = selectedLocalWorktree`, which
-            // subsumes the empty-path check.
-            if showFilePanel, let worktree = selectedWorktree, !worktree.localPath.isEmpty {
+            if showFilePanel, let worktree = selectedLocalWorktree {
                 FilePanelDivider(panelWidth: Binding(
                     get: { CGFloat(filePanelWidth) },
                     set: { filePanelWidth = Double($0) }
--- a/Sources/TBDApp/FileViewer/FileViewerPanel.swift
+++ b/Sources/TBDApp/FileViewer/FileViewerPanel.swift
@@ -144,9 +144,9 @@ private func parseBranchDiff(_ output: String) -> [BranchFileChange] {
 // MARK: - FileViewerPanel
 
 struct FileViewerPanel: View {
-    // Task 7: this stored property becomes a `LocalWorktree`, and the
-    // `.localPath` reads below become `.path` on the wrapper.
-    let worktree: Worktree
+    /// Every read below is of a directory on this disk, so the panel takes the
+    /// proven-local wrapper rather than a bare `Worktree`.
+    let worktree: LocalWorktree
     @EnvironmentObject var appState: AppState
 
     @State private var staged: [GitFileStatus] = []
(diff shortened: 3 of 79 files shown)


cheapsteak:
Both findings from the review at patch-id `31c3d470` are fixed. That comment is now stale — the branch is at patch-id `e30178ab`.

- **HIGH, `repo.remove --force` mid-cascade throw** — fixed in `53b237e3`. Unforced removal stays location-neutral (a remote lane is as much a live thing to lose, so it still blocks and is counted in the error). Under `--force`, only local rows are cascade-archived; the remote row is cleared by the same `deleteForRepo` that handles archived and main rows. No actuation row is minted for a remote lane, deliberately — none of its sessions were torn down, and the log may only claim acts that were attempted. Two new tests, both failing before the fix (repo still registered, local row stranded, remote row stranded).
- **MEDIUM, orphan GC unfenced + vacuous guard test** — fixed in `a3a7f477`. Both `reconcileScratchpads` fetches now go through `listLocal`. The finding was right that the old test asserted `worktree.status`, which OrphanGC never mutates; it is replaced by two tests that plant a real directory at the remote row's slug plus a control scratchpad behind a gone *local* row, so a sweep that reaps nothing cannot pass by doing nothing. Verified red when the fence is reverted.

The MEDIUM was self-inflicted and worth recording: the spec's "orphan GC needs no change" rationale rested on a remote row having no path, which stopped being true when an earlier fix in this same PR gave remote rows a synthetic `remote://` path. The spec now states the fence as the reason GC is safe.

A re-audit for the same class (anything else assuming an empty path) swept all four `Sources/` trees and found one further site, deliberately deferred with rationale in the PR description.

Note for the gate itself: this is the fourth review run on this PR. Two failed on the orchestrator stall (`findings-correctness.json` never written) and one on schema rejection, before and after #612. Only one run produced a verdict.

cheapsteak:
Both findings at patch-id `e30178ab` are **false positives from the review's own two-dot diff fallback**, and I want to be explicit about why rather than silently rebasing past them.

The diagnostics section disclosed the cause: `git diff origin/main...HEAD` (three-dot) failed in the shallow clone, so both specialists fell back to `git diff origin/main HEAD` (two-dot). Two-dot compares tips, so anything present on `main` but absent from this branch renders as if this branch deleted it.

Verified:

- `c4c71104` (#607) landed on `main` **after** this branch's last rebase — the branch was 1 commit behind, not reverting anything.
- This branch **never touched** `AppState+Tabs.swift` or `AppState+Terminals.swift`. Three-dot: `git diff --name-only origin/main...HEAD` lists neither file.
- Two-dot attributed #607's own edits to this PR as `25 insertions, 86 deletions` across those two files, which is where "whole test classes deleted" came from.

A three-way merge would never have reverted #607 either, since the merge base predates it and this branch has no changes in those files. Acting on the finding would have meant re-adding #607's code to a branch that never touched it, manufacturing a real conflict from a phantom one.

Rebased onto current `main` anyway (now at `5382cf04`), which is the right action for a different reason: the branch now contains `c4c71104`, so the two-dot fallback produces a correct diff on the next run even if three-dot fails again.

Worth flagging for the gate itself: the shallow-clone three-dot failure is a latent source of false "you reverted X" findings on any PR whose branch is even one commit behind `main`. The fallback is silently wrong rather than loudly unavailable.

PR #615, opened 2026-08-11
fix: preflight Claude review artifacts before handoff

Jephuff:
## Summary

- make the Stop hook run the trusted base-pipeline validator before releasing a parseable review result
- feed normal schema, completeness, and disposition failures back to the live orchestrator through the existing bounded nudge path
- pin the closed finding-key contract in the specialist prompt and require lossless correction of unsupported fields
- exercise the preflight through the real-Claude sandbox and document the in-session boundary

## Root cause

The correctness specialist invented `failure_scenario` because its prompt described required and recommended fields without stating that the schema is closed. The orchestrator saw the extra fields but attempted only JSON parsing with Python, which was correctly unavailable under its tool restrictions. The Stop hook then released immediately when `review-result.json` parsed, so no in-session boundary schema-validated the specialist artifacts. The downstream validator correctly rejected the unsupported properties and failed closed, but by then the Claude session had ended and could not repair them.

## Safety and scope

- no schema widening
- no trigger, permission, token, tool-list, job-name, or restore-path changes
- exactly one `claude-review` job remains
- preflight writes its provisional verdict only in an isolated temporary directory and discards it
- repairable artifact rejection uses a dedicated status and the bounded correction path
- validator infrastructure failures use a distinct status and release to authoritative downstream fail-closed validation
- both counted correction branches remain bounded at five nudges
- real-Claude E2E scenarios install the trusted validator and schemas, declare their specialist set, and use the same Python dependency environment as the Stop hook

This is a bug fix restoring the existing review-gate contract; no feature flag or new product design is required.

## Verification

- strict TDD RED/GREEN coverage for the observed `failure_scenario` regression, E2E pipeline installation, specialist selection, status-based validation rejection, missing-validator dependency classification, real-Claude prerequisites, and hook Python alignment
- every new gate branch covered: valid artifacts, repairable rejection, unexpected validator failure, five-nudge release, unavailable validator/Python/temp directory, no-`jq`, and no-`jq`/no-Python fallbacks
- `/tmp/tbd-review-schema-venv/bin/python -m pytest -q .github/workflows/claude-review-v2/tests/` — 310 passed, 3 skipped; skips are the real-Claude scenarios because the CLI is not installed on PATH
- Python compilation of the E2E harness and all three real-Claude scenarios — passed
- `bash -n .github/workflows/claude-review-v2/hooks/stop-hook.sh` — passed
- `swiftlint --strict --quiet` — passed
- `git diff --check` — passed
- `scripts/swift-safe build` — passed

`scripts/test.sh` was not run because no daemon or shared code changed.


Jephuff:
Closing as superseded by #616.

After rebasing onto current main, the reported `failure_scenario` incident is already handled in the validator and review prompt, with focused regression coverage. The remaining changes in this PR only add a broader Stop-hook preflight and a shared five-nudge policy that is unnecessary for the incident and would require a separate human-approved liveness decision.

The PR-template deletion noted by the latest review was stale-base drift; both `.github/PULL_REQUEST_TEMPLATE.md` and `CLAUDE.md` are present on current main. The wording and cleanup-test notes apply only to the redundant preflight path, so closing this PR removes those concerns rather than layering fixes onto obsolete machinery. No review changes were pushed.

PR #618, opened 2026-08-11
Fix: Root-cause the flake cluster that made "rerun once" standard advice

cheapsteak:
## What's broken

A recurring cluster of tests fails under machine load and passes on targeted rerun: both `ProviderEventsSupervisorTests`, the `MarkdownStylesheetTests` directory-watch test, the `ThemeStoreTests` watcher-reload test, and (analyzed but not changed here, see below) the `AppearanceDebounceTests`/`SearchQueryDebouncerTests` burst tests. Across four full runs on a loaded dev machine, every run had at least one of these red; each went green on rerun. The standing advice calcified into "rerun once", which taxes every branch and trains people to wave past red suites.

## Why it happens

Swift Testing runs every non-serialized test in one process with no concurrency cap, so per-test scheduling latency scales with the total population (now 5417 tests) plus external machine load — mined runs put p50 per-test reported duration at ~1/3 of total wall time (`Tests/CLAUDE.md`, "Population is the scheduler"). Each flaking test waited on that real scheduling behind a wall-clock bound sized for a smaller population or a quieter environment. Per-test root causes, each traced through the production pipeline (none is a race in the code under test):

- **`ProviderEventsSupervisorTests` (tier 3)** — 15s bounded-poll deadlines were sized for CI's quiet serial pass; a loaded parallel local run starves the supervision task past them (observed: pid file never read in 15s; mirror rows still empty at 15s). On top of that, one assertion had a genuine logic race: `snapshotAppliedAndRestartResyncs` demanded the *exact transient* set `{live-a, run-1}`, but the stub goes silent after its snapshot and `silenceLimit` is 5, so the supervisor's own watchdog legitimately kills and respawns it every few seconds, replacing `run-1` with `run-2` — an observer starved past the first cycle could never match, at any deadline.
- **`MarkdownStylesheetTests` watch test (tier 2)** — waited out a real 150ms `FileWatcher` debounce plus `AsyncStream` delivery under one 8s bound. `FileWatcher` already takes an injected clock, and the sibling `FileWatcherTests` already proves the real-dispatch-source + `TestClock` pattern; this test predated it.
- **`ThemeStoreTests` watcher test (tier 2)** — FSEvents cannot be virtualized (the 0.1s is a stream latency parameter, not a sleep; events are journaled by fseventsd, so none can be lost) and bounded polling is the sanctioned shape; the 30s bound predated the population re-derivation that took the shared guards to 45s/90s. Its wait rides the single MainActor executor that every `@MainActor` suite queues on.

## When it broke

Not one regression commit: each bound was correct for the population it was derived against (the ThemeStore 30s in #441 against ~3000 tests; the supervisor 15s with #514 assuming the quiet tier-3 pass). Population growth to 5417 plus multi-agent load on the dev box moved the environment out from under them. The `run-1` transient-set race has been latent since #514 — it needs a >5s observation delay to bite, so only load exposed it.

## What this PR does

- **`MarkdownStylesheetTests`**: rewrites the watch test onto the `FileWatcherTests` idiom — real dispatch source delivers the *event*, the debounce *timer* is virtual (`FileWatcher(clock: TestClock)` via an arm-counting delegating clock). No wall bound covers the debounce any more; the remaining bounded polls guard only genuinely real-time legs (kqueue registration/arming, `AsyncStream` delivery) and throw named diagnostics.
- **`ProviderEventsSupervisorTests`**: rewrites `runAssertions` against the contract instead of run indices — *a* snapshot applied (`live-a` + any `run-N`), invocation count strictly above its value at the kill, then some `run-M` with `M > maxRunAtKill` (a snapshot only a post-kill connection can produce). Re-derives deadlines with the reasoning inline: bounded-poll guards 15→90s (anchoring to the `ciSafeDeadline` figure for the same contention class), zombie-reap guards 5→30s, the `stop()` promptness bound 10→60s, and pins a 6-minute suite hang limit so a regressed `stop()` cannot wedge a whole local run.
- **`ThemeStoreTests`**: re-derives the reload hang guard 30→90s (same anchor), and upgrades the timeout to a thrown diagnostic reporting `userThemes`, `loadErrors`, and the actual directory listing — discriminating "watcher never fired" from "reload ran but decode failed" from "file never landed".

On the raised numbers, explicitly rather than quietly: every one is a **positive wait** that breaks on its first satisfying probe — healthy runs pay ~0.1s, only a genuinely failing run pays the deadline — and each carries its derivation in a comment so the next re-derivation has an anchor. This is the repo's established hang-guard doctrine (`Tests/CLAUDE.md`), not tolerance-window inflation; the transient-state fix above is the part no deadline could have fixed.

**Deliberately not changed:** the `AppearanceDebounceTests`/`SearchQueryDebouncerTests` burst tests. They are already fully virtual-time and `.serialized`; their failure is the arming handshake starving inside shared machinery (`TestClock.checkSuspension`'s background-QoS megaYield under process-global saturation), and every suite-local remedy is already measured-and-refuted in `Tests/CLAUDE.md`. The real fix is the megaYield-free virtual clock that `ClockTestSupport` names as a deliberate shared-contract non-goal — that reshapes shared test infrastructure, so it needs a human decision first rather than being bundled here. Also checked per the burn-down list: all five production types in the cluster already carry the clock seam; no `no_raw_task_sleep` suppressions are involved.

## Evidence & verification

Mutation checks (each applied to `Sources/`, observed red with its named diagnostic, restored, re-observed green):

- `FileWatcher` eventMask minus `.write`/`.extend` → `a FileWatcher on a directory armed no debounce timer for an entry being created — observed armed=0 after 4 attempts of 2.0 seconds each`
- `ThemeStore.startWatching()` without `w.start(directory:)` → thrown diagnostic at 90s whose directory listing shows `ext.json` present, pinning "watcher never fired"
- Supervisor restart loop removed → `supervisor did not restart the events process within 90.0s; observed invocation count=1, was 1 at the kill`
- SIGKILL escalation removed from `killTree` → red via the leak check at the 30s reap guard (`stop()` returned promptly; the child survived)

Targeted suites green before and after: 32 tests across the three files (the rewritten markdown watch test now runs in ~0.1s instead of riding a real debounce). `scripts/swift-safe build` clean, `swiftlint --strict` clean. Full-suite under-load soak runs are executing on the loaded dev box now; results will follow as a PR comment.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=9EEB84BC-BF9D-408C-ACA1-4543A64AAA7C)

### DIFF
--- a/Tests/TBDAppTests/MarkdownStylesheetTests.swift
+++ b/Tests/TBDAppTests/MarkdownStylesheetTests.swift
@@ -1,9 +1,13 @@
+import Clocks
 import Foundation
 import Testing
 @testable import TBDApp
+import TestSupport
 
 /// Tier 1, except for `themesDirectoryWatchSeesAStylesheetCreatedLater`, which
-/// is tier 2 (real filesystem, real dispatch source, bounded polling).
+/// is tier 2: a real filesystem and a real dispatch source deliver the *event*,
+/// while the debounce *timer* is virtual — the same split `FileWatcherTests`
+/// documents.
 ///
 /// Every test supplies its own tmp themes directory and its own
 /// `UserDefaults(suiteName:)`. Neither `~/tbd` nor `UserDefaults.standard` is
@@ -285,9 +289,27 @@ struct MarkdownStylesheetTests {
     /// That is the only way the viewer can notice a stylesheet that did not
     /// exist when it opened — there is no inode for a file watcher to open.
     ///
-    /// Tier 2 by construction (real dispatch source), so it uses the production
-    /// clock and bounded polling rather than a `TestClock`.
-    @Test("a watch on the themes directory sees a stylesheet created later")
+    /// Tier 2, split the way `FileWatcherTests` splits: a real dispatch source
+    /// delivers the *event*, while the debounce *timer* is virtual — production
+    /// takes `clock: any Clock<Duration>`, so the 150 ms window is crossed by
+    /// advancing a `TestClock` instead of being waited out on a loaded runner.
+    /// Waiting out a real debounce plus `AsyncStream` delivery under one 8 s
+    /// bound is exactly what starved here when the process carried >5000 tests.
+    ///
+    /// The bounded polls that remain wait only on legs that are genuinely
+    /// real-time — the dispatch source delivering an event and production
+    /// reaching `clock.sleep` (observed through `armed`), and the consuming
+    /// `Task` receiving an `AsyncStream` element — never on the debounce
+    /// interval itself. Each is a hang guard whose healthy path exits on its
+    /// first probe, and each timeout travels as a thrown `Failure` carrying the
+    /// observed counts (Tests/CLAUDE.md rule 4).
+    ///
+    /// One-sidedness worth naming: this watches a *directory*, so the events
+    /// are not attributable to a particular entry. A straggler probe event
+    /// could in principle supply the phase-2 arm or notification. The final
+    /// `resolve()` assertion is what pins the actual content — a watcher that
+    /// never saw `later.css` cannot make it read `body{color:teal}`.
+    @Test("a watch on the themes directory sees a stylesheet created later", .clockDriven)
     func themesDirectoryWatchSeesAStylesheetCreatedLater() async throws {
         let dir = try tempDir()
         defer { try? FileManager.default.removeItem(at: dir) }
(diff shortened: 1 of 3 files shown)


cheapsteak:
## Under-load soak results (the failure condition from the PR description)

Three consecutive full-suite runs (5411 tests in one process) on the loaded dev box, load averages 15–44 on 12 cores at run starts:

- **Fixed cluster: 12/12 green.** All four rewritten tests passed in all three runs. Their *reported durations* show the regime they survived — mostly time suspended waiting for a scheduler turn, exactly the population effect the fixes were derived against:
  - ThemeStore watcher reload: 131s / 228s / 92s reported (old bound: **30s** — every run would have been red before this PR)
  - MarkdownStylesheet directory watch: 44s / 63s / 33s reported (old end-to-end bound: **8s**)
  - ProviderEventsSupervisor: 33–55s reported (old bounds: **15s**)
- **The deliberately-unfixed burst pair reproduced its diagnosed failure in 2 of 3 runs** (`AppearanceDebounceTests` "rapid scheme changes…", `SearchQueryDebouncerTests` "a burst within one window…"): `Issue recorded` at the `advanceWhenSuspended` arming handshake, then `fired.values → []`, at ~131s and ~228s. Live confirmation of the TestClock arming-starvation root cause described in the PR body; the megaYield-free clock lands as a stacked follow-up PR.
- Out-of-cluster load flakes observed during the soak, not touched by this PR: `ReplayLiveIntegrationMatrixTests` firehose-overflow (run 1) and the `TerminalTeardownReapTests`/`ChildReaperTests` PTY-reap suites (run 2). Same family, separate work.

Also for the record: this PR's first CI run flaked on `WorktreeArchiveDeletionQueueTests.theFallbackRemovalArmsItsOwnRaisedTimeout` — a test outside this diff, same load-sensitive class — rerun in progress at time of writing (this line will not be edited again; check the checks tab). The tax this PR is about, paid one more time on its own gate.

PR #627, opened 2026-08-14
fix: reject invalid Accessibility window geometry

Jephuff:
## What's broken

An external Accessibility client can send TBDApp a non-finite window position or size. AppKit accepts the Accessibility request far enough to construct an invalid `NSWindow` frame, then raises an uncaught `NSInternalInconsistencyException` and terminates TBDApp. The observed crash attempted to apply a frame whose vertical origin was `NaN`.

## Why it happens

TBDApp relies on AppKit's default Accessibility window setters. Those setters forward position and size values into AppKit's frame machinery without rejecting `NaN` or infinite components, and AppKit treats the resulting frame as an internal-consistency violation rather than returning an Accessibility error.

## What this PR does

- Installs a one-time guard on `NSWindow`'s Accessibility position and size setters before SwiftUI creates application windows.
- Rejects geometry containing non-finite components or an unexpected Objective-C payload encoding, and records a fault-level diagnostic.
- Forwards finite and unexpected values through the original AppKit implementations unchanged.
- Adds real-AppKit regression coverage for non-finite and type-mismatched position/size values plus valid forwarding.

The interception stays at the Accessibility boundary implicated by the crash instead of changing general `NSWindow` frame behavior.

## Evidence & verification

- Standalone reproduction: sending `(-556, NaN)` through AppKit's Accessibility position setter reproduced the same invalid-frame exception outside TBDApp.
- RED: `scripts/test.sh --filter AccessibilityWindowGeometryGuardTests` failed because the guard did not exist before production code was added.
- Claude review follow-up RED: passing `NSValue(rect:)` to the position setter reproduced the uncaught Objective-C accessor exception.
- GREEN on the current PR head: the focused command passes all 6 tests, including the observed `(-556, NaN)` position, an infinite size, and mismatched Objective-C value encodings.
- `scripts/swift-safe build` passes.
- `swiftlint --strict` reports 0 violations across 721 files.
- A full `scripts/test.sh` run executed 5,865 tests. The new Accessibility suite passed; the run had one unexpected failure in the unrelated transcript row-height estimator suite. Rerunning that suite alone, without installing or executing this guard, reproduces the same existing failure.
- Independent pre-merge review found no blocking correctness, ABI, lifecycle, or isolation issues after direct IMP-level verification of the reviewer's initial `nil` concern.

### DIFF
--- a/Sources/TBDApp/Diagnostics/AccessibilityWindowGeometryGuard.swift
+++ b/Sources/TBDApp/Diagnostics/AccessibilityWindowGeometryGuard.swift
@@ -1,0 +1,73 @@
+import AppKit
+import ObjectiveC.runtime
+import os
+
+private let accessibilityGeometryLogger = Logger(
+    subsystem: "com.tbd.app",
+    category: "accessibility-geometry"
+)
+
+private typealias AccessibilityGeometrySetter = @convention(c) (
+    AnyObject,
+    Selector,
+    AnyObject
+) -> Void
+
+/// Prevents malformed geometry from external Accessibility clients from
+/// reaching AppKit's `NSWindow` frame setter, which raises an uncaught
+/// `NSInternalInconsistencyException` for NaN or infinite coordinates.
+@MainActor
+enum AccessibilityWindowGeometryGuard {
+    private static var isInstalled = false
+
+    static func install() {
+        guard !isInstalled else { return }
+        isInstalled = true
+
+        installSetter(
+            selectorName: "accessibilitySetPositionAttribute:",
+            geometryName: "position"
+        ) { value in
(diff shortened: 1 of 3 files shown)


cheapsteak (approved):
> An external Accessibility client can send TBDApp a non-finite window position [..]

@Jephuff what am I reading here? 🙏 
Why is external Accessibility client sending TBDApp anything?
What's the usage pathway that triggers this?

Jephuff:
> @Jephuff what am I reading here? 🙏
> Why is external Accessibility client sending TBDApp anything?
> What's the usage pathway that triggers this?


Window managers! for me it was Rectangle, but I think it could be any window manager

PR #647, opened 2026-08-14
Fix: Say who holds the shared build lock and keep talking while you wait (three agent sessions killed by the silence)

cheapsteak:
## What's broken

`scripts/swift-safe` serializes every Swift compile on the machine behind one
lock, so on a box with several active worktrees, waiting for the build slot is
routine. The wait is silent. The waiter prints one line — "another TBD worktree
is compiling; waiting for the shared build slot" — and then says nothing at all
for up to its 30-minute timeout.

Two costs, both observed on a five-worktree machine:

- **A human cannot tell a queued build from a hung one.** No indication of who
  holds the slot, and no indication that time is passing. The only way to tell a
  healthy queue from a wedged compiler was to go digging in `ps`.
- **Automated callers get killed mid-work.** An agent session that emits nothing
  for 600s is terminated by its harness's stall watchdog. Three sessions died
  that way in one afternoon, each losing its in-progress work; the build
  underneath was healthy every time. The silence is what killed them.

That second cost is the motivating one, and it generalizes: any wait longer than
a watchdog window destroys in-progress work, and this lock is the largest single
source of long waits in the repo. A wait that reports itself is not just nicer to
read — it is the difference between queuing and losing the session.

## Why it happens

The information needed was already on disk and simply never read. Before exec'ing
SwiftPM, the lock holder writes its `pid`, `cwd` and `command` into the lock file.
A waiter opens that exact file to take the lock — and never looks at its contents.
`_acquire` set an `announced` flag on the first blocked attempt and used it only to
suppress every subsequent line, so the spin loop ran mute until it either won the
lock or timed out.

## What this PR does

The waiter now reads the record it was already holding open:

- **Names the holder on first blocking** — the worktree by its directory
  basename, the pid, and the command. The basename is what identifies a worktree
  to a person; the full path is noise, and in a public multi-tenant repo it is
  unnecessary detail.
- **Heartbeats every 60s** with elapsed time and a re-read of the holder, so both
  a human and a watchdog see liveness, and a slot that changed hands is reported
  as it now stands rather than as it was.
- **Stays entirely on stderr**, so a caller parsing SwiftPM's stdout sees nothing
  from the wrapper.
- **Never claims a holder it cannot see.** Only newline-terminated lines are
  parsed, so a record caught mid-write (`pid=12` of an eventual `pid=12345`) is
  ignored rather than read as a plausible wrong pid. A recorded pid is reported
  only while that process is alive; a stale record is reported as stale, and a
  non-positive pid is treated as unrecorded rather than probed — `os.kill(0, 0)`
  inspects the waiter's own process group and would always "succeed" — and a pid
  past the platform's `pid_t` is reported as not running rather than crashing the
  waiter, since `os.kill` raises `OverflowError` there and that is not an
  `OSError`. The holder writes its identity *after* acquiring, so "someone holds
  this, and has not said who yet" is a real state and gets its own honest wording.
- **Rejects a wait setting that would disable the wait's own limits.** `float()`
  accepts `"nan"` and `"inf"`, and every comparison against nan is False, so the
  old positivity check passed them through: a nan heartbeat never fires (silent
  wait again) and a nan timeout makes the deadline unreachable, so the wait never
  ends at all. Both settings now require a finite number, checked before the
  positivity test since nan defeats that test. The unbounded-timeout half of this
  predates the PR; it shares the one guard.

The timeout line now carries the holder too, since that is the moment a reader
most wants it.

**Why 60s for the heartbeat.** The floor is the 600s stall watchdog that has been
killing sessions; the ceiling is not wanting to spam a 30-minute wait. 60s clears
the watchdog by 10x — enough margin that several delayed beats on a loaded machine
still cannot trip it — and costs at most 30 lines across the entire timeout. A
typical wait behind one sibling build (1-3 minutes) prints one to three lines.

**The `execv` is untouched.** The lock still rides the process that becomes the
compiler, so a killed wrapper cannot release the slot while leaving an orphaned
compiler alive for the next worktree to overlap.

## Findings for follow-up PRs

**`run` holds the slot for the running program's whole lifetime — latent, not
live.** `COMPILE_SUBCOMMANDS` includes `run`, and because of the `execv`,
`scripts/swift-safe run TBDApp` would hold the machine-global lock until TBDApp
*exits*, blocking every other worktree indefinitely. I verified it is unexercised
rather than merely believed to be: no script, workflow or doc in the repo invokes
`swift-safe run`. `restart.sh` and `mock.sh` build and then launch the built
binary; `test.sh` and the nightly stress harness pass `test`/`build` (its
`run_governed_swift` pass-through has exactly one call site, and it passes
`build --build-tests`). The path is reachable by instruction, though — the
guardrail that blocks raw SwiftPM tells users to run
`scripts/swift-safe <build|test|run> ...` — so this PR adds a stderr warning on
that path rather than leaving the trap unmarked.

I did not fix it here, deliberately. The obvious fix — build, release, then run
the built product — needs the build to happen in a child process so the wrapper
survives to release the lock, and that is precisely the structure the `execv`
exists to avoid. There is a version that preserves the property (fork, let the
child inherit the lock fd so the open file description outlives a killed parent,
then release and exec the product), but it also needs product-path derivation and
`swift run` argument semantics, which is real restructuring of the load-bearing
exec path for a path with no callers today. Worth its own PR with that sketch as
the starting point.

**Lock fairness.** The acquire loop is a 0.25s `LOCK_NB` spin with no queue, so
wake order is arbitrary and a worktree can be starved indefinitely while others
cycle through. Real, and now more visible — the heartbeat will show it happening.
A fix is a behavior change that needs its own design.

**An unrelated test is racy and flaked on this PR's first CI run.**
`theFallbackRemovalArmsItsOwnRaisedTimeout()` in
`Tests/TBDDaemonTests/WorktreeArchiveDeletionQueueTests.swift:118` was the single
real failure in a run reporting 71 issues (70 known). Its control step removes a
throwaway worktree through a `GitManager(subprocessTimeout: .milliseconds(1))`,
expects `GitTimeoutError`, and then asserts the directory **still exists** — on
the reasoning that a killed `git worktree remove` cannot have completed. That
last step is a race rather than a deduction: if git finishes the unlink inside
the 1 ms window before the kill lands, the directory is legitimately gone and the
assertion fails. It fails on a **fast** machine, not a slow one. The test's own
comment says `runBoundedProcess` "makes that discriminating rather than racy",
which holds for *reporting* `.timedOut` (the `#expect(throws:)` above it) but not
for *whether the directory survives* — two different claims. Fixing it means
making the survival assertion deterministic instead of dependent on winning a
race; that belongs in its own change rather than in a build-wrapper PR. The test
dates from #604 and has no quarantine entry, so this looks like its first
recorded flake. It passed on re-run here with no code change.

**The job budget may be obsolete.** `DEFAULT_JOBS = 2` machine-wide dates from
Swift 6.2, whose WMO debug builds hit 6-9 GB RSS per frontend. `Package.swift`
documents that Swift 6.3 fixed this (measured peak 1.35 GB), so the machine may
be sitting idle while everyone queues. That needs measurement, not a guess, and
an evidence-driven PR of its own.

## Assumptions

- A recorded pid that is alive is the holder. A pid recycled between the holder's
  write and the waiter's read would be misreported; the window is small and the
  consequence is a wrong name in a diagnostic line, not a wrong locking decision.
- 600s is the stall-watchdog window worth designing against. If a harness uses a
  materially shorter one, the 10x margin absorbs it down to about 300s; below
  that, `TBD_SWIFT_HEARTBEAT_SECONDS` overrides the cadence.

## Evidence & verification

Repro before the fix: hold the lock from one worktree and start a build in
another — one line, then nothing, for as long as you care to watch.

After, against a real holder in a directory named `acme-worktree`:

```
swift-safe: another TBD worktree is compiling; waiting for the shared build slot (held by worktree acme-worktree, pid 85653 running: swift build --jobs 2)
swift-safe: still waiting for the shared build slot after 60s of 1800s (held by worktree acme-worktree, pid 85653 running: swift build --jobs 2)
```

`scripts/test-swift-safe.py` grows from 12 to 29 tests (5s, no build), covering
the holder being reported when the record is populated, nothing fabricated when
it is empty/missing/truncated/garbage/stale/impossible, the heartbeat cadence,
non-finite settings being rejected before SwiftPM is exec'd, and stdout staying
clean. Cadence is driven by a fake clock that advances only when the code
sleeps, and the end-to-end test compresses the interval via
`TBD_SWIFT_HEARTBEAT_SECONDS`, so no test spends real time waiting. It already
runs in the `plans-guard` CI job, so the new coverage is gated without a workflow
change.

Discriminating evidence — every guard was broken, watched red, and restored:

- Parse half-written lines → the two "partially written" tests fail.
- Report a non-positive recorded pid as a holder → seven tests fail, including
  the empty-file and garbage-file cases, which start claiming "held by pid 0".
- Treat a dead pid as alive → the stale-holder tests fail.
- Print the holder's full `cwd` instead of the basename → four tests fail,
  including the end-to-end path-disclosure assertion.
- Announce once and never heartbeat → the cadence tests and the end-to-end
  heartbeat assertion fail.
- Send the reports to stdout → five tests fail on stdout no longer being
  SwiftPM's alone.
- Cache the holder description instead of re-reading it → the late-identification
  test fails.
- Drop the `run` warning → the two `run` tests fail.
- Fabricate a holder when none is recorded → seven tests fail.
- Never truncate an over-long command → the truncation test fails.
- Let `OverflowError` escape again → the impossible-pid test errors and the
  end-to-end no-traceback test fails.
- Drop the finite check → the non-finite settings test fails on nan and inf for
  both variables.
- Order the finite check after the positivity check → nan is caught by the
  `!= nan` fallback but `inf` still slips through, so the test fails on inf.

Both review findings were reproduced against the real script before being fixed
(traceback and exit 1 for the oversized pid; a nan-timeout waiter still spinning
after 4s when a 0.5s timeout exits at 75), and re-run green after.

Pyright is clean on both files.

### DIFF
--- a/scripts/swift-safe
+++ b/scripts/swift-safe
@@ -5,6 +5,13 @@ Every TBD worktree shares the same default lock under ``~/tbd/runtime``.  This
 prevents otherwise-independent Claude sessions from each starting a full
 Swift compiler swarm on the same laptop.  The file may remain on disk: flock
 ownership is kernel-managed and is released when this process exits.
+
+Waiting is reported on stderr, never stdout, so a caller parsing SwiftPM's
+output sees nothing from this wrapper.  The holder writes ``pid``/``cwd``/
+``command`` into the lock file after acquiring, and a waiter reads it back to
+say who it is queued behind; a heartbeat then repeats the wait's elapsed time
+so a human can tell a queued build from a hung one and an automated caller's
+stall watchdog sees liveness.
 """
 
 from __future__ import annotations
@@ -20,6 +27,14 @@ import time
 DEFAULT_JOBS = 2
 MAX_JOBS = 4
 DEFAULT_TIMEOUT_SECONDS = 1_800.0
+# Agent harnesses kill a session that emits nothing for 600s, and a queued
+# build used to be silent for the full 30-minute timeout.  One line a minute
+# clears that watchdog by 10x while costing at most 30 lines across the whole
+# timeout, and a typical wait (one sibling build, 1-3 minutes) prints 1-3.
+DEFAULT_HEARTBEAT_SECONDS = 60.0
+# Fields the holder records; anything else in the file is ignored.
+HOLDER_FIELDS = ("pid", "cwd", "command")
+MAX_REPORTED_COMMAND_CHARACTERS = 120
 COMPILE_SUBCOMMANDS = {"build", "run", "test"}
 SUPPORTED_SUBCOMMANDS = COMPILE_SUBCOMMANDS
 RUN_OPTIONS_WITH_VALUES = {
@@ -168,25 +183,109 @@ def _lock_path() -> Path:
     return tbd_home / "runtime" / "swift-build.lock"
 
 
-def _acquire(lock_file, timeout_seconds: float) -> None:
-    deadline = time.monotonic() + timeout_seconds
+def _report(message: str) -> None:
+    """Waiting is diagnostics: stderr only, so stdout stays SwiftPM's alone."""
+    print(message, file=sys.stderr, flush=True)
+
+
+def _holder_fields(lock_path: Path) -> dict[str, str]:
+    """Best-effort read of what the current holder recorded, or ``{}``.
+
+    The holder truncates and rewrites this file after it acquires, so a waiter
+    can catch it empty or half-written.  Only newline-terminated lines are
+    parsed — a trailing fragment such as ``pid=12`` mid-write would otherwise
+    read as a plausible but wrong pid.
+    """
+    try:
+        text = lock_path.read_text(encoding="utf-8", errors="replace")
+    except OSError:
+        return {}
+    fields: dict[str, str] = {}
+    for line in text.split("\n")[:-1]:
+        key, separator, value = line.partition("=")
+        if separator and key in HOLDER_FIELDS and value:
+            fields[key] = value
+    return fields
+
+
+def _process_is_alive(pid: int) -> bool:
+    if pid <= 0:
+        return False
+    try:
+        os.kill(pid, 0)
+    except ProcessLookupError:
+        return False
+    except PermissionError:
+        return True
+    except OSError:
+        return False
+    return True
+
+
+def _holder_description(lock_path: Path) -> str:
+    """Describe the lock holder without ever claiming one we cannot see."""
+    fields = _holder_fields(lock_path)
+    try:
+        pid = int(fields["pid"])
+    except (KeyError, ValueError):
+        return "holder has not recorded its identity yet"
+    if not _process_is_alive(pid):
+        return (
+            f"the lock file names pid {pid}, which is no longer running; "
+            "the current holder is unidentified"
+        )
+
+    parts = [f"pid {pid}"]
+    # The directory basename identifies the worktree; the full path is noise.
+    worktree = Path(fields["cwd"]).name if fields.get("cwd") else ""
+    if worktree:
+        parts.insert(0, f"worktree {worktree}")
+    description = f"held by {', '.join(parts)}"
+    command = fields.get("command")
+    if command:
+        if len(command) > MAX_REPORTED_COMMAND_CHARACTERS:
+            command = command[:MAX_REPORTED_COMMAND_CHARACTERS] + "..."
+        description += f" running: {command}"
+    return description
+
+
+def _acquire(
+    lock_file,
+    lock_path: Path,
+    timeout_seconds: float,
+    heartbeat_seconds: float,
+    *,
+    monotonic=time.monotonic,
+    sleep=time.sleep,
+    report=_report,
+) -> None:
+    start = monotonic()
+    deadline = start + timeout_seconds
     announced = False
+    next_heartbeat = 0.0
     while True:
         try:
             fcntl.flock(lock_file.fileno(), fcntl.LOCK_EX | fcntl.LOCK_NB)
             return
         except BlockingIOError:
+            now = monotonic()
             if not announced:
-                print(
+                report(
                     "swift-safe: another TBD worktree is compiling; waiting for "
-                    "the shared build slot",
-                    file=sys.stderr,
-                    flush=True,
+                    f"the shared build slot ({_holder_description(lock_path)})"
                 )
                 announced = True
-            if time.monotonic() >= deadline:
+                next_heartbeat = now + heartbeat_seconds
+            elif now >= next_heartbeat:
+                report(
+                    f"swift-safe: still waiting for the shared build slot after "
+                    f"{now - start:.0f}s of {timeout_seconds:g}s "
+                    f"({_holder_description(lock_path)})"
+                )
+                next_heartbeat = now + heartbeat_seconds
+            if now >= deadline:
                 raise TimeoutError
-            time.sleep(min(0.25, max(0.01, deadline - time.monotonic())))
+            sleep(min(0.25, max(0.01, deadline - now)))
 
 
 def main(arguments: list[str]) -> int:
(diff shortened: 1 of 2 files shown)


cheapsteak:
Both MEDIUM findings are real. The review noted it couldn't execute Python to confirm them empirically, so I did that first — both reproduce against the real script, and one is broader than the finding says. Fixed in 20997bf2.

**`correctness-1` — oversized pid crashes the waiter.** Confirmed, and the threshold is lower than "pathological": `os.kill` raises `OverflowError` at **2³¹**, not at some exotic 64-bit boundary, and `OverflowError` is not an `OSError` subclass, so none of the three except clauses caught it. Repro with a `pid=99999999999` record:

```
  File ".../scripts/swift-safe", line 214, in _process_is_alive
    os.kill(pid, 0)
OverflowError: Python int too large to convert to C int
exit status: 1
```

Fixed by catching `OverflowError` alongside `OSError` — no such process can exist, so it is simply not alive. I chose catching over range-screening the parse so there's no hardcoded `pid_t` width to be wrong about on another platform. Same record now yields exit 75 and:

```
swift-safe: another TBD worktree is compiling; waiting for the shared build slot (the lock file names pid 99999999999, which is not running; the current holder is unidentified)
```

I also changed that wording from "no longer running" to "is not running", since a pid that never existed was never running.

**`correctness-2` — nan/inf disable the heartbeat.** Confirmed, and the same hole is worse on the *timeout* variable, which the finding didn't reach: `TBD_SWIFT_LOCK_TIMEOUT_SECONDS=nan` makes `now >= deadline` permanently False, so the wait **never ends at all**. Measured against a 0.5s control:

```
  timeout=0.5 -> exited with status 75 (timed out correctly)
  timeout=nan -> STILL WAITING after 4s (unbounded)
```

That half predates this PR, but it's the same defect in the same helper and shares the one guard, so fixing it separately would have been artificial. `_positive_number` now rejects non-finite values, **before** the positivity test — ordering matters here, and I mutation-checked it: with the checks swapped, nan is caught by a `!= nan` fallback but `inf` still slips through. `-inf` was already rejected by `<= 0`.

New coverage, all mutation-checked (break it, watch it red, restore):

- impossible-pid description, plus an end-to-end assertion that the waiter exits 75 with no `Traceback` — revert the `OverflowError` catch and both go red
- non-finite rejection across both variables × `nan`/`inf`/`-inf`/`NaN`/`Infinity`, asserting stdout stays empty so it's rejected before SwiftPM is exec'd — drop the guard and it goes red

29 tests, ~5s, no build lock taken. Pyright clean on both files.

**On the two informational notes**, briefly, since neither was blocking:

- *Spec requirement* — agreed with the disposition. This restores the tool to its intended behavior (a wait that reports itself) rather than revising the system's theory, and CLAUDE.md exempts bug fixes. The heartbeat cadence is a tuning constant with its rationale stated inline and in the description, not a new contract.
- *Theory placement* — agreed the concern is weak here for the reason given: `scripts/swift-safe` is editable user-land, so the cadence is already a one-file edit away. The env var mainly exists so tests can compress the interval instead of spending real minutes.

**Still deliberately out of scope**, unchanged from the description: lock fairness (a behavior change needing its own design — the heartbeat now makes starvation visible, which is the right precondition for tackling it) and `DEFAULT_JOBS = 2` (needs measurement against Swift 6.3's memory profile, not a guess). The `run` defect stays a warning plus a documented sketch — a correct build-then-run needs the build in a child process, which is exactly the structure the `execv` exists to avoid; I'd welcome being shown wrong on that, but nothing in the review contests it.

The stray `pid_test.py` the diagnostics mention was in the review sandbox, not in this branch — the working tree here is clean.

cheapsteak:
**CI note, so the red `test` job in this PR's history doesn't get re-investigated from scratch.**

The first run of the Swift `test` job failed; the re-run passed with no code change. All five checks are green now.

The failure was a single test — `theFallbackRemovalArmsItsOwnRaisedTimeout()` in `Tests/TBDDaemonTests/WorktreeArchiveDeletionQueueTests.swift:118`. The run reported `2990 tests … 71 issues (including 70 known issues)`, so that was the one real failure.

**One-line reason:** its control step kills `git worktree remove` with a 1 ms subprocess timeout and then asserts the directory still exists, which loses the race whenever git completes the unlink before the kill lands — so it fails on a *fast* machine, not a slow one.

Worth being precise about why the test's own comment doesn't cover this: it says `runBoundedProcess` "makes that discriminating rather than racy", and that is true for *reporting* `.timedOut` (the `#expect(throws:)` immediately above) but not for *whether the directory survives*. Two different claims; only the first is guaranteed.

**Unrelated to this PR by construction.** Every change here happens *before* the `execv` in `scripts/swift-safe`. Once the lock is held the process is replaced by `swift`, so nothing added here is even resident while tests run — and the heartbeat only emits while waiting for a slot, on stderr. There is no path from this diff to a `git worktree remove` race.

Not fixed here on purpose — making that survival assertion deterministic is a real change and doesn't belong inside a build-wrapper PR. It's written up as a follow-up in the description with the mechanism, so whoever picks it up doesn't have to re-derive it.

PR #648, opened 2026-08-15
Fix file panel toolbar toggle state synchronization

Jephuff:
## What's broken

The top-right Changes-panel toolbar button can appear to do nothing even though Command-Shift-E still closes the panel. A pointer click should toggle the same panel state as the keyboard shortcut.

## Why it happens

`DetailSectionHostPager` creates the non-remote detail `NSHostingController` once so terminal views survive navigation. That stable host was initialized with `Binding` values from the surrounding representable, but its root view was intentionally never refreshed. The hosted panel could therefore keep rendering an old visibility value while the toolbar observed and changed the current `@AppStorage` value.

In the failing app, a physical click hit the expected SwiftUI toolbar button and ran its action, changing the current visibility value from `false` to `true`, while the hosted panel was already visibly open. The matching stale value made the click appear ineffective.

## When it broke

Commit `5997faa9` introduced the stable detail-section host while preserving the panel preferences as bindings across that hosting boundary. The recent PR-toolbar restructuring in `fef21eb1` did not change the file-panel button or cause the state divergence.

## What this PR does

- Makes the stable non-remote hosted view observe the file-panel `@AppStorage` preferences directly.
- Keeps the toolbar and hosted panel on independent reactive observers of the same centralized preference keys.
- Preserves the one-time `NSHostingController` creation that keeps terminal views alive.

## Evidence & verification

- Before the fix, live AppKit tracing showed a physical click hitting `SwiftUIAppKitButton` and running the toggle from `false` to `true` while the panel was already visibly open and remained open.
- After the fix, clicking the same top-right toolbar button closed the open Changes panel.
- `scripts/test.sh --filter DetailSectionHostPagerTests` — 8 tests passed in 1 suite.
- `scripts/swift-safe build` — passed (`Build complete!`).

### DIFF
--- a/Sources/TBDApp/ContentView.swift
+++ b/Sources/TBDApp/ContentView.swift
@@ -2,11 +2,16 @@ import AppKit
 import SwiftUI
 import TBDShared
 
+private enum FilePanelStorageKey {
+    static let isVisible = "filePanel.isVisible"
+    static let width = "filePanel.width"
+}
+
 struct ContentView: View {
     @EnvironmentObject var appState: AppState
     @EnvironmentObject var overlayCoordinator: TranscriptOverlayCoordinator
-    @AppStorage("filePanel.isVisible") private var showFilePanel = true
-    @AppStorage("filePanel.width") private var filePanelWidth: Double = 280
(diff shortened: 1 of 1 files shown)


cheapsteak (approved):
🙏 🙏

Issue #633, opened 2026-08-14
Slice 5 (Track B) — the hosted desk and delivery


cheapsteak:
> ### Context to load before implementing
>
> This issue is one part of the fleet-supervision redesign (#631). **The specs are normative; this issue is scope and constraints only.** Read them first — paths are repo-relative, so read them locally if you have the repo checked out.
>
> - [`docs/specs/2026-07-26-fleet-supervision-design.md`](https://github.com/cheapsteak/tbd/blob/main/docs/specs/2026-07-26-fleet-supervision-design.md) — **the design.** Normative for every slice.
> - [`docs/specs/2026-07-26-fleet-supervision-requirements.md`](https://github.com/cheapsteak/tbd/blob/main/docs/specs/2026-07-26-fleet-supervision-requirements.md) — requirements, and the Built/Enabled ratchet that governs what gets compiled.
> - [`docs/cli-supervise.md`](https://github.com/cheapsteak/tbd/blob/main/docs/cli-supervise.md) — the CLI reference.
> - [`docs/specs/2026-08-09-cross-session-messaging-design.md`](https://github.com/cheapsteak/tbd/blob/main/docs/specs/2026-08-09-cross-session-messaging-design.md) — relevant to the delivery adapter.
>
> Also read `CLAUDE.md` and `Tests/CLAUDE.md` in the repo root — they carry the build/test wrappers, the clock and date seams, the migration rule, and the no-screen-scraping rule, all of which bite on this arc.
>
> **#631 carries what the specs do not**: the dependency order, the compiled defaults, the coexistence contract with Nightwatch, and the standing traps.
>
> Where a slice causes spec drift, it **amends the spec in the same PR**, rewriting the passage to stand alone — no revision history, no dated notes.

---

**Depends on:** slices 3 and 4. Blocks 6, 7, 8, 10.

The first slice that can spawn a desk and deliver to it. Also the slice that owns the Nightwatch mutual exclusion.

## What to build

**Playbook resolution, per project, three tiers, first match wins, no merging**: operator copy → the project's designated repo file (`.agents/supervision.md`) → the shipped default. TBD resolves the path, hashes the bytes, and installs verbatim. **TBD never parses the playbook.** The seeding gesture writes the operator and repo levels exactly once.

[...]

PR #661, opened 2026-08-16
Feat: The playbook a supervisor stands on — resolved, hashed, and seeded once

cheapsteak:
## Summary

A supervision desk works for exactly one project, and it stands on exactly one
playbook — the prose that tells it what stuck means, when to escalate, and how
to behave under each of its project's modes. This PR is how that file gets
found.

An operator can now ask which playbook a project would resolve to, see the tier
it came from and the hash of its bytes, and take a customizable copy at either
the operator or the repository level:

```
tbd supervise playbook show acme-platform
tbd supervise playbook show acme-platform --content
tbd supervise playbook customize acme-platform
tbd supervise playbook customize acme-platform --repo-level
```

Nothing consumes the result yet — the desk that will stand on it arrives in the
next PR of this stack. What lands here is the resolution, the hash, and the
seeding gesture, so the desk has something to install when it exists.

## Why this shape

Resolution is **per project, three levels, first match wins, and levels are
never merged** — the operator's copy, then the project's designated repository
file, then the shipped default (design §5, "Where the files live"). A declared
project's operator copy lives beside its definition at
`~/tbd/supervision/projects/<name>/supervision.md` and its repository level is
whichever member repo the project designated; a singleton project's levels are
the existing per-repo ones, `~/tbd/repos/<id>/supervision.md` and that repo's
`.agents/supervision.md`.

**TBD never parses the playbook.** Compiled code resolves the path, hashes the
bytes, and holds them verbatim; the desk is the only structure-aware reader.
That is why a mode's *name* comes from `supervision.json` and never from the
prose, and why selecting a mode changes nothing about which file resolves.
There is no line splitting, no section extraction, no mode-name derivation
anywhere in the diff — if a later change needs structure out of this file, the
design boundary has been crossed and the change needs a spec, not a parser.

The seeding gesture writes each level **exactly once**. `customize` refuses a
file that already exists rather than overwriting it, and nothing reconciles
these files at startup. Tool-provided content lives only in the tier the tool
owns — the shipped default, which updates may freely replace.

Two decisions worth naming:

- **A level that exists and does not answer is reported, not silently skipped.**
  An empty or unreadable operator copy falls through to the next tier, which is
  correct, but falling through *silently* would swap an operator's conduct
  without saying so. `show` reports such a level with its path and its reason
  on stderr. A level nobody wrote is not reported — that is the ordinary case,
  not a finding.
- **No size cap on the read.** §5 says resolve, hash, install verbatim, with no
  ceiling, and a cap's failure mode is worse than the problem it solves: a large
  file quietly resolving to the shipped default would substitute conduct nobody
  chose, which is exactly the silence the reporting above exists to prevent.

Spec: [`docs/specs/2026-07-26-fleet-supervision-design.md`](docs/specs/2026-07-26-fleet-supervision-design.md)
§5 and §8; the standing-conduct mechanics are
[`docs/specs/2026-08-01-fleet-supervision-sweep-program-design.md`](docs/specs/2026-08-01-fleet-supervision-sweep-program-design.md)
§8. This is slice 5 of the fleet-supervision redesign (#633), which the specs
already cover — no new brainstorm was needed, and this PR causes no spec drift.

## What this PR does

- `Sources/TBDShared/Supervision/SupervisionPlaybook.swift` — `SupervisionPlaybookResolver`
  (`site(for:in:repos:)` → `resolve(_:)`), the resolved value carrying tier,
  path, bytes, conduct hash and any skipped levels. SHA-256 lowercase hex,
  following the `RemoteSessionIdentity` precedent.
- `Sources/TBDShared/SupervisionPlaybookContent.swift` — the shipped default:
  the universals from §5 (what stuck means, smallest intervention, escalate
  rather than guess, one intervention per agent per wake; the question, chat,
  permission-prompt, send-time-freshness and decision-depth universals) and the
  two baseline modes `attended` and `autonomous`. No commands, no bot names, no
  organization-specific content.
- `SupervisionStore.playbook(project:)` and `.customizePlaybook(project:level:)`,
  reached by the new `supervise.playbook` and `supervise.playbookCustomize` RPCs
  — the daemon stays the single writer of `~/tbd/supervision/`.
- `tbd supervise playbook show|customize` with `--json`, matching the
  `readout`/`ledger` rendering precedent.
- `docs/cli-supervise.md` gains a `tbd supervise playbook` section, and the
  Files list gains the operator-level playbook paths.

## Assumptions

- **A repo's playbook is read from its main checkout.** `SupervisionRepo` now
  carries `path`, fed from `Repo.path`. If a repo row's checkout is missing, the
  repository tier is skipped exactly as a missing file is — and `customize
  --repo-level` refuses rather than conjuring `.agents/` inside a directory that
  is gone.
- **Declared-ness decides which operator path applies.** `SupervisionProject`
  deliberately carries no `isSingleton` field, so the resolver reads it from
  `SupervisionFile.projects[name] != nil`. Everything downstream — resolved
  values, marks, modes — still collapses; only the storage layout differs. §5
  describes this as "the collapse property holding at the file layer, not a
  special case in the code", which is true of the resolved value and not of the
  path lookup; a spec sentence naming the branch would help the next reader, and
  rides with the next PR in this stack.
- **A repo joining a declared project moves where its operator copy is read
  from.** This follows from resolution being per project rather than per repo,
  and it is documented on the CLI page rather than worked around: reporting a
  level outside the project's resolution set would introduce a "stranded level"
  concept the design does not have.

## Every durable external resource needs a named reconciler

This PR seeds playbook copies at two levels, and they are **three different
questions** rather than one, so they get three answers.

- **The repo level** (`<checkout>/.agents/supervision.md`) writes into the
  operator's own git checkout. **An orphan of TBD's making cannot arise**: TBD
  neither owns that directory nor outlives it, git versions the file, and
  removing the repo from TBD leaves the checkout untouched. Reclaiming it would
  be the bug.
- **A singleton's operator level** (`~/tbd/repos/<id>/supervision.md`) adds one
  file to the existing per-repo family that already holds `notes.md`, the hooks
  and the settings overlay. Nothing in the daemon removes `~/tbd/repos/<id>/`
  today — `reposDir` is referenced only in `Constants.swift`. This is a **call
  site in a family whose treatment predates this gesture**, not a new kind of
  resource, so the doctrine's routine-call-site carve-out applies. The family's
  own gap is real and pre-existing; inventing an answer for it here would be
  scope this PR did not create.
- **A declared project's operator level** lives in
  `~/tbd/supervision/projects/<name>/`, which is a genuinely new directory
  family under `~/tbd` — the archetypal "new kind". Its permanence is now **a
  committed decision in design §7** with the reasoning attached, rather than an
  omission.

The binding half of that reasoning, and the reason no collector is named:
**"the project no longer resolves" is not an orphan signal.** §9 already
establishes that a member repo renamed or unregistered stops resolution with
the project's mark still standing and able to take effect again if the name
resolves once more. A sweep keyed on resolution would therefore reclaim a
*live* project's playbook, journal and proposals on a rename. The hosted desk's
collector rejects the same signal for the same reason. Two supporting facts:
one directory per human-declared project has no unbounded-growth mode — the
failure the doctrine exists to prevent is a machine-minted resource
accumulating per event — and the contents are the one thing §5 promises the
tool writes once and never touches again.

No process, tmux server, git ref, or scratch space is created anywhere in this
diff.

## Evidence & verification

`scripts/test.sh`: **6786 tests in 691 suites passed**. `swiftlint --strict`:
0 violations across 760 files. `scripts/swift-safe build`: clean.

Discriminating coverage, not just green:

- Each tier wins in order, and **no merging** — with the operator copy present,
  the repository file's content appears nowhere in the result.
- An empty or unreadable level falls through and is reported with its path and
  reason.
- `{ "operator": true }` policy skips the repository tier entirely.
- The hash is stable across resolutions and changes on a one-character edit.
- `customize` writes once and refuses the second call, leaving the first file's
  bytes untouched.
- `customize --repo-level` refuses a repo whose checkout is gone, rather than
  recreating it and then refusing forever after.
- Both new surfaces encode `path` as an explicit `null` rather than omitting it,
  pinned in `SupervisionSurfaceContractTests` alongside the other three
  supervision surfaces.

**Guards were mutation-checked, not just asserted.** The write-once refusal was
removed and its `.withoutOverwriting` write downgraded — both customize tests
went red, and the operator's edited bytes were observably clobbered by the
default. The empty-level fall-through was inverted so an empty file wins —
exactly two tests reddened, both the empty-file cases, with no blast radius
elsewhere. The missing-checkout guard was weakened to `if false` and its test
reproduced the full bad sequence (success, tree recreated, then permanently
"already exists"). All three restored and re-verified green.

[20260816-lucky-tern](https://cheapsteak.github.io/tbd/open/?worktree=727D12E2-18EB-4259-A6D8-71CDFEBA8DCE)

https://claude.ai/code/session_01PSjtUmDrA4nYbNCAFgz4f3

### DIFF
--- a/Sources/TBDCLI/Commands/SuperviseCommands.swift
+++ b/Sources/TBDCLI/Commands/SuperviseCommands.swift
@@ -375,25 +375,31 @@ struct SupervisePlaybookShow: AsyncParsableCommand {
 struct SupervisePlaybookCustomize: AsyncParsableCommand {
     static let configuration = CommandConfiguration(
         commandName: "customize",
-        abstract: "Copy the shipped playbook into the operator level, or with --repo the repo level",
+        abstract:
+            "Copy the shipped playbook into the operator level, or with --repo-level the repo level",
         discussion: """
             Writes the current shipped default and prints the path. Refused if \
             that level already exists: TBD writes it exactly once and never \
             touches it again, so an existing copy is edited directly.
 
-            Without --repo the copy goes to the operator level — beside a \
+            Without --repo-level the copy goes to the operator level — beside a \
             declared project's definition, or in a singleton's per-repo config \
-            directory. With --repo it goes to the project's designated repo \
-            file, .agents/supervision.md in that repo's main checkout, where it \
-            is committed and shared with everyone working in the repo.
+            directory. With --repo-level it goes to the project's designated \
+            repo file, .agents/supervision.md in that repo's main checkout, \
+            where it is committed and shared with everyone working in the repo.
             """
     )
 
     @Option(name: .long, help: "Project name")
     var project: String
 
-    @Flag(name: .long, help: "Write the project's designated repo file instead of the operator level")
-    var repo = false
+    /// Named for the *level*, not the repo. Everywhere else in this CLI
+    /// `--repo` takes a value naming one — `tbd worktree list --repo <name>`,
+    /// `tbd gc --repo <path>` — so a boolean `--repo` here would answer a
+    /// typed-out repo id with an opaque "unexpected argument".
+    @Flag(name: .long,
+          help: "Write the project's designated repo file instead of the operator level")
+    var repoLevel = false
 
     @Flag(name: .long, help: "Output JSON")
     var json = false
@@ -403,7 +409,7 @@ struct SupervisePlaybookCustomize: AsyncParsableCommand {
         let result: SupervisePlaybookCustomizeResult = try SocketClient().call(
             method: RPCMethod.supervisePlaybookCustomize,
             params: SupervisePlaybookCustomizeParams(
-                project: name, level: repo ? .repo : .operator),
+                project: name, level: repoLevel ? .repo : .operator),
             resultType: SupervisePlaybookCustomizeResult.self)
         if json {
             printJSON(result)
--- a/Sources/TBDCLI/Commands/SuperviseCommands.swift
+++ b/Sources/TBDCLI/Commands/SuperviseCommands.swift
@@ -19,6 +19,7 @@ struct SuperviseCommand: ParsableCommand {
             SuperviseStatusCommand.self,
             SuperviseMode.self,
             SuperviseProject.self,
+            SupervisePlaybook.self,
             SuperviseReadoutCommand.self,
             SuperviseBriefCommand.self,
             SuperviseLedgerCommand.self,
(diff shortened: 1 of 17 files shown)


cheapsteak:
Addressed in `6b96ccd7`. The finding was right, and the specific diagnosis was the useful part: my answer conceded an orphan *can* arise rather than arguing it cannot, and the design §5 sentence I leaned on is about not overwriting an operator's edits — a different question from reclamation. That is a fourth shape, and the doctrine takes three.

The reason it came out mis-shapen is that I gave one answer to what are really **three questions**, so it is now answered per level:

- **The repo level** writes into the operator's own checkout. An orphan of TBD's making genuinely cannot arise — TBD neither owns that directory nor outlives it, git versions the file, and removing the repo from TBD leaves the checkout untouched. Reclaiming it would be the bug. *(shape 2)*
- **A singleton's operator level** adds one file to `~/tbd/repos/<id>/`, beside `notes.md`, the hooks and the settings overlay. I checked rather than assumed: `reposDir` is referenced only in `Constants.swift`, so nothing in the daemon removes that directory today. This is a call site in a family whose treatment predates the gesture, not a new kind. *(carve-out)*
- **A declared project's operator level** lives in `~/tbd/supervision/projects/<name>/` — a new directory family under `~/tbd`, which is the archetypal new kind, and the one tier that genuinely had no valid answer. Design §7 now carries its permanence as a decision with the reasoning attached. *(shape 3)*

On why §7 records a decision rather than naming a collector — worth stating explicitly, because "just add an `OrphanGC` leg" is the cheap answer the doctrine expects and it is wrong here: **"the project no longer resolves" is not an orphan signal.** §9 already establishes that a member repo renamed or unregistered stops resolution with the project's mark still standing, able to take effect again if the name resolves once more. A collector keyed on resolution would reclaim a *live* project's playbook, journal and proposals on a rename. The hosted desk's collector in the next PR of this stack rejects the same signal for the same reason.

Two supporting facts, neither load-bearing on its own: one directory per human-declared project has no unbounded-growth mode — the failure the doctrine exists to prevent is a machine-minted resource accumulating per event, not per human decision — and the contents are the one thing §5 promises the tool writes once and never touches again.

Putting it in the spec rather than only in this description is deliberate: slices 6/7/8 add the journal, the proposals doc and the project's own programs to that same directory, and they should inherit a settled question instead of re-litigating it.

PR #663, opened 2026-08-16
Feat: A project turned on gets a supervisor — the hosted desk, spawned and reclaimable

cheapsteak:
> Second of four PRs for slice 5 (#633): **1** the playbook (#661, merged) →
> **2 the desk exists** → **3** the desk receives → **4** the desk acts.
> Rebased onto `main` now that #661 has landed, so this targets `main`
> directly.

## Summary

Turning a project on now gives it a supervisor. `tbd supervise on <project>`
ensures a hosted desk — a scratch space with a Claude session in it, standing on
that project's playbook, knowing which project it works for — and
`tbd supervise off <project>` stands that desk down without disposing of it, so
the context it cost real tokens to build survives until the next `on`.

**The desk receives nothing yet.** Delivery, the compiled header, the delivery
ledger line and the `supervisor.live` flip are the next PR's. Until they land,
`supervisor.live` stays `false` and `supervise brief` keeps answering
`no-live-supervisor` — jointly true, because nothing can in fact be delivered to
a desk today. `SupervisionReadoutBuilder` and `SupervisorBriefingDeliverer` are
byte-identical to the base branch, deliberately: a live desk with `live: false`
would make every program's liveness check lie, and the reverse would make the
readout the liar, so they flip together in one commit or not at all.

## Why this shape

**A desk is tracked by ID, never by a display string.** A name-keyed binding is
unreclaimable the moment someone renames the thing it names, which is how
durable resources have leaked in this codebase before. A test asserts the
persisted entry's key set is exactly `{terminal, worktree, spawnedAt,
conductHash}` and that no `name` or `displayName` key exists, so the property is
enforced rather than intended.

**`desks.json` is a new file because the design had nowhere to put this.** §7
pins supervision at exactly one DB column (the fleet brake) with everything else
in files a human can open; `supervision.json` holds operator *selections* only,
and its `supervisors` map is the **appointed** binding, not the hosted desk.
Nothing in the spec said where a hosted desk's terminal ID lives. A new
TBD-owned `~/tbd/supervision/desks.json` keeps the one-column property, stays
operator-readable, and stays out of `supervision.json` precisely because it is
derived state rather than a choice anyone made. §7 is amended to name it.

**Spawn delivers nothing.** No message, nudge, or opening prompt — the opening
material rides the first briefing, so a quiet project's desk idles at zero token
cost. This is a tested property, not a detail.

**`off` stands the desk down; nothing in this PR disposes of a desk.** The mark
is a delivery precondition, so a stood-down desk simply receives nothing while
keeping its context. A stood-down project still appears in the readout —
observability is never withheld.

**The desk forces the Claude adapter.** Standing conduct rides
`--append-system-prompt`, which only the Claude spawn path carries; a Codex desk
would launch with no conduct at all and nothing would say so. Design §9 makes
supervisor capability a per-adapter qualification, so this is that rule applied
rather than a preference — and a Codex-preferring fleet is told in the log
rather than discovering it by reading a tab label.

Spec: [`docs/specs/2026-07-26-fleet-supervision-design.md`](docs/specs/2026-07-26-fleet-supervision-design.md)
§5, §7, §9; standing-conduct mechanics in
[`docs/specs/2026-08-01-fleet-supervision-sweep-program-design.md`](docs/specs/2026-08-01-fleet-supervision-sweep-program-design.md)
§8. Slice 5 is already specced by that set, so no new brainstorm was needed;
the drift this PR does cause is amended in the same diff (below).

## What this PR does

- **`desks.json`** — `SupervisionDesks{File,Entry,Store}`, atomic
  temp-file → `fsync` → `rename(2)` → directory `fsync`, matching
  `SupervisionFileStore`.
- **`SupervisionDeskManager.ensureDesk`** → `.appointed` / `.danglingBinding` /
  `.resumed` / `.spawned` / `.replaced(successor:predecessor:)`. An appointed
  binding is never silently replaced by the hosted default; a dangling one is
  loud. Topology comes from `SupervisionStore.deskInputs(project:)` so
  `supervision.json` keeps exactly one reader.
- **The playbook as a standing layer** — a `TBD_PROMPT_SUPERVISION` layer in
  `SystemPromptBuilder`, plus `TBD_PROJECT` injected following the
  `TBD_WORKTREE_ID` precedent.
- **The supervision skill** in the plugin dir. Note this dir is **fleet-shared**
  — every Claude spawn already gets `--plugin-dir` pointing at it, so the skill
  is visible fleet-wide exactly as the Nightwatch skill is. What makes a desk a
  desk is the standing layer plus `TBD_PROJECT`, never exclusive skill access.
- **Desk lifecycle ledger lines** — spawn, and replacement linking successor to
  predecessor.
- **The Nightwatch mutual exclusion**, both directions.
- **`SupervisionDeskCollector`**, a new `OrphanGC` leg.
- **`tbd gc supervision-desks on|off`** and `docs/orphan-gc.md` coverage.

## Flag & rollout

`gc_supervision_desks_enabled` — daemon config column added by
`v80_config_gc_supervision_desks`, **default OFF**. It carries **no SQL
default**, so NULL ("nobody has chosen") stays distinguishable from an explicit
`0`/`1`, and the shipped default lives in exactly one place,
`?? Config.gcSupervisionDesksEnabledDefault`. Graduation is a one-line change to
that constant, reaching everyone who never touched the toggle while preserving
every explicit opt-out — no forcing migration.

Enable for the soak with `tbd gc supervision-desks on`. Graduate once a soak
shows the collector reclaiming dead desks and never touching live ones. Only the
*reclaiming* is gated; nothing else in this PR is behind a flag, because
spawning a desk happens on an explicit operator gesture.

## Every durable external resource needs a named reconciler

A desk is a scratch space **plus** a live process — the shape that produced this
repo's measured leaks. The orphan has two shapes and each has an owner.

- **A desk that was recorded and then died** → `SupervisionDeskCollector`, a leg
  of the hourly `OrphanGC` sweep. It drops the `desks.json` entry and hands the
  scratch worktree to the archive path the sweep already runs. Its reap decision
  is **liveness and nothing else**: a desk with a process running in its
  directory is kept whatever its project or its rows say. Every failure
  direction is toward keeping — an unreadable file, an unfetchable row, a path
  that does not look like a desk's, all keep.
- **A spawn that failed before the record was written** leaves nothing for that
  collector to enumerate, so the spawn path **archives the scratch row itself**
  on every failing exit, which puts the directory under the sweep's
  deletion-queue leg. Best-effort, as create-time cleanup always is; the
  standing guarantee is still the sweep.

I verified the existing sweeps' boundary rather than assuming it: `OrphanGC`'s
agent-worktree leg iterates `db.repos.list()` and so only walks repo-backed
worktrees, and its archived legs only touch rows already `.archived`. So a
**live** desk was never at risk from them (correct), and an **orphaned** one was
reclaimed by nothing (the gap this closes).

**A project that stops resolving does not trigger a reap**, deliberately. Design
§9 is explicit that a topology gesture ending a project takes its mark, its mode
selection and its supervisor binding — and that no coverage gesture disposes a
desk. Killing a live agent because its project was renamed is precisely what the
doctrine forbids. Once the desk is not live it is reclaimed regardless, so the
check would gate nothing while adding a second reader of `supervision.json`.

**One residual is stated rather than papered over**: a daemon *crash* (as
opposed to a thrown error) between creating the scratch space and writing the
record leaves an unreferenced scratch space that nothing reclaims. It is not
silent — the row is `.active` and visible in the sidebar.

## Assumptions

- **The two exclusion gates are not mutually atomic.** `supervise on` reads the
  Nightwatch mode before writing the mark; `nightwatch.setMode` reads the marked
  projects before writing the mode. Two RPCs issued in the same instant could
  both pass and leave both paths on. Closing it means a lock shared across two
  handlers, which is a design call rather than a fix, and it takes two operator
  gestures at the same moment to reach. If both paths ever are found on
  together, this is why.
- **Desk liveness is a process-in-directory fact**, read through the sweep's
  existing `lsof` cwd set with the same prefix-boundary matching every other leg
  uses. If a desk's process is running with a cwd outside its scratch space, it
  reads as dead.

## Evidence & verification

`scripts/test.sh`: **6846 tests in 697 suites**. `swiftlint --strict`: 0
violations in 766 files. `scripts/swift-safe build`: clean.

Both branches of every new conditional are covered. Highlights:

- Spawn delivers nothing; a stood-down project still appears in the readout.
- Both Nightwatch refusals — and **the wedge state**: with Nightwatch running
  *and* a project marked, setting the mode `off` still succeeds. `off` is
  refusable in neither direction, so an operator can always get out.
- A live desk survives a sweep; a dead one is reclaimed; the flag ships off.
- The three-state flag column: a pre-migration row reads NULL, NULL follows a
  changed default, an explicit `false` survives one.
- A replacement's predecessor is archived, and only after the successor is
  safely recorded.
- A reap landing mid-spawn survives the record write — the test drops a
  *different* project's entry during the spawn and asserts that change is not
  clobbered, so "the new entry is present" cannot pass it.

**Guards were mutation-checked, not merely asserted.** Deleting the
`TBD_PROMPT_SUPERVISION` append from `build()` turned the two composed-string
tests red while the map-only test **stayed green** — `promptLayers` and `build()`
are separately maintained, so a test asserting only the env map would have
shipped a desk with no conduct. Each of the seven review fixes was likewise
reverted, observed red, and restored: the predecessor leak, the record-write
exit, the exact-path liveness match, the failed-read-reads-as-gone collapse, the
`forget` mismatch, and the terminal broadcast.

Two bugs surfaced from writing the tests rather than from reading the code: a
check-then-create race in scratch-space allocation where two concurrent ensures
could pick the same name, and a replacement that orphaned its predecessor's
scratch space permanently — one leaked directory per replacement.

[20260816-lucky-tern](https://cheapsteak.github.io/tbd/open/?worktree=727D12E2-18EB-4259-A6D8-71CDFEBA8DCE)

https://claude.ai/code/session_01PSjtUmDrA4nYbNCAFgz4f3

### DIFF
--- a/Sources/TBDCLI/Commands/GCCommands.swift
+++ b/Sources/TBDCLI/Commands/GCCommands.swift
@@ -6,7 +6,10 @@ struct GCCommand: ParsableCommand {
     static let configuration = CommandConfiguration(
         commandName: "gc",
         abstract: "Orphan GC: list reaps, restore a reaped agent worktree, trigger a sweep",
-        subcommands: [GCList.self, GCRestore.self, GCSweep.self, GCProfileDirs.self]
+        subcommands: [
+            GCList.self, GCRestore.self, GCSweep.self, GCProfileDirs.self,
+            GCSupervisionDesks.self,
+        ]
     )
 }
 
@@ -28,6 +31,28 @@ struct GCProfileDirs: AsyncParsableCommand {
         try SocketClient().callVoid(method: RPCMethod.configSetGCProfileDirsEnabled,
                                     params: ConfigSetGCProfileDirsEnabledParams(enabled: enabled))
         print("Profile-dir GC \(enabled ? "enabled" : "disabled").")
+    }
+}
+
+/// The soak switch for the supervision-desk collector. It archives the scratch
+/// space of a hosted supervisor whose session is gone and drops its record, so
+/// it ships off and is opted into by hand — the same shape `profile-dirs` uses.
+struct GCSupervisionDesks: AsyncParsableCommand {
+    static let configuration = CommandConfiguration(
+        commandName: "supervision-desks",
+        abstract: "Enable or disable reclaiming orphaned supervision desks (default off)")
+    @Argument(help: "on | off") var state: String
+    mutating func run() async throws {
+        let enabled: Bool
+        switch state.lowercased() {
+        case "on", "true", "enable": enabled = true
+        case "off", "false", "disable": enabled = false
+        default: throw ValidationError("Expected 'on' or 'off', got: \(state)")
+        }
+        try SocketClient().callVoid(
+            method: RPCMethod.configSetGCSupervisionDesksEnabled,
+            params: ConfigSetGCSupervisionDesksEnabledParams(enabled: enabled))
+        print("Supervision-desk GC \(enabled ? "enabled" : "disabled").")
     }
 }
 
--- a/Sources/TBDDaemon/Actuation/ActuationSurface.swift
+++ b/Sources/TBDDaemon/Actuation/ActuationSurface.swift
@@ -221,6 +221,15 @@ enum ActuationRail {
     /// judge session, and the close that kills that session's windows — all
     /// acts the desk performs on its own, with no RPC behind them.
     static let nightwatchDesk = "nightwatch-desk"
+    /// `SupervisionDeskManager` spawning a project's hosted desk. `on <project>`
+    /// is the gesture behind it, but the desk is TBD's own arrangement rather
+    /// than something the caller asked for by name — the operator asked for
+    /// coverage, and a desk is how coverage is carried out — so it is a rail
+    /// with its own row rather than a branch of the `supervise.*` door. One row
+    /// per spawn, naming the scratch worktree and no terminal, the shape
+    /// `worktree.revive` uses: the terminal is minted inside
+    /// `spawnPrimaryTerminals`.
+    static let supervisionDesk = "supervision-desk"
     /// The boot-time (and post-`cleanup`) reconcile sweep: it kills the windows
     /// of worktrees that left disk, parks sessions whose window is gone, and
     /// reaps orphaned windows and dead servers. One row per act — each is an
--- a/Sources/TBDDaemon/Daemon.swift
+++ b/Sources/TBDDaemon/Daemon.swift
@@ -639,15 +639,32 @@ public final class Daemon: Sendable {
         // Skipped in mock mode like every other rail: a fixture render must not
         // read the operator's real supervision file, and `supervise.*` refuses
         // with a named condition there rather than answering from it.
+        //
+        // The ledger writer is built once and shared by the store and the desk
+        // manager: it is a single-writer append-only file, and two writers on
+        // two descriptors would each be checking a different one against the
+        // path.
+        let supervisionLedger = SupervisionLedgerWriter(
+            path: TBDConstants.supervisionLedgerPath)
         let supervisionStore: SupervisionStore? = mockMode == nil
             ? SupervisionStore(
                 files: SupervisionFileStore(),
-                ledger: SupervisionLedgerWriter(path: TBDConstants.supervisionLedgerPath),
+                ledger: supervisionLedger,
                 fleet: DatabaseSupervisionFleetReader(db: database))
             : nil
         if let supervisionStore {
             self.supervision = supervisionStore
             rpcRouter.supervision = supervisionStore
+            // The hosted-desk lifecycle (design §9). Same mock-mode gate as the
+            // store: a fixture render must never spawn a session.
+            rpcRouter.supervisionDesks = SupervisionDeskManager(
+                db: database,
+                lifecycle: lifecycle,
+                tmux: tmux,
+                desks: SupervisionDesksStore(),
+                ledger: supervisionLedger,
+                actuationLog: actuationLog,
+                subscriptions: subs)
             do {
                 try await supervisionStore.load()
             } catch {
--- a/Sources/TBDDaemon/Database/ConfigStore.swift
+++ b/Sources/TBDDaemon/Database/ConfigStore.swift
@@ -62,6 +62,11 @@ struct ConfigRecord: Codable, FetchableRecord, PersistableRecord, Sendable {
     /// Resolve it through `Config.gcProfileDirsEnabledDefault`, never through
     /// `?? false`.
     var gc_profile_dirs_enabled: Bool?
+    /// Gate for the supervision-desk collector. Tri-state for the same reason
+    /// and in the same shape as `gc_profile_dirs_enabled`: no SQL default, so
+    /// `nil` means "never chose". Resolve through
+    /// `Config.gcSupervisionDesksEnabledDefault`, never through `?? false`.
+    var gc_supervision_desks_enabled: Bool?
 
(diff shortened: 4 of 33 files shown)


cheapsteak:
Both addressed — `8222c80c` (HIGH) and `689a6f98` (MEDIUM). Full suite **6848 tests in 697 suites**, lint 0/766.

### HIGH — `abandonPredecessor` archiving a live desk

Right, and worth saying plainly: this was a bug *the previous review round introduced*. `abandonPredecessor` was the fix for "a replacement orphans its predecessor forever," and in closing a leak it opened a worse hole — archiving on a stale liveness verdict, with `WorktreeLifecycle+Reconcile` then killing the live window as untracked on the next pass.

The fix keeps **one** liveness notion rather than adding a second, by making the existing `isLive` three-state underneath: a new private `liveness(_:)` returning `.live` / `.gone` / `.undetermined`, with `isLive == (liveness == .live)` so the *replace* decision is byte-for-byte unchanged and still fails toward "not live". The DB reads moved from `try?` to `do`/`catch`, so a read that threw is `.undetermined` rather than collapsing into "gone" — the same distinction `SupervisionDeskCollector.WorktreeLookup` already draws.

`abandonPredecessor` now re-reads immediately before archiving and hands the space back **only** on `.gone`. Your path 2 is closed at the root rather than narrowed: a parked session counts as `.undetermined`, not gone, because it is wakeable by design.

**The resulting tradeoff is stated in the doc comment rather than hidden.** When the recheck declines, that scratch space stays `.active` and is enumerated by nothing — the successor holds the project's `desks.json` key — so no sweep reclaims it. It is a visible `.active` row an operator can close, the same shape as the crash-window residual, and strictly better than killing a live session. Making it reclaimable would need `desks.json` to enumerate superseded entries plus a second collector input; that is a record-schema change, so it is flagged rather than smuggled in.

Two things I did not claim: the `deskLogger.error` line itself is not mechanically asserted (there is no log-capture seam in this tree), so the operator notification is the observable stand-in; and `TmuxManager.windowExists` still answers `false` both for "no window" and "tmux unreachable", so an unreachable tmux reads `.gone` — the same evidence this type has always decided on, documented as the bound on what `.gone` proves.

Tests: a predecessor live at the recheck keeps its space (driven through a seam where the two reads genuinely differ, so it is the `isLive()` false-negative path and not a restatement); the `.undetermined` branch driven by parking the terminal through the production path; and the existing dead-at-recheck test strengthened to assert **no** notification, so the decline path cannot fire silently on the happy path. Mutation check: hardcoding the recheck to `.gone` reddened both new tests with `(old.status → .archived) == .active` while the other 15 stayed green.

### MEDIUM — the check-then-act race

Comment only, at both sites, per your read that the window is narrow. It names what the window is, what state results, and that closing it needs a lock shared across two handlers — a design decision about where supervision's writes serialize, deliberately not made in this PR. Wording is consistent between the two handlers and with the Assumptions bullet.

### One thing for whoever reviews the next PR in this stack

PR 3 introduces a shared `SupervisionSessionLiveness` predicate, and merging it with this fix is **semantic, not textual**: that predicate answers `.notLive` for a parked session, which is correct for delivery and wrong for archiving. The two consumers need opposite treatment of "parked" — §9 requires a hibernating *appointed* session to read as a dangling binding loudly, while a parked *hosted* desk must not be archivable. Resolving that conflict by taking either side wholesale would reintroduce this finding, so it gets handled deliberately rather than inside a rebase.

PR #664, opened 2026-08-16
Feat: A box spawn lands where you put it -- one-click remote lanes that nest, and stay nested

cheapsteak:
## Summary

Spawning an agent session onto a remote machine should feel like making a local worktree: you fan out lanes, they appear in the tree under the lane that spawned them, each with a PR badge, and you never think about the transport. Until now it didn't. A remote lane could only be started from a right-click most people never try, it always landed at the top level instead of nested, and the create form arrived with a blank required field and a value the provider rejected.

The measurement that motivated this: every remote session that had ever existed on the box was a test probe. Nobody had used the path for real work. This PR closes the gap between "the mechanism exists" and "you would actually use it".

Now: the `+` menu offers a remote session (repo header and nested, for repos with a registered provider); selecting it creates in one click when every required answer is already knowable, falling back to the form only when it genuinely isn't; per-repo and machine-wide defaults answer the recurring questions; and a session spawned from inside a lane nests under that lane and stays where you put it.

## Why this shape

**One malformed session must not blind TBD to the fleet.** `meta` is contract-defined as a flat string-to-string map, and a provider shipped a nested object in it. Because sessions decoded as a plain array, that one bad value failed the whole envelope and took every remote session offline for ten hours — silently, because the poll failure logged at `.debug`. Decoding is now lenient per field and per element: a wrong-typed value costs its own key, an undecodable element costs its own session, and neither costs the inventory. `id` stays fatal to its element, because a session with no identity cannot be tracked.

**Adoption may give a row its first parent, but never its second.** The contract previously said a later parent value is always ignored. That was wrong in one direction — a row adopted before its parent was knowable stayed orphaned forever — and the naive fix was wrong in the other, because `parentWorktreeID == nil` also describes a row the user deliberately un-nested, so re-reading a static stamp would silently undo their gesture within one poll. `remote_parent_assigned` distinguishes "never offered a parent" from "has none right now", so healing happens at most once per row and a user gesture always wins.

**Contract change: `slug` is now a well-known `create_params` name.** One-click needs every required field answerable; the live provider requires `slug` with no default, so without this the feature could never fire. "Well-known" is a caller-side prefill convention and the provider already declares the field, so nothing changes on the provider side. TBD generates it with the same generator that names local worktrees.

**Per-repo defaults are keyed generically, not by field name.** `permission_mode` is not a contract well-known name, so a column named for it would bake one provider's vocabulary into TBD's schema. Instead a small map is stored per repo and globally, keyed by whatever the provider declares, and replayed without interpretation — which also gives `cmd` and anything else defaults for free. Values are re-validated against the field as currently declared, so a stored enum value the provider has since dropped degrades to the next level rather than reproducing the create failure this PR also fixes.

**Precedence follows the model-profile precedent** (`repo.profile_override_id` → `config.default_profile_id`): repo map → ambient prefill → global map → provider default. Note the ambient prefill sits *inside* the repo tier rather than below both stored levels. This is deliberate: the ambient `repo` value is repo-scoped evidence — the user clicked *that* repo's `+` — so letting a machine-wide stored `repo` outrank it would create sessions against the wrong repository on a multi-repo machine.

**Specs.** Two, both committed here. The creation-path decisions — one-click, the four-level precedence chain, `repo`'s refusal to fall through, the generic defaults map, and the `slug` contract addition — are in [`docs/specs/2026-08-20-remote-lane-creation-design.md`](docs/specs/2026-08-20-remote-lane-creation-design.md). The one-shot parent policy is a theory change rather than a restoration — the prior rule was that adoption never reparents a row, and this relaxes it to "never a second time", arbitrated by a new persisted marker with its own lifecycle — so it is written up in [`docs/specs/2026-08-22-remote-lane-parent-healing-design.md`](docs/specs/2026-08-22-remote-lane-parent-healing-design.md): why one assignment and not zero or unlimited, the marker's full lifecycle, why the legacy backfill marks every pre-existing remote row, why no flag, and the rejected alternatives. The decode leniency and the create-form bugs need no spec: they are bug fixes that restore the system to its existing theory.

## What this PR does

- **`TBDShared`** — lenient per-field and per-element session decoding; `parentWorktreeID` on `RemoteCreateParams`; `remoteParentAssigned` and `remoteCreateDefaults` on the models.
- **Adoption** — first-parent-only healing behind `remote_parent_assigned`, which `WorktreeStore.move()` also spends whenever a user reparents a remote row (both to a new parent and back to root), so a lane first nested by hand and then un-nested is not re-nested by the next poll; a create-time parent override that beats the stamp; the same `validateParent` rules as `move()` (self, cycle, main, archived), with a refusal degrading to a top-level row rather than failing a create whose session already exists remotely.
- **Diagnosability** — a whole-provider poll failure now logs at `.error` rather than `.debug`; skipped sessions and dropped `meta` keys log with the provider named.
- **Auto-archive** — both adoption paths now disarm the parent's auto-archive-on-merge, matching every other reparent.
- **App** — remote-session entry on the repo-header and nested `+`; a provider drill-in page; one-click via `RemoteCreateFormLogic.launch`; enum-aware repo prefill; visible labels on string and int create fields; per-repo and global defaults editors built from `describe`, in which `repo` is repo-scope only — the global editor renders it as an explanation rather than a control, and a repo's own `Auto` entry for it names the repository the session starts in rather than promising a fall-through `resolveString` refuses.
- **Migrations** — two `.sql` files under the scheme #688 introduced: `20260816122608_worktree_remote_parent_assigned` (whose backfill marks **every** pre-existing remote row, see below) and `20260816181509_remote_create_defaults`. The frozen Swift block in `Database.swift` is untouched — that file is byte-identical to `main`.
- **Docs** — `docs/remote-provider-contract.md` updated for the parent-key semantics, `meta` degradation, and `slug`.

## Reconciler

No new durable external resource. Both migrations add columns to existing tables; the remote `worktree` rows this touches are already covered by `WorktreeLifecycle+Reconcile`, and the sessions themselves are provider-owned and reconciled by the existing `list` convergence. Nothing here creates a git ref, tmux object, process, or file that outlives its request.

## Flag & rollout

None, deliberately. The late-parent assignment is the only behavior that acts without a user gesture, and it is bounded rather than autonomous: `remote_parent_assigned` permits at most one assignment per row for the row's lifetime, every existing parenting guard still applies, a refusal degrades to top-level, and any user gesture afterwards is permanent. It introduces no standing background actor, so there is no soak state a flag would protect. Everything else is user-initiated UI. The same reasoning, at length, is in [`docs/specs/2026-08-22-remote-lane-parent-healing-design.md`](docs/specs/2026-08-22-remote-lane-parent-healing-design.md).

**This carve-out is taken with maintainer sign-off, not by default.** CLAUDE.md's
flag rule names two shapes of sufficient justification — gated behind an explicit
user action, or extending already-flagged behavior — and this is neither. The rule
introduces those with "e.g.", so the boundedness argument stands as a third form,
and the maintainer has accepted it as one for this behavior specifically. The spec
records that acceptance and its scope, so the exception is visible to anyone
deciding whether to reach for it again rather than being inherited silently.

## Assumptions

- A provider's `create_params` field names are stable enough to key stored defaults on. If a provider renames a field, its stored default silently stops applying — the value is dropped rather than misapplied, so this degrades safely.
- The parent stamp a provider reports is static for a session's lifetime. If a provider began mutating it, the marker means TBD would honor only the first value.
- One-click assumes a provider declaring a required field TBD cannot answer would rather show the form than receive a guess.

## Evidence & verification

**The original defect, reproduced and fixed live.** The mirror was frozen for over ten hours (`max(lastSeen)` stuck at `04:58:52`, sampled across four consecutive polls while the provider answered fine by hand). After deploying this branch it recovered on the first poll (`15:09:39`) and the stamped session nested under its parent in the same convergence, while four sessions predating the stamp correctly stayed top-level.

**The un-nest regression, reproduced and fixed live.** `worktree reparent --root` on the nested lane, then three successful polls with the provider still emitting the stamp: it stayed at root. Before the fix it re-nested within one poll interval.

**The backfill marks every pre-existing remote row, not just the parented ones.** An earlier revision marked only rows that already had a parent, which left a hole the review gate caught: before this PR, `move()` recorded nothing when it took a parent away, so a lane the user had deliberately un-nested is indistinguishable from one nobody could ever place. Since the provider's parent stamp is static and arrives again on every poll, the first poll after upgrade would re-nest the deliberately-un-nested lane — this PR's own regression, reached through legacy data instead of a live gesture. No pre-existing row is eligible for late assignment; only rows minted after the migration are, and their `0` is written explicitly by the insert path rather than taken from the schema default. The cost is bounded in one direction: the marker can only ever withhold a parent adoption would have imposed, never impose one.

`RemoteSessionAdoptionTests.aLegacyTopLevelLaneIsNotNestedByTheFirstPollAfterUpgrade` pins it — it migrates through the frozen block, writes the legacy row by raw SQL, reopens through the production path so the real backfill runs, then polls with the stamp still present. Against the previous backfill it fails on the parent edge, not just the marker.

**Discriminating tests, including mutation checks.** Several new tests could only "fail first" as compile errors, because they name symbols that did not exist before the implementation — so the three guards carrying real weight were mutation-checked instead. Replacing the enum projection with blind replay reddened exactly the enum-drop guard while both precedence tests held; reordering the candidate array reddened exactly the two precedence guards and nothing else. Baseline: 99 tests in 5 suites, with the count predicted before the run.

**Three false-pass modes ruled out on every run**, because `scripts/swift-safe` exits 0 on both lock timeout and source-modified-during-build, and a `--filter` matching nothing exits green. One killed run in this work produced exactly that shape — clean error/timeout/clobber counts and exit 0 from a run that executed zero tests — and was caught only by the missing test count.

**Not verified: the UI has not been seen running.** The repo-header `+` entry was confirmed visually by the repo owner; the nested `+`, the provider drill-in page, the create sheet's labels and enum prefill, and the settings editors are covered by unit tests on extracted pure logic but have not been observed in the running app.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=92B3ECD3-C1E7-402F-8D90-3C8E28FB5524)


cheapsteak:
Addressing the MEDIUM on flag policy (`docs/specs/2026-08-22-remote-lane-parent-healing-design.md:178`).

The finding is fair and I haven't argued it away. It asked for an explicit maintainer sign-off that "bounded to one shot, reuses existing validation" satisfies the flag rule, rather than the call being made silently — and it correctly noted that CLAUDE.md introduces its two example justifications with "e.g.", leaving room for a third form.

The repo's maintainer has made that call explicitly. It is now recorded in two places rather than left implicit:

- the spec's "Why no feature flag" section, which states that the two named shapes don't apply, that boundedness stands as a third form, that a human took the exception deliberately, and what it is scoped to (a write that happens at most once per row and reuses `validateParent` rather than relaxing it);
- the PR description's "Flag & rollout" section, so the exception is visible without opening the spec.

No behavior changed — the sign-off is the deliverable, since the finding was about who gets to make the call and whether it's visible, not about the mechanism.

Also in this push: rebased onto `origin/main` (24 commits, 0 behind). Two conflicts, both resolved keeping *both* sides rather than picking one — `SQLMigrationLoaderTests` (main's hand-written manifest and this branch's directory-derived bundle check catch different failures, so both are kept, with the manifest extended to all three migrations) and `ModelProfileListResult` (main's `autoCreateNotesEnabled` and this branch's `globalRemoteCreateDefaults` are both additive). `Database.swift` verified byte-identical to `origin/main`.

PR #665, opened 2026-08-16
Feat: Stop the first-message "Send immediately" box forgetting every time

cheapsteak:
## Summary

Creating a worktree opens a sheet for the first message you want the agent to work on, with a **Send immediately** checkbox: ticked, TBD presses Return and the agent starts as soon as the text lands; unticked, the message waits in the composer for you to read and send.

That box started unticked every single time. Someone who wants their first messages to send — a stable preference, not a per-message judgement — had to re-tick it on every worktree they created. A checkbox that forgets reads as broken rather than cautious.

Now the choice sticks. Tick it once and the next composer opens ticked, and there's a matching row in **Settings ▸ General ▸ Worktrees** so the preference is visible and reversible without creating a worktree to get at it.

## Why this shape

Spec: [`docs/specs/2026-08-16-send-immediately-preference-design.md`](docs/specs/2026-08-16-send-immediately-preference-design.md).

The OFF default was a deliberate decision in the [original queued-prompt design](docs/specs/2026-08-10-queued-prompt-on-create-design.md), and it survives here intact: submitting starts a turn nobody watched begin, and it's the half TBD can't report on afterwards, since an unsubmitted draft leaves no machine-readable trace. That argues for making the operator opt *in*. It doesn't argue for making them opt in repeatedly. An opt-in that persists is still an opt-in.

Three decisions worth naming:

- **The checkbox writes the preference itself, rather than Settings owning it and the checkbox being a one-shot override.** The gesture "I want this on" already happens at the checkbox; sending someone to a Settings window to make it stick adds a step to the exact interaction this exists to shorten. The cost is that ticking the box and then pressing Escape still changes the preference — accepted, because the next composer opens *visibly* ticked before anything can send.
- **`ParkedPromptReadbackView` deliberately does not read the key.** It seeds from the `submit` bit stored with that particular parked message, so it can never misreport what delivery will actually do for the message on screen. Wiring it to the preference "for consistency" is explicitly rejected in the spec, and the enum's doc comment now says so at the edit site.
- **The read goes through `object(forKey:) as? Bool ?? shippedDefault`, never `bool(forKey:)`.** That keeps "never chose" distinct from "chose off". Collapsing them is what left `auto_hibernate_enabled` unflippable: once absent and explicitly-off read alike, a later change to the shipped default either reaches nobody or overrides deliberate opt-outs.

## What this PR does

- `QueuedPromptModal.swift` — `@State` becomes `@AppStorage` on a new `queuedPromptSendImmediately` key; adds the key, a `resolveSendImmediately(defaults:shippedDefault:)` helper stating the three-state contract, and a `.help` clause telling the operator the tick is remembered.
- `SettingsView.swift` — a "Send first messages immediately" row in the Worktrees section, bound to the same key.
- `SendImmediatelyPreferenceTests.swift` — six cases: three on the resolver, one per state it must keep distinct, plus three on the mechanism the shipped views actually use — `@AppStorage`'s resolution and its write-back pinned to the resolver's answer on the same store, and a reflection check that the modal's checkbox is still `@AppStorage`-backed.
- `ParkedPromptReadbackTests.swift` — retitles a neighbouring test whose title this change falsified ("A new first message does not send itself" is no longer true for someone who opted in; the body only ever asserted the shipped default).

No flag, migration, or wire change. The key resolves OFF until someone chooses otherwise — already the OFF-first shape a flag would impose — and the composer stays gated behind the daemon's `queued_prompt_enabled`. `worktree.setPendingPrompt` still carries an explicit per-message `submit` bit; the daemon never learns this preference exists.

## Assumptions

- **Nothing registers a default for `queuedPromptSendImmediately`.** The three-state distinction holds against the persistent domain only. A future `UserDefaults.register(defaults:)` for this key would make "never chose" and "chose off" read alike again, and the tests could not catch it — they use isolated `UserDefaults(suiteName:)` instances with empty registration domains. Nothing registers it today; there's a comment on the key saying not to start.
- **Two sheets share a title and a checkbox label.** Both the creation modal and the read-back are headed "First message for `<name>`" with a "Send immediately" box, and only the first writes the preference. The Settings help text names which one; if either sheet's copy changes, that disambiguation has to keep up.

## Evidence & verification

- `scripts/test.sh --filter "^TBDAppTests\."` → **2210 tests in 232 suites passed**, 0 `error:` lines (2207 before the three `@AppStorage` cases were added).
- `swiftlint --strict` → 0 violations in 758 files.
- **Discriminating evidence.** The suite was run against a deliberately weakened resolver (`defaults.bool(forKey:)`, the collapsing read this design exists to prevent) and went red on the `shippedDefault: true` arm of `absentKeyResolvesToShippedDefault`, then green again once restored. Without that arm the absent-key case degenerates to `false == false`, which the broken resolver passes by accident.

- **Live-verified in the running app.** Built and installed via `scripts/restart.sh` (the running bundle confirmed byte-identical to this worktree's build by SHA-256, so the check was against this code and not a stale install), then driven by hand: the Settings ▸ General ▸ Worktrees row appears unticked; ticking **Send immediately** in a creation sheet and dismissing it with Escape leaves the preference set; the next creation sheet opens already ticked, and the Settings row agrees. That covers the path the tests deliberately do not reach — the real sheet, the real binding, the real write on tick.

One limit, stated rather than papered over:

- **The binding is covered by proxy, not by rendering.** `resolveSendImmediately` still has no production callers — both surfaces use `@AppStorage`, whose semantics are Apple's, not this helper's — so three tests reach the mechanism instead of the statement. `AppStorage` is a property wrapper struct, so it can be built with an injected store outside a `View`: two tests assert its `wrappedValue` agrees with the resolver on the same store across all three states (absent under both shipped-default polarities, chose-off, chose-on) and that a write through `wrappedValue` lands in the store as a stored choice. A divergence in either direction now reddens. A third mirrors a constructed `QueuedPromptModal` and asserts the `_sendImmediately` child's type is `AppStorage<Bool>`, so a merge conflict restoring `@State` fails rather than passing silently — it reads the declared type only, never `wrappedValue`, which without an injected store would resolve against the developer's real `TBDApp.plist`. What remains uncovered *by tests*: nothing renders the view or drives the real sheet, so the checkbox's binding to `$sendImmediately`, the toggle's write on tick, and the Settings row's binding rest on the live check above and on reading the source — a regression in any of them would land silently in CI.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=791069BC-9B69-466A-8B6B-945895D172E6)

### DIFF
--- a/Sources/TBDApp/Settings/SettingsView.swift
+++ b/Sources/TBDApp/Settings/SettingsView.swift
@@ -38,6 +38,8 @@ struct GeneralSettingsTab: View {
     @AppStorage(AppState.enableTranscriptKey) private var enableTranscript: Bool = AppState.enableTranscriptDefault
     @AppStorage(AppState.nightwatchExperimentalKey) private var nightwatchExperimental: Bool = false
     @AppStorage(AppState.showScratchSectionKey) private var showScratchSection: Bool = true
+    @AppStorage(QueuedPromptComposer.sendImmediatelyKey)
+    private var sendFirstMessageImmediately: Bool = QueuedPromptComposer.sendImmediatelyDefault
     @AppStorage(AppState.showClaudeTabUsageTooltipKey) private var showClaudeTabUsageTooltip: Bool = true
     @AppStorage(AppState.usageResetTimeStyleKey)
     private var usageResetTimeStyle: ProfileUsagePresentation.ResetTimeStyle = .timeOfReset
@@ -157,6 +159,9 @@ struct GeneralSettingsTab: View {
 
                 Toggle("Show Scratch section", isOn: $showScratchSection)
                     .help("Hide the repo-less Scratch section. Existing scratch spaces and their terminals keep running.")
+
+                Toggle("Send first messages immediately", isOn: $sendFirstMessageImmediately)
+                    .help("On: TBD presses Return, and the agent starts working the moment the first message is in. Off: the text waits in the composer for you to read and send. The checkbox in the first-message sheet changes this too.")
 
                 Toggle("Automatically clean up orphaned agent worktrees", isOn: Binding(
                     get: { appState.gcEnabled },
--- a/Sources/TBDApp/Settings/SettingsView.swift
+++ b/Sources/TBDApp/Settings/SettingsView.swift
@@ -161,7 +161,7 @@ struct GeneralSettingsTab: View {
                     .help("Hide the repo-less Scratch section. Existing scratch spaces and their terminals keep running.")
 
                 Toggle("Send first messages immediately", isOn: $sendFirstMessageImmediately)
-                    .help("On: TBD presses Return, and the agent starts working the moment the first message is in. Off: the text waits in the composer for you to read and send. The checkbox in the first-message sheet changes this too.")
+                    .help("Default for first messages you write while a new worktree is coming up. On: TBD presses Return, and the agent starts working the moment the message is in. Off: the text waits in the composer for you to read and send. The \"Send immediately\" checkbox in that creation sheet changes this too; the identical-looking checkbox on an already-parked message edits only that message.")
 
                 Toggle("Automatically clean up orphaned agent worktrees", isOn: Binding(
                     get: { appState.gcEnabled },
--- a/Sources/TBDApp/Worktree/QueuedPromptModal.swift
+++ b/Sources/TBDApp/Worktree/QueuedPromptModal.swift
@@ -1,3 +1,4 @@
+import Foundation
 import SwiftUI
 import TBDShared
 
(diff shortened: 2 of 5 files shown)


cheapsteak:
Addressed in `190a233c`. The finding was right, so this closes it rather than arguing precedent — a shape that's already common is a reason to cover it once, not to keep extending it.

Three tests added, taking the suite from 3 to 6. They target the two risks the finding named specifically:

- **`@AppStorage`'s resolution pinned to the contract.** An `AppStorage` built against an injected store is asserted to agree with `resolveSendImmediately` across absent / explicit `false` / explicit `true` — with the absent case looped over *both* shipped-default polarities, so a collapsing resolution can't pass by matching today's `false`.
- **Write-back.** Writing through `wrappedValue` reaches the store, and an explicit `false` lands as a stored choice rather than reverting to "never chose".
- **The binding itself.** `Mirror` asserts `QueuedPromptModal._sendImmediately` is `AppStorage<Bool>`, not `State<Bool>` — the merge-conflict revert the finding called out.

**Mutation evidence for that last one**, since a guard like it is worthless unverified. Reverting the property to `@State` and re-running:

```
Test "the modal's checkbox is bound to the persisted key, not to view state" recorded an issue:
  Expectation failed: (String(describing: type(of: backing.value)) → "State<Bool>") == "AppStorage<Bool>"
Test run with 6 tests in 1 suite failed
```

Restored, and green again at 6/6.

Two assumptions behind that test were probed in an isolated `swiftc` run rather than assumed, because both could have made it silently wrong:

- A `private` stored property does **not** make the synthesized memberwise init unreachable from the test module under `@testable` against an `-enable-testing` build.
- Constructing a storeless `@AppStorage` does **not** touch `UserDefaults.standard` — the probe key was absent before construction, after construction, after `Mirror`, and after `synchronize()`. That matters here because `UserDefaults.standard` on this unbundled executable is the developer's real `TBDApp.plist`. The test reads the property's *type* only, never `wrappedValue`, and never renders the view.

**What is still not covered, stated plainly:** nothing renders the view or drives the real sheet. The checkbox's `$sendImmediately` binding, the toggle's write-on-tick, and the Settings row's binding are verified by reading the source, not by exercising them. The PR description has been updated to say exactly this rather than implying the gap closed entirely.

Full run: `2210 tests in 232 suites passed` (2207 baseline + 3), 0 `error:` lines; `swiftlint --strict` clean at 0 violations across 758 files.

PR #672, opened 2026-08-17
Feat: Start and steer Claude cloud sessions from TBD, behind a default-off flag

cheapsteak:
## Summary

You can start a Claude cloud session from TBD and talk to it, without leaving the app for claude.ai. Pick "New cloud session…" from a repository's create menu or its `+` button, describe the work, and the session starts on Anthropic's infrastructure and appears as a row you can send messages to.

The point is that a cloud session stops being a thing you open a browser tab for. It becomes a lane in the same tree as your local worktrees, created the same way and steered the same way.

One property of this feature is unusual enough to state up front, because it looks like a bug and is not: **TBD can only ever show cloud sessions that TBD itself started on this Mac.** A session you began on claude.ai, in the mobile app, or from a terminal TBD did not spawn will never appear in the list. Settings says so in a caption, because an absence nobody explains reads as breakage.

## Why this shape

The provider is **compiled into the daemon** rather than shipped as an external subprocess like the registry-file providers. It still reaches the daemon through the same `RemoteProviderInvoking` contract, so it gets the same result envelope, failure classification, health handling and poll loop an external provider gets — a `ProviderDispatcher` picks the in-process conformance by provider name and sends everything else to the subprocess runner. Nothing downstream knows the difference.

**There is no discovery, and that is a boundary rather than a gap.** No supported interface enumerates an account's cloud sessions: the Compliance API excludes Claude Code on the web by name, the documented `/v1/claude_code/` namespace grants no read access, and the undocumented claude.ai endpoints are barred by Anthropic's Consumer Terms outside an API-key carve-out that a subscription-login cloud session can never be inside. So `list` answers from a local ledger of what this machine started and reports itself **permanently incomplete**. What would reopen this is a supported interface, not a more careful scraper.

That incompleteness is load-bearing in a way worth flagging to reviewers: under the ordinary drift rule, a provider whose snapshot is always incomplete would have its own live lanes tombstoned two polls after creating them. Snapshot completeness (`complete` on the list envelope, honored by `RemoteSessionStore.applySnapshot`) is what prevents that, and it is why a successful create also triggers one out-of-band `list`.

Two vendor facts shaped the implementation, both measured against `claude` 2.1.233 rather than assumed. `create` requires a TTY — the CLI refuses `--cloud` on a pipe — so it runs under a pseudo-terminal and TBD parses the session id out of the prose it prints. `send` must **not** use a pseudo-terminal, because under one stdout and stderr merge onto a single descriptor and would corrupt the JSON it returns.

Spec: [`docs/specs/2026-08-15-cloud-sessions-slice-1-design.md`](docs/specs/2026-08-15-cloud-sessions-slice-1-design.md).

### Who reclaims the process a create spawns

A create spawns the vendor CLI outside tmux, and its deadline is held by a watchdog living inside the daemon. So a daemon that dies mid-create leaves the child reparented to pid 1 with nothing supervising it, and no `SIGHUP` reaches it — the runner never makes the pty the child's *controlling* terminal. That was reproduced rather than assumed: a child holding a pty replica, parent `SIGKILL`ed, survived and was observed reparented to pid 1 still running. `AgentReaper` does not cover it, since it only walks children of a known `tbd-*` tmux server.

Per the reconciler doctrine this takes the third of its three answers — a committed spec that deliberately chose otherwise, in [the spec's §3](docs/specs/2026-08-15-cloud-sessions-slice-1-design.md). No sweep reclaims it, because the process is a single short-lived invocation bounded by its own work, holding no lock, lease, git ref, tmux resource, or file that outlives it — a different shape from the unbounded accumulations the doctrine was written for (2,804 leaked branches, ~7,100 dead sockets). The spec states plainly what that does **not** claim: a child that hangs before it ever writes has no `EIO` to trip over and can linger. Observed orphans during the soak reverse the decision, and the spec names the remedy's shape.

## What this PR does

- **`Sources/TBDDaemon/Remote/ClaudeCloud/`** — the provider: `describe` (static and offline), `create` (pseudo-terminal, prose parsing, idempotency), `list` (ledger-only, permanently incomplete), `send` (JSON over a plain pipe), the spawn seam, and the executable resolver.
- **`ProviderDispatcher`** — routes a verb to the in-process provider by reserved name, everything else to `ProviderRunner`.
- **`RemoteProviderManager`** — registers built-in providers alongside registry entries and describes them identically; threads snapshot completeness; refreshes once after a create.
- **`claude_cloud_session` ledger** (migration `v82`) — what this machine started, with a pending→failed window for creates that never resolve.
- **Create surfaces** — the repository create menu, the `+` button, and the Remote section all enumerate through one pure function so they cannot disagree about what is on offer.
- **Settings** — the flag toggle plus the caption describing what the list covers.

## Flag & rollout

- **Flag:** `claude_cloud_enabled` (migration `v81`), **default OFF**, nested under `remote_backends_enabled` — cloud is live only when both are on.
- The column carries **no SQL default**, so NULL means "nobody has chosen" and the shipped default lives in exactly one place, `ConfigRecord.toModel()`'s `?? Config.claudeCloudEnabledDefault`. Graduation is a one-line change to that constant that reaches everyone who never touched the toggle while preserving every explicit opt-out.
- **To soak:** enable Remote backends, then Claude cloud sessions, in Settings, and restart the daemon (the provider is constructed at boot).
- **Graduation:** flip the default constant after the soak, then delete the flag.

## Assumptions

- Assumes the vendor `claude` CLI keeps `create`'s three-line output shape and `send`'s `{"ok":true,…}` response. If the prose changes, creates fail loudly as contract bugs rather than silently — but they do fail.
- Assumes `claude` is on the daemon's PATH. If it is not, the provider does not register and Settings reports it as not running; the underlying reason is only in `os_log`. Worth improving before graduation.
- Assumes a cloud session outlives the create call. TBD records the id and never polls the vendor for liveness, because nothing supported would answer.

## Evidence & verification

- **Full suite green:** 7146 tests in 721 suites, 70 known issues, no failures. `swiftlint --strict` clean. Verified after rebasing onto current `main`.
- **Every gated branch is tested on both sides** — flag off, flag on, and ungated behavior unaffected — for the RPC gate, the boot wiring, and all three create surfaces.
- **Discriminating tests, not just passing ones.** Several tests here exist specifically because a plausible-looking implementation would pass without them: `archived` is asserted on raw JSON, because a decoded `nil` and `false` are indistinguishable through the display accessor; the create parser's CRLF test builds its input by explicit concatenation, because Swift normalizes triple-quoted literals to bare `\n` and a literal-based test cannot detect the bug it targets; the caption test pins the sentence, because an *overclaiming* rewrite passed the previous keyword assertions.
- **A final review found a Critical the tests did not.** The create parser cross-checked the session id across the whole output including the title line — which echoes the user's prompt — so a prompt mentioning something like `session_store_test.py` produced a second id match and reported a *successful* create as a contract bug, orphaning a real cloud session with no discovery to recover it. Fixed to scan only the lines that structurally carry the id, with a test that fails against the previous implementation. The same review found a same-key replay could start a second cloud session; a resolved key now returns what was recorded and a pending one refuses rather than guessing.

**Not verified live.** This has not been exercised against the real `claude` CLI in a running TBD, because doing so means restarting the daemon, which would disrupt the sessions currently running on this machine. The flag is off by default, so nothing changes for anyone until someone opts in — but the first soak run should be a live create → send round trip, and that is the gap in this evidence.

## Follow-ups, deliberately not in this PR

- `archive`, `unarchive` and `land` are **not implemented**, and `describe` no longer advertises them — capabilities are a contract the app reads to decide what to offer, so advertising an unimplemented verb means offering an action that always fails. A later PR adds each capability together with its implementation.
- The app-surfaces work — the transcript pane, landed-row rendering, the trust-on-first-use approval prompt — is a separate plan and a separate PR.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=BBA22690-7E4E-4F7F-A4F8-A71351FA5FD5)

### DIFF
--- a/Sources/TBDApp/AppState+Remote.swift
+++ b/Sources/TBDApp/AppState+Remote.swift
@@ -281,4 +281,47 @@ extension AppState {
             return "Off, but a poller from before this change is still running in the daemon — restart to fully stop it."
         }
     }
+
+    /// Persist the Claude cloud gate, then re-fetch capabilities so the
+    /// Settings toggle reflects the daemon's persisted state. Cloud requires
+    /// BOTH this flag and the outer remote-backends flag, and neither takes
+    /// effect until the daemon restarts.
+    func setClaudeCloudEnabled(_ enabled: Bool) async {
+        do {
+            try await claudeCloudFlagSetter(enabled)
+            await refreshDaemonCapabilities()
+        } catch {
+            remoteLogger.error("Failed to set Claude cloud sessions: \(error, privacy: .public)")
+            showAlert(
+                "Failed to set Claude cloud sessions: \(error.localizedDescription)", isError: true)
+        }
+    }
+
+    /// Caption for the Claude cloud toggle, telling the user which of the four
+    /// `(claudeCloudEnabled, claudeCloudLive)` states they are actually in. The
+    /// daemon registers the built-in provider only at boot, so the persisted
+    /// flag and the live provider disagree in both directions — the same shape
+    /// `remoteBackendsStatusCaption` handles for the outer flag.
+    nonisolated static func claudeCloudStatusCaption(enabled: Bool, live: Bool) -> String {
+        switch (enabled, live) {
+        case (false, false):
+            return "Off. Turning this on requires a daemon restart before cloud sessions appear."
+        case (true, false):
+            return "On, but restart the daemon before cloud sessions appear."
+        case (true, true):
+            return "On and running — cloud sessions are being polled."
+        case (false, true):
+            return "Off, but the provider registered before this change is still live in the daemon — restart to fully stop it."
+        }
+    }
+
+    /// Whether the cloud toggle can be operated at all. Cloud is reached
+    /// through the same `remote.*` verbs as every other provider, so it is a
+    /// second gate INSIDE the remote-sessions switch and never a bypass —
+    /// turning it on while the outer switch is off would promise something no
+    /// verb can deliver. A named function rather than an inline `.disabled(…)`
+    /// expression, so both branches are assertable without a view harness.
+    nonisolated static func claudeCloudToggleOperable(remoteBackendsEnabled: Bool) -> Bool {
+        remoteBackendsEnabled
+    }
 }
--- a/Sources/TBDApp/AppState+Remote.swift
+++ b/Sources/TBDApp/AppState+Remote.swift
@@ -297,21 +297,48 @@ extension AppState {
         }
     }
 
-    /// Caption for the Claude cloud toggle, telling the user which of the four
-    /// `(claudeCloudEnabled, claudeCloudLive)` states they are actually in. The
-    /// daemon registers the built-in provider only at boot, so the persisted
-    /// flag and the live provider disagree in both directions — the same shape
-    /// `remoteBackendsStatusCaption` handles for the outer flag.
-    nonisolated static func claudeCloudStatusCaption(enabled: Bool, live: Bool) -> String {
-        switch (enabled, live) {
-        case (false, false):
+    /// Caption for the Claude cloud toggle, telling the user which state they
+    /// are actually in across three inputs: the persisted cloud flag
+    /// (`enabled`), whether the built-in provider is actually live in the
+    /// daemon right now (`live`, frozen at boot), and the persisted state of
+    /// the OUTER remote-sessions flag (`remoteBackendsEnabled`, re-read on
+    /// every fetch). Cloud requires both flags simultaneously true AT BOOT to
+    /// go live (`Daemon.swift`), but `config.setClaudeCloud` persists this
+    /// flag unconditionally — it does not check the outer flag
+    /// (`RPCRouter+ConfigHandlers.swift`) — so `enabled` can be true while
+    /// `remoteBackendsEnabled` is false, and a caption built from `(enabled,
+    /// live)` alone would call that state "on" when a restart right now would
+    /// not actually start cloud sessions.
+    ///
+    /// `remoteBackendsLive` (the outer flag's own boot-frozen liveness) is
+    /// deliberately NOT a parameter here: whenever `live` is true, the outer
+    /// manager was necessarily live at that same boot (cloud's boot-time gate
+    /// checks `remoteBackendsEnabled` before constructing the manager at all),
+    /// so `live` already implies the outer manager was live — a separate
+    /// `remoteBackendsLive` input could never disagree with `live` in any
+    /// reachable state and would only add branches nothing can distinguish.
+    /// The outer flag's OWN restart/live nuance is already surfaced by the
+    /// sibling `remoteBackendsStatusCaption`, rendered directly above this
+    /// toggle — this caption only needs to say when the outer flag's CURRENT
+    /// value changes what a restart would do to cloud.
+    nonisolated static func claudeCloudStatusCaption(
+        enabled: Bool, live: Bool, remoteBackendsEnabled: Bool
+    ) -> String {
+        switch (enabled, live, remoteBackendsEnabled) {
+        case (false, false, false):
+            return "Off. Turning this on requires remote sessions above enabled and a daemon restart before cloud sessions appear."
+        case (false, false, true):
             return "Off. Turning this on requires a daemon restart before cloud sessions appear."
-        case (true, false):
-            return "On, but restart the daemon before cloud sessions appear."
-        case (true, true):
-            return "On and running — cloud sessions are being polled."
-        case (false, true):
+        case (false, true, _):
             return "Off, but the provider registered before this change is still live in the daemon — restart to fully stop it."
+        case (true, false, false):
+            return "On, but remote sessions above are off — enable both, then restart the daemon before cloud sessions appear."
+        case (true, false, true):
+            return "On, but restart the daemon before cloud sessions appear."
+        case (true, true, false):
+            return "On, but remote sessions above are now off — restarting will stop cloud sessions unless you re-enable remote sessions above first."
+        case (true, true, true):
+            return "On and running — cloud sessions are being polled."
         }
     }
 
--- a/Sources/TBDApp/AppState+Remote.swift
+++ b/Sources/TBDApp/AppState+Remote.swift
@@ -351,4 +351,21 @@ extension AppState {
     nonisolated static func claudeCloudToggleOperable(remoteBackendsEnabled: Bool) -> Bool {
         remoteBackendsEnabled
     }
+
+    /// What cloud sessions TBD can see, stated once and independently of
+    /// which flag state the user is in.
+    ///
+    /// `list` answers from the ledger of what THIS machine started, because no
+    /// supported interface enumerates an account's cloud sessions. So a
+    /// session begun on claude.ai, in the mobile app, or by `claude --cloud`
+    /// in a terminal TBD did not spawn has no row here at all — not a partial
+    /// one, not a stale one, simply none. Saying so makes the absence read as
+    /// a property of the feature rather than as a bug.
+    ///
+    /// A separate value from `claudeCloudStatusCaption` deliberately: that
+    /// function's seven arms answer "which state am I in", they are pinned
+    /// byte-for-byte by tests, and this sentence is true in all seven.
+    nonisolated static let claudeCloudScopeCaption =
+        "Only sessions started from TBD on this Mac appear here — one begun on claude.ai, in the "
+        + "mobile app, or from a terminal TBD did not spawn is not listed."
 }
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -1258,6 +1258,11 @@ final class AppState: ObservableObject {
     /// injectable for the same reason as `controlModeSetter`.
     lazy var queuedPromptFlagSetter: @MainActor (Bool) async throws -> Void =
         { [daemonClient] enabled in try await daemonClient.setQueuedPrompt(enabled: enabled) }
+    /// How `setClaudeCloudEnabled` persists the Claude cloud gate — injectable
+    /// for the same reason as `controlModeSetter`, so the Settings toggle's
+    /// success and failure branches are testable without a real daemon.
+    lazy var claudeCloudFlagSetter: @MainActor (Bool) async throws -> Void =
+        { [daemonClient] enabled in try await daemonClient.setClaudeCloud(enabled: enabled) }
     /// The worktree a queued prompt is being composed for, driving
     /// `ContentView`'s `.sheet(item:)`. Non-nil only while the modal is up, and
     /// only ever set when the daemon reports `queuedPromptEnabled`
--- a/Sources/TBDApp/DaemonClient.swift
+++ b/Sources/TBDApp/DaemonClient.swift
@@ -993,6 +993,16 @@ actor DaemonClient {
         )
     }
 
+    /// Persist the Claude cloud sessions gate (default OFF). The daemon builds
+    /// its provider manager only at boot, so this takes effect on the next
+    /// daemon restart rather than the next gesture.
+    func setClaudeCloud(enabled: Bool) async throws {
+        try await callVoidAsync(
+            method: RPCMethod.configSetClaudeCloud,
+            params: ConfigSetClaudeCloudParams(enabled: enabled)
+        )
+    }
+
     /// Park the prompt composed while a worktree was still being created, to be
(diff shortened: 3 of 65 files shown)


cheapsteak:
Both findings addressed.

**Medium — no named reconciler for the spawned `claude --cloud` process.** Correct, and stronger than stated: I reproduced it rather than reasoning about it. A child holding a pty replica, parent `SIGKILL`ed, **survived and was reparented to pid 1**, still running. There is no `SIGHUP` because `runBoundedProcess` never calls `setsid` or `TIOCSCTTY`, so the pty the child inherits is never its *controlling* terminal — closing the primary side signals it nothing. And `AgentReaper.findStructuralOrphans` only walks children of a known `tbd-*` tmux server, which this process has nowhere in its ancestry. So "an orphan cannot arise" was not an honest answer available here.

The doctrine's third answer is taken instead: a committed spec that deliberately chose otherwise, now in §3 of the design doc (`docs/specs/2026-08-15-cloud-sessions-slice-1-design.md`), with the PR description pointing at it. The reasoning recorded there is that this resource is a *different shape* from the ones the doctrine was written for — a single short-lived CLI invocation bounded by its own work, holding no lock, lease, git ref, tmux server or socket, or file that outlives it, versus the unbounded accumulations that motivated the rule (2,804 leaked branches, ~7,100 dead sockets). At most one per crash, and gone once its call returns.

The section also states plainly what it does **not** claim: a child that hangs *before* it ever writes has no `EIO` to trip over and can linger indefinitely. That is a known gap, not a hidden one. Orphans observed during the soak reverse the decision, and the spec names the remedy's shape (pid + process start time on the ledger's pending row, so a sweep can check liveness, defeat pid reuse, and gate on the binary before signalling).

**Minor — no injected clock on `BoundedProcessClaudeSpawner`.** Fixed in `d29acf76`. It now takes `clock: any Clock<Duration> = ContinuousClock()` as its last initializer parameter, existential and defaulted so no call site changed, and threads it into `runBoundedProcess` — the seam already existed on the runner; this caller simply wasn't using it.

The accompanying test is mutation-checked rather than assumed to discriminate: replacing `clock: clock` with `clock: ContinuousClock()` in the `runBoundedProcess` call makes it fail (`Time limit was exceeded: 60.000 seconds`, and the outcome comes back `.completed` instead of `.timedOut`). Restored after checking. It sits in `Tests/TBDDaemonLiveTests/` as tier 3, following `GitManagerTimeoutTests` — it spawns a real child racing a deadline.

Full suite green at 7147 tests in 721 suites, `swiftlint --strict` clean.

cheapsteak:
Fixed in `f5ff1fa4`. The finding was correct — verified against the source before acting, then reproduced.

**What was wrong.** The lookup classified the key and the spawn happened later, with nothing holding the key in between. `upsertPending` returns an existing row *unchanged*, so a `failed` row stayed `failed` across the spawn — two replays arriving after the pending window gave up both read `failed`, both fell through, and both spawned.

**The fix.** `ClaudeCloudSessionStore.claimForSpawn` decides and records in one write. GRDB admits a single writer, so a concurrent caller waits and then observes whatever state the first left. Reclaiming a row moves it back to `pending` with a compare-and-set and claims **only when that UPDATE actually changed a row** — the transition is both the claim and the truth, since an attempt is in flight again and the pending window should govern it again. A brand-new key is inserted inside the same transaction, so the never-seen-key variant of the same race is closed too. An unrecognized state refuses rather than guessing: it is not a licence to start a second cloud session.

The earlier classification is kept and now documented as a **fast path, not the guard**. It answers a settled key before this call's parameter validation can reject a replay it would have answered from the ledger — `replayingAStillPendingKeyRefusesToSpawnAgain` seeds no repository, so moving the refusal behind repo resolution would have turned its `in_progress` into `not_found`.

**The test asserts the spawn count, not the row count.** That distinction is the finding: the unique index on `idempotencyKey` keeps the ledger at one row either way, so a row-count assertion is exactly the one that cannot see this bug — which is why the existing sequential test missed it. It is also gated rather than merely concurrent: the fake parks the first spawn so the second `create` provably does its own claim while the first is still in flight. Two bare concurrent calls need not interleave at all and would pass against the very race the test exists to catch.

**Mutation-checked.** Removing the compare-and-set (handing out the claim without moving the row off `failed`) makes it fail exactly as described: `spawner.requestsSnapshot().count → 2` and the second call returns `exitCode 0` instead of the `in_progress` refusal. Restored after checking.

Full suite green at 7148 tests in 721 suites, `swiftlint --strict` clean.

cheapsteak:
Both addressed in `72c026ec`.

**MEDIUM — the spec contradicted itself.** Correct, and self-inflicted: narrowing `describe` to `["send", "attach"]` updated §3 but left the summary, the capability list, and the archive-composition passages still asserting that cloud declares `archive`/`unarchive` and that Land was done.

Aligned the document **to the code**, not the reverse — the code is right, and this review said so itself: the missing declaration is what makes those paths fail closed. The header now states what ships (`describe`, `create`, `list`, `send`, with `attach` via its separate TTY-exec path) against what is designed here but deferred (`land`, `archive`, `unarchive`), and says why a verb answering `not_implemented` must not appear in a capability list — it is the contract the app reads to decide which actions to offer.

The design sections for Land and Archive stay: that reasoning is precisely what the later slice implements. What changed is every sentence asserting cloud *already* declares those verbs, which now reads as what holds once it does. Beyond the four passages you named I found and fixed one more in §1, plus §9 and §10, which described archive/land tests and work-order steps as if they were in this PR.

**MINOR — `pendingFailureWindow` had no stated rationale.** Added one, and deliberately did not manufacture evidence: no latency measurement was ever taken here, so claiming one would have satisfied the finding and been worse. The comment says 600s is a soak-tuned starting value rather than a measured one, why that range (comfortably longer than a create should take, short enough that a key stuck behind a hung create frees up within a working session), and what would revise it — real creates observed resolving later than the window during the soak.

107 tests across the 7 cloud suites pass; `swiftlint --strict` clean. No capability list, test, or behavior was touched.

cheapsteak:
Fixed in `84cc4e35`. Confirmed against the source first — this was a real bug, and mine: `claimForSpawn` (from `f5ff1fa4`) wrote only `state` on reclaim, while the sweep judges by `createdAt` age alone.

You identified the mechanism exactly. A row reaches `failed` only once its `createdAt` is already past the window, so reclaiming it left it instantly re-failable; the next poll re-failed it while the spawn it had just authorized was still in flight, and a third replay reclaimed and spawned again. The method's own doc comment asserted the correct behavior — "the pending window governs it again" — and the code did not implement it. That gap between a comment's claim and its code is the part worth flagging.

**The fix.** The reclaim `UPDATE` now sets `createdAt` to the claiming call's `now` alongside `state`, and the returned in-memory row mirrors it. The doc comment states the two meanings `createdAt` carries — the clock the pending window measures, and the session's creation time in the `list` payload — and why both stay true: a reclaimed row that goes on to resolve did have its session created by *that* attempt. It also records the cost rather than hiding it, namely that a reclaimed row no longer preserves when the original attempt ran, which nothing reads today since `list()` projects only `resolved` rows.

Checked the other writers: `claimForSpawn`'s fresh-row insert and `upsertPending` both already stamp `createdAt: now`, and `resolve()` correctly leaves it alone. No column or migration needed.

**The test closes the gap you named.** It polls `list()` at a moment **inside** the window measured from the reclaim (+300s) but **outside** it measured from the original `createdAt` (+901s) — those two readings disagree only when the bug is present. It then asserts the downstream consequence on the **spawn count**, since the unique index pins the row count at 1 either way.

Mutation-checked by reverting just the `createdAt` half: the test fails with all four consequences in order — `polled.state → "failed"`, `thirdResult.exitCode → 0`, `decodedError?.code → nil`, and `spawner.requestsSnapshot().count → 2`, the duplicate spawn itself. Restored after checking.

Full suite: 7149 tests in 721 suites, one unrelated load flake (`CodexUsageFetcherLifecycleTests`, a real-child PID-marker timeout under a load average of ~19-30) that passes in 1.7s on re-run. `swiftlint --strict` clean.

cheapsteak:
Fixed in `2859126b`. Verified against the source first — the finding is correct, and the framing is the useful part: four rounds hardened `claimForSpawn` into a compare-and-set, and `markFailed` is the *only other* writer that transitions a row's state. It kept the original read-then-blind-write shape, so the hazard came back through the writer nobody had audited.

**`markFailed` is now a compare-and-set on `pending`.** `list()` computes its batch from a read taken in a separate transaction, so a row that resolved in the gap is now skipped rather than clobbered — a no-op for it, not an error, because the batch was always a stale snapshot by construction.

**`claimForSpawn`'s `.failed` branch gained the `sessionID` check its `.resolved` branch already had**, as a second and independent layer: a `failed` row that still names a session is answered from, never reclaimed and re-spawned.

**Three tests, and the middle one matters most.** Beyond driving the read→resolve→stale-write interleaving and the downstream replay (asserted on **spawn count**, since the unique index pins the row count at 1 either way), there is a test proving a genuinely still-`pending` expired row *is* still marked `failed`. Without it, a compare-and-set whose predicate never matched would pass the first test while silently disabling the entire pending-failure sweep — a fix that satisfies the finding and breaks the feature.

The two guards are genuinely independent, which the mutation checks confirm: reverting the `markFailed` predicate fails the gap test; reverting the `.failed`-branch `sessionID` check fails only the replay test, with the spawn count no longer empty and the returned id the respawned session rather than the original. Note test 3 has to construct its `failed`-row-that-names-a-session by direct SQL, because with `markFailed` fixed that state can no longer arise through `list()`'s sweep at all.

**On folding `list()`'s read and write into one transaction** (your other suggested shape): considered and rejected. A larger transaction around the sweep would not close this window unless `resolve` were serialized behind it, which it should not be — the sweep runs on every poll. A compare-and-set makes the write self-checking without serializing the two paths against each other.

Full suite: 7314 tests in 735 suites, two unrelated tier-3 load flakes (`CodexUsageFetcherLifecycleTests`, and a firehose-overflow test whose own diagnostic reads "harness/load, not production"); both pass on re-run. `swiftlint --strict` clean.

Also note the branch was rebased onto current `main` in `a408af6b`: `main` took `v80`, so the cloud flag is now `v81_config_claude_cloud` and the ledger `v82_claude_cloud_session`. Git reports no conflict for that on its own, since both sides are additions.

cheapsteak:
Addressed in `136af295` — with one deliberate deviation from the suggested shape, explained below.

**I did not guard `resolve()` on `state = 'pending'`.** That predicate would make `resolve()` silently *drop* the write whenever the row is not pending, and dropping it loses a real cloud session's id — the orphan this whole subsystem exists to prevent, and unrecoverable because there is no discovery. A guard that causes the harm it is protecting against is worse than none.

The invariant that actually matters here is about the id, not the state: **never lose a session id, and never replace one with a different one.** So the predicate is `WHERE id = ? AND (sessionID IS NULL OR sessionID = ?)`:

- a `pending` row resolves normally;
- a row that somehow reached `failed` with nothing recorded **still resolves**, recording the session instead of orphaning it — strictly better than a state predicate;
- a same-id retry stays the harmless no-op the doc comment already promised;
- only a genuinely **different** id is refused, and that refusal logs `.error` naming both ids, because it means two sessions exist under one ledger row and a human needs to know. The zero-row case distinguishes "row absent" (the documented no-op) from "names a different session" and only reports the latter.

That leaves all three writers guarded, each on the predicate matching its own invariant rather than a uniform one: `claimForSpawn` on state to own the claim, `markFailed` on `pending` so it cannot clobber a resolution, `resolve` on the session id so it cannot lose one.

**The constants gap is fixed as asked, and mechanically.** The 60s create timeout was a bare literal at both call sites while the 600s window that makes it safe lived in another file. It is now `RPCRouter.remoteCreateTimeout`, used at both sites and cross-referenced in both doc comments — and `ClaudeCloudTimeoutRelationshipTests` pins `remoteCreateTimeout < pendingFailureWindow` with a message explaining the coupling. A comment alone would not fail when someone raises the timeout for a slower vendor CLI, which is exactly the path you identified for reopening this.

Mutation-checked: removing the `sessionID` predicate makes `resolveRefusesToOverwriteADifferentSessionID` fail with the id silently replaced (and `title`/`resolvedAt` clobbered alongside it). Restored, and confirmed identical to the backup afterward.

Full suite: 7319 tests in 736 suites. Eleven tier-3 load flakes on that run (the box was at load ~17-25), including two `EventDrivenTestClock` arming-handshake timeouts that `Tests/CLAUDE.md` documents as the signature of exactly that saturation; all 184 tests across those suites plus the cloud suites pass in 2.6s on re-run. `swiftlint --strict` clean.

PR #684, opened 2026-08-19
Fix: Stop Codex thinking indicators from getting stuck

Jephuff:
## What's broken

Codex terminals can leave the sidebar's thinking dots running after the turn has ended, been interrupted, or resumed into a new session. The opposite failure also matters: a transcript-derived working result can hide a permission prompt or undo the immediate idle state from Ctrl+C.

The stale indicator can persist indefinitely when a transcript has an unmatched historical turn, Codex rewrites a turn ID on completion, a nested run changes the active rollout, or asynchronous hook, list, and subscription responses arrive out of order.

## Why it happens

The sidebar reads the persisted hook activity state, but Codex's hook rail is not a complete turn-lifecycle signal. A working hook can remain latched after the transcript records completion, and an interrupted turn may never append a matching abort record.

The first transcript reconciliation path also treated activity, session identity, and presentation as one loosely ordered value. A delayed list response or same-state hook could therefore overwrite a newer interrupt, permission wait, SessionStart, transcript boundary, or presentation observation. Rebuilding a long transcript on the two-second terminal-list path would also make the fix expensive at fleet scale.

## What this PR does

- Reconstructs Codex presentation activity from JSONL lifecycle records: `task_started`, `task_complete`, and `turn_aborted`.
- Correlates rewritten completion IDs by `started_at`, lets a later turn supersede an orphaned historical start, and ignores timestamped late closes for older turns. When a close lacks a usable correlation key, it resolves to idle under the explicit false-thinking-worse policy; missing-identity starts remain ignored. A sample of 1,003 recent rollouts contained 150 matching-start-time ID rewrites across CLI, exec, and subagent sources; a broader aggregate scan found 64 open-turn rewritten completions without `started_at`.
- Adds an incremental transcript tracker with bounded tail bootstrap, bounded record buffering, one shared per-request byte budget, a filesystem-step cap, and fair progress across fleet and worktree-scoped polls.
- Treats unreadable, partial, oversized, or not-yet-caught-up evidence as unknown. This deliberately prefers temporary false idle over preserving a false thinking indicator.
- Keeps persisted semantic activity separate from response-derived presentation activity and from their ordering watermarks.
- Gives Codex permission waits and explicit interrupts precedence over transcript working, while allowing genuinely later work and SessionStart events to supersede them. Claude and shell activity and SessionStart retain their main-branch semantics.
- Binds new Codex activity hooks to the required `session_id` from hook stdin, rejecting delayed old-session work or permission events before they can mutate activity, prompt state, ordering, or scheduled resumes. The optional wire field preserves compatibility with already-running legacy hook configurations, and explicit interrupts remain identity-free and authoritative.
- Applies SessionStart identity, prompt retraction, and activity atomically while ordering identity on its own durable SessionStart watermark; persists enough ordering to survive daemon restart; and establishes transcript boundaries so interrupted historical turns are not replayed.
- Fences daemon and app reconciliation against delayed snapshots, deltas, equal timestamps, same-path restarts, and session rollovers. Stable polls advance hidden ordering without republishing unchanged terminal rows.
- Adds backward-compatible optional wire fields and nullable activity/session-order database migrations.
- Records the reconciliation theory, precedence, failure policy, field evidence, and rejected alternatives in [`docs/specs/2026-08-19-codex-activity-reconciliation-design.md`](https://github.com/cheapsteak/tbd/blob/clear-stale-thinking/docs/specs/2026-08-19-codex-activity-reconciliation-design.md).

## Assumptions

- Codex rollout files are append-oriented; truncation is detected and conservatively rebuilt.
- Lifecycle records are much smaller than the 1 MiB record cap. A lifecycle record over that cap is ignored.
- Aggregate field measurements put lifecycle lines around 3 KiB at p99 and 20.8 KiB maximum. The 1 MiB record/read limit therefore leaves substantial format-growth margin while bounding each two-second list request; 64 KiB quanta provide fair progress, and the 16-step cap bounds zero-byte metadata work.
- The transcript bootstrap and catch-up budgets can temporarily omit current work after a restart or large backlog. That is intentional: ambiguous evidence must not keep the sidebar falsely animated.
- Already-running legacy Codex hook configurations may omit activity session identity until the plugin is refreshed. Those events retain legacy behavior so genuine permission prompts are not lost; newly generated hooks carry identity and receive strict session binding.

## Scope and rollout classification

This is a bug fix restoring the existing user-visible contract: the thinking indicator should represent current work and must yield to an interrupt or permission wait. It does not add a feature, autonomous action, destructive behavior, or new user-facing mode, so the repository's bug-fix exemption applies. The human-approved design is nevertheless committed because the lifecycle correlation and ordering machinery establish an important reconciliation theory across asynchronous machine signals.

This wholesale replaces the Codex working/idle presentation signal, but not the persisted activity rail. Hook-derived state remains the semantic source for Codex permission waits and explicit interrupts, and Claude/shell keep their pre-existing last-arrival snapshot, activity, SessionStart, and interrupt behavior. Ordered reconciliation and its app-side ordering sidecars are gated end-to-end as a Codex-only path; transcript evidence supplies Codex's missing turn-completion signal to the presentation value. A default-off flag would retain the known stale-indicator failure for the default path and create two competing interpretations of the same status without providing a useful soak escape hatch. The approved design therefore ships the direct Codex-scoped bug fix by default, with conservative unknown-to-idle behavior and regression coverage for both precedence branches.

## Evidence & verification

- Regression tests cover orphaned and rewritten turns, copied-parent subagent rollouts, partial and oversized records, truncation, unreadable files, bounded catch-up, shared-budget fairness, and scoped/fleet polling.
- Race tests cover permission waits, Ctrl+C, SessionStart, daemon restart, same-path boundaries, stale list responses, legacy deltas, equal timestamps, session identity rollover, and app-side presentation watermarks.
- The session-binding verification passed 109 focused and adjacent tests across shared wire, CLI parsing, Codex hook overlay, activity/session handlers, tracker, app, and sidebar suites. The final end-to-end scope verification passed its focused regressions plus 200 adjacent tests across 14 suites.
- Final build and lint: `scripts/swift-safe build` and `swiftlint --strict --quiet` passed.
- Final full fenced run: 7,533 tests across 746 suites with exactly the established baseline of eight unrelated unexpected issues and 84 known issues. The unexpected issues remain three ScratchpadCollector assertions, four Markdown web-view configuration assertions, and one transcript-estimator accuracy assertion; no changed-area test failed.
- Final independent review reported no actionable findings. A separate Codex review only raised the intentionally conservative false-idle behavior for ambiguous restart/backlog evidence.

### DIFF
--- a/Sources/TBDApp/AppState+Terminals.swift
+++ b/Sources/TBDApp/AppState+Terminals.swift
@@ -728,6 +728,7 @@ extension Terminal {
         let currentOrderObservedAt = current.activityStateOrderObservedAt
             ?? current.activityStateObservedAt
         let shouldKeepCurrent: Bool
+        let shouldPreserveCurrentSemanticFact: Bool
         if incomingIsComplete,
            currentIsComplete,
            let incomingOrderObservedAt,
@@ -737,10 +738,15 @@ extension Terminal {
                     && current.preservesActivityAtEqualOrder(
                         over: activityState,
                         source: activityStateSource))
+            shouldPreserveCurrentSemanticFact = !shouldKeepCurrent
+                && activityState == current.activityState
+                && !isMeaningfulSameStateReplacement(source: activityStateSource)
         } else if !incomingIsComplete, currentIsComplete {
             shouldKeepCurrent = activityState == current.activityState
+            shouldPreserveCurrentSemanticFact = false
         } else {
             shouldKeepCurrent = false
+            shouldPreserveCurrentSemanticFact = false
         }
 
         if shouldKeepCurrent {
@@ -748,6 +754,10 @@ extension Terminal {
             activityStateSource = current.activityStateSource
             activityStateObservedAt = current.activityStateObservedAt
             activityStateOrderObservedAt = current.activityStateOrderObservedAt
+        } else if shouldPreserveCurrentSemanticFact {
+            activityStateSource = current.activityStateSource
+            activityStateObservedAt = current.activityStateObservedAt
+            activityStateOrderObservedAt = incomingOrderObservedAt
         } else if !incomingIsComplete {
             activityStateSource = nil
             activityStateObservedAt = nil
@@ -759,7 +769,9 @@ extension Terminal {
     /// raw hook timestamps do not order it. Legacy responses with no timestamp
     /// remain arrival-ordered until a timestamped observation has been adopted.
     /// On an exact tie, a non-working result wins because false working is the
-    /// costlier error and nil is a real "could not establish activity" result.
+    /// costlier error. A timestamped nil is authoritative "could not establish
+    /// activity" evidence (unreadable, incomplete, or capped), so it clears a
+    /// prior working presentation instead of serving cached positive state.
     private mutating func reconcilePresentationObservation(against current: Terminal) {
         let shouldKeepCurrent: Bool
         switch (presentationActivityObservedAt, current.presentationActivityObservedAt) {
@@ -788,6 +800,16 @@ extension Terminal {
         if activityStateSource == .terminalInterrupt { return true }
         if activityState == .waitingForUser, incomingState != .waitingForUser { return true }
         return activityState != .working && incomingState == .working
+    }
+
+    private func isMeaningfulSameStateReplacement(source: FactSource?) -> Bool {
+        source == .terminalInterrupt || source == .hookEvent("SessionStart")
+    }
+
+    private func canReplacePresentation(observedAt incomingObservedAt: Date?) -> Bool {
+        guard let presentationActivityObservedAt else { return true }
+        guard let incomingObservedAt else { return false }
+        return incomingObservedAt >= presentationActivityObservedAt
     }
 
     /// Apply a pushed activity observation under the same ordering and legacy
@@ -824,13 +846,21 @@ extension Terminal {
             return
         }
 
+        if currentIsComplete,
+           delta.activityState == activityState,
+           !isMeaningfulSameStateReplacement(source: delta.activityStateSource) {
+            activityStateOrderObservedAt = incomingOrderObservedAt
+            return
+        }
+
         activityState = delta.activityState
         activityStateSource = delta.activityStateSource
         activityStateObservedAt = delta.activityStateObservedAt
         activityStateOrderObservedAt = delta.activityStateOrderObservedAt
         if isCodexTerminal,
            delta.activityState == .idle,
-           delta.activityStateSource == .hookEvent("SessionStart") {
+           delta.activityStateSource == .hookEvent("SessionStart"),
+           canReplacePresentation(observedAt: incomingOrderObservedAt) {
             presentationActivityState = .idle
             presentationActivityObservedAt = incomingOrderObservedAt
         }
--- a/Sources/TBDApp/AppState+Terminals.swift
+++ b/Sources/TBDApp/AppState+Terminals.swift
@@ -186,6 +186,7 @@ extension AppState {
             terminals[worktreeID, default: []].append(terminal)
             inserted = true
         }
+        replaceTerminalObservationOrder(with: terminal)
         let splitRepresentationTabIDs = Set((tabs[worktreeID] ?? []).compactMap { tab -> UUID? in
             guard let layout = layouts[tab.id],
                   case .split = layout,
@@ -277,6 +278,7 @@ extension AppState {
                   $0.id == terminal.id
               }) else { return }
         terminals[terminal.worktreeID]?[index] = terminal
+        replaceTerminalObservationOrder(with: terminal)
     }
 
     /// Adopt a daemon snapshot without allowing a response that overlaps an
@@ -296,6 +298,35 @@ extension AppState {
                   recentlyDeletedTerminalIDs[snapshot.id] == nil else { return nil }
             var merged = snapshot
             let current = existingByID[snapshot.id]
+            let incomingActivityOrderObservedAt = snapshot.activityStateOrderObservedAt
+                ?? snapshot.activityStateObservedAt
+            let currentSessionOrderObservedAt = terminalSessionOrderObservedAt[snapshot.id]
+                ?? current.flatMap { terminal in
+                    guard terminal.activityStateSource == .hookEvent("SessionStart") else {
+                        return nil
+                    }
+                    return terminal.activityStateOrderObservedAt
+                        ?? terminal.activityStateObservedAt
+                }
+            let identityChanged = current.map {
+                snapshot.claudeSessionID != $0.claudeSessionID
+                    || snapshot.transcriptPath != $0.transcriptPath
+            } ?? false
+            let shouldKeepCurrentIdentity = identityChanged
+                && currentSessionOrderObservedAt.map { currentOrder in
+                    incomingActivityOrderObservedAt.map { $0 <= currentOrder } ?? true
+                } == true
+            if shouldKeepCurrentIdentity, let current {
+                merged.claudeSessionID = current.claudeSessionID
+                merged.transcriptPath = current.transcriptPath
+            } else if current == nil || identityChanged {
+                if let incomingActivityOrderObservedAt,
+                   snapshot.claudeSessionID != nil || snapshot.transcriptPath != nil {
+                    terminalSessionOrderObservedAt[snapshot.id] = incomingActivityOrderObservedAt
+                } else {
+                    terminalSessionOrderObservedAt.removeValue(forKey: snapshot.id)
+                }
+            }
             let presentationOrderObservedAt = merged.reconcileActivityObservation(
                 against: current,
                 presentationOrderObservedAt: terminalPresentationOrderObservedAt[snapshot.id]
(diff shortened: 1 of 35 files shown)


cheapsteak (approved):

Jephuff:
Resolved the latest Claude review findings:

- **Session identity ordering — fixed in `d07c468c`.** The review was correct that prompt/activity server-receipt timestamps cannot safely order session identity. SessionStart identity now has an independent durable `sessionOrderObservedAt` watermark (nullable v86 migration and optional shared model field). Prompt and activity retain their own conservative ordering, so a delayed old-session permission event cannot suppress the new session ID/path, while newer prompt/activity facts are still preserved. Deterministic tests cover the delayed-prompt race, stale SessionStart protection, equal timestamps, app snapshot rollback, and terminal recreation.
- **Design spec — bug-fix exemption applies.** This PR restores the existing thinking/idle contract; it adds no feature, autonomous behavior, destructive action, or user-facing mode. The reconciliation machinery is the implementation required to make the existing contract reliable across incomplete and reordered machine signals. The PR description now states this classification explicitly.
- **Default-off flag — not appropriate for this bug fix.** Hook-derived semantic activity remains in use for permission waits, explicit interrupts, and non-Codex agents; transcript evidence only supplies Codex's missing completion signal to the presentation value. A flag would leave the known stale-indicator defect enabled on the default path and create two interpretations of the same status without a useful rollout boundary. The PR description now records this rationale explicitly.

Verification on the pushed fix: 206 focused/adjacent tests passed; build and strict lint passed; the final fenced run completed 7,510 tests across 744 suites with only the established eight unrelated baseline failures.

Jephuff:
Resolved the latest Claude review findings in `1c92e965`:

- **Ordinary hook traffic invalidated valid transcript evidence — fixed.** Terminal-list identity checks and restart-safe boundary selection now bind to the durable session-only watermark, not `activityStateOrderObservedAt`. Deterministic races prove an ordinary activity hook no longer changes a valid transcript result, while SessionStart still invalidates pre-boundary evidence.
- **Rewritten closes without `started_at` could leave working latched — fixed with field evidence.** An aggregate-only scan of 43,772 local lifecycle rows found 64 open-turn rewritten-ID `task_complete` records without `started_at`. An uncorrelatable rewritten close now resolves to idle under the product's explicit false-thinking-worse policy. When `started_at` is present, older closes remain unable to clear a newer turn. Tests cover both completion and abort records.
- **Budget rationale — documented beside the constants and in the PR.** Aggregate measurements put lifecycle lines around 3 KiB at p99 and 20.8 KiB maximum, so the 1 MiB cap leaves broad format-growth margin while bounding hot-path work. The 64 KiB quantum provides fleet fairness, and 16 steps bound caught-up/unreadable metadata work that consumes no byte budget.
- **Design spec — bug-fix exemption still applies.** This restores the existing thinking/idle contract and adds no product mode, autonomous behavior, or destructive action. The PR description now records the classification, signal precedence, conservative ambiguity policy, and measured tuning rationale. A separate speculative feature spec would conflict with the repository rule that bug fixes need no spec.

Verification: 86 focused/adjacent tests passed across tracker, reconciliation, session-event, and activity-event suites; build and strict lint passed. The fenced suite ran 7,512 tests; its only non-baseline issue was an unrelated Codex usage PID-marker timeout that passed an isolated rerun in 0.568 seconds.

Jephuff:
Resolved the latest review in `c349a6b3`:

- **Lifecycle records without `turn_id` — fixed.** Missing-ID `task_complete` and `turn_aborted` now clear the current turn under the explicit false-thinking-worse policy; missing-ID `task_started` remains ignored. An aggregate audit of 43,784 local lifecycle rows found IDs on all 21,350 completions, 112 aborts, and 22,322 starts, so this is forward/partial-schema hardening rather than a claim about current Codex output. Tests cover absent-ID complete, abort, start, and the conservative fallback despite a mismatched timestamp.
- **Shared activity/session theory — narrowed out of the PR.** The review correctly identified that the earlier implementation changed persisted activity semantics for all agents. Ordered activity, prompt precedence, session generation, and transcript boundaries are now enforced only for Codex terminals (plus the explicit interrupt origin used by the Codex fix). Ordinary Claude/shell hooks again use main's changed-value last-arrival replacement and same-value no-op. Non-Codex SessionStart again uses main's unconditional last-arrival identity update and prompt retraction, leaves activity/order rails unchanged, and emits no session-order generation. Tests compare these exact legacy behaviors.
- **Spec classification — now matches the code.** With the unintended cross-agent policy removed, the change is a Codex-scoped bug fix restoring the existing indicator contract. The project instructions explicitly exempt bug fixes from a brainstormed spec. The PR description has been updated to state the narrowed scope and measured tuning rationale.

Verification: 90 integrated tests and 216 adjacent tests passed; build and strict lint passed. The final fenced run completed 7,516 tests across 744 suites with exactly the established baseline of eight unrelated unexpected issues and 84 known issues; all changed and adjacent suites were green.

Jephuff:
Resolved the activity-generation race in `5b2f53d7`:

- Codex 0.147.0's SessionStart, UserPromptSubmit, PermissionRequest, and Stop hook inputs all require `session_id`, and the hook runner writes their JSON input to command stdin. The generated TBD overlay now forwards that payload safely to activity events; SessionStart and Stop reuse buffered payload without shell interpolation.
- `TerminalActivityEventParams.sessionID` is optional for wire compatibility. When supplied for Codex, the store validates it against the current session inside the same transaction and rejects mismatches before activity, prompt reason, order watermark, or scheduled-resume cancellation can change.
- Matching current-session work and permission events remain accepted. Explicit user interrupts remain identity-free and authoritative. Already-running legacy hook configurations without the new field retain prior behavior so genuine permission waits are not discarded during upgrade.
- The CLI's explicit hook-payload mode limits stdin to 1 MiB and has parser, oversize, metacharacter, payload-reuse, and legacy-decode coverage.

The prior scope correction remains in place: ordered reconciliation is Codex-only, while Claude/shell activity and SessionStart use their main-branch semantics.

Verification: 109 focused and adjacent tests passed; build and strict lint passed. Two full fenced runs each completed 7,527 tests across 746 suites with exactly the established eight unrelated unexpected issues and 84 known issues; all affected suites passed.

Jephuff:
Resolved the latest HIGH scope finding in `28ddc801`:

- AppState snapshot adoption, session deltas, activity deltas, presentation watermarks, and ordering sidecars now branch on terminal kind. Codex retains ordered reconciliation; Claude/shell use main's unconditional arrival/raw semantics and clear any Codex-only sidecars.
- Claude Ctrl+C/Esc again sends the legacy nil-origin raw-idle event, shell remains ignored, and Codex Esc remains ignored. The daemon no longer selects ordered persistence for a non-Codex `origin=userInterrupt` event.
- Regression tests cover stale Claude/shell snapshots, session and activity deltas, Claude Ctrl+C/Esc provenance/time, direct older Claude interrupt arrival, and publish frequency. The focused old-code run produced 29 expected issues; all now pass.
- The intentional reducer tradeoff is explicit and tested: start A, start genuinely open B, then receive a mismatched A close with no `started_at`; the result is idle under the stated false-thinking-worse policy.

With the client, interrupt, and persistence leaks removed, ordered reconciliation is now Codex-only end to end. Claude/shell retain their main-branch behavior.

Verification: 200 adjacent tests across 14 suites passed with no unexpected issues; build and strict lint passed. The full fenced run completed 7,533 tests across 746 suites with exactly the established eight unrelated unexpected issues and 84 known issues.

Jephuff:
Resolved the latest review at `175b4125` with the approved design and an exact subagent regression:\n\n- **Committed design:** [`docs/specs/2026-08-19-codex-activity-reconciliation-design.md`](https://github.com/cheapsteak/tbd/blob/clear-stale-thinking/docs/specs/2026-08-19-codex-activity-reconciliation-design.md) records the signal precedence, ordering rails, conservative failure policy, bounded polling, compatibility contract, field evidence, and rejected alternatives.\n- **Default-on decision:** the design explicitly classifies this as a bug fix under the rollout policy even though it wholesale replaces the Codex working/idle presentation signal. It performs no autonomous or destructive actuation, resolves ambiguity to idle, and avoids retaining the known-broken legacy interpretation as the default.\n- **Subagent lifecycle:** Codex spawned agents use separate rollout files that can copy parent lifecycle prefixes. The new regression feeds `session_meta/thread_spawn`, a copied parent start, a newer child start, and the child close, then asserts idle. A temporary stack mutation made that exact test fail by restoring the copied parent as working; production remains single-slot and unchanged.\n\nFresh verification: tracker plus reconciliation passed 57 tests; the broader affected-path audit passed 154 tests; `scripts/swift-safe build` passed; the test-only commit changed no production source.

Jephuff:
Resolved the bounded-stdin finding in `cb27b368`. `terminal-activity --read-hook-payload` now reads incrementally and stops after 1 MiB + 1 byte, returning no session identity on overflow or read failure; it no longer calls `readDataToEndOfFile()`. A real temp-file regression supplies 1 MiB + 512 bytes, asserts rejection, and verifies the file offset is exactly 1,048,577, proving the remaining bytes were never read. Focused CLI/overlay/wire verification passed 34 tests across 3 suites, and `scripts/swift-safe build` passed.

Jephuff:
I reviewed and approve the Codex activity reconciliation design. I classify this as a bug fix under the repository’s explicit bug-fix exemption. It should wholesale replace the Codex working/idle presentation signal by default; I do not want a feature flag. False thinking is worse than false idle. Ctrl+C overrides, permission waits override transcript working, and transcript lifecycle governs Codex working/idle. Proceed with this design.

PR #688, opened 2026-08-20
DX: Stop parallel branches from colliding on the same migration number

cheapsteak:
## Summary

Two people adding a database migration in the same week collide. Today every
migration appends to the tail of one 1500-line Swift function and claims the
next integer in a hand-picked sequence, so parallel branches pick the same
number and edit the same lines. The fix is a renumbering commit, and then a
rename of the test file that carried the number — `a408af6b` renumbered the
cloud migrations after main claimed v80, `0411d6fa` renumbered the pin
migration to v63, `4974beed` renamed stale v56 test names to v59. `main` still
ships two migrations numbered `v35`.

It has been worse than churn. Renumbering an additive migration bricked the
daemon twice: GRDB keys applied migrations by identifier string, so a
renamed-but-equivalent migration is unapplied under its new name while its
column already exists, and SQLite throws `duplicate column name` on the way up.

After this change a migration is one `.sql` file named by timestamp. Two
branches can neither pick the same identifier nor touch the same file, and the
migration history is legible as a directory instead of a scroll through one
function.

## Why this shape

The number was never load-bearing. GRDB's `unappliedExecutions` walks the
registered migrations and runs any whose identifier is absent from
`grdb_migrations`; there is no watermark and no ordering check. Identifier
uniqueness is the whole contract, and a hand-picked counter is a poor way to
get it. The 14-digit filename is dbmate's, adopted rather than invented.

Nothing off the shelf fits. Every embeddable engine is the wrong runtime —
Flyway is JVM, dbmate/goose/golang-migrate/Atlas are Go, Alembic is Python —
and migrations here must run in-process at daemon startup, so a developer-run
CLI cannot be the runtime path. Atlas was the closest call and is rejected on
`atlas.sum`, a committed integrity file that its own documentation says "will
raise merge conflicts in most version control systems", framed as the point.
That is a coherent position and the exact opposite of this one.

A SwiftPM build-tool plugin generating a registry was the runner-up and is
recorded as the fallback. It loses on `swiftlang/swift-build#305` — build-tool
plugins re-running every incremental build and invalidating compiled targets —
against a measured 7 seconds and no relink for the resource-bundle path.

**No feature flag, deliberately.** Two migrators cannot both run over one
database, so gating would mean shipping two code paths over one schema. The
soak is the inert cutover instead: `Migrations/` ships empty, the merged list
is empty, and `buildMigrator()` produces behavior byte-identical to today. The
first real `.sql` migration is a separate PR and is the actual soak.

**No new reconciler needed.** This introduces no durable external resource —
migration files are build inputs, not runtime-created state — so there is no
new class of orphan for a sweep to reclaim.

Spec: [`docs/specs/2026-08-19-migration-identifier-scheme-design.md`](docs/specs/2026-08-19-migration-identifier-scheme-design.md).

## What this PR does

- **Loader** (`Sources/TBDDaemon/Database/SQLMigrationLoader.swift`) — discovers
  `.sql` files from the `TBDDaemonLib` resource bundle, sorts by identifier, and
  registers them after the frozen `v1`–`v84` block. Statements are split with
  `sqlite3_complete()`, SQLite's own lexer, so semicolons inside string literals
  and `CREATE TRIGGER … BEGIN … END;` bodies are handled rather than guessed at.
- **Idempotency** — an `ALTER TABLE … ADD [COLUMN]` statement is checked against
  `PRAGMA table_info` and skipped if the column exists. That is
  `addColumnIfMissing` recovered, moved from the call site to the loader. It has
  to happen before SQLite prepares anything, because `duplicate column name` is
  raised at prepare time, not step time.
- **Locator** — an ordered candidate list, deliberately not `Bundle.module`:
  SwiftPM's generated accessor bakes in an absolute path to the building
  worktree's `.build`, so a relocated daemon silently reads that tree's
  migrations instead of failing.
- **Lint** (`scripts/lint-migrations.py`) — filename shape, identifier
  uniqueness, a statement allowlist, idempotent DDL, migration history frozen,
  and a chain-apply against a committed baseline. Runs in the existing
  build-free `lint` job and the pre-push hook. It splits with Python's
  `sqlite3.complete_statement()` — a binding to the same C function the loader
  calls, so the two cannot disagree about where a statement ends.
- **The lint's own harness** (`scripts/lint-migrations.test.sh`) — one fixture
  per guard that must fail, plus a mutation that disables the guard and proves
  its fixture then passes.
- **`migration_use_helpers`** now names the escape-hatch file. It previously
  governed only `Database.swift`, whose migrations sit inside a
  `swiftlint:disable` region — so it could never fire on anything.

## Assumptions

- **The resource bundle ships beside the daemon binary.** Verified for
  `.build/debug/TBDDaemon` and under `scripts/test.sh`. If TBD is ever
  distributed as a relocated artifact, the locator's candidate list is what has
  to cover it; a bundle it cannot find throws at startup rather than running
  under-migrated.
- **Migrations are idempotent and mutually independent.** Timestamp identifiers
  mean a developer who already applied a later stamp runs a newly-pulled earlier
  one out of order — inherent to every timestamp scheme, and safe only under
  that assumption. The statement allowlist is what holds it.
- **`origin/main` is fetched when the freeze check runs.** CI fetches fresh; the
  pre-push hook compares blob hashes as well as the merge base, so a stale local
  ref cannot let an edit to shipped history through.

## Evidence & verification

- **Full suite**: 7418 tests, 740 suites, 87 issues of which 86 are known — one
  real failure, `virtualTimeoutReturnsTimedOutAndKillsChild`, in the Codex usage
  fetcher and unrelated to this change. Its neighbours all took ~69s in the same
  run while it waited on a 10-second real-time deadline; isolated it passes in
  1.28s. Zero compiler errors.
- **Lint and harness**: migration lint clean at 7 checks; `lint-migrations.test.sh`
  green across 86 assertions with every mutation flipping. `swiftlint --strict`
  clean over 788 files.
- **The cutover is inert** — a test asserts that with `Migrations/` empty the
  migrator is unchanged, and `git diff origin/main -- Database.swift` contains
  zero deletion lines, so no existing migration body was touched.
- **Discriminating evidence, not just green tests.** The CRLF splitter bug was
  found by diffing the Swift and Python splitters over ~55 adversarial inputs;
  before the fix a CRLF file split to `[]` in Swift and one statement in Python,
  meaning the loader executed nothing while GRDB recorded the identifier as
  applied. The fixture pins raw `\r\n` in a statement interior, because
  `trimmingCharacters` eats it at the edges and an edges-only fixture would pass
  either way.
- **Every lint guard has been watched fail.** The harness disables each guard in
  turn and proves its fixture then passes — including the two-dot/three-dot
  mutation, which makes a branch cut before a main-side migration fail while
  blaming a file it never touched.

[🗄 Migration Number Conflicts](https://cheapsteak.github.io/tbd/open/?worktree=158C7D6F-A350-4986-ADFF-18F540E497A8)


cheapsteak:
On the `--no-renames` finding — I'm not applying it, because the failure scenario doesn't reproduce. Adding the regression coverage the review correctly identified as missing instead (`e9f7b3c4`).

**The scenario, measured.** Two throwaway repos, `git diff --name-status <merge-base> HEAD -- <migrations dir>`:

- Unshipped migration added on the branch, then renamed on the same branch, byte-identical body: reports `A` — **with and without** `--no-renames`. Rename detection needs a source in the base tree, and a file added after the base has none, so there is nothing to pair against and no false positive to fix.
- Shipped migration (present at the merge base) renamed on the branch: reports `R100 <old> <new>`, which the check rejects. That rejection is correct — renaming an applied migration is exactly the frozen-history violation the leg exists to catch.

**What the current code already does.** Running the real `merge_base_edits()` against the second repo:

```
…/20260819143000_shipped.sql -> …/20260819143000_renamed.sql: migration history is frozen — …
```

The three-field `R` line parses safely (`fields[0], fields[1:]`) and the message names both paths.

**Why the proposed change would be a regression.** Built a `--no-renames` variant and ran both cases against it. Case 1 is unchanged (still `A`, still passes). Case 2 still fails, but the message degrades to ``git reports `D` `` naming only the old path — the reader loses where the file went, for no gain.

**What I did take from the review**, which was right: no fixture covered either case. Two are added, +8 assertions (93 → 101):

- `test_renaming_an_unshipped_migration_on_the_branch_is_accepted` — asserts rc 0 and that the output lacks "migration history is frozen", so a future green-for-the-wrong-reason can't pass as success.
- `test_renaming_a_shipped_migration_fails_and_names_both_paths` — asserts non-zero, ``git reports `R100` ``, and both paths independently. Mutation-checked, and verified to fail against the `--no-renames` variant, so it is a live trip-wire for anyone adding the flag later.

Also worth noting the review's own diagnostics: the live git repro was blocked by sandbox path restrictions, so the premise rested on `man git-config`'s documented default plus a grep. The default is stated correctly; the step that doesn't hold is that rename detection could pair two files that never coexisted in the base tree.

PR #689, opened 2026-08-20
Fix terminal panes leaving ghost text behind as content scrolls

cheapsteak:
## What's broken

Text in terminal panes leaves ghosts behind as content scrolls. The most visible form is a column of leftover characters down the left edge — the first two columns of lines that have since been replaced by a blank indent. It also shows up mid-line as two texts drawn on top of each other, so a line reads as garbled nonsense (`the two souaces disagrte sere` where the text is "the two sources disagree here").

Anyone running a TUI that redraws as it scrolls sees it, which in practice is any agent pane. Two properties are diagnostic: the ghost text is **not selectable**, and **nudging the window size wipes it clean**.

## Why it happens

Confirmed root cause, not a hypothesis.

`TerminalView` opts *into* partial backing-store updates on macOS: it sets `disableFullRedrawOnAnyChanges` and, in `viewWillDraw`, `layer.contentsFormat = .RGBA8Uint`. That is the CoreAnimation opt-in that lets the layer keep its old pixels and update only the invalidated region. The consequence is that **AppKit no longer clears the backing store except on a full-view redraw** — on a partial redraw, whatever was on screen stays on screen unless the draw path paints over it.

At the pinned revision, nothing in the CoreGraphics draw path erases the invalidated region. Each row that *is* drawn repaints itself opaquely (every glyph run carries a background colour and fills its own cell rect), so a drawn row is fine. The ghosts are the regions the draw **never touched** — rows that were not invalidated at all. They stay visible wherever the new content is blank, and where new glyphs land on top of old ones you get the overlapping garble.

That also explains both diagnostic properties. The grid is clean, so there is nothing to select; and a window resize forces a full-view redraw, which is the one path that still clears.

## When it broke

Not a regression introduced here — it is pin staleness. The pinned revision predates the fixes, and upstream has since addressed it three separate ways:

- `d5ee56e` adds `context.clear(dirtyRect)` at the top of the CoreGraphics draw, unconditionally erasing the invalidated region. This is the decisive one: with it, nothing inside a dirty rect can survive regardless of which cells the per-run fills happen to cover.
- `9446f60` removes the DEC-2026 synchronized-output buffering that "was causing updates to not flush to the screen on time" — a repaint queued inside a sync block could be consumed with nothing invalidated. It landed **one day after** the revision we pin.
- `cc5ed6d` forces full-view invalidation when the viewport is scrolled back.

It slipped through because the artifact is invisible to every automated check we have: the terminal's own grid is correct, so any assertion on buffer contents passes.

## What this PR does

Bumps the SwiftTerm pin to current upstream `main`, rather than carrying a local patch. The defect is entirely inside the dependency, this repo adds no draw override and no layer-policy override, and upstream has already fixed it — so a workaround here would be a second implementation of something already done better upstream.

Also in the diff:

- **`ChildReaper` doc amendment.** `Package.swift` requires re-verifying it on any bump. Done: it **stays, unchanged**. Upstream main still calls `waitpid` from exactly one place — its `DispatchSourceProcess` handler — and both `deinit` and `terminate()` are unchanged, so it still does not reap its own children on the teardown paths we use. Deleting it would restore the zombie leak verbatim. One clause of its serialization note went stale: `1c3f353` now arms `setEventHandler` before `activate()` and documents that `activate()` can invoke the handler *synchronously* for an already-exited child, which runs on the activating thread rather than the dispatch queue. The conclusion still holds — both call sites start the process from main-isolated code — but for a second reason that is now written down.
- **A repro suite, committed disabled.** See below; its comment says exactly what to finish.

## Assumptions

- Assumes upstream keeps `context.clear(dirtyRect)` in the CoreGraphics path. If a future bump removes it while retaining the partial-update opt-in, this artifact returns.
- Assumes upstream continues to reap children only from its exit-source handler. If that changes, `ChildReaper` becomes a competing `waitpid` on a possibly-recycled pid — worse than the leak it fixes. That is why the re-verification note stays in `Package.swift`.

## Evidence & verification

**The repro.** `TerminalStalePixelRepaintTests.swift` reproduces the mechanism offscreen: one bitmap retained across frames, with each frame drawn only inside the rects the view actually invalidated and clipped the way AppKit clips it. It measured a fully-stale cell (420 of 420 pixels off-background) surviving a partial repaint. That establishes the mechanism — pixels persist across partial repaints into a retained backing store.

**Being explicit about what it does not prove.** The suite is committed **disabled**, because it does not yet drive the invalidation path: every test fails first on "the view invalidated nothing", *identically before and after the bump*. When nothing is invalidated, nothing is drawn, so the residue it then reports is trivially explained and says nothing about the fix. A red test that misattributes its own cause is worse than no test. Its comment names the next step: assert the terminal's update range immediately after the feed, to separate "rows were dirtied but the async display pump never fired" from "the fed bytes changed nothing".

**Live verification.** Built and installed from this branch, and confirmed the running binary genuinely contains the new dependency — it exports `cellOrdinal` and `effectiveNativeBackgroundColor`, symbols that do not exist at the old revision. The app compiles against the new revision with zero source changes.

**Still outstanding:** soak time in the running app to confirm the ghosts are gone in practice, plus a look at the areas this bump could plausibly disturb — triple-click row selection (the tip commit changes it), scrolled-back rendering (`cc5ed6d`), and cell-width rounding, which upstream changed from `ceil` to round-to-nearest and which affects the default font.


cheapsteak:
Round 1 — all three findings addressed in `b9009376`.

**`Package.swift:46` — stitched comment (MEDIUM).** Correct, and my fault: the edit replaced only the old comment's first line, so its body was left dangling under the new closer. Rewritten as one paragraph — the re-verify instruction now leads, and the reason it exists ("upstream reaps only from its `DispatchSourceProcess` handler…") follows it as justification rather than as a non-sequitur.

**`ChildReaper.swift:23` — nonexistent symbol (MEDIUM).** Also correct. `startGroupedViewer` exists nowhere in `Sources/`; I asserted a name instead of verifying one. Now cites `TerminalPanelView.Coordinator.startTmuxClient(terminalView:bridge:server:windowID:)` (`TerminalPanelView.swift:656`, `@MainActor` at `:655`) and `LocalPTYTerminalView.Coordinator.start(terminalView:argv:environment:)` (`:99`, `@MainActor` at `:98`). Since the defect was an unverified citation rather than a typo, I swept **every** backticked symbol and file path in that doc: the rest resolve.

**`Package.swift:30` — cell-width rounding drift (minor).** Good catch, and it turns out to be measurable rather than speculative — but it does not point where it looks like it points:

| | width |
|---|---|
| old pin `dae32cc` | `ceil(w * scale) / scale` |
| new pin `16c5286` | `(w * scale).rounded() / scale` |
| `cellDimensions(for:)` | **no snapping** — raw advancement |

The mirror never snapped, so it diverged from the old pin too; this is **pre-existing, not introduced here**. And the direction is favourable: round-to-nearest sits closer to the raw advancement than `ceil` did, so the bump *narrows* the gap. The bound is one device pixel per cell (≤0.5pt at 2x).

I documented it at the mirror rather than "fixing" it, because the divergence is structural: `cellDimensions(for:)` is static and view-less — it has no `backingScaleFactor()` to snap against, which is exactly why `AppState` can call it without a live view. Snapping properly means threading a target screen's scale through the call sites, which is a geometry change deserving its own soak, not a rider on a rendering fix. The comment names the removal trigger (an observed off-by-one column against a real pane) and links it to the desync class `ControlModeGeometry` already documents.

Verification for this round: `swiftlint --strict` clean; changes are comments only, no behaviour change.

cheapsteak:
Round 2 — fixed in `6e9776c9`, and the finding understated the problem.

You're right that `LocalPTYTerminalView.Coordinator` doesn't resolve — the file declares `LocalPTYTerminalRepresentable` (`:26`), whose nested `Coordinator` is at `:79`. But **the other half of that same sentence was also wrong**, including the replacement suggested in round 1 and adopted here: `Coordinator` in `TerminalPanelView.swift` is at `:396`, nested inside `struct TerminalPanelRepresentable` (`:234`) — *not* inside `struct TerminalPanelView` (`:40`), which is a sibling `View`. So `TerminalPanelView.Coordinator` doesn't resolve either. Both paths now name the `NSViewRepresentable` that actually owns the `Coordinator`, and the comment says so explicitly, since "the outer type is not the one the file is named after" is the trap in both files.

On "the immediately preceding commit claimed to have swept the rest clean, and this one slipped through" — that's the part worth fixing properly, so: the round-1 sweep was **methodologically wrong**, not unlucky. It asked *"does this file declare a `Coordinator`?"* and got yes for both files, when the question that matters is *"does this dotted path resolve to a nested declaration?"* A file-name-shaped check cannot catch a file-name-vs-type-name mismatch, which is exactly the defect class in play.

This round's sweep parses every `Outer.Inner` reference in the doc, locates `Outer`'s declaration, and confirms `Inner` is nested inside it:

```
  OK   LocalPTYTerminalRepresentable.Coordinator  (Sources/TBDApp/Remote/LocalPTYTerminalView.swift)
  OK   TerminalPanelRepresentable.Coordinator     (Sources/TBDApp/Terminal/TerminalPanelView.swift)
```

`swiftlint --strict` clean. Doc-comment only, no behaviour change.

cheapsteak:
Noting a factual slip in the approval body, since it would otherwise become the stale prose this PR spent two rounds removing.

The verdict is for the current head — its `last-reviewed-patch-id` (`fdeda6383fda1e55`) matches `git diff main...HEAD | git patch-id --stable` exactly, so it read the corrected text. But the **Correctness** paragraph says it verified `TerminalPanelView.Coordinator.startTmuxClient(...)` and `LocalPTYTerminalView.Coordinator.start(...)` and that both "exist with the cited signatures". Neither path exists, and neither is what the file says any more:

- `Coordinator` in `TerminalPanelView.swift` is at `:396`, nested in `struct TerminalPanelRepresentable` (`:234`) — `TerminalPanelView` (`:40`) is a sibling `View` that does not contain it.
- `LocalPTYTerminalView.swift` declares no type of that name at all; it declares `LocalPTYTerminalRepresentable` (`:26`), whose `Coordinator` is at `:79`.

That is the same claim this reviewer correctly **rejected** one round earlier, now asserted as verified. The line numbers it cites (`:655-662`, `:98-99`) are right, so the underlying conclusion — both call sites are `@MainActor` and reach `LocalProcess` on main — does hold for the code as it now stands. No action needed on the diff; flagging it only so nobody later reads the approval body as evidence that `TerminalPanelView.Coordinator` resolves.

PR #691, opened 2026-08-20
feat: warn before archiving a worktree with dev servers still running

cheapsteak:
## Summary

Archiving a worktree closes the terminals its sessions live in, and a dev server started inside it does not necessarily go with them. #681 now reclaims those — its sweep kills processes whose cwd is inside an archived worktree, at daemon start and hourly after.

So the process no longer leaks. What is still missing is that **you are not told**, and the decision is made for you: you archive, nothing appears to happen, and within the hour a server you may have wanted is gone. The reverse is just as opaque — a server you forgot about is reclaimed silently and you never learn it was running.

This asks first: *"`my-feature` still has storybook and dev running. Archiving removes the worktree, and its servers will be reclaimed."* Archive Anyway, or Cancel.

**This is the prevent half of #681's clean-up half, not a second cleaner.** It does not kill anything, and it does not change what the sweep reclaims. It adds the moment of choice ahead of an action that is otherwise irreversible-by-schedule, and it can name what is running — `storybook and dev` — which a cwd-keyed sweep structurally cannot, because it identifies processes by where they live rather than by what they are.

It knows because of a small cross-app convention: whatever starts a long-lived dev server writes a record describing it under `${XDG_STATE_HOME:-~/.local/state}/dev-servers/`. TBD only reads.

## Why this shape

**Read-only, and not a second reclaimer.** TBD never writes a record, and this path never stops anything — reclamation is #681's, and duplicating it would put two mechanisms with different identification rules in charge of killing the same process. This PR answers one question — *is there something here you would rather not lose?* — and leaves the rest alone.

**Declaration, where #681 uses detection, and the two answer different questions.** A cwd-keyed sweep can find a process without being able to say what it is; a record carries a label. That is also why this does not simply reuse #681's collector: it needs a name to put in a sentence, and the sweep deliberately identifies by location.

**No vendor segment in the path.** The directory is a surface several unrelated applications write to. A name owned by one of them gives the next adopter a reason to invent a second directory instead of joining the first, so the convention has to be neutral to be worth adopting.

**Liveness is asked of the kernel, never of the record.** A record cannot be trusted to be current: the common case is the writer dying without cleaning up, which is the whole reason the feature exists. So the identity is `(pid, start time)`, checked through `sysctl`:

- a **reused pid** reads stale rather than pointing at a stranger — the flaw every `.pid` file scheme has;
- a **process that exited without being reaped** reads stale rather than running forever, because a zombie still reports its original start time and would otherwise satisfy the pair check indefinitely.

**Three states, not two.** `running`, `stale`, and `indeterminate` — the last meaning the record could not be read or its schema version is not one this build understands. Only `running` prompts. A definitively dead record describes nothing an archive could strand, and one that cannot be read is not evidence either; prompting on those teaches people to click through the prompt, which costs more than the warning saves.

**A separate flag from `force`.** `force` already means "skip the archive hook" and bypasses a status check for cascade flows. Confirming the dev-server prompt passes `devServersConfirmed: true` and re-issues the *original* `force`. Folding the two together would make confirming a prompt silently skip a repo's archive hook, and would make every cascade flow skip this check.

**The record's `command` is display text and nothing reads it as an instruction.** Records are written at runtime by any process, with no review, while TBD is long-running and trusted. Executing a string out of one would let a process that cannot run arbitrary commands itself get one run on its behalf.

Spec: [`docs/specs/2026-08-20-dev-server-registry-design.md`](docs/specs/2026-08-20-dev-server-registry-design.md), written for someone implementing a *writer*. This PR originally shipped without one on the grounds that it was a single read plus a dialog; that undersold it — it commits TBD to a versioned on-disk protocol other applications write into, with a liveness rule carrying several judgment calls, which is a change to the system's theory rather than a UI addition.

## What this PR does

- `TBDShared/DevServerRegistry.swift` — the reader. Record decoding, the three-state liveness rule, a `sysctl` probe behind a protocol, and `running(inWorktree:)` for the one question the app asks.
- `TBDApp/PendingArchive.swift` — the held-back archive: which worktree, which servers, and the original `force` so confirming re-issues the same call.
- `AppState.devServerArchiveGate` — a pure seam deciding hold-vs-proceed, alongside `archiveShortcutRoute`, so every arm is testable without a daemon.
- `AppState.archiveWorktree` — runs the gate, ahead of the scratch delegation.
- `RowActionMenuActions` — the sidebar's scratch archive routes through `archiveWorktree` rather than straight to the RPC.
- `RepoSectionView` — the existing Remove-Repo confirmation now also names running servers, because that path cascade-archives every worktree in the repo server-side and never reaches the per-worktree gate.
- `ContentView` — the confirmation, naming what is running.

## Assumptions

- **Assumes records are written by the launchers that start dev servers.** Nothing here creates them, so on a machine where nothing declares anything this feature is silently inert — correctly, but invisibly. It is a warning about *declared* servers, never a census: a server started by hand from a shell declares nothing, and gets archived with no prompt (and then reclaimed by #681's sweep, which does not need a declaration to find it).
- **Assumes #681's sweep remains the thing that reclaims.** If teardown-time reaping later lands — explicitly deferred in #681's spec — this warning should be revisited alongside it rather than left to describe a sequence that changed.
- **Assumes `localPath` is a real filesystem path for local rows.** Remote rows are skipped explicitly — their `localPath` is the synthetic `remote://` URI, and a registry on this machine cannot speak for a worktree on another one.
- **Assumes the answer is about intent, not a snapshot.** Between the prompt appearing and the click, the servers may exit. Confirming archives regardless rather than re-checking; re-asking on a second click would make the button appear broken.

## Evidence & verification

28 tests (17 reader, 8 gate, 3 prompt), and the ones that matter are about refusing to be confident:

- a reused pid reads stale, not running
- a zombie reads stale, not running
- an unrecognised schema version reads indeterminate, never stale
- one second of clock skew still reads running, so a truncate-vs-round disagreement between implementations does not report every live server as dead
- stale and indeterminate records are excluded from the prompt but still visible in the full listing, so a malformed record is never indistinguishable from no record
- worktree paths are compared after resolving symlinks — comparing raw strings would report "nothing running" for every worktree reached through a symlinked ancestor, and that failure looks exactly like the feature working
- an absent registry directory is empty, not an error
- `force` round-trips through the confirmation, so confirming cannot silently turn a repo's archive hook on or off
- a scratch row is gated like any other — it is a real directory that can host a server, and it used to be delegated past the check
- a remote row is not merely excluded from the result but never *looked up*, so the exclusion cannot silently become "the comparison happened not to match"

The liveness cases run against an injected probe, because a reused pid and an unreaped zombie cannot be produced on demand from a test process. One case exercises the real `sysctl` probe against this process and an unallocatable pid, so the suite cannot pass against a probe that answers `nil` for everything.

### DIFF
--- a/Sources/TBDApp/AppState+Worktrees.swift
+++ b/Sources/TBDApp/AppState+Worktrees.swift
@@ -584,11 +584,54 @@ extension AppState {
     /// spaces are delegated to `archiveScratch(id:)` — the repo-worktree
     /// `worktree.archive` RPC rejects them with `repoNotFound` — so callers
     /// don't have to route per collection.
-    func archiveWorktree(id: UUID, force: Bool = false) async {
+    /// Archive, unless the worktree still has declared dev servers running.
+    ///
+    /// Archiving removes the checkout and kills the terminal windows the
+    /// worktree's sessions live in — but a dev server started inside it does not
+    /// necessarily die with them. A detached supervisor, or one whose parent has
+    /// already exited, is reparented to `launchd` and keeps holding memory and a
+    /// port with nothing left pointing at it. Once the directory is gone, the
+    /// only remaining pointer is a record in a registry nobody thinks to read.
+    ///
+    /// So the first archive attempt asks. Confirming calls back in with
+    /// `devServersConfirmed: true` — the question has been answered, and
+    /// re-asking would make the second click do nothing.
+    ///
+    /// `devServersConfirmed` is a SEPARATE flag from `force`, deliberately.
+    /// `force` already means "skip the archive hook" and bypasses the status
+    /// check for cascade flows; folding this into it would make confirming a
+    /// dev-server prompt silently skip the repo's archive hook, which nobody
+    /// asked for, and would make every cascade flow skip this check.
+    ///
+    /// Only `running` records prompt. A record whose process is definitively gone
+    /// describes nothing an archive could strand, and one that cannot be read is
+    /// not evidence either; prompting on those would train people to click
+    /// through the prompt, which costs more than it saves.
+    func archiveWorktree(
+        id: UUID, force: Bool = false, devServersConfirmed: Bool = false
+    ) async {
         let worktree = findWorktree(id: id)
         if worktree?.isScratch == true {
             await archiveScratch(id: id)
             return
+        }
+        // `localPath` is a filesystem path only for a LOCAL row. On a remote row
+        // it is the synthetic `remote://…` URI that exists to satisfy a NOT NULL
+        // UNIQUE column, and a dev-server registry on this machine can say
+        // nothing about a worktree that lives on another one. Skipping is the
+        // honest answer; comparing would silently never match and read as
+        // "nothing running" for a reason that has nothing to do with the facts.
+        if !devServersConfirmed, let worktree, worktree.localPath.hasPrefix("/") {
+            let running = devServerRegistry.running(inWorktree: worktree.localPath)
+            if !running.isEmpty {
+                pendingArchiveWithDevServers = PendingArchive(
+                    worktreeID: id,
+                    worktreeName: worktree.displayName,
+                    servers: running.map(\.label),
+                    force: force
+                )
+                return
+            }
         }
         let worktreeName = worktree?.displayName ?? "worktree"
         do {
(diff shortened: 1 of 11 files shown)


cheapsteak:
feel like this is unnecessarily noisy; if you made the decision to archive, it's reasonable to expect that dev servers the worktree is running will be killed

PR #693, opened 2026-08-20
Fix: `tbd worktree archive` couldn't archive a scratch space

Jephuff:
## What's broken

Archiving a scratch space from the CLI fails, and the error names the wrong
object:

```
$ tbd worktree archive 20260814-zygomorphic-ant
Error: Repository not found: E3DB7381-2215-4B1A-B7A2-DEB42E4CE4A5
```

That UUID is the *worktree* id. Nothing looked up a repo.

Anyone who works in a scratch space (a repo-less workspace under
`~/tbd/scratch/<name>`) and wants to put it away from the terminal hits this
every time. The sidebar's Archive item works, so the failure looks like a CLI
bug rather than a missing capability, and the only CLI verb that appears to
apply is the wrong one: `tbd worktree forget` hard-deletes the row and its
closed-terminal history with no inverse. `tbd worktree revive` was broken the
same way, so even a successfully archived scratch space could not be brought
back from the CLI.

## Why it happens

A scratch space is defined by having no repository (`Worktree.isScratch` is
`repoID == nil`). Archive's phase 1 resolves a repo before it does anything
else, in one guard that covered two different failures:

```swift
guard let rid = worktree.repoID, let repo = try await db.repos.get(id: rid) else {
    throw WorktreeLifecycleError.repoNotFound(worktree.repoID ?? worktreeID)
}
```

For a scratch row the *first* clause fails, so `worktree.repoID` is nil and the
`?? worktreeID` fallback puts the worktree id into a message about
repositories. `beginReviveWorktree` carried the identical guard.

The daemon has always been able to archive a scratch space: `scratch.archive`
does it, and the app reached that path by routing client-side in
`AppState.archiveWorktree`, whose doc comment recorded the workaround. Only the
app knew to route. Everything else, the CLI included, sent `worktree.archive`
and got the misleading error.

## What this PR does

`worktree.archive` and `worktree.revive` route a repo-less row to the
`scratch.*` body, in the same pre-row branch those handlers already use for
remote lanes:

- `handleWorktreeArchive` dispatches a scratch row to the new
  `archiveScratchSpace(worktreeID:surface:force:actor:)`, and
  `handleWorktreeRevive` to `reviveScratchSpace(worktreeID:)`. Both are the
  extracted bodies of the existing `scratch.archive` / `scratch.revive`
  handlers, so the two RPC doors share one implementation.
- `WorktreeLifecycleError.worktreeHasNoRepo` names the repo-less case, and the
  guard splits in three places so `repoNotFound` only ever carries a repo id.
- The app's client-side branch stays (its state settlement differs: it also
  refreshes the Scratch section's Archived tab), with its comment corrected.

Routing in the daemon rather than adding `tbd scratch archive` was the design
call. `worktree.archive` is already the general verb, not the repo-worktree
verb: the same handler dispatches a remote lane to `archiveRemoteLane` because
"the files are on another machine" makes the local teardown meaningless for it,
and "the worktree has no repo" is the same kind of row-shaped fact discovered
the same way. The CLI's `worktree` noun already covers scratch spaces
everywhere else (`tbd worktree list` prints them, name resolution finds them);
`archive` refusing them was the odd one out. And one daemon-side branch covers
the CLI, hooks, scripts, and clients that do not exist yet, instead of leaving
each to rediscover the trap.

The alternative of teaching `beginArchiveWorktree` to tolerate a nil `repoID`
was rejected on risk asymmetry: the phase-2 body it would newly have to gate is
the destructive one (archive hook, deletion-queue enqueue, directory rename,
`git worktree prune`), and a missed gate there deletes the user's scratch
folder. Delegating to a path that never had a phase 2 makes that unreachable
rather than merely guarded.

A dedicated `tbd scratch archive|delete|revive` surface is deliberately not
here: with the routing in place those would be aliases for verbs that already
work. Reasonable follow-up for discoverability.

Design, including the accepted trade-offs (teardown fidelity, actuation kind,
inert `force`) and the risks:
[`docs/specs/2026-08-20-scratch-archive-routing-design.md`](docs/specs/2026-08-20-scratch-archive-routing-design.md).

### Refusals the routed path keeps

Routing must not make `worktree.archive` a quieter verb than it was, so three
guards `scratch.archive` never had are added to the scratch body (both doors get
them):

- **Active children.** `assertArchivable` asks whether a row *is* a parent, and
  a scratch space can be one: `ParentResolver`'s `caller` arm returns the caller
  id after rejecting only `.main`/`.archived`, so `tbd worktree new` run from
  inside a scratch space parents the new worktree to it. Refused unless `force`,
  which is now forwarded rather than swallowed as inert.
- **Already archived.** An idempotent no-op instead of a re-archive.
  `WorktreeStore.archive` re-stamps `archivedAt` unconditionally, and that is
  the grace clock `OrphanProcessCollector` reaps orphaned agents from, so a
  retry pushed the reap out by another grace window and re-broadcast a delta for
  a row clients had already filed away.
- **Promoted.** Promotion leaves `repoID` NULL, so `isScratch` and revive's
  status check both still pass; only the moved folder incidentally blocked
  revive. Both sibling verbs guard the pointer explicitly, so revive does too.

Revive also re-reads the row *before* broadcasting, so the one failure that can
follow the committed write (a concurrent delete) no longer tells subscribers the
revive happened while telling the caller it did not.

Actuation records the door, not the destination: `archiveScratchSpace` takes the
surface, so a routed call is filed under `worktree.archive` per
`ActuationSurface`'s own rule that "`method` names the door a request came
through and `kind` names the act". No `ActuationBranch` is needed because both
surfaces already map to `kind: .dispose`. The routed revive records against
`.worktreeRevive` rather than being the one branch of that door to skip a
fail-closed log gate.

## Known gaps, deliberately not in this PR

Pre-existing properties of the scratch body and the app, unchanged by the
routing but newly reachable from every client. All recorded in the spec:

- No archive hook runs for a scratch space through either door. Hook resolution
  is **not** repo-scoped (`<worktreePath>/.worktree-hooks/<event>` and a global
  `~/tbd/hooks/default/<event>` both apply, and a scratch space is routinely a
  git repo), so this is a real gap rather than an absence.
- The scratch body tears terminals down before the status flip and stamps
  `.dispatched` before a throwable write, the inverse of the ordering the repo
  path documents as load bearing.
- Revive spawns no terminals, so a revived scratch space comes back with nothing
  to attach to. The GUI's Archived tab has always behaved this way.
- Claude session ids are neither captured at archive nor preserved across
  revive. The two halves mask each other, so fixing either alone would clear ids
  nothing recorded.
- No delta handler updates the app's `archivedScratchWorktrees`, so a
  daemon-originated archive or revive leaves the Scratch Archived tab stale; and
  a `tbd://` deep link to an archived scratch space is a silent no-op.
- `DeskSessionManager.closeDeskSession` is a third, drifted copy of this
  teardown (it leaks per-session overlay files).

## Assumptions

- **`repoID == nil` means "scratch space", not "inconsistent row".** The
  routing keys on it. There are exactly two writers of a null `repoID`, both
  through `WorktreeStore.createScratch`: `handleScratchCreate`, whose rows are
  the intended target, and the Watch Desk's session creation, whose row both
  shared bodies refuse explicitly. A third caller would widen this routing
  silently, and its rows would need their own answer.
- **A dangling `repoID` is not constructible through the store** (the column is
  `REFERENCES repo(id) ON DELETE CASCADE`), so `repoNotFound(rid)`'s new
  argument is covered by the error-rendering test rather than a router fixture.

## Evidence & verification

**Repro, then the fix, end to end through the CLI** against a daemon on an
isolated `TBD_HOME` (the live daemon was not restarted). Before, on the same
build with only the error-message half applied:

```
$ TBDCLI scratch new --name repro-scratch
Created scratch space: repro-scratch
$ TBDCLI worktree archive repro-scratch
Error: Worktree 0F170262-... has no repository (it is a scratch space), and this operation requires one.
$ TBDCLI worktree list --status archived
No worktrees found.
```

After:

```
$ TBDCLI worktree archive repro-scratch
Worktree archived.
$ TBDCLI worktree list --status archived
1E1F6A30-46B1-4380-A7A8-7AA284504302  repro-scratch  archived
$ ls -d $TBD_HOME/scratch/repro-scratch    # folder survived
$ TBDCLI worktree revive repro-scratch
Worktree revived: repro-scratch
```

**Discriminating tests.** `WorktreeArchiveScratchRoutingRPCTests` (11 tests) and
`WorktreeLifecycleRepoErrorMessageTests` (2) are new. Reverting only
`RPCRouter+WorktreeHandlers.swift` and re-running reds the routing tests with 18
issues, including the archive returning `success: false`, the row staying
`.active`, and the archived listing not containing it.

Both false arms of `if existing.isScratch` are covered, per CLAUDE.md's
test-per-branch rule, and both are driven through guards that refuse before any
tmux, git, or disk work: a `.main` row for archive, an `.active` row for revive.
That placement is deliberate. An earlier version of the archive-side test drove
the real repo path, whose phase 2 is an unawaited `Task.detached` that outlived
the test's own `TBD_HOME` cleanup and would have renamed directories in the
unfenced real `$TMPDIR`; it also asserted only that the error was not one
specific string, so it would have stayed green with the repo path fully
broken.

Covered properties: archive succeeds and sets `archivedAt`; the folder is still
on disk afterwards; the row survives (archived, not forgotten) and appears in
`worktree.list --status archived`; revive returns a decodable `Worktree` (its
CLI and app clients both decode one, so `.ok()` would break them) and flips
back to `.active`; terminals and tabs are torn down; `--force` is inert;
revive still refuses when the folder vanished; a scratch space with an active
child is refused while `--force` still bypasses it; a second archive re-stamps
no `archivedAt` and broadcasts no second delta; and a promoted row is refused
even with a directory recreated at its stale path, so the guard rather than the
filesystem is what refuses it.

**Suite and lint.** `scripts/test.sh --filter '^TBDDaemonTests\.'`: 4068 tests
in 362 suites passed (58.1 s). `scripts/test.sh --skip
'^(TBDDaemonTests|TBDDaemonLiveTests)\.'`: 3515 tests in 366 suites passed
(43.3 s). `swiftlint --strict`: 0 violations in 792 files.
`scripts/swift-safe build`: clean.

### DIFF
--- a/Sources/TBDApp/AppState+Worktrees.swift
+++ b/Sources/TBDApp/AppState+Worktrees.swift
@@ -581,9 +581,14 @@ extension AppState {
     }
 
     /// Archive a worktree. Scratch-aware by construction: repo-less scratch
-    /// spaces are delegated to `archiveScratch(id:)` — the repo-worktree
-    /// `worktree.archive` RPC rejects them with `repoNotFound` — so callers
-    /// don't have to route per collection.
+    /// spaces are delegated to `archiveScratch(id:)` so callers don't have to
+    /// route per collection.
+    ///
+    /// The daemon routes them too now (`worktree.archive` dispatches a scratch
+    /// row to the same `scratch.archive` body), so this branch is no longer
+    /// what keeps the call from failing. It stays because the two client-side
+    /// settlements differ: `archiveScratch` also refreshes the Scratch
+    /// section's Archived tab, which the repo-worktree leg has no reason to do.
     func archiveWorktree(id: UUID, force: Bool = false) async {
         let worktree = findWorktree(id: id)
         if worktree?.isScratch == true {
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -2057,6 +2057,14 @@ final class AppState: ObservableObject {
             applyWorktreeArchivedDelta(d)
         case .worktreeRevived(let d):
             recentlyArchivedWorktreeIDs.removeValue(forKey: d.worktreeID)
+            // A revived row must leave the Scratch section's Archived tab, not
+            // just rejoin the sidebar. `refreshWorktrees` fetches with
+            // `excludeArchived: true` and never touches
+            // `archivedScratchWorktrees`, so without this a row revived from
+            // the CLI stays listed there while also showing as active, and
+            // that stale row's Delete button trashes the folder of a scratch
+            // space that is now live.
+            archivedScratchWorktrees.removeAll { $0.id == d.worktreeID }
             Task { [weak self] in await self?.refreshWorktrees() }
         case .controlModeInputHealthChanged(let d):
             applyControlModeInputHealthDelta(d)
@@ -2173,6 +2181,15 @@ final class AppState: ObservableObject {
         let failureMessage = Self.creationFailureMessage(worktree, creationFailed: delta.creationFailed)
 
         removeArchivedWorktreeFromState(id: delta.worktreeID)
+
+        // The mirror of the `.worktreeRevived` case: an archive that originated
+        // elsewhere (the CLI, a hook, a script) must ADD the row to the Scratch
+        // section's Archived tab, or an archived scratch space looks deleted:
+        // gone from the sidebar and absent from the only view that lists it.
+        // Cheap because the listing is small and unpaginated.
+        if worktree?.isScratch == true {
+            Task { [weak self] in await self?.refreshArchivedScratch() }
+        }
 
         // The daemon tells us whether creation actually failed; we never infer
         // it from `.creating` status. A deliberate archive of a still-creating
--- a/Sources/TBDApp/Helpers/RowActionMenu.swift
+++ b/Sources/TBDApp/Helpers/RowActionMenu.swift
@@ -373,8 +373,18 @@ enum RowActionMenu {
                 .action(Action(kind: .rename, title: "Rename...")),
                 // `.destructive` for consistency with the repo-worktree Archive
                 // even though scratch archive is recoverable and leaves the
-                // folder on disk.
-                .action(Action(kind: .archiveScratch, title: "Archive", role: .destructive)),
+                // folder on disk. Gated on active children for the same reason
+                // and with the same copy as the repo-worktree Archive: a scratch
+                // space can be a parent (`tbd worktree new` run from inside one
+                // parents the new worktree on it), and the daemon refuses that
+                // archive, so the item must not look available.
+                .action(Action(
+                    kind: .archiveScratch,
+                    title: context.hasActiveChildren ? archiveHasChildrenLabel : "Archive",
+                    role: .destructive,
+                    isEnabled: !context.hasActiveChildren,
+                    disabledHelp: context.hasActiveChildren ? archiveNeedsChildrenGoneHelp : nil
+                )),
             ],
             pinActions(context: context).map(Item.action),
             hibernationActions(context: context).map(Item.action),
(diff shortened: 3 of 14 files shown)


cheapsteak (approved):

cheapsteak:
ty 🙏

PR #698, opened 2026-08-21
Compound: Carry the PR worktree-link convention in the repo, not just the session preamble

cheapsteak:
## Summary

A PR opened from a TBD worktree should end its body with a deep-link back to that worktree, so the PR can be reopened as a live session months later — the reviewer's question, the half-finished follow-up, and the terminal that produced the diff are all still there. Today that instruction exists only in the TBD session preamble, which is injected at runtime and belongs to the harness, not the repo. When it isn't injected — a Codex session, a plain terminal, a fresh agent, a contributor reading `CLAUDE.md` — the convention simply isn't there, and the link is silently dropped.

This carries the convention in the repo instead, in two places:

- **`CLAUDE.md`** gains one Workflow bullet with the recipe. That's where this repo's PR conventions are wired, because `gh pr create --body` bypasses `.github/PULL_REQUEST_TEMPLATE.md` entirely — the template never reaches an agent-opened PR.
- **A guardrail rule** (`.claude/hooks/guardrails/rules/pr_worktree_link.py`) nudges at the moment it matters: as `gh pr create` / `glab mr create` is about to run, when the link isn't already in the body.

The recipe needs spelling out because the two halves don't line up on their own: `tbd link` prints the `tbd://open?worktree=<uuid>` scheme form, while a PR body needs the shareable https redirector — GitHub strips hrefs outside its http/https/mailto allowlist, so a raw `tbd://` link renders as unclickable text.

## Why a nudge and not an auto-stamp

A `PostToolUse` hook could append the link to the body itself, with no model involvement. Rejected for now: writing to a PR body means identifying the right PR from a command that has already run, and the failure mode is stamping *someone else's* PR. That's a real hazard with a real precedent, and it is a much larger change than this one. A nudge costs one line of `additionalContext` on a command the agent runs a handful of times per branch, and it degrades to exactly nothing when the agent already did the right thing.

## The rule

`Decision.info` only — it can never deny. Blocking PR creation over a courtesy link would be badly wrong, and the framework's own contract is that a buggy guardrail must never block the agent. Verified end-to-end through `dispatch.py`: it emits `permissionDecision: "allow"` with `additionalContext`, and exits 0.

It stays silent when:

- the command isn't a real `gh pr create` / `glab mr create`;
- the link is already in the body — inline `--body` text, or a `--body-file` / `-F` target read from disk;
- the session isn't under TBD at all, i.e. no `TBD_WORKTREE_ID` and a cwd outside the worktrees root. A contributor who doesn't use TBD never sees this.

No `settings.json` change: the registry auto-discovers rule modules, and `Bash` already has a `PreToolUse` matcher.

## Detection is a port, not a re-derivation

The first draft matched the phrase with a regex, and it fired on its own `git commit -m "… \`gh pr create\` …"` call during implementation — a false positive observed live, not a hypothetical.

Phrase-matching is wrong in both directions. A `gh pr create` inside a heredoc body is prose the command *writes*; a quoted one is an argument; both must stay silent. Meanwhile `/opt/homebrew/bin/gh pr create` and `gh -R owner/name pr create` are ordinary invocations here and must not be missed.

This repo already answers that exact question in `Sources/TBDShared/PRBindingExtractor.isPRCreateCommand`, with documented defenses for every one of those cases. A stdlib-only Python hook can't call Swift, so the rule now mirrors that tokenizer — heredoc-body stripping, quote-aware segmentation, path-qualified command word, value-taking flag skip, per-CLI verb pairing (`gh pr create` and `glab mr create`; `gh mr create` is rejected). Keeping the two in step is worth more than diverging coverage: the Swift side carries consequences a false match can't be taken back from, while a miss here costs nothing.

Like the Swift original it is a pragmatic tokenizer, not a shell parser, and fails **closed**. `FOO=bar gh pr create` and `if ! gh pr create; then …` are missed, deliberately, because their first word isn't `gh`.

## `$(…)` is its own quoting context

Both scanners track `$(` nesting and start each substitution with fresh quote state,
restoring the suspended quote at the matching `)`. Without it, a literal `"` anywhere in a
heredoc-composed body closes the enclosing span early, the body truncates before its last
line, and the nudge fires at a PR that *is* correctly linked. A body that quotes a phrase or
an error message is ordinary, and `--body "$(cat <<'EOF' … )"` is the documented way to
compose one, so the two meet often.

Fidelity stops there, and everything past it fails toward the nudge rather than away from it:
backticks are not treated as substitution, an unbalanced `$(` swallows the rest of the
command into one word, and `${…}` gets no special handling. This is a tokenizer, not a shell.

Body files are read in two bounded windows, head and tail, because the convention puts the
link on the body's **last** line while a bound has to start somewhere. Reading only the head
means a long body reads as unlinked precisely when it followed the convention. A body over
twice the cap with its link in the middle still nudges.

## One decision per body, not per invocation

A single Bash call can open more than one PR, so each create segment is judged on its own
body: if any of them lacks a link, the nudge fires. ORing across segments lets one
correctly-linked PR cover for an unlinked one chained after it.

The tokenizer collapses an unquoted backslash-newline pair the way a shell does. Keeping the
pair makes the newline a word, so `gh pr \` + newline + `create` reads its subcommand as
`('pr', '\n')`, matches nothing, and the rule never fires at all — a silent miss on an
everyday multi-line invocation. Inside single quotes the backslash stays literal; inside
double quotes the pair is still a continuation, which is how the shell treats both.

## What counts as "already linked"

The test is whether the body carries a **clickable** link, not whether some marker string
appears: `https?://\S*tbd/open/\S*[?&]worktree=`. Each part earns its place — `https?://`
because that is the only scheme GitHub linkifies; the `/tbd/open/` path as a whole segment, with no host, so a
fork serving the redirector from its own Pages domain counts while an unrelated host's
`/nottbd/open/` does not; and `worktree=` required
as a query parameter of that URL rather than as a free substring, with `[?&]` so a
hand-ordered `?terminal=…&worktree=…` also counts.

The `https?://` half is the one that matters most in practice. `tbd link` prints
`tbd://open?worktree=<uuid>`, and GitHub strips hrefs outside its http/https/mailto
allowlist — so pasting that output straight into the body produces text that renders but
cannot be clicked. That is the single most likely way to get this wrong, which makes it the
case the nudge most needs to catch. A substring test cannot catch it: `tbd://open?worktree=…`
contains `?worktree=`, so it reads as satisfied under any marker set that accepts the query
fragment.

## Where the rule looks for an existing link

Body flags are read in the **invoked CLI's own spelling**, keyed off which tool the segment
matched: `--body`/`-b` and `--body-file`/`-F` for `gh`, `--description`/`-d` for `glab`.
Sharing one flat flag list across both is a false-negative generator, because the spellings
collide with different meanings — gh's `-d` is `--draft`, a boolean taking no value, so a
flat list reads the *next* word as a body and a link-shaped token there silences the nudge on
a body that has none. glab has no body-file entry: it takes its description inline, and
inventing a file flag that couldn't be verified is worse than omitting one.

A repeated body flag takes the **last** occurrence, the way both CLIs resolve it — spellings
within one family are aliases, so `-b … --body …` resolves to the `--body` value. A segment
carrying both a body-text and a body-file flag is not disambiguated: both CLIs reject that
combination outright, so the command opens nothing and there is no body for the nudge to be
right or wrong about.

Scanning the whole command for a link marker is wrong: any marker anywhere in the invocation
would suppress the nudge even when the real body has none — `git log --grep "tbd://open?worktree=…" && gh pr create --body "no link"`
would go silent. So the scan is scoped to the body, in exactly three places: the inline
`--body`/`-b` (and glab's `--description`/`-d`) value from the tokenized create segment; the
contents of any `--body-file`/`-F` target; and heredoc bodies.

The heredoc leg is what makes this more than a token lookup. `--body "$(cat <<'EOF' … EOF)"`
is the shape this repo uses to open its own PRs, and the link inside it is invisible to a
token scan. Two variants, silenced by different legs: the inline form leaves the whole quoted
span as one `--body` word (the unterminated `"` means no heredoc opener is detected), so the
body-value branch catches it; a heredoc redirected into a file is carried by the heredoc leg.

**The heredoc leg is invocation-wide, not attributable to the create segment** — a heredoc
body written to an unrelated file in the same invocation still silences the nudge. That is
the deliberate residual, kept because the alternative is a false nudge on the repo's own
PR-opening idiom, and it is the only whole-command-scoped leg left. A body passed through a
plain shell variable (`BODY="…link…"; gh pr create --body "$BODY"`) nudges; the common
`BODY=$(cat <<'EOF' …)` form is silent via the heredoc leg.

Body files are read by descriptor, because this code runs in front of *every* Bash call and
blocks it: `os.open(..., O_RDONLY | O_NONBLOCK)`, then `fstat`/`S_ISREG` before a single
capped read. Checking `isfile()` and opening as separate steps leaves a window where the path
becomes a fifo and the open blocks with no writer — hanging the guardrail, and the agent with
it. `O_NONBLOCK` makes a fifo open return instead of blocking, and the descriptor check
happens on the thing actually opened.

## Assumptions

- The redirector at `https://cheapsteak.github.io/tbd/open/` stays the shareable form. It's already the single source in `Sources/TBDShared/DeepLinks.swift`; if it moves, `_LINK_MARKERS` and the message need updating with it.
- An agent that sees `additionalContext` acts on it. Nothing enforces the link — by design. If it turns out to be ignored in practice, auto-stamping is the escalation.

## Evidence & verification

`python3 .claude/hooks/guardrails/dispatch.py --self-test` → **106 tests, OK** (was 52). CI already runs this in the `test.yml` guardrails step, so the new tests are gated on merge.

Every branch is mutation-checked — broken, the corresponding test reddened, restored. 16 targeted mutations, all caught, no gaps: heredoc strip, `/gh` suffix, value-taking flags, glab entry, verb/CLI pairing, quote tracking, read cap, `isfile` guard, the `_under_tbd` gate, the `TBD_WORKTREE_ID` branch, inline-link check, body-file check, raw-vs-stripped link scan, path truncation, `=`-joined body-file, and `info`→`deny`.

Behavior confirmed end-to-end through the real dispatcher, not just against the rule object — 14 commands, no deny in any of them:

- nudges: `gh pr create --title … --body …`, `/opt/homebrew/bin/gh pr create --fill`, `gh -R owner/name pr create --fill`, `glab mr create --title x`, `git push && gh pr create --fill`, `gh pr create --body-file <file without the link>`
- stays silent on a linked body it once nudged at: a heredoc body quoting a phrase before the link, and a 70 KB `--body-file` with the link on its last line; both still nudge when the link is absent
- nudges on a continued invocation: `gh pr \` + newline + `create --body 'no link'`; the same with a link in the body is silent
- judges each body separately: `gh pr create --body '<link>' && gh pr create --body 'no link'` nudges; both-linked is silent
- reads each CLI's own flags: `gh pr create -d '<link>'` nudges (`-d` is `--draft`), `glab mr create -d '<link>'` is silent (`-d` is glab's body), `glab mr create --body '<link>'` nudges
- takes the last repeated body flag: `--body '<link>' --body 'no link'` nudges; the reverse order is silent
- nudges on an unclickable link: `--body "see tbd://open?worktree=ABC"` — the form `tbd link` prints, and the form GitHub will not linkify
- silent on the clickable form and its variants: markdown-wrapped, `&terminal=` appended, `worktree=` not first, a fork-hosted redirector
- nudges despite a marker elsewhere: `git log --grep "tbd://open?worktree=zzz" && gh pr create --body "no link"`, `gh pr create --body "none" && echo "<link>"`, `gh pr create --body-file /tmp/tbd/open/body.md` (marker in the *path*, not the content)
- silent: `git commit -m "docs: explain \`gh pr create\` usage"`, a heredoc whose body mentions `gh pr create`, `gh pr create --body-file <file with the link>`, `gh pr create --body "see [wt](<link>)"`, `gh pr list`
- silent outside TBD: the first two above, run with no `TBD_WORKTREE_ID` and a cwd outside the worktrees root
- nudges on env alone: a cwd outside the worktrees root but `TBD_WORKTREE_ID` set — the case that matters for a subagent in an ephemeral checkout

The body-file read is capped in **bytes**, not characters: it opens in binary mode, reads at
most 64 KB, and decodes with `errors="replace"`. A text-mode read counts characters, which
lets multi-byte UTF-8 pull in several times the cap. A byte cap can land mid-character; the
link markers are pure ASCII, so a mangled trailing character can never fabricate a match.
`test_body_file_read_is_bounded_in_bytes_not_characters` pins it with a 3-byte-per-character
body sized to put the link past the byte cap but inside the character cap — it fails against
a character-counting read and passes against a byte-counting one. The ASCII bound test is
kept alongside it.

No Swift changed, so no build was required.

## No spec — maintainer's call

The repository owner reviewed this against `CLAUDE.md`'s brainstorming gate and made the
call explicitly: **no spec.** Recording the reasoning here so the decision is examinable
rather than implied.

The gate exists because code expresses a theory of the system, and a diff shows what
changed but not why the theory did. This change revises no theory. The convention it
carries already governs how PRs are opened here — it lived in the TBD session preamble,
which is injected at runtime by the harness rather than committed. What the PR changes is
the *durability* of that convention, not its content: a practice that vanished whenever the
preamble wasn't injected now survives in the repo. Nobody has to decide whether agent-opened
PRs should carry a worktree backlink; that was already settled and already practiced.

The mechanism is correspondingly narrow — info-only, no flag, no migration, no autonomous
action, no state mutation, and `scratch_git_init.py` is the standing precedent for exactly
this shape of nudge rule.

The reviewer's counter-argument is a fair one and worth answering directly: the `CLAUDE.md`
bullet makes this PR the canonical source of a repo-wide policy, which the `scratch_git_init`
precedent doesn't cover. True — but being the canonical *record* of a settled practice is
not the same as originating it. Writing down what the team already does is documentation;
the gate is aimed at decisions that change what the team does.

## Follow-up

`Sources/TBDShared/PRBindingExtractor.swift` deserves a reciprocal one-line comment naming its Python mirror. The Python side already points at Swift; the reverse pointer is what protects against drift when someone hardens the Swift detector. Left out because this worktree has no `.build`, and a comment-only edit would cost a cold machine-serialized Swift build to verify.

## Test plan

- [ ] `python3 .claude/hooks/guardrails/dispatch.py --self-test` → 64 tests pass
- [ ] `python3 .claude/hooks/guardrails/dispatch.py --list` shows `pr-worktree-link`
- [ ] Open a PR from a TBD worktree without the link in the body — the nudge appears, the command still runs
- [ ] Open one with the link already in the body — no nudge
- [ ] Run `gh pr create` with `TBD_WORKTREE_ID` unset from outside `~/tbd/worktrees/` — no nudge

[🔗 Nudge PR Worktree Link](https://cheapsteak.github.io/tbd/open/?worktree=F1752EA3-B86A-479B-AB9E-6411A15E4265)


cheapsteak:
Both findings addressed.

**Medium — spec requirement.** The repository owner reviewed this and made the call
explicitly: no spec. The reasoning is now written into the PR description under
"No spec — maintainer's call" rather than left implied, including a direct answer to the
part the `scratch_git_init.py` precedent doesn't cover.

The short version: the gate exists for changes that revise our theory of the system, and
this one doesn't. The convention already governs how PRs are opened here — it lived in the
TBD session preamble, injected at runtime rather than committed. What changes is the
convention's durability, not its content. Being the canonical *record* of a settled practice
is not the same as originating it.

Thanks for framing it as a judgment call for a maintainer instead of a violation — that was
the right read, and it's what made the decision cheap to make.

**Minor — char-vs-byte read cap.** Correct, and fixed in b2acd777. `_file_has_link` now
opens in binary mode, reads at most `_MAX_BODY_BYTES` bytes, and decodes with
`errors="replace"`. The docstring notes that a byte cap can land mid-character and that the
ASCII-only link markers mean a mangled trailing character can't fabricate a match.

You were also right that the suite couldn't catch it — the bound test was ASCII-only.
`test_body_file_read_is_bounded_in_bytes_not_characters` now builds a body from a 3-byte
character, sized so the deep-link sits past the 64 KB byte mark but inside the first 64 K
characters. Verified discriminating: it fails against the character-counting read (the rule
finds the link and stays silent) and passes against the byte-counting one. Suite is 65 tests,
green.

**On the diagnostic:** noted that the correctness specialist couldn't run
`python3 .claude/hooks/guardrails/dispatch.py --self-test` and fell back to counting
`def test_` statically. Worth fixing in the workflow's permissions — the suite is stdlib-only
and runs in ~0.03s, so letting that specialist execute it would be strictly stronger
verification than counting test definitions. Filing that thought here rather than acting on
it, since it's the review workflow's business and not this PR's.

cheapsteak:
Both fixed in 8e2f74e8.

**Medium — raw-command link scan.** Correct, and I reproduced your exact command through the real dispatcher before touching anything: silent, when it should nudge.

The scan no longer touches the raw command. It looks in exactly three places: the inline `--body`/`-b` (plus glab's `--description`/`-d`) value taken from the tokenized create segment, the contents of any `--body-file`/`-F` target, and heredoc bodies. Your case nudges now, as does a marker in a segment chained *after* the create, and a `--body-file` whose *path* contains `tbd/open/` while its content has no link.

Worth naming why this wasn't just a deletion, since you flagged the raw scan as new logic not mirrored from `PRBindingExtractor` — it was load-bearing. `--body "$(cat <<'EOF' … EOF)"` is the idiom this repo opens its own PRs with, and the link inside it is invisible to a token scan because heredoc bodies are stripped before tokenizing. Removing the raw scan would have traded your missed nudge for a false one on every correctly-formed PR.

**One residual, stated plainly rather than left for you to find:** the heredoc leg is invocation-wide, not attributable to the create segment, so a heredoc body written to an unrelated file in the same invocation still silences the nudge. It is the only whole-command-scoped leg left, and it is deliberate — the alternative is a false nudge on the repo's own PR-opening idiom. Now documented in the PR description under "Where the rule looks for an existing link", along with what each leg covers.

**Minor — TOCTOU.** Fixed rather than deferred. You scoped it as low real-world risk, which is fair, but the hang would take the gated Bash call with it, and "a buggy guardrail must never block the agent" is the framework's own contract. `_file_has_link` now does `os.open(..., O_RDONLY | O_NONBLOCK)`, then `fstat`/`S_ISREG` on the descriptor it actually opened, then one capped read — the hardening you described. `O_NONBLOCK` is what makes the fifo open return instead of blocking.

Two notes from implementing it. A read-until-cap retry loop hit `EAGAIN` on a fifo and laundered the result through the fail-closed `except`, which hid the `S_ISREG` branch from every test; a single `os.read` of the cap keeps that branch observable. And `test_fifo_contents_are_never_read` holds the fifo open `O_RDWR` with the link pre-buffered, so it is deterministic and cannot hang the suite even if the guard regresses.

Suite is 69 tests, green. All eight branches mutation-checked — including dropping `O_NONBLOCK`, which reddens with "reading a fifo hung the guardrail".

cheapsteak:
Fixed in cd81215b — and you were right to flag it as worth a second look rather than assume the tested behaviour was intentional. It was a proxy that had drifted from its goal, and the test was pinning the drift.

**The obvious repair would not have worked.** Dropping `"tbd://open"` from the tuple leaves `?worktree=`, which `tbd://open?worktree=<uuid>` also contains — so the bare scheme would have kept reading as satisfied. The substring approach was the defect, not the one entry.

The check is now a pattern, `https?://\S*tbd/open/\S*[?&]worktree=`, with each part carrying weight:

- `https?://` — the only scheme GitHub linkifies. This is the half that inverts the bug.
- `tbd/open/` with **no host**, so a fork serving the redirector from its own Pages domain still counts.
- `worktree=` as a query parameter of that URL rather than a free substring, with `[?&]` so a hand-ordered `?terminal=…&worktree=…` counts too.

`test_silent_when_only_the_bare_scheme_is_in_the_body` is now `test_informs_when_only_the_bare_scheme_is_in_the_body`, asserting the opposite, with a comment recording that an unclickable link is precisely the case the nudge exists for. Your framing sharpened this beyond the fix itself: `tbd link` prints the `tbd://` form, so pasting its raw stdout is the most likely way to get this wrong — the case the rule should catch first, not the one it waves through. That reasoning is now in `_MESSAGE`, so the nudge says why the redirector form is the one that matters.

I also swept the prose for stale descriptions of the old markers: the module docstring, `_text_has_link`, `_MESSAGE`, and the `_file_has_link` comment claiming "the link markers are pure ASCII, so a mangled trailing character cannot fabricate a match" — that claim is rewritten and still true, since every literal the pattern anchors on (`https://`, `tbd/open/`, `worktree=`) is ASCII and a replacement character spells none of them.

Verified end-to-end through `dispatch.py`, never `deny`: bare `tbd://` nudges (with and without `&terminal=`); the https form is silent markdown-wrapped, with `&terminal=` appended, with `worktree=` not first, and on a fork-hosted redirector; the redirector path with no `worktree=` param nudges. Suite is 74 tests, green — all seven pattern branches mutation-checked, including restoring the old substring marker, which reddens the inverted test.

cheapsteak:
Took the Minor too, in df51f105 — it's the same shape as the two Mediums before it (a proxy that accepts something the goal doesn't), and the fix is one character.

`/tbd/open/` now matches as a whole path segment. `https://example.com/nottbd/open/?worktree=x` nudges; the real redirector and a fork-hosted one stay silent. `test_informs_when_the_path_only_ends_in_the_redirector_segment` pins it, and removing the leading `/` reddens exactly that test and nothing else. Suite is 75, green.

Flagging the environment quirk from your diagnostics, since it recurred across rounds and could mislead a future review: you noted that both new files were "present in the commit history at HEAD but absent from the working tree (deleted, uncommitted) at checkout time". That is an artifact of the review checkout, not this PR — my worktree has been clean at every push (`git status --short` empty, verified before each). Reading them via `git show HEAD:<path>` was the right call and gave you the true content; worth knowing that the working-tree state in that job isn't a signal about the branch.

cheapsteak:
Both fixed in ba7b3c4f.

**Medium — per-CLI flag scoping.** Right, and the sharpest part of the finding is that the comment above the constant already described the correct behaviour while the code did something else. `_is_forge_create_segment` returned a bare bool, so which tool matched was computed and then thrown away one line later. That is now `_forge_create_tool`, returning the tool name, and both flag tables are keyed by it: `--body`/`-b` + `--body-file`/`-F` for gh, `--description`/`-d` for glab.

`gh pr create -d '<link>'` now nudges — `-d` is `--draft`, a boolean, so nothing there is a body. `glab mr create -d '<link>'` stays silent, and `glab mr create --body '<link>'` nudges, since `--body` isn't a glab flag. The comment now states outright that gh's `-d` is `--draft` and what flattening the tables back together would break, so the next reader doesn't tidy them into one list.

glab keeps its deliberately empty body-file tuple. It takes its description inline, and an earlier round declined to add a file flag that couldn't be verified offline — inventing one is worse than omitting it.

**Minor — repeated body flags.** Also fixed rather than deferred, since it is the same false-negative direction. `_flag_values` became `_last_flag_value`: last occurrence wins, per family. Spellings inside a family are aliases, so `-b … --body …` resolves to the `--body` value rather than last-of-each-spelling.

One case your finding didn't ask about but the change forced a decision on: a segment carrying **both** a body-text and a body-file flag. I left it undisambiguated — each family independently contributes its last occurrence, and either carrying the link goes silent. Both CLIs reject that combination outright, so the command opens nothing and there is no body for the nudge to be right or wrong about. Adding precedence there would be inventing behaviour neither CLI has. That reasoning is recorded in `_body_has_link`'s docstring.

Verified end-to-end through `dispatch.py`, never `deny`: gh `-d` nudges, glab `-d` silent, glab given `--body` or `--body-file` nudges, and both orderings of a repeated `--body` resolve to the last one. Suite is 82 tests, green. Five mutations checked — flattening the tables, giving glab gh's file flags, emptying glab's text flags, discarding the tool entirely, and first-occurrence-wins — each reddening exactly the tests that name it.

cheapsteak:
Both fixed in 872ab3a1. Reproduced each through the dispatcher first — the continuation case was silent exactly as you described.

**Medium — line continuation.** The `escaped` branch now drops a `\`+newline pair instead of appending the newline as a word. One point covers both contexts where the shell treats it as a continuation: unquoted, and inside double quotes. Single quotes never set `escaped`, so their literalness needed no code — only a test to pin it, and that test is discriminating by construction: it splits `worktree=` as `workt\`+newline+`ree=` inside `'…'`, so collapsing it wrongly would fabricate a link and go silent.

You were right that this is a mainstream shape. It survived six rounds of my own testing because a continuation *after* the verb already worked, and I had only been writing single-line commands.

**On the heredoc interaction you'd expect to be affected:** the two passes turn out to be independent. `_heredoc_openers` already skips `\`+next-char, and bodies are stripped over physical lines before tokenizing, so a trailing continuation neither opens nor suppresses an opener. `gh pr \` + newline + `create --body "$(cat <<'EOF' … EOF )"` resolves correctly, pinned in both directions. The one straddling construction — a heredoc opener on a line that also ends in a continuation — I deliberately did not build for; it resolves fail-closed, and that is now stated in `_split_heredoc_bodies`' docstring rather than left for the next reader to discover.

**Minor — cross-segment OR.** Fixed rather than deferred, same false-negative direction as the rest. `_body_has_link` split into `_segment_body_has_link` (that segment's own last body value, in its own CLI's spelling) and `_every_body_has_link`, which hoists the invocation-wide heredoc check once and then `all(...)`s across segments. The heredoc leg stays invocation-wide — a body heredoc'd into a not-yet-existing file cannot be attributed to a segment — and the docstring now ends with the explicit formula.

**One stale claim your fix exposed,** flagging it since it is in prose rather than code: the module docstring said detection is a port of `PRBindingExtractor.isPRCreateCommand`. The Swift tokenizer keeps the backslash-newline pair, so that is no longer literally true. Rewritten to say the mirror is close but not exact, naming the continuation collapse as a widening-only divergence — it recognizes every invocation the Swift side does, plus continued ones. Worth knowing that **`PRBindingExtractor.swift` carries the same defect**: a `gh pr create` split across lines won't bind a PR to its worktree either. Out of scope here (it needs a Swift build), but it is a real bug on the side that carries consequences, and I would rather name it than let the divergence read as an oversight.

Suite is 92 tests, green. Four mutations checked, including restoring the original defect and collapsing inside single quotes, each reddening exactly the tests that name it. Verified end-to-end: continued invocation nudges and stays silent when linked, two chained creates nudge when either lacks a link, and both heredoc tests still pass.

cheapsteak:
Both fixed in a7fa2713. I reproduced each through the dispatcher first — and these are the first findings in this PR that run the *other* direction: a spurious nudge at a body that does carry the link, rather than a missed one.

**Medium — embedded `"` in a heredoc-composed body.** Your trace was exactly right, including that real bash would not close the span there because `$(…)` is an independent nested context. Both scanners now track `$(` nesting and start each substitution with fresh quote state, restoring the suspended quote at the matching `)`. `_heredoc_openers` therefore reaches the `<<` on `--body "$(cat <<'EOF'` and strips the body as prose; `_command_segments` no longer lets a literal `"` inside the substitution truncate the value.

Fixing either one alone would have covered your repro — they rescue it independently — but each is load-bearing on shapes the other misses, which the mutation runs pin separately.

**Where fidelity stops**, stated rather than left to be discovered, and all failing toward the nudge: backticks are not treated as substitution, an unbalanced `$(` swallows the rest of the command into one word, and `${…}` gets no handling. This is a tokenizer, not a shell, and the file's standing rule is that unusual constructions fail closed.

**Minor — head-only body-file read.** You were right that it is real, and I confirmed it with a 70 KB body: it nudged at a correctly-linked PR. The read now takes two bounded windows, head and tail, via `os.pread` so the head offset is undisturbed. The tail window is what matters here, since the convention places the link on the body's *last* line — reading only the head meant a long body read as unlinked precisely when it had followed the convention. Worst case is ~128 KB; the `O_NONBLOCK` + `fstat`/`S_ISREG` guard is byte-for-byte unchanged. A body over twice the cap with its link in the middle still nudges.

On the constant's rationale: I gave `_MAX_BODY_BYTES` a comment saying what it bounds and nothing more, with an explicit note that no claim about any forge's body-length limit is intended. Your parenthetical that 64 KB "happens to match GitHub's body character limit" is plausible but I did not verify it, so I did not write it down as fact.

**Two existing tests changed fixtures** (same names, same intent) — flagging it so the diff doesn't read as weakened coverage. `test_body_file_read_is_bounded` and `test_body_file_read_is_bounded_in_bytes_not_characters` both placed their link at the tail, which is now read *by design*. They now put it in the unread middle of a file over twice the cap, so they still test the bound, and the multibyte one still discriminates byte-counting from character-counting — verified by a mutation that triples the head window.

**One precision loss worth naming.** `gh pr create --body "$(cat <<'EOF' … link … EOF)" && gh pr create --body 'no link'` is now silent, because the heredoc body finally reaches the heredoc leg — which `_every_body_has_link` documents as deliberately invocation-wide. It is the specified behaviour and there is already a test asserting the same for the `--body-file` shape, but it is a real loss in the two-creates-in-one-call corner, and `test_a_substituted_heredoc_body_reaches_the_invocation_wide_leg` pins it.

Suite is 106 tests, green. Eleven mutations checked; one (tripling the tail window's length) **survived and is reported as inert by construction** rather than quietly dropped — the offset is `size − cap`, so the window already ends at EOF. Beyond the suite I ran a differential of old versus new over a 92-command corpus plus 13 multi-line shapes: exactly three verdicts flip, all `NUDGE → SILENT` on bodies carrying the link, zero regressions. I also re-verified one case from every previous round of this review against the new build; all 14 hold.

PR #699, opened 2026-08-21
Feat: Click a file path in the transcript to open it

cheapsteak:
## Summary

An agent finishes a task and writes "spec committed to `docs/foo/bar-design.md`". In a TBD terminal you cmd+click that path and it opens. In the transcript viewer — the surface built for reading what the agent said — the same text is inert. You leave the transcript, find the line again in the terminal pane, and click it there.

This makes paths and URLs in the transcript clickable. A plain click opens the file in the panel's viewer slot, the same place terminal cmd+click sends it. It works in ordinary prose, inside backticks, and inside fenced code blocks, which matters because backticks are where agents put paths most of the time.

A path links only if it resolves to a real file. A path to something that doesn't exist stays plain text rather than becoming a dead link.

## Why this shape

Spec: [`docs/specs/2026-08-21-transcript-link-clickability-design.md`](docs/specs/2026-08-21-transcript-link-clickability-design.md).

**Ranges are marked at render time, not resolved at click time.** Click-time resolution costs nothing until you click, and would have matched the terminal exactly — which offers no hover affordance at all. But the requirement is a *plain* click, and `NSTextView.mouseDown` runs AppKit's modal drag-select tracking loop, so click-versus-drag would have to be reconstructed by hand, risking text selection on any prose shaped like a path. Marking `.link` ranges gets plain-click-follows and drag-still-selects from AppKit for free.

**One resolver, two callers.** `resolveAsFilePath` was a method on `TBDTerminalView`, mixing pure path logic into an AppKit view and unreachable from the transcript. It moves to `ClickedPathResolver`, and the token-boundary character set moves with it — shared, not mirrored. Duplicating the tokenizer while sharing the resolver would leave the two surfaces free to disagree about where a path begins and ends, which is the same drift the shared resolver exists to prevent, one layer down.

**Links in code underline; links in prose tint.** Not a taste call. `CodeHighlightService` re-applies `.foregroundColor` over code ranges asynchronously after the row is on screen, so a tint inside a code block gets silently overwritten a moment after it appears. Underline survives, because highlighting only sets colors.

**History reveals in Finder rather than opening.** Transcript text is agent-authored, and `NSWorkspace.open` on a resolved path *executes* it when that path is a shell script — and transcripts are full of `scripts/*.sh`. Revealing is both the safe gesture and the useful one, given History has no pane layout to route into.

**`LocalFileLinker` is deleted rather than revived.** It already existed, rewriting bare absolute paths into `tbd-file:` links, and nothing in `Sources/` ever called it. It rewrites markdown *source* before parsing, so it structurally cannot reach inside a code span — rewriting `` `docs/a.md` `` into a link would stop it being code — and it has no worktree path, so it cannot resolve relative paths.

Rejected: routing a click so the transcript survives (excluding the originating pane from swap candidates and splitting off it) — it keeps your scroll position at the cost of a third pane appearing on a single click. The transcript pane gets swapped instead, and the toolbar toggle brings it back.

## What this PR does

- `ClickedPathResolver` — extracted from `TBDTerminalView`, verbatim. Owns the shared `pathTokenCharacters`.
- `TranscriptLinkScanner` — pure tokenizer, no filesystem. A path token ends at a colon not followed by a digit, so `Sources/A.swift:17:let x = 1` yields `Sources/A.swift:17` rather than swallowing the matched line: that is the shape ripgrep and compiler output paste into code blocks.
- `TranscriptLinkPass` — marks `.link` ranges over rendered prose, styling code sub-runs and prose sub-runs separately (a token can straddle the boundary).
- `.tbdCodeContext` — set by both `MarkdownCodeBlock.attributed` and `visitInlineCode`, so the pass can tell code from prose.
- The pass runs inside the existing composed-blocks cache, keyed by `(id, contentVersion)`, so its `stat()` calls happen once per row per content change rather than once per scroll. `TranscriptLinkResolverCache` memoizes per token, keyed after the line suffix is stripped so `foo.py:10` and `foo.py:42` are one entry.
- `TranscriptBubbleTextView` gains a click delegate; `TranscriptLinkDestination` holds the pure live-vs-History routing decision.
- `LocalFileLinker` and its tests deleted (436 lines).

## Assumptions

- **Resolved paths are always absolute.** `tbd-file:` URLs are hierarchical only under that assumption; `fileURL(forResolvedPath:)` enforces it with an early return rather than leaving it implied.
- **The worktree root can arrive after the pane does.** `TableTranscriptView.makeCoordinator` captures the context once and `updateNSView` never refreshes it, so anything baked into that context at first evaluation is frozen for the pane's life. The resolver therefore reads the root through a closure at each resolve rather than closing over a string, and the resolution cache drops its memo when the root changes — otherwise a pane mounted before the worktree-list RPC resolved would memoize a miss for every relative path and never recover.
- **Paths contain no spaces.** `pathTokenCharacters` has no whitespace, so `/Users/x/My Documents/a.md` never links. The terminal's widener has the same gap, so this is parity rather than a regression.

## Flag & rollout

No flag. This is additive UI behind a user gesture: no timer, no autonomous action, no state destruction, and it adds a pass to the render path rather than replacing one. No durable external resource is created, so the named-reconciler question does not arise.

## Evidence & verification

88 tests across 6 new suites; `Transcript|Markdown|Clicked` is green at 650 tests / 70 suites; full suite 7773 tests with no failure attributable to this branch.

Tests were built to fail against the old code, and two rounds of review caught cases where they didn't:

- The code-styling half was initially non-discriminating — the inline-code and fenced-block tests asserted only that a `.link` existed, which is true whether or not `.tbdCodeContext` is ever set. Both producer edits could have been skipped with a green suite, shipping code links that get tinted and then erased by the highlight pass. They now assert underline-and-not-tint through `renderBlocks`.
- The colon-boundary test used a spaced fixture (`docs/a.md: the spec`), which the trailing-punctuation trim already handled — it passed with the rule reverted. It now uses an interior colon.

Review verified by compiled snippet rather than inspection, which is how two real bugs surfaced: a URL followed by `!` parsed as host `example.com!` and opened the wrong place (`!` is a legal hostname character, so Foundation accepts it); and `linkTextAttributes = [:]` removed the pointing-hand cursor, because AppKit ships the cursor *in* that dictionary alongside the blue and the underline.

The review gate then caught a third, in the prose rather than the code: the assumption above originally claimed an empty worktree root was always permanent (remote or archived), while this file's own comment already named a transient cause — the row not having loaded yet. The resolver now reads the root live and the cache invalidates on a root change. The gate also found that a markdown-authored `file://` link bypassed the exists-and-not-a-directory rule that scanner-derived candidates get; it now goes through the same check, and "readable file" has one definition instead of two.

**Live verification is outstanding.** Nothing about rendering, hover, drag, or click behavior is machine-verified — the automated coverage stops at the pure units, and the three injection points are compile-checked only. Pending by eye, in the running app: link styling in prose versus code, plain click opening the file, drag-select still selecting, the pointing-hand cursor, History revealing in Finder, terminal cmd+click still working, and whether copying a selection spanning a link carries a dead `tbd-file:` URL into a rich-text paste.

## Follow-ups, not in this PR

- Cmd+click on an absolute path with a `:line:col` suffix does not resolve in the **terminal**. Confirmed not to be in the resolution rules — the widener strips `:` before the resolver runs — so it lives in the grid-to-token path.
- `OrphanProcessCollectorTests.parseLiveCWDs` fails on any real macOS box: it asserts an uncanonicalized `/tmp` path against a function that canonicalizes by design, and macOS `/tmp` is a symlink. Pre-existing; this branch touches no `Sources/TBDDaemon` files.

[🔗 Transcript Clickable Links](https://cheapsteak.github.io/tbd/open/?worktree=0540A635-64B0-4F59-A45B-EE1CD64548C6)


cheapsteak:
Both findings addressed.

**MEDIUM — resolver permanently binds an empty worktree root.** Correct, and the argument the old doc comment made for the snapshot being safe was exactly the one you refuted: it claimed an empty root meant remote-or-archived, both permanent, while `TableTranscriptPaneView.worktreePath`'s own comment already named the row not having loaded yet. `makeLinkResolver` now takes `worktreeRoot: @escaping @MainActor () -> String` and calls it per resolve, so a root arriving after `makeCoordinator` is picked up. `TranscriptLinkResolverCache` records the root its memo was built against and drops the memo when the root changes — the invalidation path you noted was missing. `memoRoot` is optional so "never asked" stays distinct from a genuine empty root. (`91c5b4cc`)

**MINOR — `file://` links skip the existence check.** Also correct. `TranscriptLinkTarget.init?(url:isReadableFile:)` now gates the `file://` branch on the same exists-and-not-a-directory rule, defaulting to a new `ClickedPathResolver.isReadableFile(_:)` that the existing convenience overload also calls, so there is one definition rather than two. A `file://` URL that fails returns nil, the delegate returns `false`, and AppKit's default handling takes it — the behaviour that predates this PR. The `tbd-file:` branch is untouched and `tbdFileURL_isNotRestatted` pins that already-resolved paths are not re-stat'ted on the click path. (`ee9925f1`)

Both fixes were confirmed red first by staging the signature change with the old semantics preserved, so the failures prove the behaviour is required rather than just the parameter. 82 → 88 tests in the link suites; neighbours 650 → 656; build and `swiftlint --strict` clean.

The PR description's Assumptions section carried the same wrong claim and has been rewritten, and the spec's "the click path does no filesystem work" line was corrected, since the `file://` check makes it false.

Still outstanding and called out in the description: live verification. Nothing about rendering, hover, drag, or click is machine-verified.

cheapsteak:
Addressed in `d9e3bd5f`. You were right, and the previous round fixed the wrong layer.

`makeLinkResolver` reading the root live only helps on a `composedCache` miss, and for a row already composed the resolver is simply never consulted again — so the fix was real but unreachable in the scenario it claimed to cover. Every mechanism claim in the finding checked out against the tree: the key is `ComposedKey(id:version:)`, the only invalidation sites were the width branch and `invalidateComposed(for:)` from `.updateLast`, and no root observer existed. (The `pruneCaches` pass is content-addressed, so it never drops a live row either.)

`TableTranscriptView` now takes `let linkRoot: String` — undefaulted, so both production call sites had to be updated rather than silently keeping the old behaviour. `makeNSView` seeds `coordinator.lastLinkRoot`; `update(..., linkRoot:)` compares against it (optional, so "never set" stays distinct from `""`), drops `composedCache` on a transition, and reloads. The reload is explicit on the `.noop` early-return as well as after the `.append`/`.updateLast` switch — without that, already-realized cells keep their stale prose even after the cache is dropped.

Height caches are deliberately left alone: `TranscriptLinkPass` adds `.foregroundColor`, `.underlineStyle` and `.cursor`, none of which moves a glyph, and the block count is unchanged — so `heightCache`, `estimateCache` and `blockHeightCache` still describe the rows they measured. Reasoned in a comment and in the spec.

On the MINOR — agreed, the old test proved the wrong thing. It is renamed `linkResolverClosureAlone_readsTheRootLiveWhenItArrivesLate` with a scope comment pointing at the new integrated test, `TableTranscriptHarness.lateWorktreeRootRelinksComposedRows`, which drives the real coordinator: primes and realizes rows at root `""`, asserts no `.link`, transitions to a temp worktree containing `docs/a.md`, calls `update(...)`, asserts the link appears. It was red first against the shipped signature:

```
Expectation failed: (linked?.path → nil) == (root.path + "/docs/a.md" → "/var/folders/.../docs/a.md")
  ↳ the row must recompose against the late root; got nil
```

`Transcript|Markdown|Clicked` 656 → 657; build and `swiftlint --strict` clean. Spec gained "The worktree root is an input the cache key does not carry" under Where the pass runs.

cheapsteak:
Addressed in `f11c4745`. Confirmed and fixed — and the exemption you found was mine: the earlier commit added the readability check for `file://` and deliberately exempted `tbd-file:` on the assumption that only the render pass could produce one. `visitLink` had no scheme allowlist, so that assumption never held.

Fixed at both layers, because they defend different things:

- **`visitLink` no longer mints the internal scheme.** A markdown link whose destination is `tbd-file:` gets no `.link` attribute at all — the label still renders as ordinary prose, so the message's words are never dropped, only its claim on the scheme. This stops the URL existing rather than catching it at the click.
- **`TranscriptLinkTarget.init?` applies `isReadableFile` to every path-bearing scheme.** Both branches fold into one: `resolvedPath(from:) ?? (isFileURL ? path : nil)`, then the guard. A click sees a URL, not its provenance, so the check belongs there rather than being assumed upstream. One `stat()` per click is negligible.

`tbdFileURL_isNotRestatted` is deleted. It asserted the predicate was never consulted for `tbd-file:` — it pinned the defect, and a passing test made the gap look deliberate.

New tests, each red first: a hand-authored `tbd-file:` to a nonexistent path and to a directory are not routed; the delegate returns `false` for the former; and through `renderBlocks`, markdown `[see the fix](tbd-file:/etc/passwd)` renders the text "see the fix" carrying no `.link`. An ordinary `[label](https://example.com)` is asserted unaffected.

One thing the finding under-stated: five pre-existing tests routed fabricated paths (`/w/docs/a.md`) through the default predicate and would have gone red on the boundary check alone. They now run against real files in the temp tree rather than being handed an always-true stub — weakening them would have hidden exactly this class of bug next time.

88 → 90 in the link suites; 657 → 661 for `Transcript|Markdown|Clicked`; build and `swiftlint --strict` clean.

PR #705, opened 2026-08-24
fix: surface permission prompts as attention instead of thinking dots

Jephuff:
## What's broken

A Claude session that stops on a tool-approval prompt (a permission ask, an MCP elicitation) keeps showing the animated thinking dots in the sidebar, with a regular-weight name. It looks like the agent is making progress when it is actually blocked waiting for a human. Anyone running a fleet hits this whenever a session raises a permission prompt: the one row that most needs a glance is indistinguishable from rows that are genuinely working.

## Why it happens

The signal already reaches the daemon. Claude Code's `Notification` hook fires with `notification_type: "permission_prompt"`, and `handleTerminalNotificationEvent` classifies it as prompt-on-screen and persists it to the terminal's awaiting-input columns. But that was the end of the line: nothing was broadcast, and no app surface read those columns (only the supervision resolver does). Meanwhile `activityState` still holds `working` from the last `UserPromptSubmit` hook, and the sidebar renders `working` as the thinking dots. Deliberately so on the activity side: `activityState` is a gating field (hibernation rails read it), and the Notification hook only reports that a prompt was raised, never that it went away, so writing `waiting_for_user` from it would latch sessions un-parkable.

Observed live before the fix: a terminal with `activityState: "working"` observed at 15:09 and an `awaitingInputReason` of `permission_prompt` observed at 15:20, rendering as thinking dots.

## What this PR does

When the classification is prompt-on-screen, the handler now additionally creates an `.attentionNeeded` notification row and broadcasts the existing `.notificationReceived` delta, the same pattern the `PreToolUse:AskUserQuestion` path already uses. Everything downstream already exists: `RowStatusIndicator.shouldBoldName` bolds the name for `.attentionNeeded`, the suffix slot ranks attention above working (so the amber hand replaces the dots), and focusing the worktree clears it through the existing `notifications.markRead` path.

Alternatives rejected:

- Writing `activityState = .waitingForUser` from the hook: it is a gating field for hibernation, and `.waitingForUser` is sticky by design, so a raise-only signal would latch sessions un-parkable.
- Plumbing `awaitingInputReason` into the sidebar: needs a new delta shape plus a precedence rule against `activityState`, a much larger change for the same rendered outcome.

`activityState` is untouched, so hibernation gating and interrupt behavior are unchanged.

## Evidence & verification

- TDD red run: the new assertions in `NotificationHookRPCTests` failed 11 ways before the handler change (no notification row, no delta), then passed after it.
- The parametrized classification-branch test now asserts, per branch: the three prompt-on-screen types each create exactly one unread `.attentionNeeded` row and one `.notificationReceived` delta; every other classification (including informational `agent_completed`) creates zero rows and zero deltas; and the pre-existing assertions still hold for all branches (reason recorded verbatim, activity triple unchanged, hibernation decision unchanged).
- New tests: the broadcast delta carries the created row's id, worktree, type, and message; an empty hook message falls back to "Claude needs your input".
- `scripts/test.sh --filter 'NotificationHookRPCTests'` (16 tests), `--filter 'askUserQuestion'` (7 tests), and `--filter 'Notification'` (113 tests, 29 suites) all pass; `scripts/swift-safe build` clean; `swiftlint --strict` clean on touched files.
- Live verification is pending: exercising it end to end requires restarting the daemon, which would disrupt the running fleet, so it was not done from this session.

[🛎 Fix Tool Approval State](https://cheapsteak.github.io/tbd/open/?worktree=1AB96221-AABD-435B-AE8F-86754E6D749B)

### DIFF
--- a/Sources/TBDDaemon/Server/RPCRouter+TerminalHandlers.swift
+++ b/Sources/TBDDaemon/Server/RPCRouter+TerminalHandlers.swift
@@ -3078,10 +3078,13 @@ extension RPCRouter {
     /// policy decision TBD could not depend on, while a fork here is compiled
     /// and testable.
     ///
-    /// The handler records a fact and changes nothing else. There is no
-    /// broadcast: no delta shape carries a wait reason, and inventing one to
-    /// carry a fact no surface renders yet would ship app behavior this slice
-    /// deliberately does not have.
+    /// The handler records the reason columns, and for a prompt-on-screen
+    /// classification additionally raises an `.attentionNeeded` notification
+    /// so the sidebar swaps the thinking dots for the attention indicator and
+    /// bolds the name (the app's `RowStatusIndicator` already ranks attention
+    /// above working). There is still no delta carrying the wait reason
+    /// itself: the notification row is the fact the app renders, and focusing
+    /// the worktree clears it through `notifications.markRead`.
     func handleTerminalNotificationEvent(_ paramsData: Data) async throws -> RPCResponse {
         let params = try decoder.decode(TerminalNotificationEventParams.self, from: paramsData)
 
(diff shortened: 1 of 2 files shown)


cheapsteak (approved):
🙏

PR #706, opened 2026-08-24
Feat: Move the project disclosure chevron to the right of the name

cheapsteak:
## Summary

In the sidebar, each project (repo) section header carried its expand/collapse
chevron on the far left, ahead of the project name. That put a piece of chrome
in the position the eye scans first, and it indented every project name 18pt
further right than the Scratch and remote-provider headers directly above and
below it — three section headers, three different left edges.

This moves the chevron to sit immediately after the project name, so the name
claims the leading edge and the control follows what it acts on. Project names
now start flush with every other sidebar section header. The chevron also
becomes a hover affordance rather than persistent chrome, appearing and
vanishing in step with the `+` already on that row.

## Why this shape

The chevron trails the name directly rather than being pushed to the row's
trailing edge: it is a control *on the name*, and parking it against the far
right would divorce it from what it collapses and crowd it against the `+`.

**The hover gate is an interaction-model change, and was chosen on purpose.**
Before this PR the chevron was unconditional — always rendered, always
clickable, its glyph a persistent at-a-glance read of whether the section was
expanded. It is now absent from the view tree (`Color.clear`) unless the
pointer is inside the project header or one of its worktree rows. The trade-off
is real and goes both ways:

- **Cost.** You can no longer tell a collapsed project from an expanded one
  without moving the pointer there, and the control cannot be reached by
  pointer-free navigation while hidden.
- **Benefit.** The sidebar is quieter at rest, and the two hover affordances on
  the row now behave identically instead of one being chrome and one being
  hover-revealed.
- **Cost, precisely.** `isSectionHovered` is driven only by `.onHover`, so the
  button is absent from the focus tree — not merely invisible — until a
  pointer enters the section. Tab and Full Keyboard Access skip it, and the
  `.accessibilityLabel` added here only serves a VoiceOver user who has
  already hovered with a pointer. This is a real reduction in pointer-free
  reachability of this specific control.
- **What backstops it.** Collapse/Expand is also a header context-menu item
  (`RepoSectionView.swift:475`), which is not gated on hover — so the
  capability itself stays reachable without a pointer; what is hover-gated is
  its dedicated single-press control.

**Maintainer sign-off.** The repo owner asked for the hover gate in these terms
("hide the collapse chevron along with the + button if the user isn't hovered
into the project or its worktrees"), and, when the keyboard-reachability cost
above was put to them explicitly alongside the alternative of always-mounting
the chevron in a low-emphasis idle style, chose to keep the hover gate with the
context menu as the accepted keyboard-only path. That decision is recorded at
the call site in `chevronButton`'s doc comment so it isn't silently reverted.

This mirrors the choice the `+` on this same row already made, so the header no
longer mixes two conventions. A human weighed the discoverability cost against
the quieter sidebar and asked for the hover gate; the rationale is recorded in
the `chevronButton` doc comment so a future reader doesn't "fix" it back.

No spec: this changes the visibility rule of one existing control on one row
and adds no new capability, state, flag, or persisted data.

## What this PR does

All in `Sources/TBDApp/Sidebar/RepoSectionView.swift`:

- `headerHStack` orders `nameLabel` before `chevronButton`. The name's old
  `.padding(.leading, -2)` — which existed to close the gap to a *leading*
  chevron — is dropped, aligning project names with the Scratch and remote
  headers.
- `chevronButton` gains `HoverPressButtonStyle` (the `+`'s hover/press wash)
  and the same visibility predicate the `+` uses,
  `HoverMenuModel.shouldShowPlus(hovered:menuOpen:)`. Every worktree and
  remote-session row in the section already reports into
  `onSectionHoverChange`, so hovering a child keeps both controls up.
- The glyph goes from 11pt to 10pt so it doesn't visually outweigh the 12pt
  semibold name it now trails, with `-3` leading padding to trim the hit
  target's slack and a layout-neutral `.offset(y: 2)` to drop it onto the
  name's optical baseline.
- New `.accessibilityLabel` ("Collapse/Expand <project>") — the glyph is
  text-free, so a hidden-until-hover button would otherwise announce nothing.
- The hidden branch renders a `Color.clear` at the same 18x18, so the
  `[missing]` badge that follows the chevron doesn't slide sideways as the
  chevron comes and goes.
- The hover debounce in `onSectionHoverChange` now also clears
  `isChevronHovered`.

## Assumptions

The chevron's `foregroundStyle` is applied to the `Image`, not the `Button`,
because `HoverPressButtonStyle` sets a blanket `.secondary` on its label and
the innermost style wins — this is what keeps a `missing` repo's chevron
rendering dimmed. If that style stops applying `foregroundStyle`, the missing
state still renders correctly; if the chevron's own modifier is ever hoisted
out to the button, it stops.

## Evidence & verification

Verified live in the running app across the iterations that produced the final
values (`scripts/restart.sh --app` after each): the chevron reads as attached
to the name, its hover wash matches the `+`, and both controls appear and
disappear together whether the pointer is over the header or over one of the
project's worktree rows.

One behavior needed a guard rather than an observation. The chevron is torn
down by the same hover gate that would deliver its `onHover(false)`, so that
event can be lost — which would strand `isChevronHovered` at `true` and leave
every worktree row in the section dimmed to 0.7 opacity permanently. The
debounce that clears `isSectionHovered` now clears `isChevronHovered` alongside
it.

`scripts/swift-safe build` and `swiftlint --strict` are clean. No daemon or
shared code changed, so `scripts/test.sh` was not run.

[▸ Chevron Right Of Project](https://cheapsteak.github.io/tbd/open/?worktree=74B2B340-FF9C-4938-85CF-90ACB8BC1422)

### DIFF
--- a/Sources/TBDApp/Sidebar/RepoSectionView.swift
+++ b/Sources/TBDApp/Sidebar/RepoSectionView.swift
@@ -195,6 +195,12 @@ struct RepoSectionView: View {
     /// `newWorktreePlusButton`) so the two hover affordances appear and vanish
     /// together. The hidden branch keeps the 18pt square so the "missing" badge
     /// doesn't slide sideways as the chevron comes and goes.
+    ///
+    /// Hover-gating a disclosure control is deliberate, not incidental: it
+    /// trades the at-a-glance expanded/collapsed read for a quieter sidebar,
+    /// and it is the choice the `+` on this same row already made. The
+    /// always-available path to the same action is the header's context-menu
+    /// Collapse/Expand item, which is not gated on hover.
     @ViewBuilder
     private var chevronButton: some View {
         Group {
@@ -212,6 +218,9 @@ struct RepoSectionView: View {
                         .contentShape(Rectangle())
                 }
                 .buttonStyle(HoverPressButtonStyle())
+                // The glyph carries no text, so VoiceOver has nothing to
+                // announce without this — parity with the `+`'s label.
+                .accessibilityLabel(repo.expanded ? "Collapse \(repo.displayName)" : "Expand \(repo.displayName)")
                 .onHover { isChevronHovered = $0 }
                 .help(repo.expanded ? "Collapse" : "Expand")
             } else {
--- a/Sources/TBDApp/Sidebar/RepoSectionView.swift
+++ b/Sources/TBDApp/Sidebar/RepoSectionView.swift
@@ -55,6 +55,10 @@ struct RepoSectionView: View {
                 // swiftlint:disable:next no_raw_task_sleep - legacy sleep, see docs/specs/2026-07-24-test-hardening-design.md
                 try await Task.sleep(nanoseconds: 80_000_000)
                 isSectionHovered = false
+                // The chevron is torn down by this same gate, so its
+                // `onHover(false)` may never arrive — without this the
+                // worktree rows would stay dimmed at 0.7 forever.
+                isChevronHovered = false
             }
         }
     }
@@ -130,7 +134,7 @@ struct RepoSectionView: View {
         }
     }
 
-    /// The section header row (chevron + name + `+`) and every modifier
+    /// The section header row (name + chevron + `+`) and every modifier
     /// attached to it — extracted out of `body` alongside `expandedContent`,
     /// and further split into `headerHStack` + its own sub-pieces below, so
     /// the type checker sees several smaller expressions instead of one
@@ -171,14 +175,14 @@ struct RepoSectionView: View {
         .tag(repo.id)
     }
 
-    /// The header row's actual content (chevron + name + optional "missing"
+    /// The header row's actual content (name + chevron + optional "missing"
     /// badge + `+`), with none of `headerRow`'s trailing modifiers — see
     /// `headerRow`'s doc comment.
     @ViewBuilder
     private var headerHStack: some View {
         HStack(spacing: 4) {
-            chevronButton
             nameLabel
+            chevronButton
             if repo.status == .missing {
                 missingBadge
             }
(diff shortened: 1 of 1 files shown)


cheapsteak:
On the MEDIUM (chevron becomes hover-only) — the facts are right, and the PR description was wrong to call this purely visual. Fixed both halves:

**The behavior stays.** The hover gate is what the repo owner asked for, explicitly and in those terms ("hide the collapse chevron along with the + button if the user isn't hovered into the project or its worktrees"). So a human did weigh the discoverability trade-off; the description just failed to record that they had. It now states the cost, the benefit, and the backstop rather than claiming there was no interaction change, and the same rationale is in the `chevronButton` doc comment so a future reader doesn't quietly revert it.

Two things worth adding to the ledger:

- The always-available path is the header context menu's Collapse/Expand item (`RepoSectionView.swift:475`), which is not hover-gated. The capability was never lost — its dedicated control is what became conditional.
- The `+` on this same row already used exactly this predicate and the same `Color.clear` hidden branch. So this change makes the header consistent rather than introducing a novel pattern; if hover-gating a sidebar control is wrong here, it is equally wrong for the `+`, and that is a separate and larger conversation than this diff.

**The a11y half was a genuine miss and is fixed** in c46afd4b: the chevron now carries `.accessibilityLabel("Collapse/Expand <project>")`. You're right that it had none — the glyph is text-free, so it previously announced nothing to VoiceOver even back when it was always visible. That's strictly better than the pre-PR state.

cheapsteak:
Re: the narrowed MEDIUM (chevron missing from the keyboard/FKA focus tree) — the correction is accepted. You're right that `isSectionHovered` is driven purely by `.onHover`, so the `.accessibilityLabel` I added in c46afd4 only helps a VoiceOver user who has already moved a pointer onto the row. It does not restore what unconditional mounting gave keyboard-only users, and framing that commit as fixing VoiceOver support oversold it.

You asked for a maintainer's conscious sign-off between the two paths. It was put to the repo owner explicitly — keep the hover gate and accept the context-menu fallback for keyboard-only users, or always-mount the chevron in a low-emphasis idle style — and they chose to **keep the hover gate**, with the ungated header context-menu Collapse/Expand as the accepted keyboard-only path.

That sign-off is now recorded in two places rather than asserted in a comment thread:

- The PR description states the cost precisely (absent from the focus tree, not merely invisible; Tab and FKA skip it) and names the decision and who made it.
- `chevronButton`'s doc comment (03344997) says the same at the call site, and says explicitly not to "fix" it by always-mounting — so a future reader doesn't quietly reverse a deliberate call.

The `.accessibilityLabel` stays regardless: it serves VoiceOver whenever the button is mounted, which it never did before this PR, so it's strictly additive even though it doesn't answer the reachability point on its own.

Jephuff:
I hate this 😂 You ok if I make this a setting? :P

cheapsteak:
lol yup for sure; could you make the current way the default though? 🙏

PR #708, opened 2026-08-24
Add a setting for automatic Notes tabs

Jephuff:
If people like the notes tab opening by default, I can make this into a flag?

## Summary

TBD currently opens an empty Notes tab for every ordinary repo-backed worktree. This PR adds a global setting under Settings → Worktrees so people who do not use that tab can turn automatic creation off.

The setting defaults on to preserve the intentional existing behavior. Conversations revived on a fresh branch still receive their populated provenance note even when automatic empty notes are off.

## Why this shape

The preference is global because worktrees can be created through several paths, while the daemon is the single place that creates their initial tabs. Creation snapshots the resolved config once, so inline and pre-session flows make the same decision.

This makes an existing behavior configurable rather than introducing a new autonomous behavior, so it does not need a rollout spec. The fresh-branch revival exception remains unconditional because that note contains source context rather than being an empty convenience tab.

## What this PR does

- Adds a nullable `auto_create_notes_enabled` config column with no SQL default.
- Resolves an unset value through the shipped Swift default of `true`, while preserving explicit on and off choices.
- Adds a config RPC and reuses the existing `modelProfilesChanged` refresh path to keep app state synchronized.
- Adds “Create a Notes tab for new worktrees” under Settings → Worktrees.
- Gates empty Notes creation for ordinary inline and pre-session worktrees without changing primary focus or tab ordering.
- Keeps populated provenance notes for conversations revived on a fresh branch.
- Updates the shipped timestamp-migration manifest and adds branch coverage across persistence, RPC, lifecycle, carryover, and UI state.

## Flag & rollout

This is a permanent user preference, not a temporary soak flag.

- **Preference** — `autoCreateNotesEnabled`; persisted as `auto_create_notes_enabled`.
- **Default** — on, matching the behavior before this PR.
- **Opt out** — turn off “Create a Notes tab for new worktrees” in Settings → Worktrees.
- **Stored states** — NULL means untouched and follows the shipped default; explicit false and true remain stable if that default changes later.
- **Graduation** — none; the preference remains available.

## Evidence & verification

- Test-first failures covered the missing persistence contract, RPC, lifecycle branches, app-state wiring, and migration manifest before their implementations were added.
- `scripts/test.sh --filter 'AutoCreateNotesSchemaTests|AutoCreateNotesRPCTests|NoteTabOnCreateTests|WorktreeConversationCarryoverTests|PreSessionHookTests|ModelProfileRPCTests|AutoCreateNotesSettingTests|SQLMigrationLoaderTests|SchemaBaselineDriftTests'` — 150 tests passed in 10 suites.
- `scripts/swift-safe build` — passed.
- `scripts/lint-migrations.py` — passed, 1 migration and 7 checks.
- `git diff --check origin/main...HEAD` — passed.
- `scripts/test.sh` ran 7,862 tests. All changed and affected suites passed; the command remains nonzero from 10 issues in unchanged suites: ScratchpadCollector (3), Codex usage fetcher lifecycle (1), replay live integration (1), transcript row-height estimator (1), and Markdown web-view navigation (4). The run also reported 85 existing known issues.
- Live UI verification was not performed in this worktree.

[⚙️ Add Notes Tab Setting](https://cheapsteak.github.io/tbd/open/?worktree=ABF8B4FA-91BA-4093-A085-A41412A8D2B1)

### DIFF
--- a/Sources/TBDApp/AppState+ModelProfiles.swift
+++ b/Sources/TBDApp/AppState+ModelProfiles.swift
@@ -16,7 +16,7 @@ extension AppState {
     /// Refresh the full model profile list and global default ID from the daemon.
     func loadModelProfiles() async {
         do {
-            let result = try await daemonClient.listModelProfiles()
+            let result = try await modelProfilesFetcher()
             if result.profiles != modelProfiles {
                 modelProfiles = result.profiles
             }
@@ -37,6 +37,9 @@ extension AppState {
             }
             if result.gcEnabled != gcEnabled {
                 gcEnabled = result.gcEnabled
+            }
+            if result.autoCreateNotesEnabled != autoCreateNotesEnabled {
+                autoCreateNotesEnabled = result.autoCreateNotesEnabled
             }
             if result.nightwatchMode != nightwatchMode {
                 nightwatchMode = result.nightwatchMode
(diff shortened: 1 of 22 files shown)


cheapsteak (commented):
please do make this a settings toggle 🙏  I've been relying on the default open notes (creating a new note tab is two clicks, closing it is one 😛 )

cheapsteak:
please disregard the spec request; that prompt needs some tuning :s

Jephuff:
I was just digging and seeing this was intentional :D 
I guess I don’t take enough notes 😂 
I’ll make a settings toggle! 👍

cheapsteak (approved):

PR #709, opened 2026-08-24
fix(app): dismiss the sidebar hover menu when TBD stops being active

Jephuff:
## What's broken

Hovering the `+` next to a repo header or a worktree row in the sidebar opens the new-worktree profile picker. If you switch to another application while that menu is open, the menu keeps painting on top of the app you switched to.

Two flavors, depending on where the pointer happened to be:

- **Pointer on the `+`**: the menu lingers over the other app for roughly a second, then vanishes.
- **Pointer over the menu itself**: it never goes away. It floats over every other application until you switch back to TBD and scrub the pointer across the `+` again.

Anyone who Cmd-Tabs away with the picker open hits this, which in practice means anyone who opens the picker and gets distracted.

## Why it happens

Two independent causes, both confirmed.

**The window has nothing that orders it out.** The picker is not a popover; it is a bare borderless panel presented at the `.popUpMenu` window level. Window levels are windowserver-global rather than per-application, so any level above zero floats above *every other app's* normal windows, not just TBD's. Three things could normally pull such a window down when the app stops being active, and none of them are in play: the panel is not a child window (that link was deliberately severed in #112 to avoid a constraint-budget crash, and should stay severed), nothing observes app deactivation, and `hidesOnDeactivate` is false. That last one is the surprise. AppKit defaults it to true for panels, but only for *titled* ones; a borderless, non-activating panel opts out of the default, because that style mask exists for cross-app tool palettes that are supposed to stay up.

**The presenting model's state goes stale, which is the version that sticks.** `HoverMenuModel` decides whether the menu is open from two flags, both fed exclusively by SwiftUI's `.onHover`. Hover is an active-app tracking-area concept, and Cmd-Tab moves no pointer, so no exit event is delivered and both flags keep whatever value they had.

When the pointer was on the `+`, AppKit usually synthesizes an exit as tracking areas deactivate, which is why that case eventually closes. The delay before it visibly does is the anchor tearing the panel down on the next SwiftUI render pass, which a backgrounded app runs lazily.

When the pointer was over the menu, no exit is synthesized at all: the "pointer is inside" flag stays true forever, the close is never even scheduled, and the menu stays open. That also keeps the `+` itself force-mounted, since it stays visible while its menu is open.

## What this PR does

Gives the faux menu the dismissal contract a real menu has, at both layers.

`HoverMenuModel` now closes on app deactivation, reusing its existing close path so the hover flags are cleared and both pending transitions are cancelled. This is the half that actually fixes the sticky case. Hiding the window alone would not have been enough: the model would still believe the menu was open, so reactivating the app would pop it straight back up and leave the `+` pinned visible in the meantime. The notification center is an injected initializer parameter, defaulted, so tests can post to a private center and no call site changes.

`FloatingPanel` sets `hidesOnDeactivate = true` as a backstop, so a panel that somehow retains stale state still cannot paint over another application, and `collectionBehavior = [.transient, .ignoresCycle]` so a menu-shaped panel is not captured by Mission Control or stranded on another Space.

Alternatives considered and rejected: lowering the window level does nothing, since any level above zero still floats over other apps. Re-parenting the panel as a child window would fix ordering but reintroduces the crash #112 removed. Moving back to `NSPopover`, whose transient behavior handles all of this for free, would undo the arrow-less, animation-less presentation #408 deliberately built.

Both other `FloatingPanel` users were checked and neither needed different handling. The emoji picker overlay is only reachable while TBD is active, so hiding on deactivation is correct for it. The Cmd-K jump menu already auto-closes on losing key focus, which fires on deactivation anyway, so the new flag is redundant-but-consistent there rather than a behavior change.

## Assumptions

Assumes the picker panel is only ever presented while TBD is the active application. If some future path opens it from a global hotkey while TBD is in the background, `hidesOnDeactivate` would hide it immediately on open and the feature would appear broken.

## Evidence & verification

The repro is the report above: open the picker by hovering a sidebar `+`, Cmd-Tab away, and watch it paint over the other app, permanently if the pointer was resting over the menu.

The discriminating test is `closes when the app stops being active with the pointer parked over the menu`, which reproduces exactly that case at the model layer: open the menu, mark the pointer as inside the menu, post the deactivation notification, and assert it closed. It fails without the model change, because the stale "inside" flag means no close is ever scheduled. It also asserts the hover flags were cleared and that a following hover-exit does not resurrect the menu, since clearing state is the part that keeps reactivation from re-showing it. A companion test posts to a *different* notification center and asserts the menu stays open, which proves the injected seam is real rather than the model quietly observing the default center.

Full app test target: 2499 tests in 261 suites pass (14 pre-existing known-issue records). `swiftlint --strict` reports 0 violations across 809 files.

**Verified live**, before and after, on the same machine with the same gesture: hover a sidebar `+`, rest the pointer inside the menu, switch to another app for about ten seconds, switch back.

The measurement is not a screenshot but a window-server sample taken twice a second, recording which app is frontmost and which window levels TBD has on screen. The main window is level 0 and the picker panel is level 101, so "another app is frontmost while TBD still owns a level-101 window" is the bug stated as data.

Before, on a build without the fix, every switch held the panel up for the whole visit to the other app, and it closed only after returning to TBD:

```
[ 45.2s] front=Slack  tbdWindowLevels=[0, 101]
[ 54.7s] front=TBD    tbdWindowLevels=[0, 101]
[ 55.8s] front=TBD    tbdWindowLevels=[0]
```

After, across three switch cycles, no sample has another app frontmost with 101 present:

```
[ 54.7s] front=TBD    tbdWindowLevels=[0, 101]
[ 57.2s] front=Slack  tbdWindowLevels=[0]
```

In two of the three cycles the panel disappears while TBD is still frontmost, a beat before the other app comes forward. That is the deactivation notification firing as TBD resigns active, which is the seam this PR hooks, rather than anything incidental about window ordering.

[🎈 Fix Plus Popover Stuck](https://cheapsteak.github.io/tbd/open/?worktree=41E915B4-9DD6-4319-AE2A-9A17298E65BD)


cheapsteak (approved):
🙏

PR #723, opened 2026-08-26
Fix: Nine CI test failures on a diff of two Python files -- the waits were lying, not losing a scheduler race

cheapsteak:
## What's broken

A test can report `still not true after 30.0 seconds` about a condition that
was true the entire time.

The instance that prompted this: PR #716, attempts 1 **and** 2, nine failures
in `EventDrivenTestClockSelfTests` on a commit whose entire diff is two Python
files. Not a subset of that file's waits — **all nine** of them, both attempts.
Its base passed, and the identical content rebased onto a different base
passed. The reflex reading is CPU starvation under the fast parallel pass, and
the reflex remedy is a rerun.

That reading is wrong here, and this is where a rerun costs a day.

## Why it happens

Every bounded wait in this repo is written the same way, and the order of the
two checks is the bug:

```swift
while ContinuousClock.now < deadline {   // deadline first
    if condition() { return }            // condition second
    try? await Task.sleep(for: .milliseconds(5))
}
Issue.record(Timeout(...))               // verdict taken from the loop's exit
```

`Task.sleep` is a floor, not a ceiling. When the poller's resumption lands
after the deadline, the loop exits **without evaluating the condition again** —
so the verdict it prints is a sample that may be a full `timeout` old, taken at
the top, before the task it was waiting on had run at all. The helper stopped
looking and then reported what it had not seen.

Under the fast parallel pass that is the *expected* path, not an exotic one.
Swift Testing starts every non-serialized test in one process with no
concurrency cap; mined xUnit from CI puts **p50 per-test latency at 56-70 s**
against a 123-158 s pass, with 3,799 of 4,260 tests over 10 s. A 5 ms step
aside routinely returns its turn tens of seconds later, and a 30 s real-time
handshake budget sits *below* the median per-test latency of the process it
runs in.

**Evidence that this is a lie and not starvation**, since "it's load" is the
comfortable answer and has been wrong here five times:

- **The green comparison run was slower.** On `36c89528` the same pass ran
  p50 70.4 s / wall 158.3 s with the clock self-tests taking 82.7-99.4 s, and
  passed. The red attempts ran p50 60.4 / 56.3 s, wall 129.3 / 127.7 s, self-
  tests 71.9-85.7 s, and failed. Load is not the discriminator: the same tests
  took *longer* in the run that went green.
- **Every downstream assertion in the failing tests passed.** Each of the nine
  recorded exactly one issue, the wait. `advanceFiresInDeadlineOrder` went on
  to assert `sleeperCount == 2` after its first advance and pulled `"a"`,
  `"b"`, `"c"` out of the recorder in order — impossible unless all three
  sleepers really were registered. `requireArmedReturnsOnRegistration` and
  `armedReturnsImmediatelyWhenSleeperExists` each then ran a 50 ms hang guard
  that throws or records on a miss; neither fired. The handshakes had
  completed. Only the waits claimed otherwise.

That contrast is the general triage rule, and it separates this from genuine
starvation cleanly. Compare `AppearanceDebounceTests` in a loaded local run:
the wait fails *and* `fired.values → []` fails with it. **When a handshake
genuinely misses, the assertions after it fail too. When only the wait fails
and the rest of the test passes, the wait is lying.**

## When it broke

Not a regression — a latent flaw that started biting as the population grew.
`waitUntil` was introduced with this shape in `b9046244` (#646). The repo had
already learned the lesson elsewhere and applied it inconsistently:
`advanceUntil`, `watchForSleeper` and `SidecarTestSupport.waitUntil` all
re-read after their loop; six other helpers did not. It went unnoticed because
the defect only surfaces when a poller's resumption crosses its own deadline,
which needs a process population large enough to make that routine — 3,013 to
4,536 tests in three weeks did it.

## What this PR does

Test-only. No `Sources/` changes, so nothing about the product moves.

Since this was opened, `TestDeadlines.saturatedPass` (90 s) and
`TestGate.deadline` (120 s) landed on main and were applied to four of the six
helpers below — the *raise the number* remedy. None of the six gained a fresh
read, so the defect is unchanged on main today. That is the distinction this PR
turns on: a bigger budget cannot repair a verdict that is not derived from an
observation, it only moves the threshold at which the same lie gets told. This
change keeps main's sizing everywhere and fixes the derivation.

- **The fresh read.** Six bounded waits now take their verdict from a probe
  *after* the deadline test, matching the three helpers that already did:
  `EventDrivenTestClockSelfTests.waitUntil`, `ClockTestSupport.waitForSuspension`
  (149 downstream call sites via `advanceWhenSuspended`),
  `DeliveryVerifierTests.awaitDeliveryCycle` (13 sites),
  `RepoRemoveCascadeTests.waitUntil` and `RPCRouterRepoTests.waitUntilRemoved`.
- **The cancellation spin, in every helper that can suffer it.** `try?` around
  the poll sleep cannot tell expiry from cancellation, and a cancelled
  `Task.sleep` throws instantly — so the loop stopped suspending and busy-spun
  its whole budget on a cooperative thread, in a process where thousands of
  tests are queued behind it, then blamed the call site for a harness
  cancellation. Measured before the fix: **33,754,162 condition samples in
  30 s**. All four `try?` loops now break on cancellation and return without a
  diagnostic, matching what `EventDrivenTestClock.sleeperArmed` documents. The
  two *throwing* helpers propagate it instead, as a `CancellationError`: they
  have somewhere to send it, and returning would walk a cancelled test into the
  assertions that follow and pin the failure on those.
- **A diagnostic that discriminates.** `HandshakeTimeout` reports elapsed
  alongside the budget. "Still not true after 30.0 s" reads identically whether
  the wait polled steadily or got one turn and came back 45 s later, and only
  the second is a scheduling gap. This is the field that would have made the
  original triage a first-attempt job.
- **A gate that lied.** `RepoRemoveCascadeTests`' `PathGate.wait` self-released
  in silence, handing the test a cascade it believed was still held, with the
  failure landing on whatever it asserted next. It now reports through the
  shared `TestGateTimeout`, keeping main's `TestGate.deadline` sizing.
- **One loop, not six.** The fresh read and the cancellation guard were applied
  by hand at six sites, which is six chances to reintroduce them. The loop now
  lives once in `Tests/TestSupport/BoundedPoll.swift` as `pollUntilTrue`, and the
  helpers keep only their own diagnostic. `PollOutcome` is three cases rather
  than a `Bool` on purpose: callers must treat `.satisfied` and `.cancelled`
  alike and report only on `.timedOut`, and collapsing that to a `Bool` is what
  let the cancellation half be forgotten at five of the six sites.
- **`Tests/CLAUDE.md` gains assertion-hygiene rule 5** and
  `docs/specs/2026-07-24-test-hardening-design.md` §6 gains its spec entry, so
  the rule lives somewhere other than comments on the sites that already had it
  right, along with the lying-vs-starved triage signature above.

Why this shape rather than raising the deadlines: a bigger number cannot fix a
verdict that is not derived from an observation. It only moves the threshold at
which the same lie gets told.

## Assumptions

Assumes the condition closures are cheap and side-effect-free, since they are
evaluated once more on the failure path. Checked at every call site: most read a
lock-guarded counter or flag, and the two in `RPCRouterRepoTests` /
`RepoRemoveCascadeTests` are `db.repos.get(id:)` reads against an in-memory test
database — an extra read there costs nothing and changes nothing. If a future
caller passes a condition with side effects or real I/O, that assumption stops
holding and the double read becomes visible.

## Evidence & verification

**Fails without the fix, and both fixes were demonstrated separately.**

- `waitUntilRereadsTheConditionBeforeGivingUp` passes a condition that literally
  `return true` with a zero budget. Against the unfixed helper it reports
  `still not true after 0.0 seconds -- observed: a condition that is already
  true`, the same diagnostic shape seen in CI. Fixed: green.
- `waitUntilStillReportsAConditionThatNeverHolds` is the mutation check — a
  re-read that always returned happily would satisfy the first test and delete
  every hang guard in the file. Asserts the diagnostic still fires and still
  carries the caller's wording.
- `waitUntilIsSilentAndPromptOnCancellation` asserts on **sample count**, not
  elapsed time, because a "returns promptly" assertion would itself be the kind
  of wall-clock deadline this file exists to stop trusting. With the guard
  removed: 33,754,162 samples and a mis-attributed diagnostic. With it: 2.
- `BoundedPollSelfTests` pins the shared primitive, and is mutation-checked
  twice. Re-collapsing `pollUntilTrue` to its pre-fix shape turns three of the
  five red, including 31,093,043 evaluations in 30 s from the restored
  busy-spin, while `neverTrueTimesOut` stays green as the guard against a poll
  that reports success unconditionally. Swapping the condition read and the
  cancellation check turns `conditionBeatsCancellation` red on its own.

Full suite green apart from the load-flake families already known on this box
(`TerminalTeardownReapTests`, `AppearanceDebounceTests`,
`ArchivedWorktreeSearchTests`, `TerminalActivityEventHandlerTests`) — all 68
pass in isolation in 1.7 s, none use a helper this PR touches. `swiftlint
--strict` clean.

## Out of scope, and why

Two of the three reported flake instances are **not** this bug, and I am not
folding them in:

- `TmuxControlSupervisorTeardownTests` (five failures on main, run
  `32874930748`) uses `SidecarTestSupport.waitUntil`, which already re-reads.
  Its conditions really were false after 90 s of wall clock, in the
  highest-load pass measured (p50 69.4 s, wall 164.7 s). That is genuine
  starvation.
- The local `TerminalTeardownReapTests` / `CodexUsageFetcherLifecycleTests` /
  `ReplayLiveIntegrationMatrixTests` reds are `.timeLimit` expiry on real
  subprocess work, and collapse to sub-second in isolation.

Both are the same underlying question — whether a fixed wall-clock budget can
survive an uncapped in-process pass — and per `CLAUDE.md` that is a design
change needing `/tbd-brainstorming` with a human, not an agent's own answer.
Flagged, not attempted.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=1CA7F26F-9813-411B-8B75-DBB51FFD314A)

## Summary by CodeRabbit

* **Tests**
  * Improved asynchronous test waits to recognize conditions that become true at the deadline.
  * Added cancellation handling to prevent unnecessary polling and suppress misleading timeout diagnostics.
  * Enhanced timeout reporting with elapsed-duration details.
  * Added coverage for deadline rechecks, persistent false conditions, cancellation, and excessive polling prevention.
  * Added diagnostics when expected conditions fail to occur before the timeout.

* **Documentation**
  * Updated assertion-hygiene guidance with requirements for final condition probes, elapsed-time reporting, and cancellation-aware polling.


cheapsteak:
Rebased onto `origin/main` (88 commits) and addressed every gate finding.

**High — `waitForSuspension` cancellation spin.** Fixed. The gate was right and
the criticism was exact: the PR described the cancellation fix in general terms
while applying it to one helper. All four unguarded `try? await Task.sleep`
loops now break on `Task.isCancelled` and return without a diagnostic —
`waitForSuspension` (149 call sites via `advanceWhenSuspended`),
`awaitDeliveryCycle`, `PathGate.wait`, and the original `waitUntil`. The two
helpers that use bare `try await Task.sleep` (`RepoRemoveCascadeTests.waitUntil`,
`RPCRouterRepoTests.waitUntilRemoved`) propagate cancellation already and need
no guard.

**Medium — `PathGate.wait` / `awaitDeliveryCycle`.** Both fixed as above. The
gate's note that this PR had raised `PathGate.wait` 10 s → 60 s and thereby made
a cancellation spin 6x more expensive was correct and is now moot twice over:
the rebase adopts main's `TestGate.deadline` (120 s) with its dominance
rationale, and the loop no longer spins. The local `PathGateSelfReleased` type
is gone in favour of the shared `TestGateTimeout` that landed on main — same
diagnostic, one type.

**Minor — stale "Four rules" in the spec.** Fixed;
`docs/specs/2026-07-24-test-hardening-design.md` §6 now carries rule 5 with both
corollaries, so the `Tests/CLAUDE.md` pointer resolves.

**The rebase makes the case stronger, not weaker.** In the twelve days this sat,
`TestDeadlines.saturatedPass` (90 s) and `TestGate.deadline` (120 s) landed and
were applied to four of the six helpers this PR touches — the *raise the number*
remedy. Not one of the six gained a fresh read. That is precisely the failure
mode the PR argues about: a larger budget cannot repair a verdict that is not
derived from an observation; it only moves the threshold at which the same lie
gets told. All six helpers on `main` today still take their verdict from the
loop's exit, `Tests/CLAUDE.md` still says "Four rules", and `PathGate.wait` still
self-releases in silence. The defect is live.

Verified: `scripts/swift-safe build --build-tests` exit 0, `swiftlint --strict`
clean over 1,040 files, and the four touched suites green (72 tests). Whole-suite
verdict left to CI.

cheapsteak:
Addressed the surviving MEDIUM, and the fix is the one the finding implies
rather than five more tests.

**"Fix is demonstrated by test for only 1 of 6 modified wait helpers."** Correct,
and the reason was that the loop existed six times. Writing five more regression
suites would pin six copies of one contract and leave the seventh copy free to be
written wrong. So the loop now exists **once** —
`Tests/TestSupport/BoundedPoll.swift`'s `pollUntilTrue` — and all six helpers
delegate to it, keeping only their own diagnostic. The defects can now only be
reintroduced in one place, which is where `BoundedPollSelfTests` pins them.

`PollOutcome` is deliberately three cases rather than a `Bool`: a caller must
treat `.satisfied` and `.cancelled` alike and report only on `.timedOut`.
Collapsing that to a `Bool` is precisely what let the cancellation half be
forgotten at five of six sites in the previous round.

**Mutation-checked, not just asserted.** Re-collapsing `pollUntilTrue` to its
pre-fix shape (verdict from the loop's exit, no cancellation guard) turns three
of the five tests red — including the restored busy-spin at **31,093,043
condition evaluations in 30 s** — while `neverTrueTimesOut` stays green as the
guard against a poll that reports success unconditionally. `waitForSuspension`
keeps a closure of its own because `checkSuspension()` throws when a sleeper
*is* registered, so its probe reads inverted.

**Minor — the PathGate 10 s → 60 s claim.** Correct catch, and it was stale
prose rather than a code change: those figures described the pre-rebase draft,
where `PathGate.wait` did carry a literal `60`. The rebase adopted main's
`TestGate.deadline` and no threshold in this PR changes at all. The description
was corrected before this comment; the change there is purely the qualitative
half the finding names — silent self-release became a loud `TestGateTimeout`.

Verified: `swift-safe build --build-tests` exit 0, `swiftlint --strict` clean
over 1,040 files, 202 tests across the 12 affected suites green. Whole-suite
verdict left to CI.

cheapsteak:
All three addressed. The MEDIUM was a real regression this PR's own refactor
introduced, and the finding is exactly right.

**MEDIUM — the two throwing waiters swallowed cancellation.** Correct, and it
was a behaviour change I did not intend or notice. `RepoRemoveCascadeTests
.waitUntil` and `RPCRouterRepoTests.waitUntilRemoved` used a bare `try await
Task.sleep`, so a cancelled task ended them with a thrown `CancellationError`;
routing `.cancelled` through the same non-throwing branch as `.satisfied` made
them return successfully and walk a cancelled test into the assertions right
after it. That is the mis-attribution this PR is about, reintroduced by its own
consolidation. Both now switch on all three outcomes and throw
`CancellationError` on `.cancelled` — pre-refactor behaviour restored. The
non-throwing helpers keep returning silently: they have nothing to throw, and a
diagnostic there would blame a call site for a cancellation the harness caused.
The stale "propagate cancellation already" line in the description is rewritten
to say what the code now does.

**MINOR — `conditionBeatsCancellation` never reached its branch.** Correct: a
live budget meant the loop satisfied the wait on its first read, so the
post-loop ordering was never exercised and the test's doc comment overclaimed.
It now uses a zero budget, which skips the loop entirely so the only read is the
one directly above the cancellation check. Mutation-checked by swapping those
two lines in `pollUntilTrue`: the test goes red now and did not before.

**MINOR — the "Assumptions" blanket claim.** Correct; `db.repos.get(id:)` is an
async DB read, not a lock-guarded flag. The section now says so, notes the read
is against an in-memory test database, and states the condition under which the
assumption would stop holding.

Verified: `swift-safe build --build-tests` exit 0, `swiftlint --strict` clean
over 1,040 files, 77 tests across the five touched suites green. Whole-suite
verdict left to CI.

cheapsteak:
Fixed, and the finding was the sharpest one yet: rule 5 named those three as the
helpers that got it right while they carried half the defect. Citing an
exemplar that is half-wrong is worse than not citing one.

`watchForSleeper` and `SidecarTestSupport.waitUntil` now delegate to
`pollUntilTrue`, which brings every pure bounded poll in the suite — nine of
them now — to one implementation. `advanceUntil` cannot delegate: it *advances*
the clock rather than only reading it, so it keeps its loop and takes the guard
directly. The guard sits before both branches there, because the armed branch
never suspends and a cancelled task would spin on it just as readily as on the
sleep.

Rule 5 is rewritten accordingly: it now leads with "call the helper" instead of
"write the loop like these three", since the three exemplars no longer exist as
separate implementations. It also spells out the case a `Bool` return would
hide — report on `.timedOut` only; `.cancelled` returns silently from a
non-throwing waiter and rethrows `CancellationError` from a throwing one, never
a fabricated timeout against an innocent call site.

**MINOR — the stray doc paragraph on `Counter`.** Correct, and it was exactly
what it looked like: an elapsed-vs-budget paragraph that belonged to
`HandshakeTimeout` and drifted during iteration. Moved back.

Verified: `swift-safe build --build-tests` exit 0, `swiftlint --strict` clean
over 1,040 files, 216 tests across the 13 affected suites green.

PR #728, opened 2026-08-26
Feat: name which provider a remote surface is talking about

zionts:
## Summary

TBD lets you register several providers against the same provider binary — one pointed at a production control plane, one at staging, say. Until now every surface in the app labelled them with `describe.name`, which names the provider's *kind*: both registrations reported `agentbox`, so both sidebar rows read "agentbox", both desks were titled "agentbox", and both appeared under the same name in every create picker. The registry key — the thing that is unique by construction, and the thing the session mirror, attach, and notifications all key on — was never on screen.

The consequence is worse than a cosmetic label: a green provider badge over an empty session list reads as "staging is idle" when the truth is "you are looking at the other registration, which has no sessions." This PR makes each surface answer the question the reader actually has — *which* provider is this, is this list current, is that session working, and where will Attach connect?

## Why this shape

Spec: [`docs/specs/2026-08-26-provider-identity-and-connection-ergonomics-design.md`](docs/specs/2026-08-26-provider-identity-and-connection-ergonomics-design.md).

Four decisions worth naming:

- **The registry key is the identity.** `RemoteProviderRegistry` rejects duplicate names, so `config.name` is unique; `describe.name` is a kind and is now shown only when it differs ("reports as agentbox"). Rejected: suppressing `describe.name` entirely — it is genuinely useful when a registration is named something else.
- **`describe.identity` is a flat display map, not a schema.** Every vendor's identity has a different shape (account and environment for one, box handle and region for another), and a fixed set of typed fields forces a provider to leave them empty or stretch them. This reuses the pattern the contract already has for `meta`, with a well-known key *order* and no interpretation. Rejected: deriving an environment from the exec path or arguments — that guess is wrong precisely when it matters (an entry named `…-staging` pointed by a stale flag at production), so the command line is shown verbatim instead.
- **Attach cannot express a fallback.** `RemoteAttachPreflight` matches the registry key exactly or returns a named diagnosis. Making "attach through some other provider" unrepresentable is stronger than remembering not to do it.
- **No feature flag.** Nothing here acts autonomously, mutates persisted state, or replaces a load-bearing path. The changes are additional read-only rows, changed wording, one changed tint, and an error surfaced where a silent skip used to be.

## What this PR does

**Contract (TBD half only — no provider repository is touched)**

- `docs/remote-provider-contract.md`: adds the optional `describe.identity` map — non-secret display pairs naming the backend a registration points at — with its answer-locally, never-carry-credentials, and degrade-gracefully rules, and corrects the `describe.name` bullet to say it names a kind rather than a registration. Additive within contract major 2; every existing provider stays conformant.
- `ProviderIdentity` (new) decodes it leniently — a non-string value costs its key, a non-object costs the map, and neither can cost the provider its registration. `ProviderIdentityRedaction` drops credential-named keys and redacts secret-looking registry arguments before anything is displayed.
- `RemoteSessionPayload.pendingQuestion`: the contract's `pending_question` is decoded TBD-side for the first time (it has been riding the wire unread). Cleared by `projectedForStaleSnapshot` — it is a liveness-axis claim, and a provider that has stopped answering leaves TBD no standing to make it.

**App**

- `RemoteProviderIdentityPresentation` (new): headline, kind subtitle, and the desk's identity rows (identity pairs → command line → versions). Every provider-name display site now routes through it — sidebar header, desk masthead, session detail, both create pickers, the create sheet, the create-defaults editor, and the auth CTA.
- `RemoteAgentAttention` (new): one sentence per row — the structured question and its options, else the provider's reason string (`permission_prompt` humanized, everything else verbatim), else the bare state. Blockage is still read off `agent_state` only, per the contract.
- `RemoteProviderInventoryState` (new): five distinct readings — populated, empty-success, never-snapshotted, freshness-unreadable, stale — each with its own sentence, plus a cross-provider note ("3 sessions are registered under another provider (agentbox-staging: 3)") on an empty desk.
- `RemoteAttachPreflight` + `RemoteAttachDiagnosisView` (new): exact-match resolution with six named outcomes, rendered in the attach pane. `RemoteAttachPager` no longer `continue`s past an unresolvable selection.
- Desk: identity block, an attention section naming blocked sessions and what blocks them, and the aggregate "N running sessions report no agent state — terminal liveness is not agent progress". Terminal-Running loses its green tint; green now means the agent axis said `working`.

## Assumptions

- Assumes a provider that sends `identity` honours the contract's non-secret rule. TBD does not trust this alone — it drops credential-named keys — but a provider that puts a secret under an innocuous key would still display it.
- Assumes `config.exec` is resolved the way the preflight probes it (absolute/relative path, else `PATH`). If TBD ever spawns providers through a shell with a different resolution, the probe could report "missing" for something that would have run.

## Evidence & verification

Seven new suites, all tier 1 and deterministic:

- `ProviderDisambiguationScenarioTests` is the reproduced failure end to end: two registrations of the same kind reporting the **same box handle**, zero inventory on one, three live sessions on the other, one of them blocked at a prompt, and a missing attach dependency. It asserts the two registrations never render alike, that the empty desk names where the sessions are, that three running terminals are not three working agents, that attach routes through the selected registration and refuses the other, and that no credential reaches the screen.
- `ProviderIdentityTests`, `RemotePendingQuestionTests` (shared): decoding, lenient degradation, redaction, wire round-trip, and the stale projection clearing the question block.
- `RemoteProviderIdentityPresentationTests`, `RemoteProviderInventoryStateTests`, `RemoteAgentAttentionTests`, `RemoteAttachPreflightTests`: every string and every branch above.

**Not verified locally, and this needs a reviewer's attention:** the environment this branch was authored in has no Swift toolchain (Linux; TBD is macOS-only), so `scripts/swift-safe build` and `scripts/test.sh` could not be run — `swift-safe` exits 69 with "swift executable not found". The code is reviewed by eye against the existing types, but CI is the first actual compile. No live app verification for the same reason; the UI changes are asserted through the pure presentation types rather than the views, which is why every string lives outside the view layer.

A follow-up dependency issue is filed for the provider half (emitting `describe.identity`); TBD degrades to the registry key and command line until then.

### DIFF
--- a/Sources/TBDApp/Remote/RemoteAgentAttention.swift
+++ b/Sources/TBDApp/Remote/RemoteAgentAttention.swift
@@ -1,0 +1,94 @@
+import Foundation
+import TBDShared
+
+/// One mirror row, read for the question "does a human need to do something,
+/// and what?" — the attention axis, stated in words.
+///
+/// Pure policy with no SwiftUI or `AppState` dependency, like
+/// `RemoteAttachExitClass` and `RemoteSessionDetailGates` beside it.
+///
+/// Two rules it never breaks, both from the provider contract:
+///
+/// - **Blockage is read off `agent_state`, never off `pending_question`.**
+///   The contract says so normatively, so every path here checks
+///   `waitingInput` first and reaches for the question block second, as
+///   detail. A provider that leaves a stale question block behind can make
+///   TBD's explanation wrong; it can never make TBD claim a session is
+///   blocked that the agent axis says is working.
+/// - **`state: running` with `agent_state: unknown` is not progress.** It
+///   means the terminal exists and nothing machine-readable reported what the
+///   agent is doing. It is a distinct, named reading here for exactly that
+///   reason: it used to render as the same green as real work.
+enum RemoteAgentAttention {
+    /// Why a human is being asked to look, in one sentence, or nil when
+    /// nothing about this row asks for a human.
+    ///
+    /// Ordered most specific source first: the structured question block, the
+    /// provider's own reason string, then the bare state.
+    static func explanation(for session: RemoteSessionInfo) -> String? {
+        guard !session.gone else { return nil }
+        switch session.payload.agentState {
+        case .waitingInput:
+            if let question = questionSummary(session.payload.pendingQuestion) { return question }
+            if let reason = humanizedReason(session.payload.agentStateReason) { return reason }
+            return "Waiting for input."
+        case .unknown where session.payload.state == .running:
+            return unattributedRunningExplanation
+        case .idle where session.payload.state == .running:
+            return "Agent is idle — it reported no work in progress."
+        case .working, .idle, .exited, .unknown:
+            return nil
+        }
+    }
+
+    /// The sentence for a live terminal whose agent says nothing. Named
+    /// rather than inlined because the desk states it both per row and once
+    /// in aggregate, and the two must not drift into saying different things.
+    static let unattributedRunningExplanation =
+        "Terminal is running and no agent state was reported — terminal liveness is not agent progress."
+
+    /// Rows a human has to act on. The agent axis alone decides this; a
+    /// terminal that happens to be running or exited underneath does not
+    /// change whether the agent is blocked.
+    static func needsAttention(_ session: RemoteSessionInfo) -> Bool {
+        !session.gone && session.payload.agentState == .waitingInput
+    }
+
+    /// Rows with a live terminal and no reported agent state — counted
+    /// separately from both "working" and "waiting" because it is neither.
+    static func isUnattributedRunning(_ session: RemoteSessionInfo) -> Bool {
+        !session.gone && session.payload.state == .running && session.payload.agentState == .unknown
+    }
+
+    /// The provider's free-form `agent_state_reason`, made readable.
+    ///
+    /// Exactly one value is translated: `permission_prompt`, the contract's
+    /// own worked example. Everything else is passed through verbatim rather
+    /// than pattern-matched — the field is provider-defined, and a caller
+    /// that invents meanings for strings it was never promised will
+    /// eventually put a confident sentence over a value that meant something
+    /// else.
+    static func humanizedReason(_ raw: String?) -> String? {
+        guard let raw else { return nil }
+        let trimmed = raw.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !trimmed.isEmpty else { return nil }
+        if trimmed == "permission_prompt" { return "Blocked on a permission prompt." }
+        return trimmed
+    }
+
+    /// The structured question block as one line: the first question's
+    /// prompt, plus its option labels when it has any, plus a count when more
+    /// than one question is pending.
+    static func questionSummary(_ question: RemotePendingQuestion?) -> String? {
+        guard let question, let first = question.questions.first else { return nil }
+        var line = "Blocked on a question: \(first.prompt)"
+        let options = first.options.map(\.label).filter { !$0.isEmpty }
+        if !options.isEmpty {
+            line += " (\(options.joined(separator: " / ")))"
+        }
+        if question.questions.count > 1 {
+            line += " — and \(question.questions.count - 1) more."
+        }
+        return line
+    }
+}
--- a/Sources/TBDApp/Remote/RemoteAttachDiagnosisView.swift
+++ b/Sources/TBDApp/Remote/RemoteAttachDiagnosisView.swift
@@ -1,0 +1,48 @@
+import SwiftUI
+import TBDShared
+
+/// What the attach pane shows when `RemoteAttachPreflight` could not resolve
+/// a selection to a spawnable command.
+///
+/// Deliberately in the pane rather than an alert or a log line: the user is
+/// looking here, so the reason belongs here. `RemoteAttachPager` used to
+/// `continue` past an unresolvable selection — no tab, no error, and a blank
+/// rectangle where the terminal should have been, which reads as "attach is
+/// broken" whatever the actual cause was.
+///
+/// Every string comes from the pure diagnosis, so what a user reads is
+/// exactly what `RemoteAttachPreflightTests` asserts.
+struct RemoteAttachDiagnosisView: View {
+    let selection: RemoteSessionSelection
+    let diagnosis: RemoteAttachPreflight.Diagnosis
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 10) {
+            HStack(spacing: 8) {
+                Image(systemName: "exclamationmark.triangle.fill")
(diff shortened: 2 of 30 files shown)


cheapsteak (approved):

zionts:
Rebased onto upstream/main and pushed. Resolved conflicts in `RemoteAttachPager` (kept upstream's @Observable env injection and generation-aware mount keys, layered in the PR's attach-preflight diagnosis path), `WorktreeRowView` (kept upstream's #744/#664 cloud+registry menu split, applied the identity headline label to the registry-provider menu), and the contract doc (kept upstream's full `capabilities` list, added the PR's `name`/`identity` wording). No local build this pass (machine load); relying on CI.

zionts:
Rebased again — upstream/main had moved (PR #897, remote re-attach-after-network-change, landed and heavily touched `RemoteAttachPager.swift` in the same area). Merged #897's testable `reconcile(_:mounts:activeSelection:appState:makeItem:)` structure with this PR's attach-preflight diagnosis behavior: the diagnosis-vs-terminal decision and the coordinator's self-healing re-resolve now live in `updateNSViewController`'s `makeItem` closure, keeping `reconcile` itself unaware of `RemoteAttachPreflight` so `RemoteAttachPagerWiringTests` (which drives `reconcile` directly with a stub `makeItem`) needed no changes. Verified `upstream/main` is now a genuine ancestor (`git merge-base --is-ancestor`) and GitHub reports MERGEABLE.

zionts:
Fixed the bare-positional-secret redaction gap: `redactArguments` now also matches well-known secret prefixes (`sk-`, `ghp_`, `xoxb-`, `AKIA`, `eyJ`, etc.) and, absent a known prefix, a conservative high-entropy heuristic (>=20 chars, mixed letters/digits, no excessive repetition, not a semver string, not a UUID) on any bare positional argument. Added tests for the reported gap, several known-prefix families, the high-entropy fallback, the repetition guard (fixtures long enough to actually reach that code path), the UUID carve-out, and a regression guard for ordinary positional arguments (paths, branch names, short identifiers).

zionts:
Addressed all three findings:
- Gated the bare-positional heuristic on `!arg.hasPrefix("-")` — an unrecognized flag like `--use-http2-multiplexing` no longer gets swept up with real secret values.
- Replaced the dot-count-based semver carve-out with `isVersionLike`, which requires every dot-separated segment to be purely numeric (after an optional leading `v`) — a Discord-style `id.timestamp.hmac` or PASETO-style `v2.local.payload` token no longer rides the same exemption a real version string gets.
- Added a paragraph to the PR body documenting the heuristic's thresholds and rationale at the point they're introduced, and explaining why this is a bug fix (no spec required) rather than a new tunable.
New tests for both fixes, plus one proving the version carve-out is actually reachable (the existing `1.2.3` fixture was too short to ever exercise it). Left the MINOR hex-identifier false-positive as-is, per the review's own framing ("a false positive in the safe direction... narrower real-world trigger than the Medium findings").

zionts:
Fixed: the `=`-joined shape now also falls back to judging the value on its own merits (`looksLikeSecret`) when the flag name isn't a recognized secret key, matching what the bare-positional and space-separated shapes already did — `--bearer=eyJ...` and `--pat=ghp_...` are now redacted. Added tests for both the fix and the non-redaction case (an ordinary `--size=large`/`--region=us-east-1` stays untouched). Left the MINOR `attachUnsupported` message-conflation finding as-is, per the review's own framing (safe-direction, message-accuracy nit, not blocking).

zionts:
Addressed the mirror-lag finding: documented at the call site why `resolve()` deliberately falls through to `.ready` when a session id is absent from every provider's mirror (a freshly created or freshly republished session legitimately has no row anywhere for a beat — `sessions` exists here only to catch an actual OTHER-provider claim, never to gate the selected provider's own resolution), and pinned the intended behavior with `resolvesReadyWhenSessionIsInNoMirrorYet`, per the review's own "comment + test" alternative to a new diagnosis case.

zionts:
Fixed: added a small, reviewer-named exact-match set (`shortSecretFlagAliases = ["t","p","k"]`) that `isSecretKey` checks before the substring scan, since a single-letter flag can never contain a multi-letter word. `-t`/`-p`/`-k` now redact their value unconditionally, matching their long-form spellings. Added a regression test for the fix and a companion test proving ordinary short flags outside that set (`-n`, `-e`, `-v`) are unaffected.

Also noting: the previous `claude-review` run on this PR hit an infra error (`Reached maximum number of turns (100)`, not a REJECT verdict) — reran it and got a real review that surfaced this finding. Separately, the `test` check failed on `AttachReplayOrchestratorTests` (a real-pipe-FD EOF-timing assertion, `n → -1` instead of `0`) in a file this PR doesn't touch at all; `main`'s own CI is green on that suite, so this reads as a load-sensitive flake rather than a regression — reran that job too.

cheapsteak:
## Review note: argv redaction has become the long pole — recommend descoping to a fixed-position rule

**How redaction got into scope.** The problem this PR solves — two registrations of the same provider kind looking identical — is solved by leading every surface with the registry key, which is unique by construction. The spec then added a **Command** row (the registry entry's `exec` and args, verbatim) as a secondary disambiguator that needs no contract change. Because args are user-authored and can carry credentials, the `identity` key filter was extended to them ("belt-and-braces"). A key-name filter only catches `--token=x`, so each review round found another shape: glued short flags, `-t`/`-p`/`-k` aliases, a secret flag after a secret flag, `=` inside glued values, capitalised single-dash flags, dash-less `KEY=value`, and finally bare positionals, which needed a prefix table, a 20-character floor, and version-shape exemptions.

Of the 16 commits after the feature commit, 11 are redaction work. The latest REJECT concerns a limit the spec already documents (a letters-only or sub-20-character secret behind an unrecognised flag). That gap can't be closed: no argv heuristic can decide in general what a secret looks like, so the review will keep finding shapes for as long as we keep patching.

**Recommendation.** Replace the heuristic with a fixed-position rule, as `TmuxManager.redactedArguments` already does: show the `exec` path and flag names, and redact every value — flag-following, glued (`--x=`, `-xVALUE`), or bare positional. For example: `agentbox --env ‹redacted› --region ‹redacted›`. It still tells registrations apart by their flags, there is no shape-guessing to find holes in, and it replaces roughly 300 lines of heuristic and ~450 lines of tests with about 20 lines and a few tests. The spec's "Redacting the command line" section would shrink to match.

What this gives up is reading `--env staging` off the desk. Telling registrations apart doesn't need it (the registry key does that), and "which backend is this pointed at" is the job `describe.identity` exists to do.

**On the review gate.** If the reviewer keeps rejecting on redaction-shape findings after the fixed-position rule lands, I'm prepared to do a maintainer override (admin merge). A display-only surface that redacts every value by position is a sound place to stop, and an unbounded chase after secret shapes isn't a good use of the gate.

PR #733, opened 2026-08-26
Feat: Say which session sent a peer message, and link to it

cheapsteak:
## Summary

When another Claude session sends you a message, the transcript viewer said only that *someone* had spoken:

> **Another Claude session sent a message:**

In a fleet of dozens of concurrent sessions that is close to no information. The sending session's own terminal shows the sender's name — the transcript viewer, reading the same transcript, showed less than the TUI did. And having learned a specific session messaged you, the obvious next gesture — go look at that session — had nothing to click.

A peer message now carries a header naming the sender, and clicking that name navigates to its worktree. The bubble is tinted amber so a received message never reads as something you typed.

It also renders the *message* instead of the delivery envelope. Each peer message arrives wrapped in a framing line, a `<cross-session-message …>` tag pair, and a ~90-word anti-escalation security preamble. Those tags survived into the rendered markdown as empty blocks, so every peer message appeared pushed down by stray blank lines — the visible symptom that prompted this work.

## Why this shape

Spec: [`docs/specs/2026-08-25-peer-message-attribution-design.md`](docs/specs/2026-08-25-peer-message-attribution-design.md).

**Attribution is chrome, never text.** The header is built from the harness-written `origin` dictionary and drawn *outside* the markdown body. `origin` is written by the receiving harness; the body is agent-authored. Keeping them separate means a peer can neither author its own attribution nor mint a worktree-navigation link inside its message. A test asserts the sender name is drawn and that no drawn body string contains it.

**Verified and asserted senders are not the same thing.** `origin.from` is sender-asserted — anything running as the same OS user can set it. What the harness actually verifies is the peer socket and pid. So `verified` requires *both* `name` and `verifiedPeerPid`; a uds path in `from` alone is not verification. Only a verified sender is presented as a name or made clickable; an asserted one renders muted and marked.

**Resolution refuses to guess.** A verified sender resolves by exact display-name match against active worktrees. Zero matches or two-or-more returns nil and the name renders unlinked. A local fleet of 1,916 worktrees contained a duplicated display-name pair — navigating to the wrong session is worse than not navigating, because it looks like it worked.

**Rejected:** resolving through the peer registry (`~/.claude/sessions/<pid>.json`) would be exact and immune to name drift, but the entry exists only while the sending process lives — so it answers precisely the cases name matching already answers, stays silent on historical ones, and puts a filesystem read on a render path. Also rejected: resolving at parse time, which would break the guarantee that a tail parse produces items byte-identical to a full parse of the same lines.

**Envelope stripping is anchored, not searched.** Every step works from one end of the string; nothing scans the interior. A peer that writes `</cross-session-message>` inside its own message — or quotes the preamble — keeps its text intact, and `<cross-session-messageXYZ>` is not mistaken for the tag.

## What this PR does

- **`TranscriptItem.peerMessage`** with a `PeerSender` (name, raw `from`, verified, pid). A distinct case rather than fields on `.userPrompt`: a received message is a different thing from something you typed, and the compiler then forces all seven exhaustive switches to say what they do with it instead of silently inheriting user-prompt rendering.
- **`PeerOriginExtractor`** — one pure function of a single row, shared by the transcript parser and the session scanner so the bubble and the session-picker rows cannot drift in how they read an envelope.
- **`PeerSenderResolver`** — pure, takes the worktree list as a parameter, no I/O.
- **Amber bubble tint** reusing the sidebar's existing `attention` amber at the same 15% alpha the accent tint uses.
- **Sender header** on the native bubble cell, with the height budget entering the measure and render paths together.
- **Detail overlay** shows the clean body and keeps the full delivered payload behind a disclosure, reached by a "Show delivered message" context-menu entry on peer rows.
- **Session-picker rows** read `<sender>: <first line>` instead of raw envelope text.

`ChatBubbleView` carries the same role and tint for symmetry, but it is **not** a live path for peer messages: `TableTranscriptView.makeCell` routes every `.chatBubble` to the native cell, and the SwiftUI row is the else branch. It still serves `AskUserQuestionCard`'s sub-bubbles. Nothing there is load-bearing for this feature.

No flag: additive rendering of one item kind, no autonomous behavior, no load-bearing path replaced. No migration, no config column, no new durable resource.

## Assumptions

- **`origin.kind == "peer"` identifies a peer message.** Written by the harness, present on every peer row measured across Claude Code 2.1.237–2.1.241. If the harness stops writing it, peer rows fall back to rendering as ordinary prompts — the pre-change behavior — rather than breaking.
- **The security preamble begins `This came from another Claude session —`.** Matched as a suffix from the end. If the wording changes, the preamble stops being stripped and reappears in the bubble; the rest of the extraction is unaffected.
- **A verified peer name equals its TBD worktree display name.** True because TBD passes `--name <displayName>` at spawn, but it is fixed at spawn — a renamed worktree, or a session spawned before that behavior, resolves to nothing and renders unlinked rather than wrong.

## Evidence & verification

**The repro, on real data.** A throwaway probe ran the production parser over a real local transcript (not the fixture) and asserted structure only, then was deleted. All 26 peer messages in that session: none retained the framing line, envelope tags, security preamble, or a leading blank line; all 8 verified senders carried names, all 18 asserted senders carried none; all 26 retained their delivered payload. Before this change every one of them rendered the envelope.

**Discriminating tests.** Each task was written test-first and watched fail before implementing — e.g. the parser suite went from 68 tests / 12 issues red to 68 passing, so the 63 pre-existing tests are unchanged. Specific ones worth naming:

- `headerIsChromeNotBodyText` — the name is drawn, and no drawn body string contains it.
- `assertedSenderNeverLinks` — an asserted label that *equals* a worktree display name still does not link.
- `verifiedUnresolvedSenderIsPlainText` — duplicate display names resolve to nothing.
- `nonPeerRowHeightIsUnchanged` — non-peer rows keep their exact pre-change height (spelled out, not recomputed).
- `peerCellRealizesAtTheReservedRowHeight` — measure equals render for a peer row, in both render sites.
- `peerBubbleOffersTheDeliveredEntry` / `invokingTheEntryOpensTheOverlayOnThatRow` — the disclosure is actually reachable, driven through the production render path (a real `NSTableView` over the actual coordinator) rather than by calling the overlay's helpers directly.
- Contrast is asserted on **computed** WCAG ratios, never hardcoded hex, with a guard that the amber actually resolves differently per appearance — otherwise every ratio would measure the same swatch twice and still pass.

**Contrast, measured rather than eyeballed.** The tint composites to `#F4EBDD` light / `#39301F` dark. Body text ~13:1 light and ~10:1 dark; inline code 5.9:1 / 6.2:1 — both clear WCAG AA. Link and secondary text land at 3.3–3.6:1, which is *below* AA — but they measure 3.3–3.9:1 on today's accent-tinted user bubble too, so link contrast is a pre-existing condition of every bubble in the transcript rather than something amber introduces. The tests hold peer bubbles to no-worse-than the existing bubble rather than pretending to fix it. Raising link contrast across all bubbles is separate work; `#0F4C9E` light / `#8ABEFF` dark would reach ~7:1 on this tint if someone picks it up.

**Suite.** Full run on the final tree: 8,220 tests in 795 suites, zero compile errors, every peer and transcript suite green. Two unrelated failures, neither from this branch: one contention flake that passes in isolation, and `terminal.activityEvent handler`, which is cross-suite state pollution latent on main — it passes alone (26 tests) and in a two-suite pair, fails in a four-suite mix, and no commit in the range touches activity-state code.

**Not yet verified: the pixels.** Everything above is structural. The rendered bubble has not been looked at in the running app, and transcript work specifically has a history of headless harnesses going green while the live view is broken. Reviewer or author should eyeball a peer message before merge.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=7EF5CE24-63E7-4954-858E-9F7FAFF716E0)


cheapsteak:
Addressed in `ec5c4ca6`.

**The finding is correct and the fix is real.** I verified all four of its claims independently before acting: no `openTranscriptOverlay` / `onTapGesture` / `contextMenu` in either bubble view, `menu(for:)` offering only "Copy message", and `ActivityRowFormatter.presentation(for:)` returning `nil` for `.chatBubble`. The disclosure was unreachable exactly as described.

A **"Show delivered message"** entry now appears on peer rows, gated on an overlay closure actually existing so it is never a dead menu item, and threaded through `configure` as a required parameter so a new call site has to decide rather than silently inherit `nil`. Scoped to peer rows on purpose — wiring the overlay for user and assistant bubbles would widen a pre-existing gap that predates this PR.

The review's sharpest point was about the tests, not the wiring: the old ones called the overlay's static helpers directly and never rendered a bubble through to a click, so they were shaped in a way that could not see the gap. The new suite drives the production path — a real `NSTableView` over the actual coordinator, cells obtained via `tableView(_:viewFor:row:)` — and asserts the entry opens the overlay with *that row's* id, that non-peer rows do not gain it, that a scroll-reused cell drops it, and that "Copy message" still works.

**One correction to the secondary point.** The review says `TranscriptParser.lookupDetail`'s peer branch "is never invoked anywhere in `Sources/TBDApp`" — true as written, but it is not dead code: it is reached by `terminal.transcriptItemFullBody` (`RPCRouter+TerminalHandlers.swift:3544`) and by `fullBody` (`TranscriptParser.swift:750`). It is a daemon-side path the app does not happen to use, and the overlay reads `deliveredPayload` off the already-parsed item, which needs no round-trip. I left both as they are rather than routing the overlay through an RPC it does not need.

**Two things the review prompted me to fix in the PR description**, both my errors:

- It claimed the sender header landed in "both render sites". `TableTranscriptView.makeCell` routes every `.chatBubble` to the native cell, so `ChatBubbleView` never renders a peer message in production. Its role and tint are there for symmetry only; the description now says so instead of implying a live second path.
- Same reason I did **not** add a `.contextMenu` to `ChatBubbleView`: an affordance on a path that never runs, and unassertable headlessly — the same class of defect as the finding itself.

Still outstanding and stated plainly: the rendered bubble has not been looked at in the running app. Everything here is structural plus a probe over a real transcript. Worth an eyeball before merge.

cheapsteak:
Addressed in `289c777a` — by removing the dead branch, not by wiring the overlay to it.

**The finding is right, and my previous reply dodged it.** That reply answered "is `lookupDetail` dead code?" — no, an RPC reaches it — when the actual question was "does the *overlay* reach it?", and the answer is no. Restating a narrower true thing in place of the asked question isn't a rebuttal. Verified before acting this time: `TranscriptOverlayView.swift` contains no `daemonClient`, no `Task`, no `lookupDetail`; `lookupItem()` resolves from `appState.sessionTranscripts`.

It was also worse than the finding said. `ItemDetail.deliveredPayload` was **written but never read anywhere** — `terminal.transcriptItemFullBody` uses only `text` and `attachment`, and `TerminalTranscriptItemFullBodyResult` has no such field. So the field was dead in the strictest sense, not merely unreachable-for-this-feature.

**Chose removal over wiring** because the overlay genuinely does not need an RPC round-trip: `buildItems` already performs the identical clean extraction, and the parsed `.peerMessage` item already carries `deliveredPayload`. Wiring the overlay to `lookupDetail` would have added a daemon hop purely to make a test's claim come true, which is the wrong direction. Removing the field also removed the only reason the explicit `init` existed, so the synthesized memberwise init is back; all 14 construction sites are unchanged.

`TranscriptItemDetailPeerTests.swift` is deleted with it. Those tests asserted real behavior — the defect was the guarantee I attached to them in the PR description, which claimed a parse-only test could not miss an overlay-rendering bug. It never could, because the overlay does not call the function they covered. A test that passes while proving nothing about the path it is credited with protecting is worse than no test, so it went rather than being re-pointed.

**Root cause worth recording, since both of this PR's findings share it.** I asserted how a code path worked without reading it, wrote that assertion into the implementation brief, and the agent built exactly what I described. The High finding (an unreachable disclosure) and this one (a branch nothing reaches) are the same mistake in two places. The PR description's `itemDetail` claim is now corrected to describe what the tests actually cover.

Verification: 119 tests in 5 suites green locally after the removal, and `test` is green on CI for the previous commit. No behavior change — `buildItems`'s peer branch is untouched.

Still outstanding, unchanged: nobody has looked at the rendered bubble in the running app.

Issue #735, opened 2026-08-26
Perf: terminal output stalls up to 1s reaching the renderer; cause unidentified


cheapsteak:
> [!IMPORTANT]
> **The attribution below is retracted.** This issue originally named the `AppState` poll cycle as the cause, on the strength of one 30-second window in which twelve of fourteen stalls began during a poll. A larger sample — 363 stalls — puts the association at **1.02×**, indistinguishable from chance, because poll cycles had grown to occupy 43.7% of the timeline. See [the correction](https://github.com/cheapsteak/tbd/issues/735#issuecomment-5427881411).
>
> **What stands:** terminal output waits up to **1,096 ms** to reach the renderer against a sub-2 ms median, replicated across three windows. Rendering is exonerated (parse 0.11 ms, display pass 5–9 ms). **The cause is unidentified**, and 55% of stalls begin in code with no instrumentation at all.
>
> The measurements below are accurate; only their attribution was wrong. Kept as written so the reasoning — and the way a small-sample correlation misled it — stays legible.

---

Terminal output waits up to **162 ms** to reach the renderer, against a **0.08 ms** median. The stalls coincide with the 2-second `AppState` poll cycle, and they are what makes typing and scrolling in embedded terminals feel laggy.

## The measurement

Temporary `os_signpost` intervals around the main-thread hop in `TerminalPanelView.dataReceived` (enqueue → execute), around `feed`, and around the display pass. `rpc.pollCycle` was already instrumented via `perfRPCSignposter`. Collected over 30 s of ordinary operation — agents streaming, no user interaction — with:

```
/usr/bin/log stream --style ndjson --signpost --predicate 'subsystem == "com.tbd.app"'
```

```
mainThreadHop    n=471   p50=0.08ms   p90=2.03ms   p99=156.56ms   max=162.35ms
rpc.pollCycle    n=15    p50=231.11ms p90=538.60ms p99=604.91ms   max=604.91ms
feed             n=471   p50=0.11ms   p90=0.92ms   p99=1.21ms     max=1.36ms
displayPass      n=257   p50=5.28ms   p90=8.03ms   p99=9.07ms     max=10.60ms
```

**Twelve of fourteen stalls over 20 ms began while a poll cycle was in flight**, including all eight of the worst.

## What the correlation shows, and what it does not

[...]

Issue #719, opened 2026-08-26
Perf: TBD's wheel monitor amplifies scroll into 100+ mouse reports/sec, overriding SwiftTerm's own accumulator


cheapsteak:
TBD intercepts every scroll wheel event in a local `NSEvent` monitor and hand-encodes SGR mouse reports, emitting `max(1, Int(abs(deltaY)))` of them per event:

```swift
let lines = max(1, Int(abs(deltaY)))
for _ in 0..<lines {
    tv.terminal.sendEvent(buttonFlags: buttonFlags, x: col, y: row)
}
```

`Sources/TBDApp/Terminal/TerminalPanelView.swift:794-797`.

There is no momentum-phase filter and no fractional-delta accumulator, so a sub-line delta of 0.3 still sends a full report, and a trackpad flick's momentum tail delivers events at 60-120 Hz for a second or two. Each report makes the hosted TUI repaint, and that output returns through the PTY as a redraw — so **scrolling manufactures the output flood that the draw path then pays for**.

## The pinned dependency already has a better handler

SwiftTerm's own `scrollWheel` (`Mac/MacTerminalView.swift:2986-3062` at the pinned revision) does pixel-delta accumulation against real cell height with remainder banking, exposes a `scrollSensitivity` knob, supports shift-bypass of mouse reporting, and handles DECSET 1007 arrow-key translation on the alternate screen. It never runs, because the monitor consumes the event first.

Upstream also carries an unmerged change bounding wheel reports: one report per classic wheel notch, and a 100 reports/second token bucket with a six-event immediate burst for precise trackpad deltas.

## Proposed fix

Keep the monitor, but **forward the event** — `tv.scrollWheel(with: event)` — instead of hand-encoding reports.

The monitor must stay. It was added because scroll wheel delivery was silently dead in this embedding (`TerminalView.scrollWheel` is not `open`, so it cannot be overridden in the subclass, and the comment at `:752-770` documents the visibility filtering the monitor performs for the keep-alive pager). Deleting it and hoping the responder chain works now would regress that. The change is to what the monitor *does*, not whether it exists.

Consider also adopting the token-bucket bound, which protects against a fast wheel regardless of what the accumulator does.

## Why this matters

[...]

Issue #667, opened 2026-08-17
Tracking: main-thread render cost in TBDApp — derived-state rebuilds and AppState invalidation


cheapsteak:
## What we're tracking

Main-thread rendering cost in TBDApp: derived state (sidebar child lookup, transcript
presentation) rebuilt wholesale inside SwiftUI view bodies, amplified by `AppState`
publishing object-wide to every view that observes it.

Distinct from #129, which tracks *layout* hangs inside the transcript pane
(`LazyVStack` row measurement). This issue is about *rebuild* cost and
*invalidation frequency*. They compound, but the mechanisms are separate.

## Measured evidence (2026-08-16)

Working set at the time of measurement: ~1,700 worktree rows in the DB, of which the
app's 2 s poll holds **127** (`excludeArchived: true`) — 117 active/creating, 53
nested, 11 repos — plus 154 terminals.

Standalone `swiftc -O` benchmarks mirroring the real type shapes:

- **`AppState.children(of:)` — 79.5 ms per sidebar render pass.** It runs
  `flatMap` + `filter` + `sorted` over every worktree in every repo, returning full
  `Worktree` copies, and `WorktreeSubtreeView.body` calls it once per row,
  recursively — so a render pass is O(N²), ≈16,100 `Worktree` copies. `Worktree`
  is 28 stored properties, 11 of them refcounted (≈14 retain/release pairs per copy).
  A parent-keyed index costs 1.68 ms rebuilt-every-pass and 0.004 ms warm.
- **`TranscriptPresentation.makeIndexSections` — 2.4 / 11.3 / 70.8 ms** at
  200 / 800 / 3000 transcript items. One `JSONSerialization` parse per tool-call item,
  plus an O(categories × keys) rescan, run inside a view body on every evaluation.

Both numbers are load-independent (isolated benchmarks). Note for anyone reproducing
the original symptom: it was first noticed on a heavily loaded dev machine (load
average >150, swap near capacity, ~70 concurrent agent sessions), where machine
saturation contributes independently of anything app-side. Don't attribute all
perceived lag here.

## Refuted — please don't re-run this

**`TranscriptItem.init(from:)` and `RPCResponse.decodeResult` are not on the main
thread.** `DaemonClient` is a plain `actor`, never `@MainActor`, and every
`decodeResult` call site is actor-isolated; `sendRawAsync` further pushes the
blocking `recv` onto a detached task. These frames appear prominently in `sample`
[...]

PR #739, opened 2026-08-26
Feat: See how long terminal output waits before it reaches the screen

cheapsteak:
## Summary

TBD could not see how long terminal output waits before it reaches the screen,
and that blind spot cost this project real time. Main-thread *utilisation* was
measured repeatedly, found low, and read as "the app is not the problem" — but
utilisation cannot see queueing delay. A main thread that is idle 80% of the
time still stalls terminal I/O for a third of a second if it spends that
remaining 20% in long uninterrupted bursts.

This adds two permanent stopwatches on the path terminal output travels from the
pty to the screen, so that question is answerable in one capture instead of a
week of narrowing. It also ships the analysis tool that turns a capture into an
attribution.

There is no user-visible behaviour change. Signposts are silent unless someone is
recording.

## Why this shape

Design: [`docs/specs/2026-08-26-terminal-latency-signposts-design.md`](docs/specs/2026-08-26-terminal-latency-signposts-design.md).
Evidence it was built from: [`docs/perf/2026-08-26-main-thread-burst-attribution.md`](docs/perf/2026-08-26-main-thread-burst-attribution.md).

**`mainThreadHop` is the load-bearing one, and not only for its duration.** A
period during which one is open is a period when terminal bytes were
demonstrably sitting on the main queue undelivered. That makes it a *window
definition derived from the symptom rather than from a suspect* — which is what
let the attribution rule the poll cycle out instead of assuming it in. An earlier
round of this investigation had blamed the poll cycle on a co-occurrence that
turned out to be a duty-cycle artefact (1.02x enrichment against chance).

**No feature flag, deliberately.** `docs/diagnostics-strategy.md` already settles
this: diagnostics live in the code permanently and stay silent by default, and
`TranscriptSignposts` (`perf-transcript`, `perf-rpc`) is the existing precedent
carrying no flag. CLAUDE.md requires a default-off flag for behaviour that acts
autonomously, destroys state, or replaces a load-bearing path; emitting a
signpost does none of those. A flag here would contradict the diagnostics
principle directly — "never delete a log line to reduce noise" applies equally to
never making one conditional on a toggle nobody remembers to set.

**Rejected: an interval around the display pass.** `TerminalView.draw(_:)` is
`public`, not `open`, so the only reachable seam is `viewWillDraw()` — and
closing an interval opened there requires posting an extra block to the main
queue on every display pass, roughly nine per second. That block lands on *the
very queue whose delay `mainThreadHop` exists to measure*. A permanent instrument
must not add traffic to its own subject. The spec records this as rejected
rationale rather than as a changelog entry, because the seam is obvious and the
next person will otherwise reach for it.

## What this PR does

- **`Sources/TBDApp/Diagnostics/TerminalSignposts.swift`** (new) — `mainThreadHop`
  and `feed` on `com.tbd.app` / `perf-terminal`, beside `TranscriptSignposts`.
  Region names are `StaticString` because `OSSignposter` requires it: the name is
  baked into the binary rather than formatted per emission, which is most of why
  a signpost costs nothing when nobody is recording.
- **`Sources/TBDApp/Terminal/TerminalPanelView.swift`** — emits both intervals
  around the existing main-queue hop in `dataReceived(slice:)`. Every chunk is
  instrumented rather than sampled: the median hop is 0.05 ms in every window
  measured, and the finding lives at the 99th percentile.
- **`Tests/TBDAppTests/TerminalSignpostsTests.swift`** (new) — pins the category
  and region names, because `scripts/diag/` matches them verbatim and a rename
  would break the tooling without breaking a build.
- **`scripts/diag/main-thread-attribution.py`** (new) — records Instruments' hang
  detector, a 1 ms sampling profiler and the signposts into **one** `xctrace`, so
  all three share a timeline and correlation is exact rather than clock-matched.
  Attributes main-thread CPU inside symptom-defined windows by which runloop
  callout it hangs off.
- **`scripts/diag/render-latency-report.py`** — the no-Xcode quick check.
- **Docs** — the spec, plus an evidence record covering the finding, the #724
  before/after, and the method.

## Assumptions

- **Region names and category are a published interface.** The analysis scripts
  match them verbatim. If they are renamed without updating `scripts/diag/`, a
  future capture reports "no intervals found", which reads as "the app is not
  emitting" rather than "someone renamed a string". The test exists to make that
  fail loudly instead.
- **`OSSignposter` emission is free when nothing is recording.** Backed by the
  `StaticString` requirement rather than by assertion, but if that stops holding,
  the cost lands on the terminal hot path.
- **Instruments' `Hangs` and `Time Profiler` remain exportable via `xctrace
  export --xpath`.** The attribution tool depends on the table schemas
  `potential-hangs`, `time-profile` and `os-signpost-interval`.

## Evidence & verification

The instrument was used to answer a real open question before being proposed for
merge, which is the evidence that it works.

**The finding**, replicated across two independent 150 s windows: during periods
when terminal output sat undelivered, the main thread was 97–98% on-CPU, 87–92%
of it inside a runloop-observer callout (SwiftUI's `Update.end`), while CPU spent
servicing the main dispatch queue fell from 25–28% to 0.2–5.1%. That contrast is
the mechanism rather than a correlation — a queued block cannot run while the
thread is inside a callout, so it is one event described from both sides. The
observer was the only main-thread callout class ever exceeding ~35 ms
uninterrupted; everything else topped out at 89 ms once, against 105 and 118
observer runs over 50 ms reaching 295 ms and 366 ms.

**The same instrument then measured the fix.** After #724, across five windows at
matched process ages: stall load fell from 1.33–3.71% to 0.26–0.83% of wall
clock, worst observed wait from 228–325 ms to 58–162 ms. Process age was tested
rather than assumed — 6-minute and 59-minute post-fix windows agree almost
exactly, so uptime does not explain the improvement.

**A number in this PR's own evidence record was revised down.** An earlier draft
claimed 325 ms → 58 ms by comparing the worst before-window against the best
after-window; a third after-window contradicted it, and the record now states the
range and notes explicitly that the post-fix side is variable (p99 5.3 ms in one
window, 148.5 ms in another). Flagging it because that number had already been
quoted elsewhere before the correction.

- `scripts/swift-safe build` — clean, run again after rebasing onto current
  `main` since the rebase pulled daemon changes onto the branch.
- `scripts/test.sh --filter TerminalSignposts` — **"Test run with 2 tests in 1
  suite passed"**. Count verified rather than exit code alone, since a
  zero-match filter exits green.
- `swiftlint --strict` — clean.

**What this deliberately does not measure.** Both intervals cover output
travelling *towards* the screen. Keystroke-to-echo has still never been measured
under a validated window, and scrolling was not measured here either. Wheel-event
amplification (#719) and the 150 ms post-keystroke frame-limiter window remain
the better suspects for those, and neither is touched by this PR. A future window
intending to cover typing or scrolling must be checked for containing the action
— four windows earlier in this investigation silently captured no scrolling and
produced a confident, wrong conclusion.

Related: #735 (whose premise this refutes and which needs correcting), #667,
#719, #724.

[main thread bursts](https://cheapsteak.github.io/tbd/open/?worktree=90E977ED-D96C-4BC8-A1D0-238F625E2B46)

### DIFF
--- a/Sources/TBDApp/Diagnostics/TerminalSignposts.swift
+++ b/Sources/TBDApp/Diagnostics/TerminalSignposts.swift
@@ -1,0 +1,59 @@
+import os
+
+/// Shared `OSSignposter` for `category: "perf-terminal"` — the path terminal
+/// output travels from the pty to the screen.
+///
+/// Signposts are silent unless a recording tool is attached, so these live in
+/// the code permanently rather than being added and deleted per investigation
+/// (see `docs/diagnostics-strategy.md`, and `TranscriptSignposts` for the
+/// sibling on `perf-transcript`).
+///
+/// Region names emitted today — **these are load-bearing**: the analysis
+/// scripts under `scripts/diag/` match on them verbatim, and
+/// `TerminalSignpostsTests` asserts them so a rename cannot silently break the
+/// tooling:
+///
+/// - `mainThreadHop` — begins on the pty reader thread immediately before the
+///   `DispatchQueue.main.async` in
+///   `TerminalPanelRepresentable.Coordinator.dataReceived(slice:)`, and ends
+///   inside that block. The duration *is* the main-thread queueing delay for
+///   terminal output — the thing average CPU utilisation cannot show, since a
+///   mostly-idle main thread still stalls output behind any long run-loop
+///   callout. Metadata: byte count of the chunk.
+///
+///   This interval also *defines* an analysis window: a period during which one
+///   is open is a period when terminal bytes were demonstrably sitting
+///   undelivered. That makes it a window derived from the symptom rather than
+///   from a suspect, which is what let the 2026-08-26 attribution rule the poll
+///   cycle out instead of assuming it in.
+///
+/// - `feed` — the nested `TerminalView.feed(byteArray:)` call, separating parse
+///   cost from queueing delay. Measured at 0.11 ms p50, which is what exonerates
+///   the parser.
+///
+/// Every chunk is instrumented rather than sampled: the median hop is 0.05 ms in
+/// every window measured, and the finding lives at the 99th percentile.
+///
+/// A third interval, `displayPass`, was deliberately **not** kept. Closing it
+/// required posting an extra block to the main queue on every display pass —
+/// the very queue whose delay `mainThreadHop` exists to measure. See
+/// `docs/specs/2026-08-26-terminal-latency-signposts-design.md`.
+enum TerminalSignposts {
+    /// Subsystem/category pair. Exposed so tests can assert it, since the
+    /// capture recipes and analysis scripts filter on it.
+    static let subsystem = "com.tbd.app"
+    static let category = "perf-terminal"
+
+    /// Interval names, referenced by the analysis scripts.
+    ///
+    /// `StaticString`, because `OSSignposter` requires it — the name is baked
+    /// into the binary at compile time rather than formatted at each emission,
+    /// which is a large part of why a signpost costs nothing when nobody is
+    /// recording.
+    enum Region {
+        static let mainThreadHop: StaticString = "mainThreadHop"
+        static let feed: StaticString = "feed"
+    }
+
+    nonisolated static let signposter = OSSignposter(subsystem: subsystem, category: category)
+}
--- a/Sources/TBDApp/Terminal/TerminalPanelView.swift
+++ b/Sources/TBDApp/Terminal/TerminalPanelView.swift
@@ -1245,9 +1245,30 @@ struct TerminalPanelRepresentable: NSViewRepresentable {
         }
 
         func dataReceived(slice: ArraySlice<UInt8>) {
+            // `mainThreadHop` begins here, on the pty reader thread, and ends
+            // inside the block below once the main thread actually picks it up.
+            // That duration IS the main-thread queueing delay for terminal
+            // output — the quantity average CPU utilisation cannot show, and the
+            // one that identified the SwiftUI view-graph flush as the stall.
+            // See `TerminalSignposts` for why every chunk is instrumented.
+            let signposter = TerminalSignposts.signposter
+            let hop = signposter.beginInterval(
+                TerminalSignposts.Region.mainThreadHop,
+                id: signposter.makeSignpostID(),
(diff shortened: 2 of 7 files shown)


cheapsteak:
Closing as overtaken by events — twice over, and neither is a criticism of the work.

**1. It instruments the tmux arm only.** `mainThreadHop` and `feed` are placed in `TerminalPanelRepresentable.Coordinator.dataReceived(slice:)`, which is the `LocalProcessDelegate` callback. A pty-holder session has no `LocalProcess` — the coordinator says so in as many words ("it has no `LocalProcess` and no control-mode attach") — and its bytes reach the view through `HolderStreamReader` in `startHolderClient`, a different call site entirely. So on a holder session these signposts would never fire.

**2. The delay it measures has since been removed from that path.** `dataReceived` now reads:

```swift
// Feed synchronously on the calling (IO) thread — this is the
// off-main parse the 2.0 bump exists for
viewHolder.withView { $0.feed(byteArray: slice) }
```

The `DispatchQueue.main.async` that `mainThreadHop` was built to time is gone from the feed path; it survives only for the viewer-MRU signal. The holder arm feeds off-main in the same shape. So the queueing delay this PR exists to expose was engineered away rather than measured away.

**What should not be lost.** The reasoning here was good and outlived the patch:

- `docs/specs/2026-08-26-terminal-latency-signposts-design.md` — why a symptom-derived window beats a suspect-derived one, which is what let the attribution rule the poll cycle *out* instead of assuming it in.
- `docs/perf/2026-08-26-main-thread-burst-attribution.md` — the evidence, including the 1.02x-against-chance finding that killed an earlier wrong conclusion.
- `scripts/diag/render-latency-report.py` and `main-thread-attribution.py`.

Anyone instrumenting the holder arm should start from that spec rather than from scratch.

**The gap this leaves is real and worth naming.** There is now *no* pty-to-screen latency measurement on either transport — `TerminalSignposts` has zero references on `main`. That matters because the holder migration's whole justification is latency: the transport design argues tmux costs per-keystroke process wakeups whose latency tracks load. Nothing currently proves the migration delivered that. Instrumenting `HolderStreamReader`'s feed, and the daemon-side emulator drain beside it, is the version of this work that is still worth doing.

PR #746, opened 2026-08-28
Fix: A raised hand never comes back down after a permission prompt is answered

cheapsteak:
## What's broken

When a Claude session raises a permission prompt, TBD puts a raised hand next to it in the sidebar. Nothing reliably puts that hand back down. You answer the prompt, the session carries on working for hours, and the hand stays up — so the one indicator that means "a human is needed here" degrades into noise, and a fleet of sessions accumulates hands that no longer mean anything.

Probing the running daemon on a development machine found it live: of 153 terminals, one had been showing a raised hand for roughly 25 hours while writing to its transcript all day. A second showed the healthy case for contrast — a prompt genuinely pending, its transcript frozen for six minutes, the hand coming down on its own once answered.

Separately, two of Claude Code's prompt notification types raised no hand at all.

## Why it happens

The hand goes up on Claude Code's `Notification` hook and comes down only when some *later* observation of the session's activity reaches the daemon and retracts it. That retraction is the only take-down path, and it depends on a hook firing again. When the hook rail goes quiet — and it does — nothing else ever revisits the standing reason. The row sits there indefinitely, because no read path asks whether it is still true.

There is no hook to lean on instead. Claude Code has no permission-*decision* event: `PermissionRequest` fires before the decision, `PermissionDenied` is auto-mode-classifier only, and a user rejection produces a tool_result string rather than an error, so neither `PostToolUse` nor `PostToolUseFailure` fires either. Scoping a `PostToolUse` matcher to "the permission prompt" is not possible — the matcher key is `tool_name`, and a permission prompt is not a tool.

The second bug is narrower: `AwaitingInputClass` enumerated the notification types Claude Code's *documentation* lists, but the binary emits a wider set. `worker_permission_prompt` and `elicitation_url_dialog` fell through to `.unrecognized` and raised nothing.

## What this PR does

Uses the signal TBD already has. While a prompt is on screen the session is blocked, so its transcript file does not move; once the prompt is answered the transcript starts growing again. That makes a transcript change a sound retraction signal, and it needs no hook to fire.

- When a `.promptOnScreen` reason is recorded, capture a fingerprint of the transcript as it stood: path, size, mtime. Other classes store nothing and pay no `stat`.
- On each read path, compare a fresh fingerprint against the stored one. Different — the file moved — and the hand comes down.
- Fingerprints are compared, not ordered against a clock. This asks "did the file change since the prompt was raised", which depends on no ordering between Claude Code's debounced notification timer and its own transcript flush, and on no two clocks agreeing.
- Rows written before this existed carry no fingerprint. Those get one *adopted* on first sight rather than being retracted on a guess, so nothing standing is taken down speculatively.
- The classifier now enumerates the types the binary actually emits, so the two missing prompt types raise a hand and six quota/computer-use types are filed as informational.

A transcript change is not by itself an answer, and this is the part worth reading closely. A parallel `Task` subagent appends **sidechain** records to the same JSONL while the prompt is still on screen, so raw file growth would drop the hand on a session that is still blocked — the exact inference `SessionStateResolver` already refuses, for the same documented reason. So the check is two-tier: an unchanged fingerprint costs one `stat` and the hand stays up; only a *changed* one pays to look, and then only at the region appended since the stored size. A non-sidechain record there means the parent session itself wrote, so the prompt was answered and the hand comes down. Sidechain-only means a nested agent wrote while the human is still blocked — the hand stays up, and the stored fingerprint is refreshed so the next pass is cheap again. The delta read is capped at 64 KiB; over the cap only the tail is examined, which fails toward leaving the hand raised.

`terminal.list` and the supervision readout reconcile. A third reader, the `session.states` RPC, deliberately does not: it never wrote the row, `terminal.list` heals it on its own cadence, and `SessionStateResolver` independently degrades a stale prompt to `unknown` rather than asserting it confidently. Net cost is one `stat` per raised hand per pass, with no timer, no sweep, and no new background work. Design rationale and rejected alternatives: [`docs/specs/2026-08-27-awaiting-input-transcript-supersession-design.md`](docs/specs/2026-08-27-awaiting-input-transcript-supersession-design.md).

No feature flag. This is a bug fix — an exempted category — and the write it makes is narrow, conditional and idempotent: it nils or rewrites one reason row on a read path rather than adding a timer or a sweep. Every failure direction is the status quo (a hand that stays up), not a new one.

## Assumptions

- **A blocked session's *parent* does not write to its transcript.** Nested agents do, which is why the appended region is examined rather than the file's size alone. If Claude Code ever wrote a non-sidechain record while a prompt was pending, the hand would come down early.
- **`transcriptPath` identifies the file the prompt was raised against.** A terminal retargeted by a `/clear`, a compaction, or a resume gets a fingerprint whose `path` no longer matches, which reads as changed — the hand comes down. That is the intended reading: a retargeted session is not sitting on the old prompt.
- **Fingerprints are not persisted across a schema change.** The field rides inside the existing JSON reason blob, which is why this needs no migration; a reason written by an older build simply has the key absent.

## Evidence & verification

- **The repro** is the live daemon probe described above: a standing `prompt_on_screen` reason whose transcript mtime was a day newer than the observation that raised it, with no activity observation in between.
- **Discriminating tests.** Each new behavior has a test that fails without the fix. The classifier change was run red first and produced exactly the two intended failures before the implementation landed. The supersession helper's tests assert both directions — a moved transcript retracts, an unmoved one leaves the hand raised — and two of them assert the fingerprinter was *never called* for non-prompt classes, so the "pays no stat" claim is enforced rather than described.
- **Full suite:** 8285 tests in 806 suites. Two failures remain, both pre-existing load flakes unrelated to this change — a child-process PID-marker timeout, and a live-tmux replay test whose own failure message reads "harness/load, not production". Repo-wide `swiftlint --strict`: 0 violations across 835 files.
- **Not live-verified in the running app.** Confirming the hand visibly drops requires restarting the daemon, and this machine is running a live fleet of sessions. The behavior is covered by tests at both read paths, but a reviewer should know the pixels have not been watched.

## Review history

Two findings from the review gate are addressed here rather than argued with. The `session.states` completeness gap was real — the description did claim two readers where there are three — and is corrected above and in the spec. The no-flag justification did argue from a false premise (that the feature deletes no persisted state); the conclusion stands but the reasoning is rebuilt on the applicable carve-out.

Checking the first of those turned up something neither review raised: `SessionStateResolver` carries a committed argument against reading transcript growth as an answer, naming the sidechain case. That was the load-bearing assumption of the original implementation, and it was wrong. The two-tier check above is the fix.

## Note for commit-by-commit review

The commits were produced by parallel agents and landed interleaved: `b405e655` and `c0e590d5` reference `AwaitingInputReason.transcriptFingerprint` but sit *before* `21eba5be`, which introduces it. The branch tip builds; those two intermediate commits do not build in isolation. Review the diff as a whole, and note it if you bisect.

[20260827-quiet-ferret](https://cheapsteak.github.io/tbd/open/?worktree=D66E687D-3E8C-4E24-8E78-027028EC2D07)

## Summary by CodeRabbit

* **New Features**
  * Awaiting-input prompts now clear automatically when parent conversation transcripts change.
  * Prompts remain visible for unchanged, unavailable, unreadable, or sidechain-only transcript updates.
  * Expanded recognition of Claude Code permission, URL, computer-use, quota, and related notification events.
  * Prompt notifications now retain transcript information for more accurate status updates.

* **Documentation**
  * Added design documentation covering transcript-based prompt reconciliation and notification handling.

### DIFF
--- a/Sources/TBDDaemon/Database/TerminalStore.swift
+++ b/Sources/TBDDaemon/Database/TerminalStore.swift
@@ -854,6 +854,69 @@ public struct TerminalStore: Sendable {
         }
     }
 
+    /// Retract a standing prompt whose transcript no longer matches the
+    /// fingerprint taken when it was recorded, WITHOUT asserting an activity
+    /// state — the same entitlement `clearAwaitingInputReasonIfNotNewer` has,
+    /// for the same reason: this caller knows the prompt is gone, not what the
+    /// session is doing instead.
+    ///
+    /// `expected` is re-checked inside this transaction. The stat that decided
+    /// to call happened outside the database and
+    /// `handleTerminalNotificationEvent` writes on its own connection, so a
+    /// prompt raised in that window carries a fresh fingerprint — and clearing
+    /// it would drop a prompt nobody has answered.
+    ///
+    /// A terminal that vanished between the caller's read and this write
+    /// reports `false` rather than throwing: the callers are read paths.
+    public func clearAwaitingInputReasonIfFingerprintMatches(
+        id: UUID,
+        expected: TranscriptFingerprint
+    ) async throws -> Bool {
+        try await writer.write { db in
+            guard var record = try TerminalRecord.fetchOne(db, key: id.uuidString),
+                  let standing = FactColumnJSON.decode(
+                      AwaitingInputReason.self, from: record.awaitingInputReason),
+                  standing.transcriptFingerprint == expected else {
+                return false
+            }
+            record.awaitingInputReason = nil
+            record.awaitingInputObservedAt = nil
+            try record.update(db)
+            return true
+        }
+    }
+
+    /// Attach a fingerprint to a standing prompt that has none — a reason
+    /// recorded before fingerprints existed.
+    ///
+    /// Adoption rather than retraction, because an absent fingerprint says
+    /// nothing about whether the prompt was answered. Adopting makes the row
+    /// answerable to the transcript's next write without inventing a comparison
+    /// against a clock, and it leaves a genuinely pending prompt raised.
+    public func adoptTranscriptFingerprint(
+        id: UUID,
+        fingerprint: TranscriptFingerprint
+    ) async throws -> Bool {
+        try await writer.write { db in
+            guard var record = try TerminalRecord.fetchOne(db, key: id.uuidString),
+                  let standing = FactColumnJSON.decode(
+                      AwaitingInputReason.self, from: record.awaitingInputReason),
+                  standing.classification == .promptOnScreen,
+                  standing.transcriptFingerprint == nil else {
+                return false
+            }
+            record.awaitingInputReason = FactColumnJSON.encode(
+                AwaitingInputReason(
+                    message: standing.message,
+                    hookEventName: standing.hookEventName,
+                    raw: standing.raw,
+                    notificationType: standing.notificationType,
+                    transcriptFingerprint: fingerprint))
+            try record.update(db)
+            return true
+        }
+    }
+
     /// Retract a standing wait reason WITHOUT asserting an activity state.
     ///
     /// The mirror of `recordAwaitingInputReason`, and it exists for the same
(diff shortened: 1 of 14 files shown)


cheapsteak:
@Jephuff heads up this should fix it

PR #750, opened 2026-08-29
Perf: Lift the fixed 16.67ms paint floor and move terminal parsing off the main thread

cheapsteak:
## Summary

Every TBD terminal repaints through SwiftTerm's legacy scheduler: damage coalesces behind a fixed `DispatchQueue.main.asyncAfter(+16.67ms)` — a floor on key-to-paint latency, not a rate limit — and parsing plus rendering both run on the main thread, so a handful of streaming TUI sessions peg main (upstream issue migueldeicaza/SwiftTerm#658, which matches TBD's measured symptom). Upstream has rebuilt this outright on `main`: a `FrameDriver` on `CADisplayLink`/`CVDisplayLink` replaces the fixed delay, Metal-layer rendering moves off the main thread, and the `Terminal` grid becomes lock-guarded shared state. This PR adopts that revision and migrates TBD's terminal-access code to the new concurrency model, additionally moving PTY parsing onto the IO thread so streaming output stops costing main-thread parse CPU. (The upstream Metal-off-main rendering is present but not yet engaged: Metal stays gated behind setUseMetal, which TBD wires up only on the separate metal-renderer branch — at this revision drawing remains on the Core Graphics path, scheduled by the FrameDriver. Verified in the field by sampling the running app: FrameDriver ticks + swiftterm-io threads present, no Metal/render-loop threads.)

No tagged release contains any of this — `v1.19.0` and `v1.20.0` still carry the fixed-delay scheduler — so this continues the repo's existing practice of pinning an unreleased revision, now via a fork (below).

## Why this shape

Spec: `docs/specs/2026-08-28-swiftterm-2-locked-terminal-access-design.md` (brainstormed + design-reviewed). The three decisions, telegraphically:

- **All non-snapshot terminal access goes through `withTerminal` under upstream's public `TerminalLock`** (per-cell click reads, mouse forwarding, mutations); pure cols/rows reads use the `terminalDimensions` snapshot. Every access is gesture-frequency; FIFO fairness bounds a contended wait at one ~2ms parse batch. Rejected in the spec: snapshot-plus-throttle (priority inversion at the OSC 8 dedup guard from #113, plus a fragile normalization invariant) and text-only degradation (reintroduces the #113 double-open).
- **Parsing moves to the IO thread**: `LocalProcess(delegate:, dispatchQueue: .main, directDelivery: true)` — upstream's own shipped configuration. `feed()` parses on the calling thread, so the delivery thread decides where parse cost lands.
- **The pin is a one-commit fork.** SwiftTerm 2.0 removed every public route from a view to its `Terminal` (`getTerminal()` deleted, `terminal` internal, delegate callbacks pass the view, and its migration guide's claimed route does not exist). The fork `cheapsteak/SwiftTerm`, branch `tbd/public-with-terminal`, revision `62d0be6` = upstream `9c2518e` (main, 2026-08-27) + one commit making the existing self-locking `withTerminal` helper public (plus the migration-doc correction). Exit criterion, recorded in `Package.swift`: upstream merges an equivalent accessor → re-pin to upstream, delete the fork. The upstream PR is prepared but deliberately not yet opened.

Pin provenance: old `16c5286` (upstream main) → new `62d0be6` (fork, = upstream `9c2518e` + accessor commit).

## What this PR does

- `Package.swift`: pin moved to the fork revision; comment block rewritten (pin rationale, fork exit criterion, ChildReaper re-verification for `9c2518e`).
- `TBDTerminalView`: per-cell click reads (`hasOSC8Payload`, `extractFilePath`, `extractHyperlinkURL`), click passthrough, and `setCursorStyle` migrate to `withTerminal` copy-out blocks; `gridPosition` to `terminalDimensions`; the theme-repaint `updateFullScreen()` workaround is **deleted** — 2.0's `installColors` performs the invalidation itself (`colorsChangedLocked()` ends in `updateFullScreen()` + `frameDriver.markDirty()`); the `notify` override is replaced by `observeOscEvents` filtering OSC 777 (observer token stored; user-registered OSC *handlers* would preempt built-ins, the observer does not).
- `TerminalPanelView` / `LocalPTYTerminalView` coordinators: `directDelivery` init; `dataReceived` feeds synchronously on the IO thread through a new lock-guarded `TerminalViewHolder` (set before `startProcess`, cleared during cleanup before terminate/release, so a batch racing teardown reads nil and drops); `getWindowSize` reads `terminalDimensions`; scroll forwarding under `withTerminal`.
- `ChildReaper`: **doc comment only.** Re-verified at the new revision: `waitpid` is still called from exactly one place in the whole package (`LocalProcess.swift:384`, reachable only from the `DispatchSourceProcess` `.exit` handler at `:589`), and neither teardown path TBD uses reaps — `deinit` (`:356`) and `terminate()` (`:604`) both cancel the monitor without waiting (`terminate()` now cancels *before* sending `SIGTERM`, `:624`). So TBD must keep reaping, and ChildReaper stays. The exit monitor arms on `dispatchQueue`, which both call sites now pass as `.main` explicitly, keeping the serialization analysis intact.
- Tests: new production-path round-trips (OSC 8 payload at the exact cell, OSC 777 → `onNotification`, off-main `feed`, holder teardown + race); two pre-existing suites that used removed APIs migrated.

## Flag & rollout

Ships unflagged; the spec's Gating section carries the full argument. Short form: the pin itself cannot be flagged (two SwiftTerm revisions cannot coexist in one build), and `directDelivery` — the one mechanically flaggable piece — has no meaningful off state (its off branch still runs the new FrameDriver, lock, and renderer; it only moves parse back onto main). Soak = dogfooding on the development fleet; the separately validated key-to-paint harness provides the before/after judgment before merge.

## Assumptions

- `withTerminal` bodies copy values out and never re-enter — enforced by convention and a comment block, per the spec.
- Terminal-diagnostics code that hooks `viewWillDraw()` may go quiet at this revision (a frame loop off the main thread may never call it). Expected, noted in `Package.swift`; not a regression.
- The fork is temporary; each future bump costs one rebase of one commit until upstream merges the accessor.

## Evidence & verification

- Build green (`scripts/swift-safe build`, exit 0); SwiftLint `--strict` clean on touched files.
- Full suite: 8378 tests / 813 suites; the only unexpected reds were 60s-timeout waits during a run on a machine ~80× over its load baseline, and a targeted rerun of every affected suite passed 19/19 in 0.37s — the waits starved, no value assertion failed in any run.
- Discriminating tests: OSC 8 asserts payload at the anchor cell AND absence one cell past it; OSC 777 resumes only through the full registration → token → parse → main-hop chain; off-main `feed` asserts the bytes landed via `getBufferAsData`.
- **Performance before/after has NOT yet been measured** — that comparison runs on the separately validated key-to-paint harness and is explicitly out of scope here.
- **Live verification (in progress)**: the fleet daemon+app are running this branch. Verified so far, with log evidence: OSC 777 notification interception end-to-end in BOTH directions — delivered to `onNotification` while the terminal's window is not key (`OSC-DIAG event code=777 → parsed → delivering → sink log line`, via a tmux-passthrough-wrapped sequence into an attached viewer), and suppressed at the key-window guard while focused. Zombie scan: 0 `<defunct>` children under the new app. Still pending eyes-on: OSC 8 cmd-click single-open, plain-path click, theme-switch repaint, scroll inside tmux mouse mode. (Note for future readers: the production sink for terminal OSC 777 is a debug log line, not a UI banner — same as before this PR; tmux only forwards OSC 777 to an attached client when wrapped in the `ESC Ptmux;` passthrough, and TBD attaches viewers lazily, so a probe needs the tab visible.)

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=C1A2C601-5B14-4BF5-A7B3-7428FBEFD49F)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved terminal stability during background input, resizing, scrolling, and process shutdown.
  - Fixed terminal notifications so OSC 777 title and message alerts are reliably parsed and delivered.
  - Improved hyperlink detection and terminal content handling.
  - Reduced stale display artifacts during terminal repainting.
  - Improved cleanup and child-process handling to prevent hangs and stale terminal activity.
- **Tests**
  - Added coverage for locked terminal access, notifications, hyperlinks, concurrent access, and process cleanup.
- **Documentation**
  - Added design documentation for the updated terminal integration and lifecycle behavior.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "05b9d3a267e1b37004a9b00fcb3ceddf883f76c393cae24289146452eda0a363",
+  "originHash" : "6399cb1349f3ec270134d9239262667a01d51cf6d0ece6a4baf1c1507e9d2979",
   "pins" : [
     {
       "identity" : "grdb.swift",
@@ -139,9 +139,9 @@
     {
       "identity" : "swiftterm",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/migueldeicaza/SwiftTerm",
+      "location" : "https://github.com/cheapsteak/SwiftTerm",
       "state" : {
-        "revision" : "9c2518e21bcc6933a4fc82ce6586f926517d3045"
+        "revision" : "62d0be6d4c9641a4b27f311c55e1489b024271c3"
       }
     },
     {
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "e06da08a9cd41529dba112b2abb2f47e48bc462bd520f5c09a925efc034c3411",
+  "originHash" : "05b9d3a267e1b37004a9b00fcb3ceddf883f76c393cae24289146452eda0a363",
   "pins" : [
     {
       "identity" : "grdb.swift",
@@ -8,6 +8,15 @@
       "state" : {
         "revision" : "36e30a6f1ef10e4194f6af0cff90888526f0c115",
         "version" : "7.10.0"
+      }
+    },
+    {
+      "identity" : "h",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/rarestype/h",
+      "state" : {
+        "revision" : "aa3626829160917d4378330617971977cbd78f5b",
+        "version" : "1.0.1"
       }
     },
     {
@@ -110,6 +119,15 @@
       }
     },
     {
+      "identity" : "swift-png",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/tayloraswift/swift-png",
+      "state" : {
+        "revision" : "8a0bcd4df5e4b307c804937776a56dd6ecdf6396",
+        "version" : "4.5.1"
+      }
+    },
+    {
       "identity" : "swift-system",
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-system.git",
@@ -123,7 +141,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/migueldeicaza/SwiftTerm",
       "state" : {
-        "revision" : "16c52867763121b121f3b29a4d439052e7e76034"
+        "revision" : "9c2518e21bcc6933a4fc82ce6586f926517d3045"
       }
     },
     {
--- a/Package.swift
+++ b/Package.swift
@@ -27,25 +27,37 @@ let package = Package(
         .package(url: "https://github.com/groue/GRDB.swift", from: "7.0.0"),
         .package(url: "https://github.com/apple/swift-argument-parser", from: "1.3.0"),
         .package(url: "https://github.com/apple/swift-nio", from: "2.65.0"),
-        // Pinned to a main-branch revision for two reasons. The `fontSmoothing`
-        // public property (PR #531) still hasn't shipped in a tagged release.
-        // And the macOS CoreGraphics draw path needs `context.clear(dirtyRect)`
-        // (upstream d5ee56e / PR #582): SwiftTerm opts into PARTIAL backing-store
-        // updates (`disableFullRedrawOnAnyChanges` + `layer.contentsFormat =
-        // .RGBA8Uint`), so AppKit erases only on a full redraw and the draw path
-        // must erase the dirty region itself. Without that clear, any region the
-        // draw does not touch keeps the previous frame's glyphs — visible as
-        // stale text in default-background cells that a window resize wipes.
-        // Switch back to `from: "1.14.x"` only once a tagged release contains
-        // BOTH #531 and #582.
+        // Pinned to a main-branch revision because no tagged release contains
+        // upstream's display-link frame scheduler. Through v1.20.0, SwiftTerm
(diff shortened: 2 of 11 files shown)


cheapsteak:
Both findings from the review round at patch-id `a29886cb`:

**MEDIUM (repaint workaround deleted on an unverified premise)** — addressed in `a6ffc639`. The premise is now pinned by a discriminating production-path test (`TerminalLockedAccessTests`): feed real bytes, `clearUpdateRange()`, require the update range reads nil, call `installColors(_:)`, require `getUpdateRange() == (0, rows)` — no manual `updateFullScreen()` anywhere in the test. Discrimination was mutation-checked: with the `installColors` call commented out, the suite fails on exactly that assertion. If a future SwiftTerm bump stops invalidating on palette install, this test goes red instead of users seeing stale cells.

**MINOR (no-flag justification leans on a not-yet-run soak)** — the soak is now underway: the fleet daemon+app were restarted onto this branch and the live checklist is being executed (notification interception verified in both directions — suppressed while the terminal's window is key, delivered when not). Remaining items (OSC 8 single-open, theme repaint, scroll forwarding) and the key-to-paint before/after are still gates on *merge*, per the Evidence section — the PR description will be updated as they complete.

PR #763, opened 2026-08-31
Compound: Spec the tmux replacement that stays flat under load (per-session pty holders)

cheapsteak:
## Summary

Design spec for migrating TBD's terminal sessions off tmux onto per-session pty holders. Field measurement showed a raw pty's keystroke echo never exceeded 5 ms at any system load (it happens in the kernel line discipline, with no process wakeup), while tmux's p90 ran 9.3 → 12.7 → 139 ms as load rose — it degrades superlinearly because each keystroke costs two extra server wakeups whose latency is scheduling delay. The justification is scaling headroom, not current pain: at quiet load tmux is imperceptible, but it is the component that breaks first as the fleet grows. The spec targets ~2× today's fleet (~150 concurrent sessions at sustained load ≥ 100).

## Why this shape

The core precedent is iTerm2's session-restoration server: a supervisor that `forkpty()`s each job and hands the pty master to the app over `SCM_RIGHTS` costs zero process wakeups on the interactive path. The spec adapts that to TBD's advantage — an always-on daemon that can drain output while nothing is attached, covering exactly the window iTerm2 cannot. Key decisions, each with rejected alternatives written out in the spec:

- **Per-session holder** (never reads the pty; tiny, versioned, essentially frozen) over per-repo/global holders or holding masters in the daemon — blast radius one session, and holder upgrades become a non-event.
- **Daemon as arbiter and default reader**, draining detached sessions into a headless SwiftTerm emulator; attach is daemon-brokered over the existing FD sidecar with a snapshot preamble so reattach paints instantly.
- **Two-store model with pull-on-demand** over streaming attached output back to the daemon — machine reads stay current (pull from the app when attached, bounded, hibernation veto fails closed) at zero steady-state cost; the accepted loss is scrollback gaps across attached periods, which transcripts already cover.
- **Single-reader arbitration** that errs toward not reading (backpressure is recoverable; double-reads silently corrupt), checked against the restart script's real daemon-then-app sequencing.
- **Holder death is session death by policy** — prompt, identity-verified kill with a distinct exit reason and a transcript-based recovery ladder.

A fresh-context directional review ran against the committed spec; its findings (sidecar disconnect vs app death, wedged-app fleet stall, reaper blind spot for launchd-reparented children, and others) are folded in as the second commit.

## What this PR does

Adds `docs/specs/2026-08-30-pty-holder-session-transport-design.md`. No implementation, no flag, no migration — the spec itself defines the rollout (`pty_holder_enabled`, default-off, no SQL `DEFAULT`, spawn-time-only gating so both transports coexist through the soak) for the implementation PRs to follow. Reconciler duties are assigned per the named-reconciler doctrine, and a "Learnings from iTerm2" section directs implementers and reviewers to the prior-art evidence before re-deriving its scar tissue.

## Evidence & verification

Docs-only. The evidentiary basis is the measurement and research records under `docs/research/2026-08-30-terminal-session-persistence/` (landing separately from the branch they were produced on); the decision-critical numbers are restated in the spec so it stands alone. Design decisions were answered by a human in a brainstorming session per the repo's spec gate.

## Summary by CodeRabbit

* **Documentation**
  * Added an approved design for dedicated per-session terminal persistence.
  * Documented reconnection, reader handoff, snapshots, scrollback, resizing, recovery, rollout, and coexistence with existing persistence.
  * Added research comparing terminal persistence approaches, including reattachment, state retention, failure recovery, and cleanup.
  * Documented performance measurements and requirements for responsive sessions and reliable reattachment.

* **Chores**
  * Added diagnostic tools for measuring terminal latency, round-trip performance, and system contention.

### DIFF
--- a/docs/research/2026-08-30-terminal-session-persistence/iterm2-session-restoration.md
+++ b/docs/research/2026-08-30-terminal-session-persistence/iterm2-session-restoration.md
@@ -407,6 +407,57 @@ rediscovered by anyone building the same thing:
   "disclaim children" advanced setting
   (`sources/Tasks/iTermFileDescriptorMultiClient+MRR.m:245-250`).
 
+## Forcing a repaint after reattach: the size jiggle
+
+A reattached session inherits a live pty but no screen contents, so anything
+already drawn is gone unless the program redraws it. iTerm2's lever for that is
+a deliberate window-size perturbation, and it is worth reading closely because
+what it does *not* cover is as instructive as what it does.
+
+- **The mechanism.** `WinSizeController` (ObjC name `iTermWinSizeController`,
+  `sources/Tasks/WinSizeController.swift:17-18`) exposes `jiggle()` (`:116-124`)
+  and `forceJiggle()` (`:110-114`) over the private `reallyJiggle()`
+  (`:126-140`), which grows the last-known grid by one column, sets it, then
+  sets it straight back. Two size changes, so the child gets a SIGWINCH and
+  full-screen programs repaint.
+- **Same-size resizes are suppressed at two layers**, which is why the
+  perturbation is necessary at all rather than just re-asserting the size.
+  `WinSizeController.set(_:)` returns early when the request matches the current
+  grid, view size and scale factor (`WinSizeController.swift:171-177`), and
+  below it `iTermSetTerminalSize` only calls through to
+  `iTermForceSetTerminalSize` when the requested size differs from what
+  `ioctl(TIOCGWINSZ)` reports (`sources/Tasks/iTermTTYState.c:104-109`); the
+  sole `TIOCSWINSZ` lives at `:101`. A reattach re-inherits the same master fd
+  with its kernel-tracked winsize intact, so asking for the same geometry
+  issues no ioctl and delivers no SIGWINCH.
+- **The ordinary reattach path does not jiggle.** Restoring a saved window
+  arrangement goes through `-tryToAttachToServerWithProcessId:tty:`
+  (`sources/PTYSession.m:2657-2671`),
+  `-tryToAttachToMultiserverWithRestorationIdentifier:` (`:2673-2687`) and
+  `-tryToFinishAttachingToMultiserverWithPartialAttachment:`, none of which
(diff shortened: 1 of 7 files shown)


cheapsteak:
Addressed both Mediums and the Minor in 64452c67 — all three were internal-completeness gaps in the spec text, and all three are now specified rather than dropped.

**Advisory lock file (Medium).** The mechanism is genuinely wanted, so it got contract text rather than a retracted claim. The concrete hazard it guards: a socket file alone cannot distinguish a live holder from the corpse of a SIGKILLed one, and `bind` refuses an existing path — so "unlink the stale socket, then bind" is a race two spawners can both win, leaving one holder bound to a path the other already unlinked. The spec now specifies a zero-byte lock file beside each socket on the same UUID, an exclusive `flock` taken before any unlink or bind, the descriptor passed across exec (`FD_CLOEXEC` cleared) with the daemon closing its own copy so the holder alone holds it for its life, and back-off-rather-than-clear for a spawner that cannot take it — which is what makes the stale-daemon-from-another-checkout hazard safe rather than merely detectable. Reclaimer: `OrphanGC`'s socket sweep now unlinks the sibling lock file too. It cannot leak a *lock* (the kernel drops it on holder death), only an empty file, and the spec says so explicitly so a reader doesn't go looking for a liveness reclaimer that isn't needed. The iTerm2 learnings bullet now points at that section instead of asserting adoption in the abstract.

**Detach ordering (Medium).** Correct that this was asymmetric with Attach. Detach now states close-before-notify explicitly and says why: a notify-first detach puts the daemon on the fd while the app's last `read()` is still outstanding, which is the double-reader corruption in miniature. Also noted that an app dying mid-detach needs no special case — the fd closes with the process and the app-death path covers it.

**Identity verification on app-liveness (Minor).** Implemented rather than deferred. App death now names the same recorded-pid-plus-executable-and-start-time check used by the holder-death kill and the `AgentReaper` leg, with the failure mode spelled out (a reused pid making a dead app look alive stalls the fleet), and daemon re-adoption states up front that every app-liveness question in that bullet is that check, never a bare `kill(pid, 0)`.

PR #764, opened 2026-08-31
Decide the shared module cache once, in scripts/swift-safe

cheapsteak:
## Summary

Every TBD worktree keeps its own precompiled-module cache — 510 to 818 MB each, around
4.7 GB across the worktrees currently holding a `.build`. A shared cache already existed
to collapse those into one copy, but only `scripts/restart.sh` selected it, so most
builds never used it and the machine paid for both: the per-worktree caches *plus* 1.5 GB
of shared cache.

The disagreement also cost compile time. SwiftPM writes the module-cache path into every
compile command in the build plan, so a worktree that alternates between `restart.sh`
(shared path) and `scripts/swift-safe` or `scripts/test.sh` (default path) re-plans and
**fully recompiles on every transition, in both directions**. That is the normal
iteration loop — tests while working, restart to verify live — so the trees most actively
worked on paid it most often.

This moves the decision into `scripts/swift-safe`, the one wrapper every governed build
already passes through, so all three entry points plan identically. The per-worktree
caches stop being regenerated and about 4.7 GB comes back.

A second, unrelated bug fix rides along because it was found while investigating this
one; it is described separately below and is safe to review independently.

## Why this shape

Spec: [`docs/specs/2026-08-30-shared-module-cache-design.md`](docs/specs/2026-08-30-shared-module-cache-design.md).

The wrapper is the right home because it is already the mandatory path to SwiftPM — the
repo guardrail rejects raw SwiftPM invocations, and `scripts/test.sh` routes through it
too. A build-wide compile setting duplicated across callers drifts apart, and that drift
was the bug.

Three findings shaped it, all recorded in the spec with evidence:

- **The path is always in the plan.** SwiftPM emits a module-cache path into every compile
  command even with no flags, naming the worktree's own directory. The flags don't *add*
  an argument, they change the *value* of one that is always present — which is why the
  recompile is symmetric, and why no environment variable can select the cache instead.
- **A symlink cannot work.** Cached artifacts embed the *spelled* path of the worktree
  that created them and clang validates that spelling, not the inode. A second worktree
  sharing a directory this way fails outright with a PCH path mismatch and does not
  self-heal. The same embedded bytes defeat APFS cloning.
- **The path must come from the passwd database, not `$HOME`.** `scripts/test.sh` fences
  `HOME` and `CFFIXED_USER_HOME` to a scratch directory. A `$HOME`-derived path would
  resolve *inside the fence* on every test run, minting an empty cache each time — so
  every test run would pay a full rebuild and regrow the caches this removes, failing
  silently as nothing worse than slow tests. This is the bug class CLAUDE.md records as
  "a path hand-built from `$HOME`".

**Rejected:** deleting the flags from `restart.sh` instead. It converges just as
completely and costs about a third of the build-slot time, and it was the right answer
while the evidence was in doubt. It forfeits the 4.7 GB and discards a warm shared cache.
The spec keeps this as a genuinely close call rather than presenting the chosen design as
obvious.

**Honest scoping:** this is a disk change, not a speed change. It removes the alternation
tax, but so would the rejected alternative. Any cold-build speedup from warm system
modules is unmeasured and is not claimed. For scale, the `.build` trees these caches sit
inside total 28 GB — six times the module caches — so this is a tidy, not a fix for disk
pressure.

## What this PR does

**Shared module cache (`165d7c83`)**

- `scripts/swift-safe` gains the module-cache flags for `build`, `test` and `run`,
  resolving the directory from `pwd.getpwuid(os.getuid()).pw_dir` — never `$HOME`, never
  `Path.home()`.
- Skipped entirely when: the subcommand does not compile; `CI` is set; the caller passes
  `TBD_SWIFT_SHARED_MODULE_CACHE=0`; or the caller already supplies a module-cache flag
  (a second, differently-spelled copy would itself become a third plan variant and
  reintroduce the alternation). `TBD_SWIFT_MODULE_CACHE_PATH` overrides the location.
- A cache that cannot be had reports on stderr and adds no flags rather than failing the
  build — covering both a directory that cannot be created and a home directory that
  cannot be resolved (`getpwuid` raises for a uid with no passwd entry, as minimal
  containers and arbitrary-UID environments produce). The explicit-path override is read
  *before* the passwd lookup, so it remains a working escape hatch there.
- `scripts/restart.sh` and `scripts/update.sh` stop passing the flags. Both build the
  same runtime products through the same wrapper, so either one naming a cache of its
  own is the disagreement this removes; the invariant test sweeps every caller rather
  than `restart.sh` alone, so a third decider fails the harness.
- `docs/reclaim-build.md` rewritten. It previously claimed flag alternation was free and
  that the path was benignly sticky; both are false, and the measurements are in the spec.

**Unrelated bug fix (`4a59fb16`) — restart.sh shipped binaries it had not built**

`scripts/restart.sh` sets `set -e` but piped the build to `tail -3` without `pipefail`, so
the pipeline's status was `tail`'s and the real one was discarded. When the wrapper timed
out waiting for the machine-global build slot and exited 75 — *nothing was compiled* —
`restart.sh` read success, assembled the bundle, and relaunched the daemon and app from
whatever stale binaries were already present, reporting `RESTART_EXIT=0`. That is this
repo's documented "stale daemon reverts the fleet" failure, machine-wide, silently.

The build now runs through `scripts/restart-build-lib.sh`, which captures the real status,
keeps the output trimming, and refuses to assemble or relaunch anything when the build did
not run. 75 and 76 are reported as "the slot was not obtained, nothing was compiled,
nothing was shipped, retryable"; other non-zero statuses as a build failure. No automatic
retry — a silent 30-minute retry is worse than a clear failure. `--skip-build` semantics
are unchanged.

## CI

`swift-safe` adds no flags when `CI` is set, so CI keeps exactly today's behavior. Runners
are ephemeral, so a shared cache buys nothing there, and `test.yml` ends its job with a
bare SwiftPM build that does not route through the wrapper — flagging only the wrapped
steps would make every CI job alternate against itself and poison the cached `.build`.
Both branches of that condition are tested.

## Rebased onto `main` (2026-09-07)

The branch was 84 commits behind and `CONFLICTING`. Two files conflicted, both resolved
by keeping `main`'s newer structure and re-applying this branch's intent on top:

- **`scripts/restart.sh`.** `main` rewrote the build block since the merge base: it now
  builds each product in `RUNTIME_PRODUCTS` separately (#811), takes `-c "$build_config"`
  for `--release` (#621), stamps the build identity before compiling and delegates bundle
  assembly to `restart-bundle-lib.sh` (#805). All of that is preserved. `main` had also
  fixed the status-through-a-pipe bug independently and inline; the loop now calls
  `run_governed_build` instead, which keeps `main`'s per-product structure while restoring
  this branch's 75/76-versus-compile-failure vocabulary. It propagates the wrapper's real
  status rather than flattening it to 1, so "nothing was compiled, retry later" stays
  distinguishable from "the compiler said no".
- **`.github/workflows/test.yml`.** A line-adjacent conflict in the ubuntu
  guardrail-and-harness job only. Both sides kept; the macOS test job — its fast parallel
  passes, count floors and `--skip` regexes (#817) — is untouched by this PR.

The static assertions in `restart-build-lib.test.sh` and `reclaim-build.test.sh` that
pinned the old one-line invocation were rewritten against the new multi-line call and
mutation-checked: each one goes red when the thing it guards is removed.

`scripts/update.sh` did not exist at the merge base and had duplicated the cache decision
— from `$HOME`, not the passwd database. It is converged here for the same reason
`restart.sh` is.

Harnesses re-run on the rebased head, all green: `test-swift-safe.py`,
`restart-build-lib.test.sh`, `reclaim-build.test.sh`, `restart-bundle-lib.test.sh`,
`restart-guard-lib.test.sh`, `update.test.sh`, `test.test.sh`. shellcheck reports one
pre-existing warning, unchanged.

## Assumptions

- Assumes `pwd.getpwuid(os.getuid()).pw_dir` is the real home even when `HOME` is fenced.
  If a future harness fences the passwd database too, the cache silently relocates and
  every build becomes a full rebuild — `test_the_shared_module_cache_is_selected_only_by_the_wrapper`
  guards the code path but cannot see the environment.
- Assumes clang's own pruning keeps the shared cache bounded. `modules.timestamp` is
  present and the defaults prune weekly, evicting entries unused for 31 days; steady state
  is about 1.5 GB. If pruning stops, it grows without a reclaimer — it sits outside every
  `.build`, so `reclaim-build.sh` never sweeps it.

## Rollout

Converging costs a one-time full recompile in each worktree not already on the shared
path — six of the eight with a plan at time of writing; two already carry it and keep
their plan. Those rebuilds land against an already-warm cache and are spread across
whenever each tree next builds. Not gated behind a flag: a default-off flag would deliver
nothing since the flip is the entire change, and the shipped opt-out plus the one-line
revert cover the same ground. The spec argues this explicitly rather than skipping the
convention.

## Evidence & verification

**A test per branch**, red-then-green verified by reverting each production change
individually and confirming the named test goes red — path from `$HOME`, `CI` check
deleted, `CI` tested by presence rather than value, opt-out deleted, dedup always-true,
dedup scanning a program's own args, non-compile exclusion deleted, explicit-path override
ignored, `mkdir` deleted, `mkdir` failure made fatal, `run` appending after the executable
name, and flags never appended. `scripts/test-swift-safe.py`: **90 tests, exit 0** (was 77).

One fixture change is load-bearing: the harness now clears `CI`, because it runs inside
GitHub Actions where `CI=true` is exported — which would have silently switched every new
case onto the no-flags branch and passed green while testing nothing.

**The real build plan**, verified on a throwaway 3-target package (2 Swift, 1 C) built
through the modified wrapper under a private lock, so no machine-global build slot was
taken. A cold build with `.build` deleted produced **no local `ModuleCache` at all**, and
the shared cache did not grow — pure reuse of warm modules. The `-Xcc` half provably
reaches the C target: the shared path appears on the C node's compile command. With `CI=1`
and with the opt-out set, the plan reverts to the per-worktree path.

Every build log was checked for the queue-wait line before being trusted: a build that sat
in the queue reports zero errors having compiled nothing, which is the same shape as the
`restart.sh` bug above. One intermediate verification here *was* a false positive for
exactly that reason — a 0.23 s "Build complete" that compiled nothing — and was redone as
a genuine cold build.

`scripts/reclaim-build.test.sh` 120 ok / 0 fail; `scripts/restart-build-lib.test.sh` all
passed, including the assertion that a non-zero build status exits `restart.sh`.
shellcheck reports only two findings, both pre-existing and identical on `HEAD`.

**Not verified:** nothing was compiled at TBD scale. Local resources are constrained and
the machine-global slot was deliberately not taken, so the compile verdict for this repo
comes from this PR's CI run rather than from a local build.

[20260826-foreign-fish](https://cheapsteak.github.io/tbd/open/?worktree=DF792900-B763-4F2D-8A4E-7596563700DF)

## Summary by CodeRabbit

- **New Features**
  - Added shared Swift and C module caching for supported build, test, and run commands, with configuration overrides and safe fallbacks.
  - Improved build status handling so failed builds stop packaging and installation while providing clearer diagnostics.
  - Preserved quick and skip-build workflows.

- **Bug Fixes**
  - Improved behavior in environments without a resolvable user home directory.

- **Documentation**
  - Added detailed guidance and design documentation for shared module caching.

- **Tests**
  - Expanded coverage for cache selection, build failures, status propagation, and restart integration.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -193,6 +193,7 @@ jobs:
           bash scripts/sweep-preflight-refs.test.sh
           bash scripts/nightly-flake-stress.test.sh
           bash scripts/restart-guard-lib.test.sh
+          bash scripts/restart-build-lib.test.sh
           python3 scripts/diag/daemon-footprint-watch.py --self-test
 
   review-scripts-tests:
--- a/docs/reclaim-build.md
+++ b/docs/reclaim-build.md
@@ -138,6 +138,9 @@ governor.
   machine owner's consent — e.g. `8` on a 12-core laptop. A `-j`/`--jobs` on
   the command line may lower it but not raise it; raise this instead.
 - `TBD_SWIFT_LOCK_TIMEOUT_SECONDS` sets the wait timeout (default `1800`).
+- `TBD_SWIFT_SHARED_MODULE_CACHE=0` restores SwiftPM's per-worktree module
+  cache, and `TBD_SWIFT_MODULE_CACHE_PATH` moves the shared one. See
+  "Shared module cache" below.
 - `TBD_SWIFT_LOCK_PATH` overrides the shared lock path for isolated testing.
 - `TBD_SWIFT_ALLOW_ORPHAN=1` keeps a queued build waiting even after the
   process that requested it exits. A queued wrapper otherwise records its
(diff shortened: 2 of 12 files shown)


cheapsteak:
Addressed two of CodeRabbit's three findings in `ed506eea`, and skipped the third:

- **Unwritable existing cache directory** — valid, and it contradicted this PR's own stated contract that an unavailable cache is a slower build and never a failed one. `mkdir(parents=True, exist_ok=True)` succeeds on a directory that already exists whatever its mode, so the wrapper would name a cache the compiler could not write into. It now checks writability after creating, and degrades exactly as an uncreatable directory does. Both directions are asserted: a writable directory still gets the flags, an unwritable one gets none.
- **Ambient `TBD_SWIFT_MODULE_CACHE_PATH` in the default-resolution case** — valid in kind. `in_process_environment` now scrubs the whole `TBD_SWIFT_` prefix rather than the opt-out alone, matching what its subprocess twin `runner_env` already did. The two halves of that helper disagreeing is the exact shape of the `CI=true` leak this harness was already burned by once.
- **The skip-build form of `run`** — skipped, no change. Skipping the build means SwiftPM writes no build plan, so the appended flags are parsed and unused: there is no plan for them to vary and no recompile to provoke. Excluding it would add a branch with no observable behavior behind it.

Note for the rebase: the `test` job failure on the previous head was `TerminalTeardownReapTests` with `ENXIO` / "produced no pid for probe child" — the known pty-exhaustion flake on the runner. This PR changes **zero** Swift files, so a Swift test failure cannot originate in this diff; the suite carries purpose-built `kern.tty.ptmx_max` / `/dev/ttys*` diagnostics for exactly that condition, and they fired.

PR #768, opened 2026-08-31
Feat: Add the pty holder binary

cheapsteak:
## Summary

Task 5 of the pty-holder transport ([spec](../blob/main/docs/specs/2026-08-30-pty-holder-session-transport-design.md)). The supervisor process itself: it `forkpty()`s the job, owns the pty master for the session's life, and hands it to whoever should be reading.

**Stacked on #767 → #766 → #765.** Merge in order.

## Why this shape

**The holder never `read()`s the master.** A byte the holder consumed is a byte no reader can ever see again, and there is no way to put it back. The only operations it performs on that descriptor are `dup` (hand-over), `ioctl` (size), `write` (input) and `close` (forget).

**It is deliberately single-threaded.** `forkpty` is a `fork`, and a `fork` in a multithreaded process gives the child one thread plus whatever locks the others happened to hold — which is how a child deadlocks inside `malloc` before reaching `execve`. Reaping is a `WNOHANG` `waitpid` inside the same `poll` loop that serves clients: no thread means no window in which one could exist.

## What this PR does

Startup order, each step load-bearing: `setsid` → ignore SIGHUP/SIGPIPE → adopt the inherited lock descriptor (passed by number, never acquired here) → bind and listen → `forkpty` → serve. In the child, between fork and exec: reset SIGHUP/SIGPIPE to `SIG_DFL`, then close every descriptor above stdio.

## Evidence & verification

22 tests. Mutation-checked, each mutation reddening only its own test:

- Delete the child-side fd-closing block → `theJobDoesNotInheritTheCreationLock` and `theJobDoesNotInheritStrayDescriptors`.
- Delete only the sweep → `theJobDoesNotInheritStrayDescriptors`.
- Delete the two `SIG_DFL` resets → `theJobDoesNotStartWithSIGPIPEIgnored`.

Eight consecutive green runs of all three holder suites; `swiftlint --strict` clean across 901 files; zero leaked holder processes and no `/tmp/tbdh-*` residue after runs.

## Three findings that constrain later tasks

**1. A job's exit blocks on an undrained pty.** The job is the pty's session leader, and XNU's `proc_exit` calls `ttywait` before revoking the terminal, so it sits in `P_WEXIT` until the tty output queue is empty — four bytes of canonical echo are enough. `waitpid(WNOHANG)` correctly reports nothing, so the holder never sees a status. Since the holder must never be the drainer, **the daemon-side reader's drain is a liveness requirement, not a convenience**: if it stops draining, jobs cannot finish dying. This was measured by driving the built binary by hand, not inferred.

**2. A frame reader must queue what it decodes.** One `recvmsg` can carry a response and the unsolicited exit push together; returning the first and dropping the rest makes the next read hit EOF and report `peerClosed` for a report that did arrive. Task 6's `HolderClient` needs the same pending queue.

**3. Anything a spawner leaks reaches the job.** Measured: before the fd sweep existed, a `sleep` job inherited `scripts/swift-safe`'s machine-global `flock` and blocked every other worktree's build after its holder, its test process, and its harness were all gone. That is why the child closes everything above stdio rather than the two descriptors it knows about.

## Assumptions

`--owner <token>` was added: the documented flags carried no source for `HolderChildDescription.owner`. It defaults to an empty token so a hand-run holder still starts; Task 6 must pass it.

The unsolicited exit push reuses `.described` and is guarded on `status != .alive`. Without that guard it fires on every accept, and the stray frame answers the client's first request — so `handOverPTY` silently returns a plain `.described` with no descriptor. That bug was live until the hand-over test caught it.

### DIFF
--- a/Sources/TBDHolder/Holder.swift
+++ b/Sources/TBDHolder/Holder.swift
@@ -1,0 +1,736 @@
+import Darwin
+import Foundation
+import TBDShared
+import os
+
+/// One holder process owns one session's pty master for that session's whole
+/// life, and hands `dup`s of it out over a Unix socket.
+///
+/// The single invariant everything else rests on: **the holder never `read()`s
+/// the master.** The only operations it performs on that descriptor are `dup`
+/// (hand-over), `ioctl` (size), `write` (input) and `close` (forget). A byte
+/// the holder consumed is a byte no reader can ever see again, and there is no
+/// way to put it back.
+///
+/// The holder is deliberately single-threaded. `forkpty` is a `fork`, and a
+/// `fork` in a multithreaded process gives the child one thread plus whatever
+/// locks the others happened to hold — which is how a child deadlocks inside
+/// `malloc` before it ever reaches `execve`. Reaping is therefore a `WNOHANG`
+/// `waitpid` inside the same `poll` loop that serves clients, not a thread
+/// blocked in `waitpid`: no thread means no window in which one could exist.
+final class Holder {
+    private static let logger = Logger(subsystem: "com.tbd.daemon", category: "holder")
+
+    /// How long a holder whose child has exited keeps its socket bound waiting
+    /// for somebody to come and collect the status. Long enough for a daemon
+    /// that is restarting to reconnect, short enough that nothing lingers.
+    static let defaultExitReportTimeout: Duration = .seconds(10)
+
+    private let arguments: HolderArguments
+    private let exitReportTimeout: Duration
+    private let clock: any Clock<Duration>
+
+    init(
+        arguments: HolderArguments,
+        exitReportTimeout: Duration = Holder.defaultExitReportTimeout,
+        clock: any Clock<Duration> = ContinuousClock()
+    ) {
+        self.arguments = arguments
+        self.exitReportTimeout = exitReportTimeout
+        self.clock = clock
+    }
+
+    /// Runs the holder to completion and returns the process exit code.
+    ///
+    /// Ordering here is the contract, not a preference:
+    ///
+    ///   1. `setsid()` first, so a Ctrl-C aimed at whatever spawned us cannot
+    ///      reach the holder and strand every session on the machine.
+    ///   2. `SIGHUP`/`SIGPIPE` to `SIG_IGN`, so the spawner's death — or a
+    ///      `sendmsg` into a peer that just went away — is not ours.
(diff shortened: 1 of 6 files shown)


cheapsteak:
Pushed a fix for the HIGH finding. Root cause confirmed by mutation rather than only traced: re-adding the accept-time push reproduces it exactly and reddens only the three new tests.

Two halves had to change, because widening the exit-report guard alone would not have been enough — the loop-exit condition is the other half:

- `exitReported` became `exitCollected`, redefined as *the terminal status actually reached a client*, set only on a successful send.
- **The accept path now sends nothing.** The contract is that the holder never speaks first to a client arriving after the child is gone. The unsolicited push happens only from the reaping branch, for a client that was mid-session when the child died — which is what `reportsTheJobsExitCodeToAConnectedClient` still relies on.
- `describe` and `handOverPTY` mark collection when their answer carried a terminal status, so the wind-down is driven by a real answer to a real request.

**`handOverPTY` after exit still hands the master over**, deliberately. The holder never read it, so everything the job wrote and nobody collected is still queued and would be destroyed when the holder closes it — and the description in the same frame carries the terminal status, so no reader can mistake the child for alive.

Three regression tests, all of which fail without the fix and none of which existed before because every prior hand-over test connects while the child is alive: `handsOverThePTYToAClientThatArrivesAfterTheChildExited`, `describesAnAlreadyExitedChildOnlyWhenAsked` (a new helper distinguishes *silence* from a torn-down connection, so an EOF fails rather than reading as quiet), and `forgetsForAClientThatArrivesAfterTheChildExited`.

**One consequence worth recording:** a post-exit client gets one round trip, not a session. The holder winds down as soon as its answer carried the terminal status, so `describe` followed by a separately-written `forget` loses the second request. That was chosen over serving until the client closes, which would leave every holder lingering for the full exit-report window after a normal session ends. `forget` as a first request works, and a holder that exits and unlinks its rendezvous reaches the same end state `forget` guarantees.

**MEDIUM:** exit codes now separate the cases the comment claimed they did — 2 bad invocation, 3 environment, 1 unclassified, switched exhaustively with no `default` so a new error cannot be added without deciding which side of the retry question it falls on. **MINOR:** the exit-report timeout comment now states both directions of the trade-off and what expiry costs.

43 tests green across five runs by the implementer and three more independently; `swiftlint --strict` clean; no leaked holder processes.

PR #769, opened 2026-08-31
Feat: Add the daemon-side holder client and spawner

cheapsteak:
## Summary

Task 6 of the pty-holder transport ([spec](../blob/main/docs/specs/2026-08-30-pty-holder-session-transport-design.md)). The daemon's half of the holder protocol: a client that speaks it, and a spawner that brings holders into existence safely.

**Stacked on #768 → #767 → #766 → #765.** Merge in order.

## Why this shape

`HolderClient` is an `actor` so the single-reader discipline on the socket is enforced by the type system rather than by convention.

**It keeps a pending-message queue**, which is not incidental: one `recvmsg` can carry a response *and* the holder's unsolicited exit push together. Returning the first and dropping the tail makes the next read hit EOF and report a closed peer for a message that actually arrived. That was a real load-dependent flake in the holder's own harness before it was understood.

The spawn order is the contract. The lock is taken first; a socket present with **no** lock file is handshaked before anything is unlinked, because absence of a lock is not evidence of absence of a holder. Only a connect failing `ECONNREFUSED`/`ENOENT` licenses clearing the path.

## What this PR does

- `HolderClient` — `describe`, `handOverPTY`, `forget`, with the pending queue.
- `HolderSpawner` / `HolderHandle` — the seven-step spawn, bounded on an injected clock.
- `HolderFraming` hoisted from `TBDHolder` into `TBDShared` so both sides share one implementation rather than two that can drift.
- `Package.swift`: `TBDDaemonTests` gains its `TBDHolder` dependency, which no earlier task had actually added.

## Assumptions

Three decisions taken beyond the plan, each recorded in the code:

- **`POSIX_SPAWN_CLOEXEC_DEFAULT`**, and the lock relocated with `fcntl(F_DUPFD_CLOEXEC, 10)` rather than `makeInheritableAcrossExec()`. Clearing `FD_CLOEXEC` on the daemon's own descriptor would expose it to every other concurrent spawn in the daemon, leaving a session UUID unreclaimable until an unrelated child dies. This also sidesteps the `dup2(fd, fd)` trap unconditionally.
- **A holder that fails to bind is killed and reaped.** An unbound-but-live holder holds the creation lock forever. Safe because the holder binds *before* it `forkpty`s, so there is no job to orphan.
- **`describe()` is the handshake** — the holder does no version exchange on connect, and `.rejected(version: busySentinel)` is how a second client learns it is second.

## Known gap, deliberately not closed here

**Nothing reaps a holder that exits while the daemon lives.** The daemon `posix_spawn`s it directly, so it becomes a zombie; only the never-bound path reaps today. The general case belongs to Milestone B's `AgentReaper` leg, and it is documented in `HolderSpawner`'s doc comment rather than left to be discovered.

## Evidence & verification

38 tests across 6 suites, green on five consecutive runs (independently re-run four more times before opening this). Mutation-checked:

- `unlink(socketPath)` before throwing `.lockHeldByLiveHolder` → only `refusesToClearASocketWhoseLockIsHeld` reddens, on both legs (file gone, bytes nil).
- Truncating the client's pending queue to one message → only `queuesAResponseAndAnExitPushDeliveredInOneRead` reddens, with a receive timeout — it went back to the wire for a message it had already been handed.

Full suite: 4 failures, all pre-existing. Proved rather than assumed by holding out the new test file — the same run gives the identical four failures (`TerminalLockedAccessTests` ×3 time-limits, one replay test whose own message reads "harness/load, not production"). `swiftlint --strict` clean.

## Also fixes a Task 4 flake

`HolderLockTests.lockIsReacquirableAfterTheHoldingProcessExits` (#767) closed its duplicated lock descriptor in a *trailing* `defer` — after the reacquire. `flock` lives on the open file description and a `dup` shares one, so the parent still held the child's lock and the reacquire threw `.alreadyHeld` against a claim userspace never dropped. It only fires when the lock lands on fd 9, which depends on how many descriptors sibling suites hold open — and this PR's new tests changed that landscape enough to expose it. The duplicate is now closed at the point the child becomes sole holder, and `dup` became `fcntl(F_DUPFD_CLOEXEC, 10)`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -240,6 +240,12 @@ let package = Package(
                 "TBDDaemonLib",
                 "TBDShared",
                 "TestSupport",
+                // Load-bearing for the same reason `TBDPeerHelperTests` depends
+                // on `TBDPeerHelper` below: `Holder/HolderClientTests` spawns
+                // the real `TBDHolder` binary through the real `HolderSpawner`,
+                // so the product has to be built and sit in the same products
+                // directory as the test bundle. Nothing here imports it.
+                "TBDHolder",
                 .product(name: "GRDB", package: "GRDB.swift"),
                 .product(name: "Clocks", package: "swift-clocks"),
             ]
--- a/Sources/TBDDaemon/Holder/HolderClient.swift
+++ b/Sources/TBDDaemon/Holder/HolderClient.swift
@@ -70,17 +70,34 @@ actor HolderClient {
     private var inbox = Data()
     /// Frames decoded but not yet returned to a caller.
     ///
-    /// **This queue is the point.** One `recvmsg` routinely carries a response
-    /// and the holder's unsolicited exit push together — the holder answers a
-    /// request and reports a status microseconds apart, and the kernel
-    /// coalesces them. A client that returned the first frame and dropped the
-    /// tail would then read EOF on its next call, because the holder closes
-    /// right after reporting an exit, and would report a closed peer for a
-    /// message that did arrive. That was a real load-dependent flake in the
-    /// holder's own harness, not a hypothetical.
+    /// **The queue exists because one `recvmsg` routinely carries two frames.**
+    /// The holder answers a request and reports a status microseconds apart,
+    /// and the kernel coalesces them; a client that decoded the first and threw
+    /// the rest of the buffer away would lose a message that did arrive, then
+    /// read EOF looking for it, because the holder closes right after reporting
+    /// an exit. That was a real load-dependent flake in the holder's own
+    /// harness, not a hypothetical.
+    ///
+    /// What the queue must never do is hand one request's tail to the *next*
+    /// request as its answer. Whatever is in it when a request goes out
+    /// therefore arrived before that request existed and is retired as an
+    /// unsolicited push rather than served — see `raiseBarrier`.
     private var pending: [HolderResponse] = []
     /// Descriptors received but not yet handed to the frame they rode with.
     private var carriedFDs: [Int32] = []
+    /// The most recent description the holder sent **without being asked**.
+    ///
+    /// The holder pushes a `.described` frame carrying the terminal status when
+    /// the child exits while a client is connected, and the kernel routinely
+    /// delivers that push in the same read as the answer it trails. Such a
+    /// frame answers nobody's request, so it is retired from the queue rather
+    /// than handed to whoever asks next — and kept here, because it is the one
+    /// place a child's exit status arrives unsolicited and throwing it away
+    /// silently would lose the fact entirely.
+    ///
+    /// Deliberately survives `close()`: it is an observation about the child,
+    /// not connection state.
+    private(set) var lastPushedDescription: HolderChildDescription?
 
     init(socketPath: String, receiveTimeout: Duration = HolderClient.defaultReceiveTimeout) {
         self.socketPath = socketPath
@@ -101,7 +118,7 @@ actor HolderClient {
     func describe() async throws -> HolderChildDescription {
         try connectIfNeeded()
         try send(.describe)
-        let (response, fds) = try receive()
+        let (response, fds) = try receive(answering: .describe)
         closeAll(fds)
         switch response {
         case .described(let description), .handedOverPTY(let description):
@@ -109,7 +126,7 @@ actor HolderClient {
         case .rejected(let version):
             throw Error.rejected(version: version)
         case .forgotten:
-            throw Error.unexpectedResponse("forgotten")
+            throw unexpected(response)
         }
     }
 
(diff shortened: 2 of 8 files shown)


cheapsteak:
All three findings fixed in `05dd8bbb`. Each was real; the two HIGHs in particular.

**No request/response correlation.** Confirmed by tracing `Holder.serve()`: `pushExitToConnectedClient` writes only from the reaping branch, at an already-connected client, so a coalesced push always *trails* the answer it rode with — meaning the leftover is always misattributed to the **next** verb. `handOverPTY` then `forget` dequeued the stale `.described(exited)` as the `forget` answer: `unexpectedResponse` for a verb the holder had performed correctly, with the real `.forgotten` left on the wire for whoever asked next.

One thing the finding didn't say, and which changed the fix: **a queue-only barrier is insufficient.** The push can be on the wire but not yet decoded — written a moment after the answer was read — so `send()` now raises a barrier over both the decoded queue *and* the socket receive buffer. Retired frames are recognised as unsolicited pushes and surfaced as `lastPushedDescription` rather than served to another verb or dropped silently, and their stray descriptors are closed. `receive(answering:)` shape-checks and drops the connection on a mismatch, so a desync cannot outlive the call. This also caught a latent `pending = …` assignment that would have clobbered queued frames.

**Bind-timeout kill.** You were right that the justification described something the code never checked, and right about why it matters: `bindListener` → `forkpty` → `serve()`, with `accept()` only inside `serve()`, so a pending connect at expiry implies the fork almost certainly ran. The kill is now gated on `fileExists(socketPath)` — no socket means never bound, therefore never forked, therefore safe. A holder that bound but did not answer is **left running** and reported via a new `holderBoundButUnresponsive(holderPID:socketPath:diagnostics:)` so the situation stays recoverable instead of having its evidence destroyed. The old code also killed on `.rejected` — a holder that had *answered*; that is now impossible.

**Untested headline contract.** Correct, and my PR description overclaimed it as mutation-checked — the test threw from `.alreadyHeld` and returned before the probe branch. All three of that branch's cases are now covered: a live holder answering after its lock file was swept (back off), `ECONNREFUSED` (licenses clearing), and `ENOTSOCK` (must not clear — empirically confirmed as Darwin's errno for a regular file at that path).

Four mutation checks, each reddening only its own test: removing the barrier; forcing `socketExists = false`; deleting the listener guard; and never licensing clearing.

**A contract change for Milestone B:** a coalesced exit push is no longer delivered as the answer to a later `describe()`. The reader must read `lastPushedDescription` or reconnect. The old test had encoded the misattribution as the contract.

49 tests green across five runs by the implementer and three more independently; `swiftlint --strict` clean; no leaked holder processes.

cheapsteak:
Covered in `6b97d623` — tests only, `Sources/` byte-identical to `05dd8bbb`.

New test `drainsAnExitPushStillBufferedInTheSocketBeforeTheNextRequest`. The peer's trailer write is gated behind a semaphore rather than a sleep: the peer writes the `handOverPTY` answer and then blocks; the test asserts `lastPushedDescription == nil` (proof the push is not yet even written), releases the gate, waits for the peer to signal it has written, and only then calls `forget()`. Since `HolderClient` reads the socket only from inside a verb, nothing drains it in that interval — the push is provably in the kernel receive buffer, undecoded, when `forget`'s `raiseBarrier()` runs. Both waits are deadline-bounded and the peer's wait also polls the stub's stopped flag, so a failing test cannot strand the thread.

**Your premise was slightly optimistic, and the real picture is worse than the finding.** The existing coalesced test's two writes mostly land in one `recvmsg`, but that is a scheduling race, not a guarantee — it failed 2 of 7 runs under the isolating mutation. A standalone socketpair probe (reader parked in `recvmsg`, two back-to-back writes, 20 trials) measured plain writes coalescing 20/20, and `sendmsg` with `SCM_RIGHTS` followed by a plain write 19/20. `FDChannel.sendFDMinimal` splits the fd-carrying answer into three writes, so that test has several boundaries where the client's read can land early.

So half 2 was not uncovered — it was covered **nondeterministically**, roughly a third of the time under load. That is arguably worse than uncovered: a regression would have surfaced as an intermittent red that a rerun clears, which is exactly the kind of signal that gets dismissed. The new test makes it deterministic (7/7 under the mutation) and leaves the existing test owning half 1.

Isolating mutations, both reverted:
- Neuter only the `hasBufferedInput()` poll loop → new test fails **7/7**; `keepsAnExitPushCoalescedWithAnAnswerInsteadOfLosingIt` passes 7/7; the older coalesced test passes 5/7 (the nondeterminism above).
- Neuter only `retireQueuedFrames()` → both coalesced tests fail 3/3.

50 tests green across five runs by the implementer and three more independently; `swiftlint --strict` clean across 904 files; no leaked holder processes.

PR #773, opened 2026-08-31
Feat: Drain holder masters into a headless emulator in the daemon

cheapsteak:
## Summary

Task 7 of the pty-holder transport ([spec](../blob/main/docs/specs/2026-08-30-pty-holder-session-transport-design.md)). The daemon's `HolderReader`: a headless SwiftTerm `Terminal` fed by a dedicated drain thread, so unattended sessions never stall. Plus two transport fixes and one repo-wide test fix found while building it.

**Stacked on #769 → #768 → #767 → #766 → #765.** Merge in order.

## Why this shape

**Draining is a liveness requirement, not a convenience** — and that reframing came from measurement, not theory. A job is its pty's session leader, and the kernel's process teardown waits on the terminal's output queue before revoking it, so a job that exits with anything unread stops half-exited and `waitpid` correctly reports nothing. A few bytes of ordinary echo are enough. Since the holder must never be the drainer, a reader that stops reading is not falling behind — it is holding jobs open. The drain loop therefore does not exit early on transient errors and is never gated on anyone wanting the output.

`Terminal` is not `Sendable` and carries no `@MainActor`, so every `feed` and every grid read holds `terminalLock`. Getting that wrong garbles output rather than crashing, which means it fails late and gets misdiagnosed as an escape-sequence bug.

## What this PR does

- `HolderReader` — drain thread, headless emulator, `renderScreen`, scrollback, `write`, `resize` (which also drives `TIOCSWINSZ`, so the child actually gets `SIGWINCH`).
- **The post-spawn `.rejected` race, fixed at the spawner.** `spawn` now returns the connection its handshake ran on, still open, so no reconnect exists and the busy window cannot be entered. The alternative — a bounded retry in the client — was rejected because it makes every caller pay for a condition transient in one window, and erodes `.rejected`, which is the genuine "somebody else is attached" signal protecting against two readers on one pty. A new test asserts a real second client is still refused.
- **A pushed exit is observable without `close()`.** `raiseBarrier` decoded the push, then threw `peerClosed` on the EOF behind it before retiring — so a poller doing `describe()` + `lastPushedDescription` never saw the status. Retire is now a `defer`.
- **A busy holder no longer reports as a broken pipe.** `.rejected` is written unsolicited on accept-then-hang-up, so it routinely landed *before* the first request, where the barrier discarded it as noise; which of three errors you saw was decided by microseconds. The refusal is now recorded and outranks whatever else went wrong.
- `Package.swift`: `TBDDaemonLib` links SwiftTerm. This pulls an AppKit-importing target into the daemon binary — SwiftTerm is a single target with no core/views split, so there is no headless-only product. The daemon never touches the view types.

## The repo-wide fix, which is not about holders at all

`Tests/TBDPeerHelperTests/SpawnedPeerHelper.swift` used a bare `Process.waitUntilExit()` in `tearDown()`. It blocks until Foundation *observes* the exit and has no deadline, so a SIGKILL it never observes parks the caller forever — and because `tearDown()` runs from a `defer` in an async test, the parked thread is one of Swift Testing's cooperative-pool threads, so **the entire test process cannot finish**.

Measured: a full-suite run sat at **0.0% CPU for 24 minutes holding the machine-global build lock**, with two of these frames the only live test threads. This wedged this machine three separate times today and blocked every other worktree behind it. The file is byte-identical on `main` and this branch touches no other peer-helper file — it is pre-existing, and every other wait in that fixture was already bounded.

Now a bounded poll against the fixture's own `waitLimit`. Before: stalled, sampled, killed. After: 9,057 tests in 881 suites in 276 s.

## Evidence & verification

84 tests green across eight consecutive runs by the implementer plus four more independently; peer-helper suite 22/22; `swiftlint --strict` clean; zero leaked holder, `swift-test`, or helper processes.

Mutation checks, all run:
- Revert the spawn hand-off → 4 of 16 runs red with `rejected(version: -1)` across five different tests.
- Revert the barrier `defer` → `surfacesAPushedExitWithoutWaitingForClose` red 3/3.
- Re-discard the refusal → `refusesASecondClientWhileTheSpawnersConnectionIsHeld` red 3/3 with the broken-pipe error.
- Treat a short read as EOF in the drain loop → exactly the two wedge tests red, nothing else.

## Known gaps, deliberately not closed here

- `HolderClient.connectIfNeeded` uses a **bare blocking `Darwin.connect`** — the same class of defect as the wedge above. Pre-existing and untouched, but it should be bounded before Task 10 puts it on the session-creation path.
- After EOF the drain thread parks on its wake pipe rather than exiting, keeping the fd close exclusively on the stop path so `write()` cannot race a reused descriptor. The cost is one parked thread per finished session until someone calls `stop()`; Task 10 and Milestone B own that call.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -113,6 +113,16 @@ let package = Package(
                 .product(name: "NIO", package: "swift-nio"),
                 .product(name: "NIOPosix", package: "swift-nio"),
                 .product(name: "NIOHTTP1", package: "swift-nio"),
+                // The daemon is the default reader for a holder-backed session:
+                // it drains that session's pty master into a headless SwiftTerm
+                // `Terminal` so an unattended job never wedges behind a full tty
+                // queue (Sources/TBDDaemon/Holder/HolderReader.swift).
+                //
+                // This pulls an AppKit-importing target into the daemon binary.
+                // SwiftTerm ships one target with no core/views split, so there
+                // is no headless-only product to depend on instead; the daemon
+                // simply never touches the view types.
+                .product(name: "SwiftTerm", package: "SwiftTerm"),
             ],
             path: "Sources/TBDDaemon",
             exclude: ["main.swift"],
--- a/Sources/TBDDaemon/Holder/HolderClient.swift
+++ b/Sources/TBDDaemon/Holder/HolderClient.swift
@@ -64,7 +64,11 @@ actor HolderClient {
     }
 
     let socketPath: String
-    private let receiveTimeout: Duration
+    /// Mutable because a connection can outlive the budget it was made under:
+    /// `HolderSpawner` handshakes on a short handshake timeout and then hands
+    /// the same connection on as a session's long-lived one. See
+    /// `adoptReceiveTimeout`.
+    private var receiveTimeout: Duration
     private var fd: Int32 = -1
     /// Bytes read but not yet parsed into a whole frame.
     private var inbox = Data()
@@ -98,6 +102,21 @@ actor HolderClient {
     /// Deliberately survives `close()`: it is an observation about the child,
     /// not connection state.
     private(set) var lastPushedDescription: HolderChildDescription?
+    /// The busy sentinel this connection was refused with, if it has arrived.
+    ///
+    /// **`.rejected` is the other frame the holder writes without being asked**,
+    /// and unlike an exit push it is not noise: the holder writes it the moment
+    /// it accepts a connection it cannot serve, and then hangs up. So it
+    /// routinely lands *before* the first request is even written, where the
+    /// barrier finds it — and a barrier that discarded it left the following
+    /// write to fail with `EPIPE`, reporting a perfectly ordinary busy holder
+    /// as `transportFailed("Broken pipe")`. Which of the two a caller saw was
+    /// decided by microseconds.
+    ///
+    /// Kept, so the verb that was about to go out fails with the refusal the
+    /// holder actually gave. Cleared by `close()`: it is a fact about one
+    /// connection, and the next one gets its own answer.
+    private var refusal: Int?
 
     init(socketPath: String, receiveTimeout: Duration = HolderClient.defaultReceiveTimeout) {
         self.socketPath = socketPath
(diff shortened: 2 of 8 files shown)


cheapsteak:
Fixed in `341726df`, and I departed from the suggested shape deliberately — worth explaining, because the difference matters.

The finding proposed a bounded retry before giving up. **Giving up puts the descriptor exactly where the old code left it** — permanently unread, job unable to finish exiting — only later. Since draining is a liveness requirement rather than a convenience, what deserves bounding is the *cadence*, not the attempt count.

So the catch-all is split instead. `isPermanentReadFailure` names exactly two errnos as genuinely terminal — `EBADF` (not an open descriptor) and `ENXIO` (device gone) — and those end the drain as before. **Every other errno retries indefinitely** on a doubling backoff (10 ms → 250 ms ceiling), logged once at first failure and once at an 8-failure streak. Four attempts a second forever costs nothing and picks the session up the instant the descriptor recovers.

Two mechanics that keep that from costing responsiveness: `drainEverythingReadable` now returns a `DrainOutcome` (`.readable`/`.exhausted`/`.retryable`) rather than a `Bool`, and the backoff is spent **inside `poll` on the wake pipe alone** — the pty is dropped from the watch set for the pause — so `stop()` is still answered immediately instead of waiting out a sleep.

You were right that the branch had no coverage. It now has a seam: `HolderReadFault`, an optional `@Sendable () -> Int32?` consulted before each read, defaulted to `nil` so production never constructs one. The injected errno runs through the same `classify(errno:)` a real one does, so a fault cannot exercise a shape production never produces.

Three tests: recovery after three injected `ENOMEM`s (asserting a marker written *after* the failures reaches the emulator, and that the fault actually fired three times); retrying past any plausible give-up point without spinning (>12 attempts, <400); and stopping at once on `EBADF` (exactly one attempt while the job keeps writing).

Mutation: forcing `isPermanentReadFailure` to `true` reddens the two retry tests and nothing else — `stopsDrainingOnAPermanentReadFailure` correctly still passes, since it asserts the branch the mutation makes universal.

87 tests green across five runs by the implementer and three more independently. Happy to bound the attempt count instead if you disagree with the reasoning.

PR #790, opened 2026-09-02
Fix: A pull request you didn't open is invisible to TBD

cheapsteak:
## What's broken

A pull request opened on a worktree's branch by anyone other than you is invisible to TBD. No chip appears, nothing binds, and auto-archive never fires — the worktree looks like it has no PR at all, forever.

Anyone whose PRs are opened on their behalf hits this on every single PR: a bot or GitHub App that opens the PR so a human can approve it, a teammate who opens it for you, a PR raised through the web UI, or an agent session running `gh pr create` on a remote box where no local daemon sees the hook. Concretely: a worktree whose branch carried two PRs — one you opened and later closed, one a GitHub App opened and merged — showed the closed one indefinitely and never learned the merged one existed.

## Why it happens

TBD has three ways to discover a worktree's PR, and a PR someone else opened misses all three.

The hook that watches for `gh pr create` never fires, because the command that ran was a wrapper driving `gh` on another machine. That gate is deliberately fail-closed and is left alone here — a false bind hands `allResolved` an auto-archive nobody asked for.

Provenance seeding only applies to worktrees created *from* a PR row, which this is not.

That leaves branch matching, and this is the actual bug: it asked GitHub `viewer { pullRequests(first: 100, …) }` — *the authenticated account's own pull requests* — and then matched branch names locally. A PR you didn't author is not in that list, so it is invisible by construction rather than by omission. The 100-PR cap silently lost older PRs on top of that.

The premise behind this shape was written down in the code — that GitHub can only be asked for the viewer's pull requests, and that author-blind branch matching was a GitLab-only capability. That premise is false, and this same file already disproved it twice: `prByBranchQuery` (the single-PR refresh) and `openPRsQuery` (the PR picker) are both repo-scoped and author-blind. The viewer batch was simply the older shape, and the GitLab work fixed the asymmetry on one forge without going back for the other.

## What this PR does

Asks the repo instead of the account. One repo-scoped query per repo per pass, with one aliased field per candidate head branch:

```graphql
b0: pullRequests(headRefName: $b0, first: 10, orderBy: {field: CREATED_AT, direction: DESC}) { … }
```

This is the same question the GitLab path has always asked, so the two forges now read as siblings differing only in how the branch list is carried. Author-blind, no fleet-wide page cap, and `first: 10` newest-first covers a branch reused across a closed-and-reopened pair.

Every branch is a declared `String!` **variable** bound with `-f`, never interpolated into the query text. A git ref may legally contain `"`, and interpolating one produced a malformed query — a regression that had a comment recording it and no test. It has a test now.

**Judgement moved from the fleet to the repo.** The old `batchSucceeded` was a single boolean speaking for every worktree everywhere: one failed call suspended the by-number fallback and the head-ref heals for the entire fleet. It is replaced by `answeredIDs`, mirroring the GitLab side's existing per-project tracking. A repo whose query fails contributes no matches *and* no answered ids, so its worktrees keep their cached status and record `.undetermined` — never "no PR". This also fixes a case the fleet-wide predicate got wrong: a worktree whose repo can't be resolved was read as "no PR", when nobody could ask about it at all.

**Queries are chunked by worktree and fanned out concurrently.** Chunked because `pullRequests` is a non-null connection, so a GraphQL error on one aliased field null-propagates to `repository` and reads as "the forge did not answer" for the whole repo; a cap bounds that blast radius. Chunked *by worktree* specifically, never by branch: `matchUnnumbered` takes a worktree's first matching candidate, so splitting one worktree's candidates across chunks could attach it to its second-choice branch's PR while its own branch's PR sat in the failed chunk — and mis-attachments here reach `onMergedTransition` and auto-archive. Fanned out because the GitHub path went from one fleet-wide call to one per repo against a 30-second foreground poll.

`runGHGraphQL` and `parsePRNodes` had no callers left and are deleted, along with every comment asserting the retired premise.

This is a bug fix — it restores the system to the theory it already held, that branch discovery asks the forge what exists on a branch — so it ships without a spec.

## Assumptions

- **`pullRequests(headRefName:)` matches on head ref name regardless of which repository the head lives in**, so a cross-repository (fork) PR opened into this repo is returned too. Cross-repo nodes are deliberately *not* filtered out: `resolveNameWithOwner` resolves the base repo for a fork checkout on purpose, so for a fork user every PR of theirs is cross-repository and filtering would remove branch discovery for them entirely. The residual is that an outside fork opening a PR whose head ref exactly equals a local worktree's branch name could be matched to it. Accepted because TBD's branches are `tbd/<slug>`-shaped, merge-triggered actions default off, and the identical exposure already exists in `prByBranchQuery` (used by the interactive refresh) and the GitLab `sourceBranches` path — it is a property of asking a forge by head ref, not something new here. A real fix needs the worktree's own head-repository identity, which the poll does not resolve today. Documented as a `KNOWN RESIDUAL` in the code.
- The alias cap assumes GitHub keeps serving a few hundred aliased connection fields inside its request timeout. If that stops holding, the symptom is a repo going `.undetermined` every poll — safe, self-healing, and visible in the log.

## Evidence & verification

**The repro, before the fix.** A worktree on branch `tbd/<slug>` in a repo where that branch carried two PRs: one authored by the viewer (closed) and one authored by a GitHub App (merged). TBD showed only the closed viewer-authored one; the App-authored PR had never been seen. Its stored binding table held only the closed PR.

**Discriminating tests** (all fail without the fix):
- A PR exposed *only* through the repo-scoped branch query is discovered — and the stub asserts no query text ever contained `viewer`, so resolving it while still issuing an author-scoped query cannot pass.
- A branch carrying a closed PR and a newer open one resolves to the open one.
- Per-repo degradation: one repo's query fails while another's succeeds — the failing repo's worktree keeps its cached status and records `.undetermined`, its sibling resolves and records `.observed`. Mirrored on the by-number path: the dark repo is never re-queried by number while the answered repo's unmatched worktree still is.
- A worktree whose repo cannot be named records `.undetermined`, while a worktree whose repo answered with nothing records `.none`.
- A branch name containing `"` is bound as a field value and appears nowhere in the query text; `-F` is never used.
- Chunking splits at the cap and never splits one worktree's candidates; one failed chunk leaves the other chunk's worktrees resolved; chunk queries are in flight together (the probe reads `maxInFlight > 1`, which is exactly 1 under a serial loop).
- `parseBranchPRNodes` separates "answered with nothing" (`[]`) from "did not answer" (`nil`) across four response shapes.

**Suite.** 459 tests in 21 PR suites pass. Full local suite: 9257 tests, all PR suites green; the failures present are wait-timeouts in unrelated PTY/holder/ControlMode/subprocess-timeout tests on a box that was running another worktree's live tests concurrently — CI is the authoritative whole-suite verdict.

**Live verification**, against the real case, with the daemon rebuilt and restarted from this worktree. Before: the worktree's only binding was the closed viewer-authored PR (tombstoned), and its cached status named that PR. After one poll tick, with no manual attach:

```
number = 24926
source = branch            ← discovered by the author-blind matcher
detached = 0
headBranch = tbd/…         ← matched on the worktree's own branch
prStatus = {"state":"merged","reason":"Merged", …}
```

The App-authored PR is bound, on the right branch, with the right state, and precedence picked it over the closed PR on the same head ref. The worktree stayed `active` — the merged transition fired and auto-archive correctly did not, since it defaults off. The closed PR's `detached=1` tombstone was left untouched, so the coordinator's tombstone rule held.

**Cost, measured against the real forge** rather than assumed, since a per-repo query carries far more aliases than the old fleet-wide batch (the largest repo on the test box has 97 active worktrees): 96 aliases returned in 3.96s, 288 aliases in 5.25s at a GraphQL rate-limit cost of 3 points out of 5000/hr.

https://claude.ai/code/session_016dDWV996GHuDakWtiupDjf

[20260902-hissing-giraffe](https://cheapsteak.github.io/tbd/open/?worktree=5A13765D-A97A-47E2-8EDF-FFB3035AC1D8)

## Summary by CodeRabbit

* **Improvements**
  * Pull request discovery now works across repositories and does not depend on the current account’s authored pull requests.
  * Fork-based pull requests are resolved more reliably using stored pull request information.
  * Status polling and refreshes better isolate repository-specific failures while preserving valid cached results.
  * GitLab-only worktrees avoid unnecessary GitHub lookups.
* **Documentation**
  * Updated pull request discovery and fork handling documentation.

### DIFF
--- a/Sources/TBDDaemon/Database/Database.swift
+++ b/Sources/TBDDaemon/Database/Database.swift
@@ -964,8 +964,9 @@ public final class TBDDatabase: Sendable {
 
         // Number of the GitHub PR a worktree was created from, if any. Nullable —
         // NULL means "not created from a PR". Lets PRStatusManager resolve status
-        // by direct number lookup instead of viewer-authored/branch-name matching,
-        // which is the only way fork PRs (no matching local branch) get tracked.
+        // by direct number lookup instead of head-branch matching, which is the
+        // only way fork PRs (whose head ref lives in the fork, so no query
+        // against this repo's head refs reaches it) get tracked.
         migrator.registerMigration("v54_worktree_pr_number") { db in
             try db.addColumnIfMissing(table: "worktree", column: "pr_number", type: .integer)
         }
--- a/Sources/TBDDaemon/PR/PRBindingCoordinator.swift
+++ b/Sources/TBDDaemon/PR/PRBindingCoordinator.swift
@@ -77,9 +77,10 @@ public actor PRBindingCoordinator {
     ///
     /// Without this such a worktree owns a number and no binding, so its PR is
     /// invisible to `tbd pr list`, to the toolbar dropdown and to the status-bar
-    /// chips. Fork PRs are the sharpest case: a fork head never appears in the
-    /// viewer-authored batch, so branch matching is structurally unable to find
-    /// them and the stored number is the only handle that exists.
+    /// chips. Fork PRs are the sharpest case: a fork PR's head ref lives in the
+    /// fork, so no query against this repo's head refs can reach it, branch
+    /// matching is structurally unable to find them, and the stored number is
+    /// the only handle that exists.
     ///
     /// The source is `.manual` — naming a PR row at creation is as explicit as
     /// an attach — but seeding is **reconciled on every poll**, and `.manual` is
--- a/Sources/TBDDaemon/PR/PRStatusManager.swift
+++ b/Sources/TBDDaemon/PR/PRStatusManager.swift
@@ -49,7 +49,12 @@ enum PRUndeterminedCause {
 
 /// In-memory cache of GitHub PR status per worktree.
 ///
-/// `fetchAll` runs one batch GraphQL call for all viewer PRs, plus one combined per-PR
+/// `fetchAll` runs repo-scoped GraphQL calls — one per repo the fleet spans,
+/// split again when a repo's worktrees carry more candidate head refs than one
+/// query may ask about, and all of them concurrently. Each carries one aliased
+/// `pullRequests(headRefName:)` field per candidate head branch, so the forge
+/// does the matching and a pull request opened by a bot, a teammate, or
+/// the web UI is found — plus one combined per-PR
 /// GraphQL call (run concurrently) for each OPEN PR whose aggregate rollup isn't SUCCESS —
 /// that call returns the aggregate state, every check context with its `isRequired`
 /// flag, and a pagination flag in a single round trip. `refresh` runs
@@ -388,7 +393,11 @@ public actor PRStatusManager {
     /// Whether a forge CLI could be launched at all — the difference between
     /// "asked and got nothing back" and "there was nobody to ask". An injected
     /// runner counts as available: it is what stands in for the binary.
-    private var forgeCLIAvailable: Bool {
+    ///
+    /// `nonisolated` because the GitHub branch-match path is, and it reads only
+    /// immutable state — a `let` runner and a static path. Isolated callers
+    /// still read it unchanged.
+    private nonisolated var forgeCLIAvailable: Bool {
         ghRunner != nil || Self.resolvedGHPath != nil
     }
 
@@ -595,7 +604,7 @@ public actor PRStatusManager {
         public var disproved: [(worktreeID: UUID, parsed: ParsedPRURL)] = []
     }
 
-    /// Fetch all viewer PRs in one GraphQL call and update cache for all known worktrees.
+    /// Fetch every worktree's pull request state and update the cache for all known worktrees.
     /// worktrees: one `PollWorktree` per active non-main worktree.
     ///
     /// Returns the branch-matched PRs as `ParsedPRURL`s in `discovered` — the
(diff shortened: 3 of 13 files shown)


cheapsteak:
Addressed the review note: the `@{push}` claim in `PRPollReconcileTests` poll 2 was wrong (the harness primes `pushBranch` once and never changes it, so it stays `.noPushDestination`). The heal actually fires because the branch query answers with no nodes, sending the worktree through `cachedNumberFallback` to a re-resolved PR whose head is the branch it merely tracks. Corrected in 72d72d10.

For the record, that inaccuracy predates this PR — it is on `main` today, where the same sentence reads "the batch no longer offers the PR". This change only reworded "batch" to "branch query", but the line was in the diff and wrong, so it is fixed here.

The CodeRabbit docstring-coverage warning is not actionable against this repo's conventions: the new code is documented in this file's prose-comment style rather than per-function docstrings, which is what the surrounding code does.

cheapsteak:
The `Lint` failure on this PR is not from this change — it is a repo-wide breakage that also fails on `main`.

SwiftLint 0.65.1 adds `legacy_swiftui_aspect_ratio`, and the `Lint` job installs whatever `brew install swiftlint` currently serves. The version moved between this PR's two pushes, thirteen minutes apart, over the same tree:

- `17:53` — `Pouring swiftlint--0.65.0` → `Found 0 violations`
- `18:06` — `Pouring swiftlint--0.65.1` → `Found 3 violations`

All three are in `Sources/TBDApp/Panes/*`, which this PR does not touch (its diff is `PRStatusManager`, PR tests, and two specs).

Unblocked separately in #791. Once that lands I will update this branch and re-run.

PR #794, opened 2026-09-03
Feat: Show merge-queue position on status-bar PR chips instead of a dot reading "pending"

cheapsteak:
## Summary

A pull request sitting in a merge queue is in a mode of its own: it is no longer waiting on its author, and the only thing worth knowing about it is how many PRs are ahead of it. GitHub cannot express that through `mergeStateStatus` — a queued PR reports `UNKNOWN`, which the daemon maps to the ordinary "Checks pending". The sidebar worktree row has drawn a 🚌 with its 1-indexed queue position badged into the corner for a while; the status bar's PR chips did not, so a queued PR showed the same olive dot as one genuinely waiting on its author.

The chips now draw the same bus, from the same image, and say "In merge queue, position 3" wherever they previously said "Checks pending".

## Why this shape

**No spec.** This is a minor UI change on an existing surface — it revises no theory of the system, adds no flag, and introduces no durable resource. It borrows a glyph the sidebar already ships.

**One image, not a second bus.** `PRStatusPresentation.busImage` already existed and already caches per `(position, side)`. Both surfaces call it at the same side, so they share one `NSImage` rather than two that could drift.

**One sentence, not three literals.** Writing the chip's wording inline would have made a third copy — the sidebar had its own literal, already drifting in capitalization, and the two menu surfaces had none at all. `PRStatusPresentation.stateDescription` composes it once for all three.

**The queue clause supersedes only the decayed reason.** `mapGitHubStateAndReason` computes state independently of queue membership, so `(.checksFailed, "Checks failing")` and a non-nil position arrive together in one GraphQL response. Dropping the reason unconditionally would erase the one fact that says the PR is about to be *evicted* from the queue. Only `.pending` — the `UNKNOWN` artifact — is dropped; everything else rides alongside: `In merge queue, position 2 · Checks failing`.

**The slot width depends on which chip, never on hover.** A queued chip's icon slot is 12pt against a dot chip's 9pt. That preserves the invariant the surrounding code is built on: hovering a chip cannot change its width, so it cannot shove its neighbours or slide the xmark out from under the cursor that summoned it.

## What this PR does

- `StatusBarView.PRChip` carries `mergeQueuePosition`; the chip's icon slot draws the badged bus at rest and the untrack xmark on hover, unchanged.
- `PRStatusPresentation.stateDescription(state:reason:mergeQueuePosition:separator:)` is the single composer, used by the chip, the sidebar row, and `PRBindingPresentation.menuRows`. `separator` exists because the sidebar's spoken accessibility label joins with `", "` — an interpunct reads badly aloud.
- `PRStatusPresentation.busSide` replaces the drawn-size literal in both render sites.
- The `+N` overflow menu gains an `.id` keyed on its composed rows. That file's own comment cites the absence of one as the reason PR titles are kept out of these rows: AppKit materializes an NSMenu once, so a changed field reads stale for as long as the menu lives — and the queue position is precisely the field that ticks 3 → 2 → 1.
- `StatusBarView.iconSlotSide(for:)` is lifted out of the private `View` member so the width invariant is assertable.

## Scope note

The two menu surfaces — the status bar's `+N` overflow menu and the toolbar dropdown — had never been taught about the merge queue, so they rendered the literal "Checks pending" for the very PR whose chip beside them said "In merge queue"; on a multi-PR worktree both are on screen at once. That is past the chip this PR was asked for, but it is the same defect, and the alternative was leaving a third dialect of one sentence. The sidebar row carried the unconditional-supersession bug too and is fixed by the same change.

## Assumptions

- Only GitHub reports a queue position. `GitLabQueries` hardcodes `mergeQueuePosition: nil`, so the forge-neutral wording is safe today — but GitLab's concept is a **merge train**, and wiring it up later needs a forge branch here and in the sidebar.
- `.pending` is dropped by a **state whitelist**, not by matching the `UNKNOWN` artifact. An earlier `requiredChecksPending` branch reaches the same state with a real observation; it produces the identical string today, so nothing distinguishable is swallowed. Adding a second `.pending` reason would need that guard narrowed — noted in the code.

## Evidence & verification

- `scripts/swift-safe build --build-tests` clean; `swiftlint --strict` clean (0 violations, 914 files).
- Discriminating test: `queuedChipKeepsANonPendingReason` expects `PR#412 (In merge queue, position 2 · Checks failing)` and **fails** against the first commit's unconditional supersession, which yielded `PR#412 (In merge queue, position 2)`.
- `chipSentenceComesFromTheSharedHelper` compares each surface's output against the shared helper across 48 state × position × reason combinations, so a surface that stops calling it reddens.
- `menuRowsIDTracksTheRenderedRows` tripwires the overflow menu's refresh key; `iconSlotSideDependsOnTheChipNotTheHover` pins 12/9.
- `sentenceKeepsThePrecisePositionTheBadgeClamps` records a deliberate divergence: the badge clamps to `99+` because it is width-constrained, the sentence prints the literal position because it is not.
- **Not live-verified in the queued state** — that needs a PR actually sitting in a merge queue, which I could not stage. The unqueued paths are byte-identical to before (verified by reading each call site and by regression tests), and the queued glyph goes through the same `busImage` the sidebar has been shipping.

Three review passes ran over this branch; the second caught the supersession defect described above, which the first commit had introduced.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=B88A79EC-068D-4580-8C63-B1A90B81EE04)

https://claude.ai/code/session_0183e1N6ybrBpZSJu2Hi39UM

## Summary by CodeRabbit

* **Enhancements**
  * Pull requests in the merge queue now display their queue position and pending status consistently across status-bar chips, menus, accessibility labels, and related views.
  * Queued items use a shared bus indicator and queue-aware sizing.
  * Status descriptions now preserve relevant failure reasons and support queue positions beyond the `99+` badge limit.
  * Overflow menus refresh when displayed pull request details change.
* **Tests**
  * Added coverage for queued pull request presentation, accessibility text, menu updates, sizing, and status descriptions.

### DIFF
--- a/Sources/TBDApp/ContentView.swift
+++ b/Sources/TBDApp/ContentView.swift
@@ -1025,7 +1025,14 @@ struct PRButtonLabel: View {
     /// distinct. `nil` maps to `\0`, which escaping can never produce (its only
     /// outputs are `\\`, `\-` and `\|`), so a literal `"nil"` reason no longer
     /// reads as an absent one either.
-    static func escapedIDField(_ value: String?) -> String {
+    ///
+    /// `nonisolated` because it is a pure string transform and the other
+    /// once-materialized menu in the app — the status bar's `+N` overflow —
+    /// keys itself the same way from nonisolated
+    /// `PRBindingPresentation.menuRowsID`. `PRButtonLabel`'s `View` conformance
+    /// infers whole-type `@MainActor` isolation, which would otherwise put this
+    /// out of reach there and invite a second, subtly different copy.
+    nonisolated static func escapedIDField(_ value: String?) -> String {
         guard let value else { return #"\0"# }
         return value
             .replacingOccurrences(of: #"\"#, with: #"\\"#)
--- a/Sources/TBDApp/Helpers/StatusBarView.swift
+++ b/Sources/TBDApp/Helpers/StatusBarView.swift
@@ -134,6 +134,18 @@ struct StatusBarView: View {
         /// come from `binding.status`, so they are absent together — and a chip
         /// with no status carries no `mergeQueuePosition` either, so nothing is
         /// left unsaid by returning nil on that branch.
+        ///
+        /// It keeps the default interpunct even on the path that ends up spoken
+        /// (`openLabel`, which is this chip's `.help` AND its
+        /// `.accessibilityLabel`), where the sidebar row splits into two
+        /// joinings. The difference is that the sidebar's spoken label is a
+        /// comma-joined list this sentence is spliced INTO, so an interpunct
+        /// inside it collides with the commas around it; the chip's label is
+        /// `"Open PR #412 — <sentence>"`, already punctuated with an em-dash
+        /// and carrying no list. One string per chip is also what lets the
+        /// headline, the tooltip and the label be the same words by
+        /// construction — a spoken variant would be a second stored field and a
+        /// second thing to keep in step.
         var stateDescription: String? {
             guard let state, let reason else { return nil }
             return PRStatusPresentation.stateDescription(
@@ -360,11 +372,6 @@ struct StatusBarView: View {
         isHovering ? untrackLabel(chip) : openLabel(chip)
     }
 
-    /// The bus glyph's drawn side on a status-bar chip, kept identical to the
-    /// 12 `WorktreeRowView.prGlyph` draws at so the bar's bus and the sidebar's
-    /// are the same image at the same size.
-    nonisolated static let prChipBusSide: CGFloat = 12
-
     /// The side of one chip's leading icon slot — the square its resting glyph
     /// and the untrack xmark share.
     ///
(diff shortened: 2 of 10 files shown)


cheapsteak:
Dispositions for the two reviewer notes.

**CodeRabbit — "refresh identity does not follow the project's complete-status convention; add the remaining status fields."** Not acted on; I believe this is a false positive from applying the toolbar split button's rule to a different control.

`prSplitButtonID` has to key on the *whole* `PRStatus` plus `colorScheme` because its **label** is a baked non-template `NSImage` compositing state color, the queue badge and the auto-archive badge — fields the label renders without appearing in any row. The `+N` overflow menu has no label of that kind. It renders exactly `row.title` and `row.url` (`StatusBarView.swift`, `Button(row.title)` / `.disabled(row.url == nil)`), and `menuRowsID` keys on `id` + `title` + `url`. `title` already carries the state description, the reason and the head branch, so every field the menu draws is in the key. Adding `MenuRow.state` would key on something this menu never renders and rebuild it for changes it cannot show.

**claude-review — the `.id()` could dismiss an open menu if a poll lands while it is up.** Real, accepted, not verified live (it needs a queue position to tick during the second or two the menu is open, which I could not stage).

Worth stating the scope precisely: it is not only the queue position. `title` embeds `reason`, so *any* status poll that changes the reason now rebuilds this menu, where before the PR nothing did. The trade is against the alternative, which is the failure this PR exists to remove — an `NSMenu` materialized once and never refreshed would show "In merge queue, position 3" beside a chip whose baked badge already reads `1`, for as long as the menu lives. A menu that closes on a rare background poll is recoverable in one click; a menu that confidently states a stale position is not. The toolbar's split button has shipped with the same rebuild-on-status-change behavior via `prSplitButtonID`, so this is an existing pattern rather than a new failure mode.

If the dismissal turns out to bite in practice, the narrower fix is to key on the binding set and queue positions only, accepting staleness for `reason` — but that trades a visible wrong number for an invisible one, so I would rather see it happen first.

https://claude.ai/code/session_0183e1N6ybrBpZSJu2Hi39UM

PR #805, opened 2026-09-04
Feat: Know which TBD is running, be told when main moves, and update a live fleet without parking it

zionts:
## What

TBD binaries now know what commit they are, the daemon can tell whether `main` has moved, and `tbd update` moves a running installation to the latest `main` without parking the fleet. Spec: `docs/specs/2026-09-04-automatic-version-updates-design.md`.

- **Build identity** (`TBDShared/BuildIdentity.swift`): every binary reads a `TBDBuildIdentity.json` sidecar next to itself (or in the app bundle's `Contents/`), written by `restart.sh` and `update.sh` before the compiler runs; a build with no sidecar falls back to the derived worktree's `HEAD`, marked as such. `daemon.status` carries it; `tbd --version` prints `0.1.0 (<short>)`; the app's status bar shows its own commit.
- **Update check** (`TBDDaemon/Update/UpdateChecker.swift`): `git ls-remote` against the source worktree's `upstream` (else `origin`) `main`, on an injected clock, hourly by default (`TBD_UPDATE_CHECK_INTERVAL` overrides). Result on `daemon.status.update` with a relation of `upToDate` / `behind` / `unknown`. `daemon.checkForUpdate` runs one synchronously.
- **`update_mode` setting**: timestamped SQL migration adding a nullable `TEXT` column, resolved through `Config.updateModeDefault = .off` (house tri-state style, no SQL default). `tbd config set update-mode <off|check|auto>`, `tbd config get`, and a Settings picker.
- **Surfaces**: `tbd version [--check] [--json]`; `tbd daemon status` gains `Build:` and `Update:`; the app shows a one-line "update available" banner in the build-skew slot, dismissible per commit, plus a "Check for Updates…" menu item.
- **`tbd update`**: one daemon call to find the source worktree, then `execv` of that worktree's `scripts/update.sh` with the user's args and environment.
- **`scripts/update.sh`** (the procedure, user-land): lock; dedicated clone at `~/tbd/updates/src`; fetch + detach `origin/main`; self-re-exec once if the script changed; stamp sidecar; build `TBDDaemon`/`TBDApp`/`TBDCLI` release via `swift-safe`; assemble/sign/install the bundle through the new `scripts/restart-bundle-lib.sh` (shared with `restart.sh`); handover via `TBD_HANDOVER_FROM_PID` (set per invocation, `0` when nothing runs); wait for `daemon status` to report the new binary; restart the app; wake whatever reconcile still parked, 3 at a time, 2 s apart; timestamped log at `~/tbd/updates/update.log` and a summary. Flags: `--check`, `--dry-run`, `--debug`, `--no-app`, `--no-wake`, `--wake-only`, `--remote`, `--auto`.
- **`auto` mode** launches `update.sh --auto` under `nohup`, once per new `main` commit, with a PATH floor so a daemon relaunched with launchd's bare PATH can still build.
- **Docs**: `docs/updating.md` (operator page) and a `## Updating` README section.

## Flag

`update_mode`, default `off`: the daemon runs no timer and spawns nothing. Enable the soak with `tbd config set update-mode check` (notice only) or `auto`. Graduation edits `Config.updateModeDefault` and nothing else; a forcing `UPDATE` is not needed because the column has no SQL default.

## Depends on

- **#804** (`tbd/daemon-handover-safety`) supplies the successor-first handover and the reconcile hardening the script relies on. This PR builds and its check/notice half works without #804; the handover step would exit at the old single-instance gate until #804 lands, so merge #804 first.
- **#803** fixes the launch-PATH bug that the `auto` launcher also floors against.

## Decisions taken from the request, confirmed on review (spec §13)

The request itself came from a human and fixed the shape of the feature: opt-in, default off, out-of-place build, hand over without killing sessions, paced wake, never lose state, a clear log. The brainstorming rule also asks a human to answer the open design questions before implementation; none was available in the session that wrote the spec, so §13 records the five defaults it chose rather than hiding them: latest = `upstream/main` else `origin/main`; clone at `~/tbd/updates/src`, release build; hourly check; `auto` restarts the app too; wake pacing as script constants. This PR, with §13 in it, was then read and approved by the repository owner (review of 2026-09-04) before merge, which is the human confirmation the rule exists to obtain. Those five remain the defaults to revisit before anyone enables `auto`; `update_mode` ships off, so nothing acts until somebody does.

`tbd version [--check]` exits zero for every relation on purpose: it is a report, and `tbd update` is the command whose exit status carries the verdict; scripts that need to branch read `--json`.

Three implementation calls worth a look: a daemon *ahead of* or diverged from `main` reports `upToDate` (nothing to install); `behindBy` is counted in the source worktree and omitted when the objects are not local; `update.sh` also re-links an existing `~/.local/bin/tbd` to the new CLI (never creates one).

## Durable resources under `~/tbd/updates/`

This PR creates files that outlive the request that made them, and none of the three named reconcilers covers them. That is deliberate, because none of them can accumulate: every entry is a fixed singleton path each run reuses or overwrites, never a per-request resource.

- `src` — one long-lived clone every update reuses. It is in nobody's `git worktree list` and `scripts/reclaim-build.sh` never scans it; a run that dies mid-fetch or mid-build leaves the same clone the next run fetches into.
- `update.lock` — one file naming the running update's pid. A lock left by a crashed run names a dead pid and the next run takes it over. Acquisition is atomic (a noclobber create, so simultaneous first runs cannot both hold it). A takeover happens inside a critical section guarded by a `update.lock.takeover` mutex directory and replaces the lock by renaming a `update.lock.new.<pid>` file over it, so the path is never vacant. A run killed between that write and the rename leaves the temp file; the next takeover, which holds the mutex and so knows every such file is a stray, sweeps them. A mutex abandoned by a crash is reclaimed by age. The mutex directory records its creator's pid, and is reclaimed only when it is both older than a minute and its creator is gone (or never got to record itself); a live holder stalled inside the section keeps it. Residual risk, accepted: staleness of both the lock and the mutex is decided by `kill -0`, so a recycled pid would block takeovers until it exits; an operator can remove the lock by hand, and the harness's fan-out and takeover cases cover the rest.
- `update.log` — one append-only log.
- `wake` — one scratch directory for the wake step's per-terminal results, emptied at the start of every wake and removed at its end; a run killed mid-wake leaves it for the next run to empty. (It replaces a per-invocation `mktemp -d` under `$TMPDIR`, which the singleton argument did not reach.)
- `previous/TBD.app` — exactly one bundle, the one the last update replaced, overwritten by the next update and kept as the no-rebuild route back.

Removing any of them is an operator gesture and the next update recreates it. The same statement is in `CLAUDE.md` ("Every durable external resource needs a named reconciler") and spec §7.

## Tests

- Swift: `TBDDaemonTests` 4986/439 suites and the complement 4697/471 pass; `swiftlint --strict` 0 violations; `scripts/lint-migrations.py` OK (18 migrations). New suites: build-identity loader table, relation table, `UpdateChecker` (off runs nothing; check never launches; auto launches once per commit; mode change honored next tick), launch plan, `config.setUpdateMode` RPC both values, `UpdateModeFlagTests` schema suite, both CLI renderers, app presenter.
- Shell: `scripts/update.test.sh` (137 checks: lock, remote order, single re-exec, sidecar, wake filter + batching, opt-outs, dry-run, summary), `scripts/restart-bundle-lib.test.sh` (66), plus the guard and environment harnesses, all wired into the `lint` CI job.

https://claude.ai/code/session_01SwPTd69A1awTHB2TA3suCC

[📦 Archive Agent Box Sessions](https://cheapsteak.github.io/tbd/open/?worktree=9C81D24D-12C8-4945-88A7-431D2116B526)

## Summary by CodeRabbit

- **New Features**
  - Added version checking and update support through the app and CLI.
  - Choose Off, Check only, or Check and install update modes in Experimental settings.
  - View build identity, update status, and dismissible update notices.
  - Added a “Check for Updates…” keyboard command.
  - Added `tbd version` and `tbd update`, including dry-run, logging, rollback, and recovery options.
  - Automatic update monitoring now checks periodically and can install available updates.
- **Documentation**
  - Added guidance for configuring updates, monitoring progress, reviewing logs, and rolling back installations.


zionts:
On the §13 sign-off finding: agreed, and it is not something this session can fix. No human was available to answer the brainstorming questions, so the spec transcribes the decisions in the request and lists the five defaults chosen here. This PR should not be merged until Adam has read spec §13 and either confirmed those five items or asked for changes; `update_mode` stays `off` by default either way, so nothing acts until an operator opts in. The install-before-handover ordering finding is being fixed in the script (previous bundle kept aside and restored if the handover fails), and the self-re-exec lock window is being closed by carrying the pid-keyed lock through `exec`.

cheapsteak (approved):

PR #806, opened 2026-09-04
Fix: Stop a dying daemon from unlinking a live successor daemon's socket

cheapsteak:
## What's broken

Stopping a TBD daemon deletes the Unix socket of a *different*, live daemon that
has since taken over the same path.

Concretely: daemon A is shutting down while daemon B has already bound
`~/tbd/sock`. A's shutdown unlinks the path. B keeps running and keeps
listening — on a socket file that no longer has a name — so every client gets a
connection failure against a path that isn't there, and B never notices anything
is wrong. Recovery is a second restart.

I hit this for real while verifying an unrelated feature: killing an orphaned
daemon took the live daemon's socket with it.

## Why it happens

The daemon's shutdown removed the socket file unconditionally — it never checked
whether the file at that path was still the one it created.

**There were two unlinks, and the one in our code was not the one that bit.**
Fixing only the obvious one left the bug fully alive, and the test caught it:
NIO unlinks the socket too. `ServerBootstrap.bind(unixDomainSocketPath:)` builds
its `ServerSocket` with `cleanupOnClose = true`, and closing that socket calls
`unlink(2)` on the bound *path*, checking only that the file is a socket — never
that it is the same socket. So closing the channel deleted the successor's file
before our own cleanup reached any guard at all.

## What this PR does

- **Captures the socket file's identity at bind time** — `(st_dev, st_ino)` via
  `lstat(2)` immediately after `bind(2)` — and unlinks on shutdown only if the
  path still resolves to that same inode. A successor's `bind(2)` always
  produces a different inode, so a takeover is unambiguous.
- **Takes the unlink back from NIO.** The socket is now created and bound
  directly and handed to `bootstrap.withBoundSocket(_:)`, whose descriptor NIO
  documents as owner-cleaned. NIO still sets it non-blocking and calls
  `listen(2)`.
- Moves `chmod` to immediately after `bind(2)`, before anything can connect.
- **Puts the failed-start cleanup under the same check.** `start()` captures the
  identity and then *awaits* NIO adopting the descriptor; that await is a
  suspension point a successor can bind inside. Removing the file on existence
  alone there deletes the successor's live socket — the same bug, reached from a
  failed startup instead of a shutdown. Both reclaim sites now call one
  `unlinkSocketFile(ifStillIdentity:)`.

This follows the same shape as #802 — prove ownership, then unlink — with the
file's inode standing in for that fix's pid, because a socket file carries no
identity of its own. That fix's residual race transfers verbatim: there is a
microsecond window between the `lstat(2)` and the `unlink(2)`, and darwin has no
unlink-if-inode primitive to close it.

- **Takes the claim on the file exactly once.** The identity is read and
  cleared in one step under a lock. Nothing serializes `stop()` — SIGTERM and
  SIGINT each fire an independent, undeduplicated `Task { await daemon.stop() }`
  in `main.swift`, which a supervisor escalating signals will do — so with a
  plain read-then-nil two shutdowns could both come away holding the same
  identity, and the loser then unlinks against a path a successor may hold by
  then. That is this fix's own protection, defeated.
- **Stops two startup paths from unlinking a socket they are not about to
  bind.** `PIDFile.cleanupIfStale()` and `AppState.startDaemonAndConnect()`
  each removed the socket file once the pid file's recorded pid was judged not
  a live `TBDDaemon`, with no check of the socket's current occupant. Both
  removals are gone. See "Clearing the path is the binder's act" below.
- **Runs each server's shutdown exactly once, because a second one hangs.**
  `SocketServer.stop()` and `HTTPServer.stop()` now go through a shared
  `ShutdownLatch`: the first call runs the body, and every later call —
  sequential or concurrent — `await`s that same run. See "A second `stop()`
  used to hang" below, which also corrects what round two of this PR said was
  hanging.

The pre-bind removal in `start()` is deliberately left alone: that is the
intended takeover, and it is exactly what makes the successor's inode differ.

## Clearing the path is the binder's act

The rendezvous path is cleared by whoever is about to bind it, and by nobody
else. `SocketServer` is that code: `start()` clears the path immediately before
`bind(2)`, and `stop()` reclaims the file only while the path still resolves to
the inode it bound.

Two startup paths broke that rule on the strength of a stale pid file, and a pid
file reads as stale for exactly as long as a successor has bound the socket
without having rewritten it yet. Deleting the socket there costs a live daemon
its rendezvous and puts nothing in its place: the path stays empty for the whole
of the next daemon's startup, and for good if that startup throws first.

Removing them is not quite free, and an earlier round of this PR overstated the
case for it. The claim was that `SocketServer.start()`'s pre-bind unlink always
eventually runs on every path that used to reach the removed code. It does not.
`AppState.startDaemonAndConnect()` returns early when no `TBDDaemon` binary is
found and when `process.run()` throws, before the daemon it would spawn exists
to bind anything; and `Daemon.start()` can throw between its own
`cleanupIfStale()` (step 2) and the socket bind hundreds of lines later. On
those paths a stale socket file now outlives the cleanup that used to remove it.

The removals are still right, because the two failures are not comparable. An
unbound socket file behaves to a client exactly like a missing one —
`connect(2)` fails either way, and the next daemon to bind clears the path
regardless — whereas unlinking a *live* successor's socket silently strands a
working daemon that keeps accepting on a listener no client can reach, with
nothing to restore the path. The comments at both sites now say that, rather
than claiming the bind always follows.

That is also why the guard sketched for these sites in an earlier round is not
what shipped. Porting `SocketFileIdentity` into `TBDShared` to reach the
app-side call site would have made the removals *safe*; deleting them makes them
*unnecessary*, which is a smaller change and a stronger invariant. (A
`connect(2)` probe remains unsound for the same reason as before: #802 measured
darwin refusing instantly once the accept queue fills, so refused does not mean
gone.)

`HolderSpawner.swift:252` and `PeerHelper.swift:190` are a different rendezvous
and stay out of scope.

## A second `stop()` used to hang, and it was not `shutdownGracefully()`

Nothing serializes `Daemon.stop()`: SIGTERM and SIGINT each fire an
independent, undeduplicated `Task { await daemon.stop() }` in `main.swift`,
both reach the servers through `Daemon.stop()`, and a supervisor escalating
signals sends both. Round two of this PR recorded that a second `stop()` wedges
and attributed it to `MultiThreadedEventLoopGroup.shutdownGracefully()`. **That
attribution was wrong**, and the correction changes the fix.

- `shutdownGracefully()` is safe to call again, and safe to call concurrently.
  Its own documentation says so — "Even though calling `shutdownGracefully`
  more than once should be avoided, it is safe to do so and execution of the
  `handler` is guaranteed" — and the implementation backs it. Every call takes
  `shutdownLock` and switches on `runState`: `.closing` appends the handler to
  the pending-callback list, `.closed` dispatches it immediately, and only a
  caller that found `.running` drives the shutdown, completing the rest from its
  `g.notify` block (`MultiThreadedEventLoopGroup.swift:404-511`, swift-nio
  2.96.0). Concurrent callers serialize on that lock.
- **`channel.close()` is what hangs.** Off the event loop it is
  `eventLoop.execute { self.close0(mode:promise:) }`
  (`ChannelPipeline.swift:788`), and `SelectableEventLoop._schedule0` on a loop
  whose group has shut down takes the `!validExternalStateToScheduleTasks`
  branch: it prints "Cannot schedule tasks on an EventLoop that has already
  shut down. This will be upgraded to a forced crash in future SwiftNIO
  versions." and returns *without enqueuing anything*
  (`SelectableEventLoop.swift:480-492`). The close promise is never fulfilled,
  so `try await channel.close()` suspends for good — and the event-loop threads
  have already exited, so the wedged task leaves a process at 0% CPU with
  nothing to see. This is the same mechanism already recorded further down for
  why the failed-start window needs an injected seam.

Both NIO-backed servers in the teardown have that shape, so both go through a
shared `ShutdownLatch`. `HTTPServer` is included deliberately rather than as
scope creep: `Daemon.stop()` stops the socket server and then the HTTP server,
so fixing only the first would hand the same hang to the next line and the
second shutdown still would not reach its `exit(0)`.

The latch settles both halves at once. The body runs exactly once, so no
`close()` is ever submitted to a dead loop and `shutdownGracefully()` is never
called twice or concurrently at all; and a later caller `await`s that same run
rather than returning early, so `stop()` still means "the shutdown has
finished" for everyone who called it. The task is unstructured on purpose — it
does not inherit its creator's cancellation, so a cancelled signal handler
cannot abandon a shutdown other callers are waiting on.

One thing measurement corrected here too, and the tests are shaped by it: two
`stop()` calls that merely *start* together are not the dangerous case. Both get
their `channel.close()` in while the event loop is still alive, so that ordering
survives even without the latch. The case that hangs is a `stop()` that arrives
after another has passed `shutdownGracefully()`, so the overlapping test stages
exactly that.

## Why the identity comes from the path, not the descriptor

`fstat(2)` on the bound descriptor looks like a free tightening of the window
against `lstat(2)` on the path — the bind and the stat are back-to-back
syscalls, but an OS-level race could in principle let a successor unlink and
rebind between them, so this process would record the *successor's* inode as
its own.

It does not work on darwin, where the two do not name the same object. `fstat`
on a bound `AF_UNIX` descriptor reports the socket's own in-kernel vnode, with
`st_dev` of -1, not the filesystem entry `bind(2)` created. Measured on darwin
25.1, over both a C and a Python probe:

```
fstat dev=-1        ino=15302332
lstat dev=16777235  ino=863366896
```

An identity captured that way could never equal the one read back from the
path, so every ownership check would fail closed and the socket file would never
be reclaimed at all — a strictly worse failure than the misattribution it was
meant to prevent. The measurement is recorded on `SocketFileIdentity` so it is
not re-proposed.

## Assumptions

- Assumes `(st_dev, st_ino)` identifies a socket file across the process's
  lifetime. If the path is moved and restored, or the filesystem reuses an inode
  after deletion within one daemon's life, the check could read a foreign file
  as ours. Neither is reachable through any path TBD takes.

## One behavior change to weigh deliberately

`withBoundSocket` calls `listen(2)` during channel construction, *before*
server-channel options are applied, so the explicit `.backlog` of 64 no longer
reaches the socket and it takes NIO's default of **128**. This is forced by the
seam rather than chosen. It moves in the safe direction — and #802 measured why
a fuller accept queue matters, since darwin refuses a connect instantly once the
backlog fills rather than blocking.

Verified against the vendored source, swift-nio 2.96.0 (`Package.resolved`
revision `b315658`):

- `cleanupOnClose` is off for a pre-bound descriptor —
  `ServerSocket.init(socket:setNonBlocking:)` sets `cleanupOnClose = false  //
  socket already bound, owner must clean up` (`ServerSocket.swift:83`), and
  `ServerSocket.close()` computes the path to unlink only when that flag is set
  (`ServerSocket.swift:148`). NIO's own bind path takes the other branch:
  `cleanupOnClose = true` for `.unix` (`ServerSocket.swift:56`).
- NIO still calls `listen(2)`. `ServerBootstrap.withBoundSocket(_:)`
  (`Bootstrap.swift:344`) builds `ServerSocketChannel(socket:eventLoop:group:)`
  (`Bootstrap.swift:353`), whose initializer ends in `try
  self.socket.listen(backlog: backlog)` (`SocketChannel.swift:253`).
- The backlog is NIO's default 128. `ServerSocketChannel.backlog` is
  `private var backlog: Int32 = 128` (`SocketChannel.swift:208`), and that
  `listen` runs inside `makeServerChannel` (`Bootstrap.swift:403`), which
  `bind0` invokes *before* `serverChannelOptions.applyAllChannelOptions(to:)`
  (`Bootstrap.swift:413`) — so a `.backlog` option would be recorded by
  `setOption0` (`SocketChannel.swift:273`) with no second `listen` to apply it.

## Evidence & verification

- **The repro is the test**: server A binds, server B rebinds the same path, A
  stops. B's inode must still be at the path *and* a real `connect(2)` must
  succeed.
- **Discriminating proof** — with the source fix reverted and the tests kept,
  the shutdown cases fail, reporting the path resolving to no inode at all and
  `connect(2)` returning `-1`. Reverting only the failed-start gate to the old
  unconditional remove reddens the failed-start successor test alone, with the
  other four green. With the fix, 6 tests in 2 suites pass.
- Two further tests pin that the check cannot degenerate into "never unlink": a
  server with no successor still reclaims its own file at shutdown, and a start
  that fails after binding still reclaims the file it just bound.
- **The claim-once guarantee has its own discriminating test.**
  `onlyOneOverlappingShutdownClaimsTheFile` runs eight concurrent claimants and
  requires exactly one to come away owning the file. The read-and-clear window
  is a few instructions wide, so the race cannot be reached by starting threads
  and hoping: a `duringIdentityClaim` seam — `nil` in production, alongside the
  existing `beforeAdoptingBoundSocket` one — holds the window open until every
  thread has reached the call site. Under the lock the losers wait outside it
  and come away with nothing; with a plain read-then-nil all eight read the
  same identity.
- **The startup rule is pinned too.**
  `onlyTheBinderClearsTheRendezvousPath` scans TBDDaemon and TBDApp and
  requires that nothing outside `SocketServer` removes
  `TBDConstants.socketPath`. It is a source-text pin because that path resolves
  from a process-global this test target must not mutate — `ConstantsTests`
  pins that no suite in this process sets `TBD_SOCKET_PATH`, and
  `scripts/test.sh` sets it for the whole run. `cleanupRemovesAStalePidFile` is
  the positive control that the pid-file cleanup still cleans.
- The failed-start window is reached through an injected step in `start()`,
  `nil` in production. The real failures there are NIO refusing the descriptor,
  and the in-process ways to force one either hang (a shut-down event loop drops
  the submitted task and never completes the future —
  `SelectableEventLoop.execute` swallows the enqueue failure,
  `SelectableEventLoop.swift:461`) or trip NIO's own debug assertions on a
  half-constructed channel (`BaseSocket.deinit` asserts `!isOpen`,
  `BaseSocket.swift:261`).
- `swiftlint --strict`: 0 violations across 925 files.
- **The second round's two tests were run against the fixes reverted, and both
  fail there.** 12 tests in 2 suites, 2 issues:
  `only one of many overlapping shutdowns claims the socket file` fails with
  `(claims.count -> 2) == 1`, and
  `nothing outside SocketServer removes the rendezvous socket file` fails naming
  both offenders, `PIDFile.swift:63` and `AppState.swift:2953`. With the fixes
  in place the same 12 pass in 0.330 s.
- **The overlapping-shutdown path is now tested end to end, through the real
  `stop()`.** Three tests, and all three go through production entry points:
  - `overlappingStopsBothFinishAndClaimTheFileOnce` starts a server and stages
    two genuinely overlapping `stop()` calls — the `duringIdentityClaim` seam
    holds the first shutdown at its reclaim step, past
    `shutdownGracefully()`, until the second caller is at `stop()` — then
    requires both to return, the socket file to be claimed exactly once, and
    the file to be gone. The seam now carries the identity the claimant came
    away with, which is how "claimed exactly once" is observed from inside a
    real `stop()`; it has no other outward sign.
  - `repeatStopLeavesTheSuccessorAlone` covers the sequential half a signal
    escalation actually produces: shut down, let a successor bind the path,
    `stop()` again. It must return, and it must leave the successor's socket
    alone.
  - `repeatHTTPServerStopReturns` pins the sibling server in the same teardown,
    and `latchRunsItsBodyOnceAndEveryCallerWaits` pins the latch itself — eight
    callers, one run, and nobody returning before that run has finished.

  All of them wait on detached tasks with a deadline rather than awaiting
  directly or gathering a task group, because the failure under test is a
  shutdown that never returns: a task group would wedge the whole run alongside
  it instead of reporting it.
- **Discriminating proof for the latch**: with only the two `stop()` bodies
  un-latched (called directly again) and everything else in the branch kept,
  **19 tests in 3 suites, 3 failures**, one
  per shutdown path: the overlapping test reporting `finished.count == 2` never
  satisfied at its 15 s deadline, `repeatStopLeavesTheSuccessorAlone` the same
  for its second `stop()`, and `repeatHTTPServerStopReturns` the same for the
  HTTP server. `latchRunsItsBodyOnceAndEveryCallerWaits` still passes there,
  because it exercises the latch directly. Restored, the same 19 pass in
  0.399 s.

  An earlier shape of the overlapping test — two `stop()` calls simply started
  together — passed against the un-latched code, which is what turned up the
  ordering point above and is why the overlap is staged instead.

## Follow-ups not taken here

Two sibling unconditional unlinks are left as out of scope:
`HolderSpawner.swift:252` and `PeerHelper.swift:190`. They are a different
rendezvous — the per-session pty-holder sockets under `~/tbd/holders/`, which
have their own named reconciler — but the same "unlink only what you still own"
question applies to them, and the holder one is worth a look.

https://claude.ai/code/session_01TMmwK6VzokeGobKs8YunXZ

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Unix socket lifecycle handling to prevent older daemon instances from removing successor sockets.
  - Socket cleanup now occurs only when the server owns the socket, including during shutdown and startup failures.
  - Stale PID recovery no longer removes a socket that may belong to another daemon.
  - Improved reliability when multiple daemon instances use the same socket path.
  - Repeated or concurrent shutdown requests now complete safely without duplicate cleanup.
  - Startup failures now provide specific errors for oversized paths, socket creation failures, and binding failures.

### DIFF
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -2970,12 +2970,24 @@ final class AppState {
             // The socket file is deliberately left where it is. Clearing the
             // rendezvous path is the job of whoever is about to bind it, and
             // the daemon spawned below does exactly that in its own
-            // `SocketServer.start()` — so removing it here adds nothing when
-            // the pid file really is stale, and deletes a live daemon's socket
-            // when it is stale only because a successor has already bound and
-            // has not rewritten the pid file yet. That successor would keep
-            // accepting on a listener no client can reach, with nothing to
-            // restore the path if the spawn below then fails.
+            // `SocketServer.start()`.
+            //
+            // What removing it here can cost is a live daemon's rendezvous:
+            // the pid file reads as stale for as long as a successor has bound
+            // the socket without having rewritten the pid file yet, and
+            // deleting the socket there leaves that successor accepting on a
+            // listener no client can reach.
+            //
+            // What it buys, against that, is bounded. It is not true that the
+            // bind below always eventually runs: this method returns early
+            // when no TBDDaemon binary is found and when `process.run()`
+            // throws, and even on the spawned path `Daemon.start()` can throw
+            // between its own pid-file cleanup and the socket bind hundreds of
+            // lines later. So a stale socket file can outlive this branch. It
+            // is left anyway, because an unbound socket file behaves to a
+            // client exactly like a missing one — `connect(2)` fails either
+            // way — while an unlinked live one silently strands a working
+            // daemon.
             logger.warning("""
             Stale daemon pid file: pid \(pid, privacy: .public) is not a live \
             TBDDaemon — removing the pid file before spawning
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -2963,15 +2963,24 @@ final class AppState {
                 await connectAndLoadInitialState()
                 return
             }
-            // Stale artifacts — dead pid, or pid recycled by another process
-            // (e.g. after a reboot). Remove them so the spawn below starts
-            // from a clean slate instead of tripping over a dead socket.
+            // Stale pid file — dead pid, or pid recycled by another process
+            // (e.g. after a reboot). Remove it so the spawn below starts from
+            // a clean slate.
+            //
+            // The socket file is deliberately left where it is. Clearing the
+            // rendezvous path is the job of whoever is about to bind it, and
+            // the daemon spawned below does exactly that in its own
+            // `SocketServer.start()` — so removing it here adds nothing when
+            // the pid file really is stale, and deletes a live daemon's socket
+            // when it is stale only because a successor has already bound and
+            // has not rewritten the pid file yet. That successor would keep
+            // accepting on a listener no client can reach, with nothing to
+            // restore the path if the spawn below then fails.
             logger.warning("""
             Stale daemon pid file: pid \(pid, privacy: .public) is not a live \
-            TBDDaemon — removing pid file and socket before spawning
+            TBDDaemon — removing the pid file before spawning
             """)
             try? FileManager.default.removeItem(atPath: pidPath)
-            try? FileManager.default.removeItem(atPath: TBDConstants.socketPath)
         }
 
         // Find TBDDaemon binary using sibling and source-worktree candidates
--- a/Sources/TBDDaemon/PIDFile.swift
+++ b/Sources/TBDDaemon/PIDFile.swift
@@ -62,10 +62,27 @@ public struct PIDFile: Sendable {
         return true
     }
 
+    /// Remove a stale pid file. **The socket file is deliberately left alone.**
+    ///
+    /// Removing it here buys nothing: the only thing that needs the rendezvous
+    /// path free is `bind(2)`, and `SocketServer.start()` clears the path
+    /// itself immediately before binding — on every path that gets that far,
+    /// including this one.
+    ///
+    /// What it can cost is a live daemon's rendezvous. A pid file reads as
+    /// stale whenever the pid it names is not a live `TBDDaemon`, and a
+    /// successor daemon writes its pid file (step 4 of `Daemon.start()`) well
+    /// before it binds the socket, so "stale pid, live socket" is a state the
+    /// path really passes through. Deleting the socket there leaves the
+    /// successor accepting on a listener no client can reach, and nothing
+    /// restores the path until this daemon reaches its own bind — the whole of
+    /// a startup later, or never, if that startup throws first.
+    ///
+    /// Clearing the rendezvous path is the act of whoever is about to bind it,
+    /// and of nobody else. See `SocketServer.unlinkSocketFile(ifStillIdentity:)`.
     public func cleanupIfStale() {
         if isStale() {
             remove()
-            try? FileManager.default.removeItem(atPath: TBDConstants.socketPath)
         }
     }
 }
--- a/Sources/TBDDaemon/PIDFile.swift
+++ b/Sources/TBDDaemon/PIDFile.swift
@@ -64,22 +64,26 @@ public struct PIDFile: Sendable {
 
     /// Remove a stale pid file. **The socket file is deliberately left alone.**
     ///
-    /// Removing it here buys nothing: the only thing that needs the rendezvous
-    /// path free is `bind(2)`, and `SocketServer.start()` clears the path
-    /// itself immediately before binding — on every path that gets that far,
-    /// including this one.
-    ///
-    /// What it can cost is a live daemon's rendezvous. A pid file reads as
-    /// stale whenever the pid it names is not a live `TBDDaemon`, and a
-    /// successor daemon writes its pid file (step 4 of `Daemon.start()`) well
+    /// What removing it here can cost is a live daemon's rendezvous. A pid file
+    /// reads as stale whenever the pid it names is not a live `TBDDaemon`, and
+    /// a successor daemon writes its pid file (step 4 of `Daemon.start()`) well
     /// before it binds the socket, so "stale pid, live socket" is a state the
     /// path really passes through. Deleting the socket there leaves the
-    /// successor accepting on a listener no client can reach, and nothing
-    /// restores the path until this daemon reaches its own bind — the whole of
-    /// a startup later, or never, if that startup throws first.
+    /// successor accepting on a listener no client can reach.
     ///
-    /// Clearing the rendezvous path is the act of whoever is about to bind it,
-    /// and of nobody else. See `SocketServer.unlinkSocketFile(ifStillIdentity:)`.
+    /// What it buys, against that, is bounded. The only thing that needs the
+    /// rendezvous path free is `bind(2)`, and `SocketServer.start()` clears the
+    /// path itself immediately before binding — but this runs at step 2 of
+    /// `Daemon.start()` and that bind is hundreds of lines later, with plenty
+    /// of `throw`s in between, so it is not true that the bind always
+    /// eventually runs. A startup that fails first can leave a stale socket
+    /// file behind.
+    ///
+    /// It is left anyway, because an unbound socket file behaves to a client
+    /// exactly like a missing one — `connect(2)` fails either way — while an
+    /// unlinked live one silently strands a working daemon. Clearing the
+    /// rendezvous path is the act of whoever is about to bind it, and of
+    /// nobody else. See `SocketServer.unlinkSocketFile(ifStillIdentity:)`.
     public func cleanupIfStale() {
         if isStale() {
             remove()
--- a/Sources/TBDDaemon/Server/HTTPServer.swift
+++ b/Sources/TBDDaemon/Server/HTTPServer.swift
@@ -18,6 +18,13 @@ public final class HTTPServer: Sendable {
     private let group: MultiThreadedEventLoopGroup
     private nonisolated(unsafe) var channel: Channel?
 
+    /// Keeps the shutdown to one run, however many `stop()` calls arrive.
+    /// `Daemon.stop()` is not deduplicated and this server is torn down right
+    /// after `SocketServer`, so an escalated SIGTERM-then-SIGINT reaches it
+    /// twice; a second run would wedge here instead and the shutdown would
+    /// never reach its `exit(0)`. See `ShutdownLatch`.
+    private let shutdownLatch = ShutdownLatch()
+
     /// The port the server is bound to, available after start().
     public nonisolated(unsafe) var port: Int = 0
 
@@ -60,15 +67,18 @@ public final class HTTPServer: Sendable {
         logger.info("Listening on http://127.0.0.1:\(self.port, privacy: .public)")
     }
 
-    /// Stop the server and clean up.
+    /// Stop the server and clean up. Safe to call any number of times, from
+    /// any number of tasks at once.
     public func stop() async {
-        do {
-            try await channel?.close()
-        } catch {
-            // Already closed
+        await shutdownLatch.run {
+            do {
+                try await self.channel?.close()
+            } catch {
+                // Already closed
+            }
+            try? await self.group.shutdownGracefully()
+            try? FileManager.default.removeItem(atPath: self.portFilePath)
         }
-        try? await group.shutdownGracefully()
-        try? FileManager.default.removeItem(atPath: portFilePath)
     }
 }
 
--- a/Sources/TBDDaemon/Server/ShutdownLatch.swift
+++ b/Sources/TBDDaemon/Server/ShutdownLatch.swift
@@ -1,0 +1,57 @@
+import Foundation
+
+/// Runs an `async` shutdown exactly once, and lets every caller await that one
+/// run.
+///
+/// Nothing serializes daemon shutdown: SIGTERM and SIGINT each fire an
+/// independent, undeduplicated `Task { await daemon.stop() }` in `main.swift`,
+/// and a supervisor escalating signals sends both. Both reach every server's
+/// `stop()` through `Daemon.stop()`.
+///
+/// For a NIO-backed server, running the shutdown body twice does not merely
+/// repeat work — it hangs. The body's first act is `channel.close()`, which off
+/// the event loop is `eventLoop.execute { ... }` fulfilling a promise, and a
+/// `SelectableEventLoop` whose group has already shut down *discards* submitted
+/// work: `_schedule0` takes the `!validExternalStateToScheduleTasks` branch,
+/// prints "Cannot schedule tasks on an EventLoop that has already shut down"
+/// and returns without enqueuing anything. The promise is never fulfilled, so
+/// `try await channel.close()` suspends for good — and the event-loop threads
+/// have already exited, so the wedged task leaves a process at 0% CPU with
+/// nothing to see.
+///
+/// (`MultiThreadedEventLoopGroup.shutdownGracefully()` itself is *not* the
+/// problem, and it is worth saying so because it looks like the culprit: it
+/// documents that a repeat call is safe and its handler still runs, and its
+/// implementation switches on `runState` under a lock — `.closing` appends the
+/// handler to the pending list, `.closed` dispatches it immediately — so
+/// repeated and concurrent calls both complete. Once the body runs once,
+/// though, the question stops arising at all.)
+///
+/// Callers `await` the one run rather than returning early, so `stop()` still
+/// means "the shutdown has finished" for everyone who called it. The task is
+/// unstructured on purpose: it does not inherit its creator's cancellation, so
+/// a cancelled signal handler cannot abandon a shutdown other callers are
+/// waiting on.
+final class ShutdownLatch: Sendable {
+    private nonisolated(unsafe) var task: Task<Void, Never>?
+    private let lock = NSLock()
+
+    /// Run `body` if this is the first call; otherwise await the run the first
+    /// call started. Later `body` arguments are never invoked — every caller of
+    /// a given latch passes the same shutdown.
+    func run(_ body: @escaping @Sendable () async -> Void) async {
+        await started(body).value
+    }
+
+    /// Hand back the single run, starting it if this is the first call.
+    /// Synchronous and lock-held so two callers arriving together cannot both
+    /// start one.
+    private func started(_ body: @escaping @Sendable () async -> Void) -> Task<Void, Never> {
+        lock.lock()
+        defer { lock.unlock() }
+        if let task { return task }
+        let created = Task { await body() }
+        task = created
+        return created
+    }
+}
--- a/Sources/TBDDaemon/Server/SocketServer.swift
+++ b/Sources/TBDDaemon/Server/SocketServer.swift
@@ -17,6 +17,14 @@ public final class SocketServer: Sendable {
     private let group: MultiThreadedEventLoopGroup
     private nonisolated(unsafe) var channel: Channel?
 
+    /// Identity of the socket file this server actually bound, read from the
+    /// path right after `bind(2)`. `nil` until a successful bind, and again
+    /// once `stop()` has settled the file.
+    ///
+    /// This is what licenses the unlink in `stop()`. See
+    /// `unlinkOwnedSocketFile()`.
+    private nonisolated(unsafe) var boundSocketIdentity: SocketFileIdentity?
+
     /// Number of currently connected clients. Updated atomically.
     private let _connectedClients = ManagedAtomic<Int>(0)
 
@@ -47,8 +55,14 @@ public final class SocketServer: Sendable {
         let connectedClients = self._connectedClients
         let limiter = self.limiter
 
+        // The backlog is deliberately not set here: this bootstrap is handed an
+        // already-bound descriptor (below), and NIO listens on it while
+        // constructing the channel — before any server-channel option is
+        // applied — so the accept queue takes NIO's own default of 128. That is
+        // strictly more forgiving than the 64 it replaces, which matters because
+        // a full accept queue is indistinguishable from a dead socket at the
+        // client's `connect(2)`.
         let bootstrap = ServerBootstrap(group: group)
-            .serverChannelOption(.backlog, value: 64)
             .childChannelInitializer { channel in
                 channel.eventLoop.makeCompletedFuture {
                     let handler = SocketRPCHandler(
@@ -60,14 +74,40 @@ public final class SocketServer: Sendable {
                 }
             }
 
-        let ch = try await bootstrap.bind(
-            unixDomainSocketPath: socketPath
-        ).get()
-
-        // Set socket permissions so any local user can connect
-        chmod(socketPath, 0o700)
+        // Bind the socket here and hand NIO the descriptor, rather than calling
+        // `bind(unixDomainSocketPath:)` and letting NIO do it.
+        //
+        // NIO's own bind marks the server socket `cleanupOnClose`, and that
+        // cleanup runs `unlink(2)` on the *path* the socket was bound to when
+        // the channel closes — its only check is that something at that path is
+        // a socket, never that it is still the socket this process created. So
+        // a daemon shutting down deletes the rendezvous file that a successor
+        // daemon has since bound, and the successor keeps accepting on a
+        // listener no client can reach. Handing NIO a descriptor that is
+        // already bound sets `cleanupOnClose` to false — "socket already bound,
+        // owner must clean up" — which is what puts the unlink under the
+        // ownership check in `unlinkOwnedSocketFile()`.
+        let boundFD = try Self.bindListeningSocket(at: socketPath)
+        // Read the file's identity as close to `bind(2)` as possible, before
+        // any await gives a successor the chance to replace it. A successor
+        // daemon's `start()` unlinks whatever is at this path and binds a fresh
+        // socket, so the inode is what separates our file from theirs.
+        let identity = SocketFileIdentity(path: socketPath)
+        let ch: Channel
+        do {
+            ch = try await bootstrap.withBoundSocket(boundFD).get()
+        } catch {
+            // Do not close `boundFD` here: NIO takes ownership of the
+            // descriptor as soon as it wraps it, and closing a descriptor
+            // another layer already closed can later shut down an unrelated
+            // reused one. The socket file is unambiguously ours, so reclaim
+            // that and let a failed start surface.
+            try? fm.removeItem(atPath: socketPath)
+            throw error
+        }
 
         self.channel = ch
+        self.boundSocketIdentity = identity
         logger.info("Listening on \(self.socketPath, privacy: .public)")
     }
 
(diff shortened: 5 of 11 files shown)


cheapsteak:
Addressed the MEDIUM finding, and verified the three NIO claims against the vendored source the review sandbox lacked. Pushed as `ed083a6a`.

## The finding is correct — fixed

`start()` captured `identity` right after `bind(2)` and then `await`ed `bootstrap.withBoundSocket(boundFD).get()`. That await is a suspension point: a successor daemon's `start()` can run inside it, unlink our file, and bind its own under the same name. The catch block's `try? fm.removeItem(atPath: socketPath)` would then delete the successor's live socket — the same bug this PR exists to fix, reached from a failed startup instead of a shutdown. The comment justifying it was true only at the instant the identity was read, not when the cleanup ran.

Both reclaim sites now go through one `unlinkSocketFile(ifStillIdentity:)`, which unlinks only while the path still resolves to the captured inode. `unlinkOwnedSocketFile()` is now a thin shutdown wrapper that consumes `boundSocketIdentity` and delegates. The stale comment is gone, replaced by one that says what is actually true — the identity was true when read, the await can invalidate it, so the removal is gated.

## The discriminating test, and why it needs a seam

`SocketServerSocketOwnershipTests` gains two tests. The load-bearing one, "a start that fails after binding leaves a successor's socket alone", binds a real successor socket at the path from inside the window (unlink + `bind(2)` + `listen(2)`, exactly what a successor's `start()` leaves) and then fails the adoption. It asserts the successor's inode is still at the path and that a real `connect(2)` still reaches it. The companion test pins the opposite mutation: a failed start with nobody behind it still reclaims its own file, so the gate cannot degenerate into "never unlink".

Mutation proof — reverting only the failure-path gate to the old `try? fm.removeItem(atPath: socketPath)`:

```
Test "a start that fails after binding leaves a successor's socket alone" recorded an issue at
  SocketServerSocketOwnershipTests.swift:178:9: Expectation failed:
  (socketFileInode(at: socketPath) -> nil) == (successorInode -> 863338155)
    the failed start deleted the live successor's socket
  SocketServerSocketOwnershipTests.swift:185:9: Expectation failed: (clientFd -> -1) >= 0
    clients can no longer reach the live daemon at /tmp/tbd-sockown-EE41426C.sock
Test run with 5 tests in 1 suite failed
```

The other four tests stayed green under that mutation, so the new test discriminates this call site specifically.

Reaching that window costs a seam, and I want to be upfront that it is one: `start()` takes an injected step that runs between `bind(2)` and the handover, `nil` in production. The real failures there are NIO refusing the descriptor, and every in-process way I found to force one is worse than the seam: with a shut-down event loop the submitted task is silently dropped and the future never completes (`SelectableEventLoop.execute` discards the enqueue failure, `SelectableEventLoop.swift:461`), so the test hangs rather than fails; and any path where `ServerSocketChannel`'s initializer throws after `super.init` trips NIO's own debug assertions on the half-constructed channel (`BaseSocket.deinit` asserts `!isOpen`, `BaseSocket.swift:261`; `BaseSocketChannel.deinit` asserts `canBeDestroyed`, `BaseSocketChannel.swift:564`) and crashes the test process. The seam is handed the descriptor so the test closes it instead of leaking it.

## NIO claims — all three hold

Read from `.build/checkouts/swift-nio`, swift-nio 2.96.0, `Package.resolved` revision `b31565862a8f39866af50bc6676160d8dda7de35`.

1. **`cleanupOnClose` is disabled by `withBoundSocket` — confirmed.** `ServerSocket.init(socket:setNonBlocking:)` opens with `cleanupOnClose = false  // socket already bound, owner must clean up` (`Sources/NIOPosix/ServerSocket.swift:83`). `ServerSocket.close()` reads the path to unlink only through that flag: `let maybePathname = self.cleanupOnClose ? (try? self.localAddress().pathname) : nil` (`ServerSocket.swift:148`), and only then calls `BaseSocket.cleanupSocket(unixDomainSocketPath:)`. The other constructor — the one `bind(unixDomainSocketPath:)` uses — takes the opposite branch: `case .unix: cleanupOnClose = true` (`ServerSocket.swift:53-56`). So the PR's account of both halves is accurate.

2. **NIO still calls `listen(2)` on the pre-bound descriptor — confirmed.** `ServerBootstrap.withBoundSocket(_:)` (`Sources/NIOPosix/Bootstrap.swift:344`) builds the channel via `ServerSocketChannel(socket:eventLoop:group:)` (`Bootstrap.swift:353`), whose convenience initializer is `let sock = try ServerSocket(socket: socket, setNonBlocking: true); try self.init(serverSocket:...); try self.socket.listen(backlog: backlog)` (`Sources/NIOPosix/SocketChannel.swift:250-254`). So both "sets it non-blocking" and "calls `listen(2)`" are literal.

3. **The backlog falls back to NIO's default of 128 — confirmed.** `ServerSocketChannel` stores `private var backlog: Int32 = 128` (`SocketChannel.swift:208`), and the `listen` above happens inside `makeServerChannel`, which `bind0` runs at `Bootstrap.swift:403` — *before* `serverChannelOptions.applyAllChannelOptions(to: serverChannel)` at `Bootstrap.swift:413`. A `.backlog` option would still be recorded by `setOption0` (`SocketChannel.swift:273`) but there is no second `listen` to apply it, so the socket keeps the 128 it was listened with. The old `.serverChannelOption(.backlog, value: 64)` would therefore have become dead configuration, which is why it was removed rather than left in place.

The PR description now carries these citations and the corrected test counts (6 tests in 2 suites).

## Verification

- `scripts/swift-safe build --build-tests` — Build complete (60.89s)
- `scripts/test.sh --filter SocketServer` — exit 0, 6 tests in 2 suites passed
- mutation run above — exit 1, 5 tests in 1 suite, 1 failure (the new test)
- `swiftlint --strict` — exit 0, 0 violations in 925 files

cheapsteak:
Second review round: both MEDIUMs are fixed, and the minor is answered with a measurement rather than taken. All in ba3de812.

**Finding 2 — the "called once" invariant is now enforced.** `boundSocketIdentity` is read and cleared in one step under an `NSLock`, through `takeBoundSocketIdentity()`; `start()` records it through the matching `storeBoundSocketIdentity(_:)`. You were right that the doc comment was asserting something nothing held up: two undeduplicated `Task { await daemon.stop() }` handlers could both observe the identity, and the loser's unlink is aimed at a path a successor may hold by then.

The discriminating test is `onlyOneOverlappingShutdownClaimsTheFile`: eight concurrent claimants, exactly one may come away owning the file. The read-and-clear window is a few instructions wide, so threads-and-hope proves nothing — a `duringIdentityClaim` seam (`nil` in production, alongside the existing `beforeAdoptingBoundSocket` one) holds the window open until every thread has reached the call site. Under the lock the other seven wait outside the window and get nil; with the plain read-then-nil all eight read the same identity.

**Finding 1 — fixed, not deferred. Both unlinks are gone.** I took the fix route, and the reason is that the sound fix turned out to be smaller than the deferral: neither site needs to remove the socket at all. The only thing that needs the rendezvous path free is `bind(2)`, and `SocketServer.start()` clears the path itself immediately before binding — on every path that reaches it, including these two. In the daemon that bind is the same startup a few seconds later; in the app it is the daemon it goes on to spawn. So removing the socket buys nothing when the pid file really is stale, and when it is stale only because a successor has bound without having rewritten it yet, it costs a live daemon its rendezvous and puts nothing in its place — empty for the whole of the next startup, and for good if that startup throws first.

That also retires the reason the PR gave for deferring. Porting `SocketFileIdentity` into `TBDShared` would have made the removals *safe*; deleting them makes them *unnecessary*, which is less code and a stronger invariant: clearing the rendezvous path is the act of whoever is about to bind it, and of nobody else. (A `connect(2)` probe stays unsound for the reason #802 measured.)

Pinned by `onlyTheBinderClearsTheRendezvousPath`, which requires that nothing in TBDDaemon or TBDApp outside `SocketServer` removes `TBDConstants.socketPath`. It is a source-text pin because that path resolves from a process-global this test target must not mutate — `ConstantsTests` pins that no suite in this process sets `TBD_SOCKET_PATH`, and `scripts/test.sh` sets it for the whole run. `cleanupRemovesAStalePidFile` is the positive control that the pid-file cleanup still cleans.

**Minor — `fstat(boundFD)` is not free, it is wrong on darwin.** Worth checking before taking: `fstat(2)` on a bound `AF_UNIX` descriptor does not report the filesystem entry `bind(2)` created. It reports the socket's own in-kernel vnode, with `st_dev` of -1. Measured on darwin 25.1 with both a C and a Python probe on the same socket:

```
fstat dev=-1        ino=15302332
lstat dev=16777235  ino=863366896
```

An identity captured that way could never equal the one read back from the path, so every ownership check would fail closed and the socket file would never be reclaimed at all — worse than the misattribution it was meant to prevent. The measurement is now recorded on `SocketFileIdentity` so it is not re-proposed. The residual lstat-window remains documented and accepted.

**Verification.** `/opt/homebrew/bin/swiftlint --strict` — 0 violations, 0 serious, 925 files, exit 0.

The suite ran on CI rather than locally: the machine-global build slot was held for this whole session by a wedged SwiftPM test process from another lane (0% CPU, no compiler leaves, an idle run loop and nothing else on its stack), and it was not this lane's process to kill. Two runs, both the full suite:

- [33934559476](https://github.com/cheapsteak/tbd/actions/runs/33934559476) — this branch at ba3de812.
- [33934365862](https://github.com/cheapsteak/tbd/actions/runs/33934365862) — `preflight/socket-claim-discriminator`, which is ba3de812 with the three production fixes undone and the tests kept. That is the discrimination proof: `onlyOneOverlappingShutdownClaimsTheFile` and `onlyTheBinderClearsTheRendezvousPath` must fail there.

One thing to flag for whoever merges: GitHub did not schedule the `pull_request` Test run for this push (only `Claude Code Review` fired), so the required `test` check still points at the previous commit. Both runs above are `workflow_dispatch`, which does not report as that check — the PR will want a re-run before merge.

cheapsteak:
Verification correction and follow-up, now that the machine-global build slot freed up — the CI runs linked in my previous comment were cancelled and replaced by local runs, which are better evidence anyway. Head is now 30502805.

**Local, through `scripts/test.sh`:**

- Fixes in place — `--filter 'SocketServerSocketOwnershipTests|PIDFileStaleTests'`: **12 tests in 2 suites passed** in 0.330 s, exit 0.
- Fixes reverted, tests kept — same filter: **12 tests, 2 issues**, exit 1.
  - `only one of many overlapping shutdowns claims the socket file` — `Expectation failed: (claims.count → 2) == 1`.
  - `nothing outside SocketServer removes the rendezvous socket file` — `offenders → ["PIDFile.swift:63: try? FileManager.default.removeItem(atPath: TBDConstants.socketPath)", "AppState.swift:2953: ..."]`.
- `scripts/swift-safe build --build-tests`: exit 0. `/opt/homebrew/bin/swiftlint --strict`: 0 violations, 0 serious, 925 files, exit 0.

**One thing found along the way that is worth knowing.** The claim-once tests as first written wedged the whole test process, three runs in a row, and it was not the environmental teardown hang it looked like: `MultiThreadedEventLoopGroup.shutdownGracefully()` never completes when the group is already shut down, so a **second `stop()` on one `SocketServer`** — overlapped in a task group, or merely sequential — suspends its task forever. By then the event-loop threads have exited, so what is left is a process at 0% CPU with a parked main run loop, two threads, and no output at all, which is indistinguishable at a glance from the Foundation `waitUntilExit` teardown hang.

What production overlaps is two shutdown *tasks* racing for the identity, and the contested step is the claim — so the tests contend for `takeBoundSocketIdentity()` and `unlinkOwnedSocketFile()` directly and call `stop()` at most once per server. Recorded in the tests and on `unlinkOwnedSocketFile()` so it is not re-introduced.

Also: GitHub did not schedule the `pull_request` Test run for either push on this branch (only `Claude Code Review` fired), so the required `test` check still points at 1f155b01. It will want a re-run before merge.

cheapsteak:
Round three addressed in `a342c83a`. Both findings were correct — and the MEDIUM one was correct about more than it claimed, because the mechanism round two recorded was wrong.

## MEDIUM — fixed, and it was never `shutdownGracefully()`

I went to find which call actually wedges before writing a guard, and round two's attribution does not survive contact with swift-nio 2.96.0:

- **`MultiThreadedEventLoopGroup.shutdownGracefully()` is safe repeated *and* concurrent.** Its own doc comment says so ("Even though calling `shutdownGracefully` more than once should be avoided, it is safe to do so and execution of the `handler` is guaranteed"), and the implementation backs it: every call takes `shutdownLock` and switches on `runState` — `.closing` appends the handler to the pending-callback list, `.closed` dispatches it immediately, and only a caller that found `.running` drives the shutdown and completes the rest from its `g.notify` block (`MultiThreadedEventLoopGroup.swift:404-511`). Concurrent callers serialize on that lock. So the review's open question — "is truly concurrent `shutdownGracefully()` even safe?" — resolves the benign way, but it was never the thing hanging.
- **`channel.close()` is what hangs.** Off the event loop it is `eventLoop.execute { self.close0(mode:promise:) }` (`ChannelPipeline.swift:788`), and `SelectableEventLoop._schedule0` on a loop whose group has shut down takes the `!validExternalStateToScheduleTasks` branch: it prints "Cannot schedule tasks on an EventLoop that has already shut down. This will be upgraded to a forced crash in future SwiftNIO versions." and returns **without enqueuing anything** (`SelectableEventLoop.swift:480-492`). The close promise is never fulfilled, so `try await channel.close()` suspends for good — and the event-loop threads have already exited, which is exactly the 0%-CPU-and-nothing-to-see symptom round two saw and misfiled. (The branch already recorded this same mechanism, on `beforeAdoptingBoundSocket`, for why the failed-start window needs an injected seam.)

**The fix.** `SocketServer.stop()` and `HTTPServer.stop()` now run through a shared `ShutdownLatch`: the first caller runs the body, every later caller — sequential or concurrent — `await`s that same run. Awaiting rather than returning early keeps `stop()` meaning "the socket file has been settled" for everyone who called it, and the task is unstructured on purpose so a cancelled signal handler cannot abandon a shutdown other callers are waiting on.

`HTTPServer` is in deliberately, not as scope creep. `Daemon.stop()` stops the socket server and *then* the HTTP server, which had the byte-identical body — so fixing only `SocketServer` would have handed the same hang to the next line and the second `Daemon.stop()` still would never have reached `Foundation.exit(0)`. That inference is the one the review flagged as unproven; only fixing both makes it true.

**One measurement that changed the test's shape, and is worth flagging.** Two `stop()` calls that merely *start* together are **not** the dangerous case: both get their `channel.close()` in while the event loop is still alive, so that ordering passes even un-latched. I found this the hard way — my first version of the overlapping test passed against the reverted fix. The case that hangs is a `stop()` arriving after another has passed `shutdownGracefully()`, so the test now stages exactly that: the `duringIdentityClaim` seam holds the first shutdown at its reclaim step until the second caller is at `stop()`.

**Tests, all through production `stop()`:**

- `overlappingStopsBothFinishAndClaimTheFileOnce` — the staged overlap above; both calls must return, the socket file must be claimed exactly once, and the file must be gone. The seam now carries the identity the claimant came away with, which is how "claimed exactly once" is observed from inside a real `stop()`; it has no other outward sign.
- `repeatStopLeavesTheSuccessorAlone` — the sequential half a signal escalation actually produces, and now a real second `stop()` rather than `unlinkOwnedSocketFile()` (which is private again).
- `repeatHTTPServerStopReturns` and `latchRunsItsBodyOnceAndEveryCallerWaits` — the sibling server, and the latch itself (eight callers, one run, nobody returning before that run finished).

All of them wait on detached tasks with a deadline rather than awaiting directly or gathering a task group, because the failure under test is a shutdown that never returns: a task group would wedge the whole run alongside it instead of reporting it.

**Discriminating proof** — with only the two `stop()` bodies un-latched and everything else kept: **19 tests in 3 suites, 3 failures**, one per shutdown path, each its 15 s deadline expiring on a `stop()` that never returned; `latchRunsItsBodyOnceAndEveryCallerWaits` still passes there because it exercises the latch directly. Restored, the same 19 pass in 0.363 s.

## MINOR — premise corrected, removals kept

Both counterexamples confirmed and now written down at both sites and in the description, in place of the blanket claim:

- `AppState.startDaemonAndConnect()` returns early when no `TBDDaemon` binary is found and when `process.run()` throws — before the daemon that would bind exists.
- `Daemon.start()` runs `pidFile.cleanupIfStale()` at step 2 and binds the socket hundreds of lines later, with plenty of `throw`s in between.

So a stale socket file can outlive the cleanup. The removals stay, and the comments now say *why* that is acceptable rather than claiming the bind always follows: an unbound socket file behaves to a client exactly like a missing one — `connect(2)` fails either way and the next daemon to bind clears the path regardless — whereas unlinking a *live* successor's socket silently strands a working daemon with nothing to restore the path.

## Commands

- `scripts/swift-safe build --build-tests` — Build complete
- `scripts/test.sh --filter 'SocketServerSocketOwnershipTests|PIDFileStaleTests|ServerShutdownLatchTests'` — exit 0, **19 tests in 3 suites** passed in 0.363 s
- same filter, latches reverted — exit 1, 19 tests in 3 suites, **3 issues**
- `swiftlint --strict` — exit 0, 0 violations in 939 files

cheapsteak:
## The two CI failures were scheduling latency, not a wedged shutdown

Pushed `66a5db81`. Diagnosis first, since the elapsed times in the log are misleading.

### What the failing run actually says

Run 33937652520, `fast parallel pass 1/2`: **5,059 tests in 445 suites, one process**, cooperative pool three threads wide on the runner.

- **`the latch runs its body once…` failed with exactly ONE issue.** Its `runs.count == 1` expectation *passed*. So the shutdown body did start and did take the latch — what it did not do inside 15 s was finish the hundred `Task.yield()`s that follow. Each yield is a full trip to the back of a run queue holding thousands of runnable tasks. Reported state: `only 0 of 8 callers returned`.
- **`two overlapping shutdowns…` failed with `claims.count == 0`** and the socket file still on disk. The shutdown never reached its reclaim step at all.
- The two tests that exercise the second-`stop()` NIO wedge head-on — `a second shutdown of the HTTP server returns` and `a second shutdown after a successor took over…` — **passed**.

That last pair, plus `claims.count == 0`, refutes the NIO wedge as the cause. The wedge produces the opposite signature: the first `stop()` runs to completion and claims the file, and only a *second* one hangs. `claims.count == 0` means no `stop()` got as far as the claim. And the latch test touches no NIO whatsoever.

### Reproduced locally

Restored the pre-fix code and ran the whole daemon target the way CI does, on a 12-core box, so the process carried the same 5,059-test population:

```
Test "the latch runs its body once, and every caller waits for that one run"
  recorded an issue at ServerShutdownLatchTests.swift:61:9:
  Expectation failed: await waitUntil({ returned.count == callers }, timeout: .seconds(15))
  ↳ only 0 of 8 callers returned
Test run with 5059 tests in 445 suites failed after 162.355 seconds
```

Same line, same message, same single issue. Those same tests take **0.4 s** when the run is filtered to just these two suites, and **56–62 s** inside the broad pass — a ~150x scheduling-latency swing on a machine four times wider than the runner. The socket test did not go red locally (12 cores vs 3), but its CI signature is unambiguous on its own.

### The fix

The 15 s bound was the binding constraint. This repo has sized that budget once already for the same reason — `ciSafeDeadline` went 30 s → 90 s as the single-process population grew — and the number lives in `TestDeadlines.saturatedPass`. These waits now go through `waitFor`, which defaults to it and puts the observed state on the primary failure line instead of the trailing `↳`.

Two things change besides the number:

- **The latch body yields 20 times, not 100.** One yield already opens the window the test needs (a caller that failed to wait for the run would return while the body is in flight); a hundred only made the test's cost track how loaded the machine is.
- **`two overlapping shutdowns…` no longer blocks a cooperative-pool thread.** It staged its overlap by blocking inside `duringIdentityClaim` reached from a plain `Task.detached` — the hazard `Tests/CLAUDE.md` names under "Thread-blocking gates run off the cooperative pool". Both `stop()` calls now start with `gateHoldingTask`, since either may be the one that wins the latch and runs the body, and every semaphore wait goes through `waitForGate`, so a handshake that never completes names itself instead of owning a thread for good. That is also the stronger half of the fix: the shutdown now gets a thread the tests own, rather than queueing behind the pass.

### Verification

- `scripts/swift-safe build --build-tests` — exit 0
- `scripts/test.sh --parallel --filter 'ServerShutdownLatchTests|SocketServerSocketOwnershipTests'` — exit 0, **11 tests in 2 suites** passed
- `scripts/test.sh --parallel --filter '^TBDDaemonTests\.'` — exit 0, **5,059 tests in 445 suites** passed (167 s, 71 known issues, the run's usual baseline)
- `swiftlint --strict` — exit 0, 0 violations in 939 files

cheapsteak:
## The staging gate was starved, not wedged — and `gateHoldingTask` could never have covered it

Run `33940811838` recorded the same issue **12 times**, all within 5 ms of each other, at `SocketServerSocketOwnershipTests.swift:282` — the staging gate in *"two overlapping shutdowns both finish, and the socket file is claimed once"*. The gate's own text blamed a plain `Task` pinning a cooperative-pool thread. That text was written for a different failure, and it sent the last fix to the wrong place.

### It was not wedged

Everything else in that test passed in the same run. The failing run's tally is `83 issues (including 71 known issues)` — the 12 non-known issues are exactly the 12 copies of this one gate. So:

- both `stop()` calls returned inside the outer **90 s** wait (no failure at the observation),
- exactly one shutdown came away owning the socket file,
- the file was gone afterwards.

`claims.count == 1` can only be true if the first shutdown actually reached `takeBoundSocketIdentity()`, and the gate is signalled on the way in. So the first shutdown *did* reach its reclaim step — after 30 s, before 90 s. The handshake was late, and the inner bound was the only thing that noticed.

### Why it is late, and why the previous fix could not help

`SocketServer.stop()` hands its body to `ShutdownLatch`, which runs it in an **unstructured `Task`** — deliberately, so a cancelled signal handler cannot abandon a shutdown other callers are awaiting. [SE-0417](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0417-task-executor-preference.md) carries a task executor preference into child tasks (`async let`, task groups) and into default actors, and **not** into `Task {}` or `Task.detached`.

So the work that signals the gate is scheduled on the cooperative pool however the test starts `stop()`. Wrapping both `stop()` calls in `gateHoldingTask` covered their own frames and nothing past the latch. Against a parallel pass that admits all ~5,000 tests at once on a three-thread runner, 30 s is simply not enough scheduling latency.

That is a *scheduling* cost, not a pinned thread, and no call-site change can move it. Only the bound was ours to get right — and it was the asymmetry: the outer waits had been raised to `TestDeadlines.saturatedPass` (90 s) while this inner gate stayed at 30 s, so the inner one reported starvation the outer one was sized to tolerate.

### Why «unknown», and why twelve copies

One cause, both symptoms. `Test.current` and `Configuration.current` are **task-locals**, and the wait ran on a bare `Thread.detachNewThread`. With no `Test.current` Swift Testing can name no test; with no `Configuration.current` it will not drop the event either, so it posts it to *every* registered event handler — one expiry, several identical lines. (Contrast the `EventDrivenTestClockSelfTests` entries in the same log: 32 recordings for 32 genuine task-group children, attributed to a named test.)

### What changed

- The staging gate takes `TestDeadlines.saturatedPass` (90 s); the observation that must dominate it takes `TestGate.deadline` (120 s). That ordering keeps a genuinely lost handshake reporting as a lost handshake rather than as a wedge.
- The waiting side, and the eight claimant threads in the sibling test, moved from `Thread.detachNewThread` to `gateHoldingTask` — still off the cooperative pool, and being tasks they carry the two task-locals, so an expiry names the test and prints once.
- `BoundedGateWaitTests` now **pins the SE-0417 split** rather than trusting anyone to remember it: inside a `gateHoldingTask`, the operation's own frames run on a thread named `GateExecutor.threadName`, and a nested `Task {}` does not. If Swift ever changes that, the assertion goes red and every bound sized for the cooperative pool is re-derivable.
- The gate's timeout text now names both causes — a holder that pinned a pool thread, and a release that was merely starved because an unstructured `Task` sits on its path — with the different fix each needs. `Tests/CLAUDE.md` carries both rules.

### Verification

- `scripts/swift-safe build --build-tests` — exit **0**, `Build complete!`
- `scripts/test.sh --parallel --filter '^TBDDaemonTests\.'` — exit **0**, `Test run with 5060 tests in 445 suites passed after 122.755 seconds with 71 known issues`; zero `✘` lines, zero «unknown». The socket-ownership suite passed in 27.8 s and the new SE-0417 pin in 22.4 s.
- `swiftlint --strict` — exit **0**, `0 violations in 939 files`.

**Not reproduced locally, and that is expected.** The pre-fix head ran the same whole target green here (5059 tests, 118 s): this machine's cooperative pool is far wider than the runner's three threads, which is the entire variable. The CI evidence above establishes the mechanism without needing a local red.

To be explicit about the framing: CI is under real strain right now, but this is **not** somebody else's environmental failure. The bound is one my own work added, it was under-sized against a limit the repo had already measured and named, and it is mine to fix.

PR #807, opened 2026-09-05
Feat: Run a terminal session on its own pty holder instead of a tmux window

cheapsteak:
## Summary

TBD can now run a terminal session on its own **pty holder** instead of a tmux
window. A tiny holder process calls `forkpty`, owns the pty master, spawns the
job on the slave, and hands duplicates of the master to whoever needs to read or
write it — never reading it itself. A viewer panel opens on a live agent
session; closing the panel hands the session back to the daemon, which keeps a
headless emulator running so nothing is lost while nobody is watching.

This is the transport half of moving off tmux. It ships behind
`pty_holder_enabled`, which defaults **off**: with the flag off, sessions take
the tmux path exactly as before.

## Why this shape

Design: [`docs/specs/2026-08-30-pty-holder-session-transport-design.md`](docs/specs/2026-08-30-pty-holder-session-transport-design.md).

The invariant everything else follows from: **reading a pty master is
exclusive.** Two readers each get an arbitrary half of the byte stream, with no
way to recover the interleaving. So exactly one process reads a session at a
time, ownership moves by an explicit handshake, and the holder — which could
have read it — deliberately never does. Writing is not exclusive, which is why
the write path gets its own arbitration rather than inheriting the read path's.

A second consequence worth naming: **draining is a liveness obligation.** A job
that exits with unread output cannot finish exiting, because the kernel waits on
the terminal's output queue. That is why the daemon keeps reading a detached
session rather than simply letting it sit.

## What this PR does

- **`TBDHolder`** — the holder executable: `forkpty`, own the master, spawn the
  job, hand descriptors over `SCM_RIGHTS`, answer a handshake, and outlive the
  daemon that spawned it.
- **`HolderRegistry`** — adoption, attach and handback, generation-guarded
  viewer attachments, and seizing a session back from an app confirmed dead
  (and only a confirmed-dead one).
- **`HolderReader`** — the daemon's headless emulator for detached sessions.
- **`TerminalPanelView`** — a holder-backed panel takes its own `dup` of the
  master, tracks panel geometry onto the pty, and detaches on close.
- **`OutgoingInputQueue`** — the single serialization point for everything one
  panel writes: the person at the keyboard and the daemon's injections. While a
  paste is open it holds an injection rather than writing it, so a paste that
  closes within the hold's bound never has one land between its markers, where
  the child would absorb it as pasted text. The bound is what the guarantee is
  traded against: a paste that never closes gets the injection written into it
  and logged, because absorbed-and-visible beats held-forever, and the bound
  stays strictly shorter than the daemon's ack deadline so the daemon never
  starts writing into open pastes systematically.
- **`ProcessIdentity`** — one shared identity verdict for "is that still the
  process I think it is", used by the reaper and the liveness arbiter with
  different tolerances and the reasoning for each stated.
- **`SocketSIGPIPE`** — per-socket `SO_NOSIGPIPE` at both socket-creation
  funnels. Process-wide `SIG_IGN` was rejected: it inherits across fork/exec,
  and the app forks shells.
- **Close-on-exec on every descriptor that names a session.** A session's pty
  master must not survive an `exec` in either process. Both processes stamp the
  flag at their *receive chokepoint*, before the descriptor reaches any waiter:
  `FDSidecarClient`'s receive loop in the app — which covers the holder attach
  and the control-mode attach alike, because both register through `expectFD`
  and are settled by that one loop — and `HolderClient.readMoreFrames` in the
  daemon. Stamping it there rather than at each waiter is what makes the
  guarantee structural: it also covers the stale-vend path, the leftovers
  drained on disconnect, and any waiter added later. The flag stays out of the
  shared `FDChannel` transport deliberately, so a caller that means to hand a
  descriptor to a child keeps that option; a canary test in each target pins
  that `SCM_RIGHTS` still delivers an inheritable descriptor, so neither
  receive-side assertion can pass vacuously. Beyond the receive chokepoints, the
  app takes its write copy with `F_DUPFD_CLOEXEC`, and the daemon does the same
  on the copy `HolderReader` makes for a viewer, on the holder connection
  itself, and on the sidecar's listener and accepted connections. `dup(2)` is
  never used for any of them, because it clears the flag on the copy whatever
  the original carries. The failure it prevents is quiet: `HolderSpawner` is the
  only spawn path that asks for `POSIX_SPAWN_CLOEXEC_DEFAULT`, and every other
  child the daemon or the app starts — git, tmux probes, usage and PR-status
  fetchers, peer supervisors, a local-PTY panel's `forkpty` — inherits whatever
  is open and inheritable at that instant. One such child holding a copy keeps
  the session open for as long as it lives, so the reader sees no EOF when the
  job exits, death-detection-by-EOF stops working, and the handback never
  completes.
- **Close-on-exec on the app's two daemon-facing sockets**, for the same reason
  at the other end of the same wire. `DaemonClient.makeConnectedSocket` (one per
  RPC, ~200 call sites plus a two-second poll) and `FDSidecarClient` both minted
  plain `socket(2)` descriptors, and SwiftTerm's `Pty.fork(andExec:)` — started
  for every tmux-backed panel — is a `forkpty` whose child does `chdir` then
  `execve` with **no close sweep**, so every viewer process held a copy. The
  sidecar is the sharp one: EOF on the accepted sidecar connection is the
  daemon's *only* app-death trigger (there is no periodic liveness sweep), so
  after an app crash seizure waited on the last inheritor to exit. Bounded
  rather than permanent — those children die when their masters close — but it
  is the same defect this PR fixed on the daemon's end of that connection, and
  it was left unfixed on the app's. `AppSocketInheritanceTests` covers both,
  with a canary row pinning that `socket(2)` still returns an inheritable
  descriptor so neither assertion can pass vacuously.
- **The handback suite creates its own descriptors close-on-exec.** Every test
  target links into one process, so `HolderDetachHandbackTests` runs beside two
  suites that keep `/bin/sleep 120` children alive for the rest of the run. The
  attach's `fcntl` cannot run before the fixture has built its `socketpair`, so
  a fixture that left the flag to the attach left a window in which a sibling
  spawn inherited the vended end and held it — and the ordering row then read
  `.stillOpen` about somebody else's child rather than about the panel. The
  fixture sets the flag at creation and the one row whose subject *is* the flag
  clears it deliberately.

## Also on this branch: agent-authoring tooling

Two files here belong to no part of the transport and change no product
behavior: `.claude/agents/decision-explainer.md` and
`.claude/skills/explaining-decisions/SKILL.md`. They are a subagent definition
and the methodology it follows for writing an HTML explainer that lets a human
make one technical decision. They were added at the repo owner's explicit
request while this branch was open, as authoring tooling for how technical
decisions get explained, and they ride along here rather than in a branch of
their own for that reason alone.

They are additive and inert unless invoked: nothing the daemon, the app, or the
CLI runs reads them, and no shipped behavior changes whether they are present or
absent. They carry no spec deliberately. The brainstormed-spec rule exists for
changes to our theory of the system or the product; a prompt file that shapes
how an agent writes a document is neither, so there is nothing here for a
reviewer to be asked to accept about how TBD works.

## Flag & rollout

Five flags, **all default off**, each added as a tri-state column with no SQL
`DEFAULT` so "nobody has chosen" stays distinguishable from "somebody chose no":

- **`pty_holder_enabled`** — the transport itself. Everything else here is inert
  while this is off.
- **`gc_holder_rendezvous_enabled`** — `OrphanGC`'s leg over holder rendezvous
  files (the socket a SIGKILLed holder could not unlink, plus its lock and log).
  Keep-biased for young sockets.
- **`gc_rowless_holders_enabled`** — `OrphanGC` deleting holder sessions with no
  database row.
- **`reap_holder_children_enabled`** — `AgentReaper`'s holder leg. It sweeps by
  each session's recorded child pid rather than by enumerating a tmux server's
  children, because a holder's job re-parents to launchd and no tree walk
  reaches it. It verifies process identity before signalling and keeps whenever
  identity is uncertain.
- **`holder_row_reconcile_enabled`** — reconciling holder rows against reality.

**Graduation is gated on the write-completion work in #808**, which stacks on
this branch and finishes a daemon injection that the pty refuses part-way.
After that lands, the plan is to soak `pty_holder_enabled` on one machine, then
the reconciler flags individually, and flip defaults one at a time.

## Reconcilers

Per the named-reconciler doctrine, the durable resources this introduces and who
reclaims them:

- **Holder rendezvous files** (`~/tbd/holders/`) — `OrphanGC`, behind
  `gc_holder_rendezvous_enabled`.
- **Holder child processes** — `AgentReaper`'s holder leg, behind
  `reap_holder_children_enabled`.
- **Holder sessions with no row** — `OrphanGC`, behind
  `gc_rowless_holders_enabled`, for a holder that is **still alive**: it
  recovers the child pid from the holder's own handshake, then kills the job and
  the holder.

**One case is genuinely uncovered, and no reconciler is claimed for it.** A
row-less holder's job, *after the holder itself has died*, is reclaimed by
nothing. The handshake that would name the child pid is exactly what is gone:
`RowlessHolderCollector` reads `.noListener` and keeps,
`HolderRendezvousCollector` unlinks that holder's files and signals nothing, and
both `AgentReaper`'s holder leg and the reconcile arm work from session rows —
the half that by definition does not exist here. The job re-parents to launchd
with no record of it anywhere. Reaching it would need a pid recorded somewhere
outside both the row and the holder (the rendezvous directory is the obvious
candidate), which is a design change rather than an extra sweep, so it is
disclosed here instead of half-built. `HolderSpawner` and
`AgentReaper.sweepHolderChildren` both say so in their own comments.

## Assumptions

- Assumes a holder that completes a handshake and returns a matching owner token
  is ours. The default `TBD_HOME` is shared by every checkout on a machine, so
  "reachable but absent from my database" is also what a foreign healthy session
  looks like — the owner token is what separates them. If that token stops being
  minted per installation, reclamation starts eating other checkouts' sessions.
- Assumes the app's main actor makes progress. A wedged-but-alive app keeps its
  claim on a session; the seize path deliberately fires only on confirmed death.

## Known gaps

Stated here rather than discovered later:

- **`--keys` refuses on the holder transport.** No shipped rail regresses —
  every internal named-key actuator is already gated off holder upstream — but
  it is a real capability gap for humans and future scripted rails.
- **A dead app's disconnect can be missed.** If a crash and relaunch land inside
  one actor hop, the dead app's identity is gone and there is no backstop; those
  sessions stay claimed for the daemon's lifetime. Low reachability, no fix
  attempted.
- **A holder-backed agent session that ends is deleted where a tmux one is
  parked.** With `gc_rowless_holders_enabled` also on, a wrong deletion is worse
  than losing an inventory entry. A job-still-running gate stands behind it.
- **No CLI surface for the transport flag itself.** The four reconciler flags
  each have a switch — `tbd gc holders`, `tbd gc rowless-holders`, `tbd gc
  holder-children`, `tbd gc holder-rows`. `pty_holder_enabled` has none: it has
  an RPC method (`config.setPtyHolderEnabled`) and nothing that calls it, and
  `tbd config set` is not where it would live anyway — that command takes three
  keys (`auto-archive-on-merge`, `auto-hibernate-on-merge`, `update-mode`), none
  of them a feature gate. Starting the transport soak therefore means speaking
  to the daemon socket directly.

## Evidence & verification

Every task in this milestone went through a fresh implementer, an independent
review, and a scoped re-review of the fixes. Beyond that:

**Live verification, on a real session rather than a harness.** With
`pty_holder_enabled` on, a holder-backed panel rendered a live Claude Code TUI
— banner, model line, cwd, composer, status line — where a placard used to be.
Three things were measured rather than inferred:

- **Geometry tracks the panel.** Read from the child's own tty
  (`stty -f /dev/ttysNNN size`): launched at 220×50, attached with the sidebar
  open at 170×47, sidebar collapsed at 216×47. No mis-wrapping after switching
  away and back at the wider geometry.
- **Ownership moves on detach.** `terminal.output` returns "No live holder
  reader" exactly while the app is attached — panel open, and still open after
  navigating away, which is correct because the panel stays mounted. Closing
  the pane returns a live screen, proving the daemon re-adopted.
- **The session survives the round trip.** Row present, holder process alive,
  child process alive afterwards.

**Six defects in the implementation plan were caught by implementers checking
rather than complying** — a wrong test path four times, a step that referenced
a later pass that did not exist, an instruction that would have stranded live
sessions, a step that should not have been built, and one instruction not
implementable as written.

**Why there is no whole-process descriptor audit.** Close-on-exec has now been
found missing in five places by four different methods, so a single test that
enumerates this process's descriptors (`proc_pidinfo(PROC_PIDLISTFDS)`) and
requires every pty and `AF_UNIX` socket to carry `FD_CLOEXEC` was designed and
then rejected as unsound. Every test target links into one process and Swift
Testing runs suites in parallel across all of them, and at least five sibling
suites in three targets — `FDChannelTests`, `FDVendingServerTests`,
`AttachRPCTests`, `HolderProtocolTests`, `SocketSIGPIPETests` — hold `AF_UNIX`
`socketpair` ends with no close-on-exec discipline, while several more hold
SwiftTerm `forkpty` masters, which SwiftTerm returns inheritable and this repo
does not control. Those descriptors cannot be attributed to a suite: a
socketpair has neither a local nor a peer path in `PROC_PIDFDSOCKETINFO`, a pty
master's `/dev/ttysNNN` names the device rather than its creator, and the kernel
reports no per-descriptor owner. The only allowlist broad enough to stop the
audit reddening on other suites' descriptors is "unnamed sockets and all ptys",
which is precisely the shape it was meant to catch — a test that cannot fail.

A sound version needs the audit to be the only thing in its process: a small
executable target that stands up the fixture and audits its own table, spawned
by a test. Run-level `--no-parallel` plus close-on-exec discipline in every
suite of every target would also do it, but neither half is enforceable from
inside a test and both rot silently.

What is sound is narrowing the audit to descriptors a test can *name*, and that
is what `AppSocketInheritanceTests` does: `DaemonClient(socketPath:)` takes a
socket path unique to the test, and `FDSidecarClient.adopt` takes the descriptor
itself. Both rows redden if the production line is deleted and neither can be
tripped by a concurrent suite. That is per-site rather than class-closing. The
mechanism that would close the class is a SwiftLint custom rule over
`socket(AF_UNIX`, `accept(`, `openpty(`, `posix_openpt(` and `dup(` in
`Sources/` — deterministic, static, immune to test scheduling, and running
already in the `lint` job and the pre-push hook. It is worth building, and it is
not built here: it would flag roughly fifteen live sites across `TBDHolder`,
`TBDPeerHelper`, `TBDDaemon` and `TBDCLI`, several of which hand a descriptor to
a child *deliberately* (`BoundedProcessRunner`'s and `TmuxControlConnection`'s
`openpty` replicas, the holder's own `dup`). Each needs a real judgment about a
fork path this PR does not otherwise touch, which belongs in a change of its
own rather than as a rider here.

**Not covered:** streaming content under sustained load. Nothing currently
exposes a per-session environment hook on the holder path, and setting a global
override on a machine running many live agent sessions was not a reasonable
thing to do for a test.

## Summary by CodeRabbit

* **New Features**
  * Added holder-backed terminal sessions with live attachment, resizing, snapshot replay, and session handback.
  * Added paste-aware terminal input delivery with ordered queuing and fallback handling.
  * Added terminal snapshot serialization for screen content, modes, scrollback, and cursor state.
  * Added opt-in holder-session reconciliation controls and a command-line switch.
  * Added an AI decision-explainer workflow that produces evidence-based HTML explainers.

* **Bug Fixes**
  * Improved recovery from disconnected clients and closed sockets.
  * Prevented duplicate terminal readers and unsafe session reclamation.
  * Improved terminal input delivery reliability and failure reporting.
  * Improved descriptor cleanup and terminal state preservation during handoffs.


cheapsteak:
Addressed the MEDIUM finding on `OutgoingInputQueue.swift:90` (`pasteHoldBound` vs `HolderInjectionCourier.ackDeadline`) in 59fa16ff.

**Mechanism: both, because neither alone is enough.**

A shared constant removes the drift between two literals but cannot stop someone editing the one line that now holds both; a test detects drift but cannot stop two modules from disagreeing about which values they are even talking about. So:

1. **`HolderInputTiming` in TBDShared** now owns `pasteHoldBound` (2 s) and `injectionAckDeadline` (5 s), side by side, with the full statement of the invariant on the type. Both are the *defaults* of the existing initializer parameters — the injection seams are untouched, and every test that wants a bound of its own still supplies one.
2. **Three tests, each catching the drift direction that is actually unsafe.**
   - `HolderInputTimingTests` asserts `pasteHoldBound < injectionAckDeadline` on the production constants. No local `let`, nothing supplied by the test.
   - `OutgoingInputQueueTests.unclosedPasteDoesNotStrandInjectionForever` now **default-constructs** the queue (it previously supplied `.seconds(5)`, which pinned nothing) and advances by the shared constant: a shipped hold longer than the constant leaves the injection held and hangs out to `.clockDriven`; a shorter one releases during the short-of-the-bound advance and fails there.
   - `HolderInjectionRoutingTests` gets the same two-sided pin on the courier — the missing-ack test default-constructs it, and a new *"The ack deadline is not reached early"* answers one instant short of the deadline and requires `.viewerWrote` with `lateAcksObserved == 0` and no daemon write. The harness lost its own `ackDeadline: Duration = .seconds(5)` default for the same reason: a harness-supplied value would have made those tests prove nothing about the shipped one.

Together those pins say the two shipped defaults *are* the shared constants, which is what makes the ordering assertion bind real code rather than two numbers in a test file.

**Proof the tests discriminate** — one run with three deliberate mutations applied (constants inverted to 5 s / 2 s, plus both call sites regressed to literals of their own), filtered to the three suites:

- `The paste hold expires strictly before the injection-ack deadline` — failed: `(HolderInputTiming.pasteHoldBound → 5.0 seconds) < (HolderInputTiming.injectionAckDeadline → 2.0 seconds)`. This one is reddened by the inversion alone; it reads nothing else.
- `The ack deadline is not reached early` — failed with 3 issues (delivery was `.daemonWrote(.ackDeadlineElapsed)`, a late ack was counted, the daemon wrote).
- `A paste that never closes does not strand an injection forever` — failed with 3 issues (the injection was released during the short-of-the-bound advance).

Run: `21 tests in 4 suites failed ... with 7 issues`, exit 1. Reverted, same filter: `21 tests in 4 suites passed`, exit 0.

**Prose rewritten rather than left stale.** The "shortening the hold bound is not an alternative way to satisfy the invariant" argument moved intact onto `HolderInputTiming.pasteHoldBound`; both file headers now point at the shared type instead of restating each other's magnitudes (the courier's section heading is now "what it must be **longer** than", which is what it actually governs); and `docs/specs/2026-09-03-single-typist-injection-design.md`'s `OutgoingInputQueue.swift:109` pointer follows the constant to TBDShared.

Verification: `swift-safe build --build-tests` complete; `swiftlint --strict` 0 violations in 964 files; filtered suite green as above.

cheapsteak:
The `HolderDetachHandback` CI failure — the ordering row reading `.stillOpen` where it requires `.closed` — was descriptor inheritance, not timing.

The panel took the vended pty master as it arrived (inheritable) and made its write copy with `dup(2)`, which clears close-on-exec on the copy whatever the original says. This app spawns children constantly: every local-PTY panel is a `forkpty`, and sibling suites keep `/bin/sleep 120` children alive for a whole run. Any child that starts while a holder panel is attached inherits a copy of that session's master and holds the session open for as long as it lives — so the far end reads no EOF after the handback even though the panel closed both of its own copies in the right order.

That is a product bug as much as a test artifact: an inherited copy keeps a live session open past the handback and leaves the panel that closed its tab no longer the last writer.

Fix in `344b8d9d`: `FD_CLOEXEC` on the attach in `TerminalPanelView`, and `F_DUPFD_CLOEXEC` instead of `dup(2)` for the write duplicate. It adds a deterministic row that asserts the flag directly rather than leaving it to whichever test the spawn timing happens to redden — and first checks the fixture really hands over an inheritable descriptor, so it cannot pass vacuously.

Verified locally: build with test targets clean, `HolderDetachHandbackTests` 9/9 green (including the previously failing ordering row), SwiftLint `--strict` clean.

Shared with the stacked PR — the same commit landed there, since it is milestone-B-only code.

cheapsteak:
Two review findings addressed, plus the CI failure that turned out to be a third thing.

## HIGH — the daemon had the same fd-inheritance defect, in three more places

Fixed, following `344b8d9d`'s shape so both sides read the same way:

- **`HolderClient.readMoreFrames`** now sets `FD_CLOEXEC` on every descriptor a `recvmsg` returns. Darwin has no `MSG_CMSG_CLOEXEC`, so a descriptor crossing `SCM_RIGHTS` arrives inheritable however carefully the sender flagged its own — and this one becomes `HolderReader`'s fd and lives for the session's whole life.
- **`HolderReader.duplicate()`** now takes the viewer's copy with `F_DUPFD_CLOEXEC` instead of `dup(2)`, which clears the flag on the copy whatever the original carries.
- **`HolderClient.connectIfNeeded`** — same family, and load-bearing on its own: a holder learns its single client slot is free by reading EOF, so an inherited copy of that socket would have every later verb answered with the busy sentinel by a holder whose real client let go minutes earlier.

### What the sweep for other sites found

- `Darwin.dup(` in `Sources/TBDDaemon` — **one** occurrence, `HolderReader.duplicate()`. Now gone.
- `recvmsg` / `SCM_RIGHTS` — one receive path, `FDChannel.receiveMessage`, reached from `HolderClient` (daemon) and `FDSidecarClient` (app). Left as-is on purpose: a receiver that always set the flag would be deciding for callers that might legitimately want to pass a descriptor on. The two callers that hold session ptys both set it themselves.
- `FDVendingServer` — its listener and its accepted client connections are the channel session ptys are vended over and live as long as the daemon does. **Fixed** (`fcntl` after `socket` and after `accept`; Darwin has no `accept4`).
- `SSHAgentResolver.canConnect`, `ShadowPeerAttribution`, `ShadowPeerReconciler` — short-lived probe sockets, `defer close(fd)` in the same function, connect-and-drop. A spawn landing inside that window would inherit one, but the socket names nothing durable and nothing depends on its EOF. Named rather than fixed.
- `SocketServer` / `DaemonClient` — NIO and the app's RPC dial; not session descriptors.
- `Foundation.Process()` — 23 call sites across 18 files (git, tmux probes, usage and PR-status fetchers, SSH-agent resolution, `PeerLinkSupervisor`, `ProviderEventsSupervisor`, `ShadowPeerHelperProcess`, `TmuxControlConnection`, `UpdateLauncher`, `DaywatchRunner`). Several are long-lived. `HolderSpawner` remains the only path asking for `POSIX_SPAWN_CLOEXEC_DEFAULT`. These are the inheritors, not the bug — the fix is at the descriptors.

**Verdict:** closed for descriptors that name a session. Narrower, not closed, for the short-lived probe sockets above, which are named here rather than skipped silently.

### The test

`HolderDescriptorInheritanceTests` (3 rows, `TBDDaemonTests`, no spawned holder, 0.008 s). Both assertion rows fail when the corresponding production line is reverted — measured, not asserted:

```
✘ the handed-over pty is close-on-exec by the time a caller sees it
✘ the copy vended to a viewer is close-on-exec
Test run with 3 tests in 1 suite failed with 2 issues
```

Neither can pass vacuously. A third row pins the platform fact they rest on — a descriptor sent with `FD_CLOEXEC` explicitly set arrives at the receiver **without** it — so if the transport ever started delivering close-on-exec descriptors, that row reddens and says the other two have become tautologies.

## The CI failure was neither of those — it was the fixture's own window

`theDescriptorIsClosedBeforeTheDetachIsSent` failed on run 33939878330 at head `344b8d9d`, which already carried the panel-side fix. The daemon-side fixes above cannot explain it and are ruled out structurally, not by luck: that test's descriptor comes from a `socketpair` created inside its own fixture and handed to the panel by a stub. No daemon, no `HolderClient`, no `HolderReader`, no `SCM_RIGHTS`.

The panel fix was right and incomplete. It sets `FD_CLOEXEC` inside the attach — and the fixture builds the pair before the attach is even scheduled. Every test target links into one process, Swift Testing runs their suites in parallel, and two suites in that same target (`TerminalTeardownReapTests`, `QuietIngestTests`) keep `/bin/sleep 120` children alive for the rest of the run, started through `LocalProcess.startProcess`, which is a `forkpty` that closes nothing. A spawn landing in that window inherits the vended end and holds it until the run ends; the far end then never reports a close however correctly the panel ordered its own. The row asks whether the panel let go and is answered about somebody else's child.

The fixture now sets the flag on both ends at creation, and the one row whose subject *is* the flag clears it deliberately so it still starts from the state a real hand-over arrives in.

`anInheritableDescriptorIsHeldOpenByAChildSpawnedBesideIt` pins the mechanism instead of leaving it inferred: the same `/bin/sleep` spawn run twice, and the far end reports the close only when the descriptor was close-on-exec. Each leg is the other's control, and it reproduces with no load at all — which is the point, since a filtered run of the ordering test is green and says nothing.

## MEDIUM — the agent/skill files

Named in the PR description now, under **Also on this branch: agent-authoring tooling**: what the two files are, why they are on this branch, and why they need no spec. They are a subagent definition and its methodology for writing decision explainers, added at the repo owner's explicit request; they are additive, inert unless invoked, and change no product behavior. The brainstormed-spec rule is for changes to our theory of the system or the product, and a prompt file that shapes how an agent writes a document is neither. My read is that they genuinely need no spec; if a reviewer disagrees, that is a call for a human rather than something I should write unprompted.

## Verification

- `scripts/swift-safe build --build-tests` — exit 0.
- `scripts/test.sh --filter 'HolderDetachHandbackTests|HolderDescriptorInheritanceTests'` — exit 0, **18 tests in 3 suites** passed.
- `scripts/test.sh --filter 'HolderDescriptorInheritanceTests'` with each production fix reverted — exit 1, 2 issues, as quoted above.
- `scripts/test.sh --filter '^TBDAppTests\.'` — **3175 tests in 331 suites**, 14 issues of which 11 are known-issue rows. The 3 unknown are all `TerminalLockedAccessTests` 60 s time-limit exceedances, the load artifact this box is known for. `theDescriptorIsClosedBeforeTheDetachIsSent` passed in that pass, alongside the `sleep 120` suites.
- `swiftlint --strict` — exit 0, 0 violations in 964 files.

cheapsteak:
## HIGH — the app did not set `FD_CLOEXEC` on a vended pty — fixed in `21908fe5`

The finding was correct and the description overstated what was verified. `handleFDVend` passed the raw descriptor to its waiter with no `fcntl` anywhere in `FDSidecarClient`; on the holder-attach path the flag was not applied until `TerminalPanelView.startHolderClient:810`, after a `CheckedContinuation.resume`, an async return and a hop onto `@MainActor`. `HolderClient.readMoreFrames` did it correctly, which is what made the symmetric claim look true.

**Where the flag now goes, and why there.** `FDSidecarClient.receiveLoop`, immediately after `FDChannel.receiveMessage` returns and before any frame is dispatched — mirroring `HolderClient.readMoreFrames` at the daemon's chokepoint. Stamping it in the loop rather than at each waiter is what makes it structural: it covers the routing-key demux, the stale-vend path that just closes the fd, the leftovers drained on disconnect, and any waiter added later.

**Control-mode: yes, covered.** `DaemonClient.openAttach` and `HolderAttachClient.attach` both register through `fdSidecar.expectFD`, land in the same `waiters` dictionary, and are settled by that single receive loop. There is one receive path, so one stamp covers both. The test drives both key shapes.

**The judgment call — kept out of `FDChannel`, deliberately.** The premise behind the original placement ("every session-pty caller sets it itself") was false, so it deserved re-deciding rather than restating. Three things decided it:

- **No current caller depends on inheriting a *received* descriptor** — checked every `receiveMessage` call site in `Sources/` and `Tests/`. So the transport option is not blocked by a dependency; it is declined on cost, not necessity. The places that genuinely want inheritance across `exec` (`HolderLock.makeInheritableAcrossExec`, `HolderSpawner`'s fixed lock slot) open their descriptors locally and never receive them over `SCM_RIGHTS`.
- **Setting it in the transport makes both receive-side assertions vacuous.** `HolderDescriptorInheritanceTests` already carries a canary whose failure message says exactly that, and it exists because the daemon-side assertion is only meaningful while the transport stays neutral. Moving policy into `TBDShared` would silently change the holder process too, for a fix whose whole scope is one app-side loop.
- **The forgettability that caused this is addressed at the right level.** The bug was not "a caller forgot" so much as "the flag was set at a *use site* instead of a *receive site*". There are exactly two receive loops that carry session ptys; both now stamp in the same place, with the same reasoning written down, and each is pinned by a test.

**Discriminating test** — `Tests/TBDAppTests/FDSidecarClientTests.swift`:

- `a vended descriptor is close-on-exec by the time its waiter sees it` — vends two descriptors, one per production caller's key shape (holder session, control-mode pane), with `FD_CLOEXEC` explicitly **cleared on the sender's copy** first, and asserts both arrive close-on-exec at their waiters.
- `the transport delivers what it was handed, inheritable` — the anti-vacuity companion: sends a descriptor flagged close-on-exec and receives it through bare `FDChannel.receiveMessage`, asserting it arrives inheritable. `FD_CLOEXEC` is a property of the descriptor, not the open file, and `SCM_RIGHTS` does not carry it — so a pass on the row above can only be the client's doing.

Proof it discriminates: with the one-line `fcntl` removed from the receive loop and nothing else changed, the suite fails with exactly 2 issues, both new assertions:

```
Expectation failed: isCloseOnExec(holderFD → 19)
  a holder session's pty reached its waiter inheritable, …
Expectation failed: isCloseOnExec(controlFD → 6)
  a control-mode attach's pty reached its waiter inheritable — …
Test run with 14 tests in 1 suite failed after 0.364 seconds with 2 issues.  (exit 1)
```

The line restored, the same 14 tests pass (exit 0).

**Description corrected.** The sentence "The app sets `FD_CLOEXEC` on the descriptor it is vended…" now reads that both processes stamp the flag at their *receive chokepoint* before the descriptor reaches any waiter — naming `FDSidecarClient`'s receive loop and `HolderClient.readMoreFrames`, saying that it covers the control-mode attach for the shared-loop reason, and recording that the flag stays out of `FDChannel` with a canary in each target pinning that.

**Verification** — every command with its real exit code:

- `scripts/swift-safe build --build-tests` — exit 0.
- `scripts/test.sh --filter 'FDSidecarClientTests|HolderDescriptorInheritanceTests|HolderDetachHandbackTests'` — **32 tests in 4 suites passed**, exit 0.
- The without-the-fix run above — 14 tests, 2 issues, exit 1.
- `swiftlint --strict` — 0 violations, 0 serious, 964 files, exit 0.

No live-tier suites were run for this change; it touches one receive loop and its unit tests. `ArchivedWorktreeSearchTests` / `AppearanceDebounceTests` remain red on this branch and are untouched by this commit — pre-existing and load-related, not caused here.

cheapsteak:
Addressed the two MEDIUM findings plus the independent review, in `e8287965`.

## Prose that described unbuilt design as shipped

**The transport spec's "Input is not arbitrated, but it is serialized"** described a single-injector mechanism with "no acknowledgement, deadline, or fallback writer" whose frames and verbs exist nowhere in `Sources/`. It now describes `HolderInjectionCourier`: route by attach state, `injection` frame and `injectionAck` while attached, fail open on `injectionAckDeadline` and duplicate rather than drop, and the one attached state that does *not* fail open (a confirmed attach has closed the daemon's descriptor, so the fallback reports rather than writes). The single-typist design is now cited explicitly as the target, marked as not the mechanism above.

**`HolderSpawner` vs `AgentReaper` — the reaper's comment was the false one.** Worked out from the code rather than split: `RowlessHolderCollector` recovers the child pid from the holder's *own handshake*, so a `.noListener` socket is kept and nothing is reclaimed; `HolderRendezvousCollector` unlinks that holder's files and signals nothing; `AgentReaper.sweepHolderChildren` and the reconcile arm both read session rows. So a row-less holder's job **after the holder has died** is covered by nothing. `HolderSpawner`'s statement was accurate and is now sharper; `AgentReaper`'s claim that `OrphanGC` covers the case is corrected to name the live-holder qualifier it was missing. The case is disclosed in the PR description's Reconcilers section as genuinely uncovered, with the reason a sweep cannot reach it.

**"Graduation is gated on the write-completion work in this PR (see below)"** — that work is #808, which stacks on this branch, and "see below" pointed at nothing. Fixed.

**"`tbd config set` accepts only two keys"** — it takes three (`auto-archive-on-merge`, `auto-hibernate-on-merge`, `update-mode`; `update-mode` arrived in #805), and the gap was overstated in the other direction too: four of the five flags have `tbd gc` switches. Only `pty_holder_enabled` has no CLI surface. Rewritten.

**`OutgoingInputQueue`'s "an injection is never written between a paste's markers"** — `forceDeliver` writes it anyway once `pasteHoldBound` elapses with the paste still open. The guarantee now states the bound it is traded against, and why the bound is not a weakening: a paste that never closes would otherwise strand the injection forever, and absorbed-and-visible beats held-forever. The strictly-shorter-than-the-ack-deadline relationship is stated in the same place.

## The file-descriptor gap

`DaemonClient.makeConnectedSocket` and `FDSidecarClient` both minted plain `socket(2)` descriptors. **Verified the claim about SwiftTerm rather than trusting it** — `.build/checkouts/SwiftTerm/Sources/SwiftTerm/Pty.swift`, `fork(andExec:args:env:desiredWindowSize:)`:

```
let pid = forkpty(&master, nil, nil, &desiredWindowSize)
if pid < 0 { return nil }
if pid == 0 {
    if let cCurrentDirectory { _ = chdir(cCurrentDirectory) }
    _ = execve(cExecutable, cArgs.base, cEnv.base)
    _exit(127)
}
```

Optional `chdir`, then `execve`. No close sweep, no `closefrom`, no `FD_CLOEXEC` applied to anything, and no signal reset either. So the child inherits every inheritable descriptor open in the app at that instant, and one is started per tmux-backed panel. Both sockets now get `FD_CLOEXEC` at creation — before `connect`, which can block — and the sidecar gets it again in `adopt`, the funnel every sidecar socket passes through. `makeConnectedSocket` became internal so `AppSocketInheritanceTests` can read the flags off the descriptor it returns against a listener of its own; that also retired the "deliberately untested" note on its `SO_NOSIGPIPE` line.

## The whole-process descriptor audit: designed, then rejected as unsound

Not shipped, and the reasoning is in the PR description so it does not have to be rediscovered.

Every test target links into one process and Swift Testing runs suites in parallel across all of them. At least five sibling suites in three targets — `FDChannelTests`, `FDVendingServerTests`, `AttachRPCTests`, `HolderProtocolTests`, `SocketSIGPIPETests` — hold `AF_UNIX` `socketpair` ends with no close-on-exec discipline, and several more hold SwiftTerm `forkpty` masters that SwiftTerm returns inheritable and this repo does not control. Those descriptors cannot be attributed: a socketpair has neither a local nor a peer path in `PROC_PIDFDSOCKETINFO`, a pty master's `/dev/ttysNNN` names the device rather than its creator, and the kernel reports no per-descriptor owner. The only allowlist broad enough to stop it reddening on other suites' work is "unnamed sockets and all ptys" — exactly the shape it was meant to catch. That is a test that cannot fail, so it was not written.

A sound version needs the audit to be the only thing in its process: a small executable target that stands up the fixture and audits its own table, spawned by a test. Run-level `--no-parallel` plus close-on-exec discipline in every suite of every target would also do it, but neither half is enforceable from inside a test.

What *is* sound is narrowing to descriptors a test can name, which is what `AppSocketInheritanceTests` does — a socket path unique to the test, and a descriptor handed in directly. Per-site rather than class-closing.

**On the SwiftLint rule: yes, worth building — as its own change.** It is deterministic, static, and immune to test scheduling, which is everything the audit test is not. It would flag roughly fifteen live sites across `TBDHolder`, `TBDPeerHelper`, `TBDDaemon` and `TBDCLI`, and several of them hand a descriptor to a child *deliberately* (`BoundedProcessRunner`'s and `TmuxControlConnection`'s `openpty` replicas, the holder's own `dup`). Each needs a real judgment about a fork path this PR does not otherwise touch, and getting one of them wrong breaks a spawn path silently. That belongs in a change a reviewer can look at on its own.

## Verification

- `scripts/swift-safe build --build-tests` — exit 0. The seven warnings are pre-existing (unhandled-resource notices and a SwiftTerm `Color: Sendable` restatement).
- `scripts/test.sh --filter` over `AppSocketInheritanceTests`, `SocketSIGPIPETests`, `FDSidecarClientTests`, `HolderDetachHandbackTests`, `HolderInjectionDeliveryTests`, `OutgoingInputQueueTests`, `OutgoingInputQueuePasteMarkerTests`, `HolderDescriptorInheritanceTests` — exit 0, **61 tests in 9 suites**.
- `swiftlint --strict` — exit 0, 0 violations in 965 files.

PR #808, opened 2026-09-05
Fix: A paste into a holder-backed session is finished rather than truncated at 1 KB

cheapsteak:
## Summary

Paste a long prompt into an agent session that TBD renders through the pty
holder — a full-screen TUI, so the pty is in raw mode — and until now only the
first kilobyte arrived. A raw-mode pty master accepts 1,022 bytes and then
refuses; the app wrote once, waited 20 ms for room that a busy agent was never
going to make, and reported the whole write a failure with that first kilobyte
already committed. What the person saw was a prompt cut off mid-sentence, and,
because the paste's closing `ESC[201~` was part of what got dropped, a composer
left stuck in bracketed-paste mode.

After this change the panel finishes every write it starts. What the kernel
refuses is kept in an outbox, every later byte from every stream — the rest of
the paste, the keystrokes typed after it, a daemon injection released from the
paste hold — queues behind it whole and in arrival order, and the queue drains
whenever the pty says it can take more. A paste is no longer truncated, the end
marker is never lost, and a keystroke typed behind a stalled paste lands after
it rather than inside it.

The visible trade is that keystrokes are now *held* during a stall where they
were previously *dropped*. That is the intended trade, and a panel-level
backpressure banner is what makes it legible: "This session is not accepting
input — 4 KB waiting", shown only once a stall has outlasted a second, and
cleared when the outbox empties. It deliberately says nothing about loss,
because nothing is lost.

## Why this shape

Spec: [`docs/specs/2026-09-04-pty-write-completion-design.md`](docs/specs/2026-09-04-pty-write-completion-design.md).
All ten of its "decisions that need a human" were settled before implementation
and each resolved to its own recommendation.

The load-bearing decisions:

- **The bound is the descriptor, never a clock.** An episode ends on `EIO` (the
  last slave closed) or `EBADF` (teardown) and on nothing else. A clock that
  expired on a full queue could not write the released bytes either — it could
  only cut the payload with somebody else's bytes in the gap, which is the
  failure this change exists to remove.
- **"Written" means accepted by a writer that will complete or report.** That
  is the meaning the seam already carried on its `localProcess` and sidecar
  arms, where bytes go to `DispatchIO` and to a client send queue and neither
  can report a failure that happens after the call returns. The pty arm now
  joins them: a kernel refusal acks `written: true` because the app is that
  writer. Only a genuinely absent transport — no descriptor, an exited child,
  a disconnected sidecar — acks `false`.
- **No interrupt bypass.** `0x03` queues with everything else. It could not
  have been written any sooner in any case: the queue that refused everything
  else refuses it too.
- **The outbox is `[Data]`, not one buffer.** Contiguity comes from the FIFO;
  concatenating would copy a multi-megabyte paste on every append.
- **The drain notifier is armed only while the outbox is non-empty.** See the
  measurements below — this is a correctness requirement, not tidiness.

## What this PR does

- `PTYWrite.all` collapses to one non-blocking `write(2)` with no `poll`, and
  reports three outcomes: `.complete`, `.partial(written:)`, and
  `.failed(errno:written:)`. The 20 ms budget parameter is deleted rather than
  defaulted to zero — a parameter nobody passes, over a loop nothing reaches,
  is a hazard waiting for someone to restore it.
- `OutgoingInputQueue` gains the outbox, the hold, `drain()`, the episode
  diagnostic and the backpressure edges. It stays a `@MainActor final class`
  with synchronous, hop-free `enqueueUserBytes` / `beginUserPaste` /
  `endUserPaste`; appending to the outbox is O(1) and is the only new work a
  keystroke does.
- New `OutgoingDrainNotifier` seam plus `WriteSourceDrainNotifier`, a
  `DispatchSource` write source on the main queue over the panel's own pty
  descriptor.
- `TerminalPanelView`: `performOutgoingWrite` and `writeToHolderPTY` return an
  attempt outcome instead of `Bool`; the notifier's lifecycle is wired into
  `startHolderClient` / `stopHolderReader`, cancelled before the descriptor
  closes; the backpressure banner and its plumbing.
- `TerminalBackpressurePresentation` — the banner's copy as a pure function of
  the byte count, so a test can pin it without driving SwiftUI.
- Tests: `RawPTYPair` (a real pty pair, raw-mode and unread, in-process with no
  child), `OutgoingOutboxTests` (the outbox against a capacity-bounded fake —
  no pty, no timing), `HolderWriteOutboxTests` (the panel end to end against a
  real pty), `OutgoingDrainNotifierTests`, and the `PTYWrite`-level tests in
  `HolderInjectionDeliveryTests` following the outcome collapse.
- Prose: `holderWriteFD`'s doc, and the courier's `viewerReportedNothingWritten`
  doc, which described `false` as covering "every refusal the app can know
  synchronously" — a kernel refusal now acks `true`.

**Deliberately not changed:** `HolderReader.writeAll` keeps its 250 ms budget
and stays honest. The daemon's detached path still writes through
`HolderReader.write` on the daemon's own descriptor, and gives up on a payload
that does not land inside that budget — so the truncation this PR removes for
an attached panel remains real for a detached session. The daemon does not get
an outbox in this branch, and there is no holder `write` verb: the single-typist
spec designs one, and nothing in `Sources/` implements it
(`HolderProtocolVersion.current` is 1 and `HolderProtocol.swift` has no such
case). Closing the detached side is that spec's work, not this PR's.

## Flag & rollout

**No new flag.** This ships under `pty_holder_enabled`
(`Config.ptyHolderDefault`, `Sources/TBDShared/Models.swift:1698`), which is
default-off and has never shipped on. No new config column, no new
`UserDefaults` key.

Both branches of that gate stay testable, because both still exist: a tmux
panel's write path is untouched, and a holder panel's is this one. Adding a
second column would have created a broken quadrant — "holder on, outbox off" is
the truncating arrangement this change removes — and would double the soak for
a path with no users to protect.

Graduation is `pty_holder_enabled`'s own graduation; this changes nothing about
that plan.

## Reconcilers

No new kind of durable resource, and no new creation path to one. The outbox is
memory in the app, owned by a descriptor whose close drops it, inside a panel
whose teardown is already ordered: the drain notifier is cancelled by the same
teardown that closes the descriptor it watches, and `stopHolderReader` cancels
before the close rather than after. An app that dies takes its outbox with it,
which is the interim residue named below and not an orphan for a sweep to find.

## Assumptions

- **The panel's pty descriptor is `O_NONBLOCK`.** The flag lives on the open
  file description the daemon opened and rides the `dup`. A blocking descriptor
  would park the main actor inside `PTYWrite.all` instead of returning a
  refusal. If that ever stops being true, keystroke latency breaks, not
  correctness.
- **The kernel cuts a write wherever it runs out of room**, including inside a
  multi-byte UTF-8 sequence or a paste marker. Nothing tries to choose the cut.
  What heals it is the hold: the next accepted bytes continue the same stream
  with nothing written in between.
- **A viewer that stops draining stalls its own writes.** While attached, the
  panel is the session's reader, so an agent blocked on output does not read
  its input. Completion is coupled to the app's liveness for everything past
  the first kilobyte; delivery and non-duplication are not.

## The interim ack residue, stated plainly

With the interim ack, a kernel refusal reports `written: true` to the daemon,
because the bytes are with a writer that will complete them. **If that writer
goes away before the outbox drains, those bytes are lost and the daemon has
already been told they landed.** The actuation row reads dispatched for a
message that did not fully arrive.

**It is not only app death.** `stopHolderReader` calls
`outgoingQueue.shutdown()`, which drops the outbox on *every* panel teardown —
closing the tab, detaching, or a refused `attach.ready`. So an injection acked
`true` with its remainder queued behind a stalled child is lost when a person
closes that tab, with the app perfectly alive and the daemon re-adopting a
session that never learns. Supervision nudges are aimed at busy agents, and
closing a busy agent's tab is an ordinary thing to do, so this is the more
reachable half of the residue rather than a corner of it.

The event is not silent — the daemon's app-death path seizes the session
(`HolderRegistry.swift:1457`, `:1519`), so app death itself is loud — but the
row is wrong, and nothing reconciles it. This is the same residue the
`localProcess` and sidecar arms already carry under the same `true`, now
extended to cover a remainder rather than only an in-flight handoff.
`2026-09-03-single-typist-injection-design.md` deletes it outright, by deleting
the viewer's injection write.

### The spike, and what it settled

Before any of this was built, a throwaway spike (compiled with `swiftc`
directly, nothing committed) measured a real pty on this kernel. Verdict A:
**write-readiness fires cleanly on a `dup` of a pty master.** The shipped
notifier is therefore `WriteSourceDrainNotifier`, a `DispatchSource` write
source on the main queue.

- **Raw-mode ceiling: 1,022 bytes** (`TTYHOG − 2`), then `EAGAIN` immediately —
  identical across six runs of six. This is what makes `RawPTYPair`
  deterministic. No test hard-codes the number; each learns its own kernel's
  ceiling.
- **The source fires after the child drains: 3/3.** The load-bearing question.
- **The source is quiet while the queue is full and nothing drains: 0 fires,
  3/3** — but only a *fresh* source armed after the queue is already full
  answers that question. Arming a source that was suspended while the queue was
  empty reports exactly one fire, a resume-edge artifact that is neither
  "silent" nor "spinning".
- **An armed source over an idle, writable pty with a handler that writes
  nothing fires 44,486 / 182,060 / 47,845 times per second.** This is why
  disarm-on-empty is a correctness requirement rather than an efficiency one:
  an outbox left armed while empty is the main queue at 100% for as long as the
  panel is open. Every path that can empty the outbox disarms on that edge, and
  a test walks all four of them (nothing was ever refused, drained to empty,
  transport died, teardown).
- **With a production-shaped handler** — drain until `EAGAIN` on every fire —
  it self-limits to a few hundred main-queue hops per second carrying 1.3–1.9
  KB each: 247 hops / 428 KB, 187 / 239 KB, 507 / 939 KB. That confirms the
  spec's cost claim of about one hop per KiB the child reads.
- **`poll(POLLOUT)` reports readiness on the same `dup`** (`revents = 4`). The
  `poll`-thread notifier therefore stays viable as a fallback. It was not
  needed and is not built; the `OutgoingDrainNotifier` protocol is the seam
  that would take it.
- **The suspend/resume balance.** Only `suspend → cancel → resume → release`
  survives. Releasing a source whose suspend count is non-zero traps with
  SIGTRAP 133, and so does over-resuming. After `suspend → cancel`, the cancel
  handler does not run until a rebalancing `resume()` lands — a deferred-close
  ordering hazard the teardown path respects.

### Tests

Three tiers, each discriminating:

- **No pty, no timing** — the outbox against a fake transport with a byte
  capacity, which models a full tty input queue exactly. The remainder is kept,
  later chunks are held in order and never overtake it, a drain delivers every
  byte exactly once, the arm and disarm edges fire once per episode, an
  injection released by `endUserPaste` lands after an end marker still queued,
  a dead transport drops the outbox and logs once, a short write never logs a
  transport failure, and the banner appears only after the threshold and clears
  on drain.
- **A real pty** — `RawPTYPair`, raw-mode and unread, in-process with no child
  (all test targets compile into one process and Swift Testing runs suites in
  parallel, so `forkpty` here would be a multithreaded fork and a stray process
  on a shared machine). The three-chunk paste, the keystroke behind it, the
  interim ack matrix, teardown with a remainder outstanding.
- **The notifier itself** — arm/disarm idempotence and the teardown balance.

Three properties are deliberately untested, each with a comment saying why
rather than an assertion that greens either way: the absence of the `poll` (any
threshold separating 0 ms from 20 ms is inside the noise of a loaded machine —
what replaces the test is the type, which has no budget parameter and no
`poll`); and two named in the plan at Task 3 and Task 6.

### Follow-up filed, not built

**A detach with a pending end marker leaves the child mid-paste.** Decision #5
accepts this for now: the remainder is dropped at teardown, and if it held the
closing `ESC[201~` the child stays in bracketed-paste mode until something else
closes it. File it as an issue if the soak shows it is felt.

**An explicit "still draining" answer for the single-typist paste lease.** The
lease's expiry now has a second reading — a viewer that is alive with a large
paste still draining into a stalled child. It resolves the same way as the
first (the daemon writes on its bound, the injection is absorbed into the
person's paste, visible and never doubled), and the refinement waits on a soak
showing the residue is felt. Recorded in
`docs/specs/2026-09-03-single-typist-injection-design.md`.

---

## Summary by CodeRabbit

- **New Features**
  - Terminal input now preserves ordering and completes partially accepted writes, including large pastes and injected input.
  - Added a passive backpressure indicator showing pending terminal input.
  - Launch data is transferred through a protected channel without exposing request contents in process arguments.

- **Bug Fixes**
  - Improved handling of stalled, disconnected, and closing terminal sessions, with clearer completion, failure, and exit-status reporting.

- **Documentation**
  - Added specifications covering terminal write completion, input sequencing, backpressure, and injection behavior.

### DIFF
--- a/Sources/TBDApp/Terminal/OutgoingDrainNotifier.swift
+++ b/Sources/TBDApp/Terminal/OutgoingDrainNotifier.swift
@@ -1,0 +1,98 @@
+import Darwin
+import Dispatch
+import Foundation
+import os
+
+private let logger = Logger(subsystem: "com.tbd.app", category: "outgoingDrain")
+
+/// Tells `OutgoingInputQueue` when the session's pty can take more.
+///
+/// **Armed only while the outbox is non-empty**, and that is measured, not
+/// stylistic: readiness notification is level-triggered, so a notifier left
+/// armed over a writable descriptor and an empty outbox fires tens to hundreds
+/// of thousands of times per second (44,486 / 182,060 / 47,845 across three
+/// measured runs). The queue disarms on the edge where the outbox empties, and
+/// the implementation below makes `arm`/`disarm` idempotent so that edge can be
+/// signalled more than once without unbalancing anything.
+@MainActor
+protocol OutgoingDrainNotifier: AnyObject {
+    /// Start reporting readiness. Idempotent.
+    func arm()
+    /// Stop reporting it. Idempotent.
+    func disarm()
+    /// Permanent teardown. After this the notifier never calls back, whatever
+    /// the descriptor does. Idempotent.
+    func cancel()
+}
+
+/// A `DispatchSource` write source on the main queue. Its handler is main-actor
+/// by construction (the source's queue is `.main`), so `assumeIsolated` is
+/// sound and the drain runs on the same executor as every append.
+///
+/// ## The suspend count is the whole safety story
+///
+/// A `DispatchSource` is created suspended. `resume()` and `suspend()` are a
+/// **counted** pair, and libdispatch traps (SIGTRAP 133) on the release of a
+/// source whose count is non-zero — as it does on an over-resume. Every one of
+/// these was measured fatal: releasing while suspended, `suspend → cancel →
+/// release` with no rebalancing resume, two suspends against one resume, and
+/// `resume` on an already-running source.
+///
+/// So this type keeps the count in {0, 1} and tracks which it is. `isArmed`
+/// is not a convenience: it is what makes a second `arm()` or a second
+/// `disarm()` a no-op instead of a crash, and it is what tells `cancel()`
+/// whether a resume is owed.
+@MainActor
+final class WriteSourceDrainNotifier: OutgoingDrainNotifier {
+    private var source: (any DispatchSourceWrite)?
+    /// `true` when the source is running (suspend count 0), `false` when it is
+    /// suspended (count 1). A source starts suspended, so this starts `false`.
+    private var isArmed = false
+
+    init(fileDescriptor: Int32, onWritable: @escaping @MainActor () -> Void) {
+        let source = DispatchSource.makeWriteSource(
+            fileDescriptor: fileDescriptor, queue: .main)
+        source.setEventHandler {
+            MainActor.assumeIsolated { onWritable() }
+        }
+        self.source = source
+    }
+
+    func arm() {
+        guard let source, !isArmed else { return }
+        isArmed = true
+        source.resume()
+    }
+
+    /// Called on the edge where the outbox empties, and it must not be skipped:
+    /// see the fires-per-second measurement in this file's protocol comment.
+    func disarm() {
+        guard let source, isArmed else { return }
+        isArmed = false
+        source.suspend()
+    }
+
+    /// **Cancel, then rebalance to a suspend count of zero, then release.**
+    ///
+    /// A notifier at teardown is usually *suspended* — the ordinary state is an
+    /// empty outbox — and releasing it there is the SIGTRAP. The resume after
+    /// the cancel is what returns the count to zero; it cannot deliver an
+    /// event, because a cancelled source's event handler does not run and this
+    /// call is already on the source's own serial queue.
+    ///
+    /// It also has a second job. After `suspend → cancel`, a cancel handler
+    /// does **not** run until the rebalancing resume lands (measured: still
+    /// unrun 0.5 s later). This type installs no cancel handler precisely so
+    /// nothing important waits on that; if one is ever added — to close a
+    /// descriptor, say — it will not run until this resume, and anything
+    /// ordered after it must account for the deferral.
+    func cancel() {
+        guard let source else { return }
+        self.source = nil
+        source.cancel()
+        if !isArmed { source.resume() }   // count back to zero before release
+        isArmed = true                    // running, cancelled, and safe to drop
+        logger.debug(
+            "writeSourceDrainNotifier: cancelled, suspend count rebalanced before release")
+    }
+}
--- a/Sources/TBDApp/Terminal/OutgoingInputQueue.swift
+++ b/Sources/TBDApp/Terminal/OutgoingInputQueue.swift
@@ -41,6 +41,34 @@ private let logger = Logger(subsystem: "com.tbd.app", category: "outgoingInputQu
 /// called, and nothing can interleave between them, because there is no
 /// suspension point for anything to interleave at.
 ///
+/// ## The outbox, and what it holds behind it
+///
+/// A pty master takes 1,022 bytes in raw mode and then refuses, so a paste
+/// above a kilobyte into a child that is not reading is written short. Nothing
+/// can un-write the prefix and nothing can ask the pty how much room it has,
+/// so the only repair is to finish the write: **whoever committed the prefix
+/// owns the remainder, and nothing else may reach that pty until it has
+/// landed.**
+///
+/// The remainder goes in `outbox`, and every later chunk — the rest of the
+/// paste, the keystrokes after it, an injection released from the paste hold —
+/// is appended behind it, whole, in arrival order. `drain()` finishes it when
+/// the transport says it can take more. Two consequences worth stating
+/// outright:
+///
+/// - **Ordering is FIFO across both streams.** A keystroke typed after a short
+///   paste lands after the paste's end marker. That is what keeps a child from
+///   being left in bracketed-paste mode with the person's typing absorbed into
+///   the paste.
+/// - **Held, never dropped, and never bypassed.** Nothing the person typed is
+///   discarded while the panel owns the pty, and no byte jumps the queue —
+///   including `0x03`, which could not be written any sooner anyway, since the
+///   queue that refused everything else will refuse it too.
+///
+/// The outbox has no size cap: the paste that filled it was already accepted at
+/// the view, and the bytes are a copy in this process's memory. Its bound is
+/// the descriptor — `EIO` or `EBADF` — never a clock.
+///
 /// ## What is guaranteed, and what is not
 ///
 /// The guarantee is a **safety** guarantee, not an ordering one: **while a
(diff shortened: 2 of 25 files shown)


cheapsteak:
## `drainsAFinishedJobsOutputBeforeReleasingIt` — diagnosis and fix

The quiet-pass failure (`(status → .exitedStatusUnknown) == .exited(code: 7)`,
run 33947165902) is a **real bug, and a pre-existing one** — not a regression
from this branch, and not load noise. Fixed here.

### What `.exitedStatusUnknown` means, and where it comes from

It is the holder protocol's "the session is over and nobody collected how"
(`HolderChildStatus.exitedStatusUnknown`). Four producers; only two can reach
this test:

- `Holder.status(fromWaitpidStatus:)` — the child was killed by a signal, so
  there **is** no exit code and one is not invented.
- `HolderRegistry.exitProbeOutcome(for:)` — the reclaimer's probe got `ENOENT`
  or `ECONNREFUSED` at the rendezvous, so the holder is gone and nothing will
  ever be learned about that child again.

The failure is the second.

### The interleaving

A holder pushes the terminal status at whichever client is connected the moment
it reaps, and the *very next thing* it does is break its serve loop, close its
listener and unlink its rendezvous (`Holder.serve`: `pushExitToConnectedClient`
sets `exitCollected`, and `window.isArmed && exitCollected` breaks). The
reclaimer's probe connects the instant the drain runs dry — a moment *before*
the child is reapable — so it is the client that push lands on, and it then
finds nothing underneath it.

That push answers no request, so `HolderClient`'s send barrier retires it
rather than serving it as a reply: into `lastPushedDescription`, whose own doc
comment says throwing it away "would lose the fact entirely".
`confirmChildExit` never read it. It classified the failed round trip
(`peerClosed`) as `.retry`, reconnected to a path the holder had already
unlinked, read the `ENOENT` as absence — correctly, in isolation — and recorded
`exitedStatusUnknown` for a child that had exited 7 and said so.

Measured, deterministically, by holding the pushed-to connection by hand:

```
socket file still present: false
describe on the pushed-to connection: peerClosed
lastPushedDescription: exited(code: 7)
fresh probe: cannotConnect(..., errno: 2)      # ENOENT
exitProbeOutcome(fresh): established(exitedStatusUnknown)
```

### Not from this branch

`HolderRegistry`, `HolderClient`, `HolderReader` and the holder's `serve` state
machine are **byte-identical** to `main` on this branch
(`git diff origin/main e15b49bd -- …` is empty). The `--launch-fd` change moved
the launch payload off argv and touches neither the reclaim path nor the
descriptor the child ends up with. The defect arrived with the reclaimer itself
(#786) and reaches production, not only the test: an ordinary exit code is what
the activity rail's liveness and the reconcile verdict read.

### Reproduction rate

- Isolated, idle: **0 / 8** runs of `HolderReclaimTests`; **0 / 120** iterations
  of the scenario.
- Under the whole tier-3 pass, serial: **1 / 120** iterations — same wrong
  value, same shape.
- With the fix: **120 / 120** clean under the same load.

### The fix

`confirmChildExit` reads the push. A holder that pushed has answered — it
observed its own child, said how it ended, and wound down, which is exactly why
the connection is dead by the time the caller looks. `exitProbeOutcome` gains a
push-aware form so the rule stays pure and pinnable; the errno classification
underneath is untouched, and a push carrying another installation's owner
token, or one that says `.alive`, still establishes nothing.

Tests: `recordsAnExitTheHolderPushedRatherThanAnswered` (tier 3, deterministic —
it holds the pushed-to connection itself rather than racing for it) and four
cases in `HolderExitProbeClassificationTests`.

PR #811, opened 2026-09-05
Fix: Stop the "run without tmux" toggle from silently doing nothing after a restart

cheapsteak:
## What's broken

Enable "Run new sessions without tmux" in Settings, restart TBD with `scripts/restart.sh`, create a session: it runs in a tmux window exactly as before, and nothing says why. On a daemon that has never had the holder binary the toggle is greyed out with a caption blaming the missing helper, but once the flag is on, every restart from then on silently degrades: the daemon comes up without the helper and routes each new session to tmux. `tbd update` produces the same daemon.

Measured on one machine before the fix: after a clean `scripts/restart.sh`, `.build/debug/TBDHolder` did not exist, and `daemon.capabilities` reported `ptyHolderEnabled: true` alongside `ptyHolderSupported: false`. Building that one product by hand (about 17 seconds) and restarting flipped `ptyHolderSupported` to true.

Two more helpers have the same shape of failure and were also never built by a restart:

- **The CLI.** Every replacement agent process (hibernation resume, terminal replacement) is handed the daemon's sibling `TBDCLI` path as `TBD_CLI_PATH`, and the Claude and Codex hook lines run that binary with errors swallowed. With no sibling, those sessions' hook events never reach the daemon.
- **The peer helper.** With remote peer messaging enabled, a remote lane spawns `TBDPeerHelper` beside the daemon and fails to arm when it is missing. Its own doc comment already claimed the restart script stages it.

## Why it happens

The daemon does not run holder-backed sessions, hooks, or remote lanes itself. It spawns small helper binaries, and it finds each one in exactly one place: beside its own executable. That choice is deliberate, so a stray binary on `PATH` can never own the daemon's sessions.

`scripts/restart.sh` built two products, `TBDDaemon` and `TBDApp`, under a comment saying restarting needs exactly those two. That was true before any sibling helper existed. `scripts/update.sh` kept its own list, `TBDDaemon`, `TBDApp` and `TBDCLI`, added for the CLI's version report. Two lists in two files drifted independently, and none of the sibling lookups fails loudly enough for a restart to notice.

## When it broke

The `TBDHolder` target arrived in #767 and the daemon began spawning through it in #777. `TBD_CLI_PATH` arrived in #751 and `TBDPeerHelper` in #762. The restart script's product list was never revisited, and nothing exercises the restart script end to end, so the mismatch stayed invisible until the toggle in #810 gave a person a reason to expect the feature to work.

## What this PR does

- The product list lives once, as `RUNTIME_PRODUCTS` in `scripts/restart-bundle-lib.sh`, which both scripts already source. It names the daemon, the app, and the three sibling helpers, with a comment per product saying what breaks when it is missing.
- `scripts/restart.sh` and `scripts/update.sh` loop over that list. The per-product `swift-safe` invocation and its exit-status capture are unchanged. The restart loop now stops at the first failed product: with five products, building the rest behind the same machine-global lock only delayed the abort and scrolled the failure off the screen.
- The `tbd update` self re-exec compares the two libraries the script sources, not just `update.sh`. The product list now lives in one of those libraries, which `update.sh` reads from the running tree, so a change that landed only there would otherwise build the old list against new sources with nothing to trigger the hop.
- `docs/updating.md` and the automatic-updates spec describe the product list and the widened re-exec trigger. `scripts/mock.sh` says why its own two-product list deliberately stays put: a mock daemon wires no holder registry and no peer bridge.

The `~/.local/bin/tbd` hard link is deliberately not touched. The app's launch-time installer already detects an inode mismatch against the sibling `TBDCLI` and prompts to refresh, and that prompt is a user gesture. Building `TBDCLI` here makes the sibling the daemon hands to hooks exist; refreshing the hard link is the existing prompt's job and stays a separate concern.

`scripts/mock.sh` builds only the daemon and app. Its mock daemon wires no holder registry and no peer bridge, so it needs neither helper.

## Assumptions

**A helper that fails to compile blocks the restart or update, on purpose.** Before this change neither script touched the holder or the peer helper, so a compile error in one of them could not stop a restart. Now it does, exactly as a daemon compile error does. The alternative, building helpers as a warning and continuing, was rejected: each helper degrades gracefully at runtime when it is absent, and that graceful path is precisely how a never-built helper went unnoticed, so a warning would put the same silent degradation one scroll above the "Daemon ready" line. The holder and peer helper depend on `TBDShared` and nothing else, and the CLI on `TBDShared` plus packages the daemon also links, so a helper failing while the daemon builds means the tree is broken, and a broken tree is what a restart should refuse. If a helper ever grows a dependency the daemon does not share, this assumption is the one to revisit.

## Evidence & verification

- The update harness (`scripts/update.test.sh`) runs the real `scripts/update.sh` against a stubbed `swift-safe`. It now asserts that `TBDHolder` and `TBDPeerHelper` are built, that five builds run per configuration, and that a fetched tree whose only change is in `restart-bundle-lib.sh` re-execs exactly once and completes the dry run. Against the pre-change `update.sh` each of those assertions fails; on this branch the whole harness passes (202 ok, 0 fail).
- Not covered, and worth a follow-up: nothing pins the sibling lookups in `Sources/` to the list. The next helper that lands with a sibling lookup still depends on someone adding it here. A test enumerating those lookups would be grep-shaped over Swift source, so it was not added.
- The bundle-lib harness (`scripts/restart-bundle-lib.test.sh`) pins every name in the list to an executable target in `Package.swift`, so a typo or rename fails there rather than on the next restart. A mutated list with a bogus name reddens it; the real list passes (59 ok).
- `scripts/restart.sh`'s own build loop is not reachable from any harness (the `restart-*-lib.test.sh` files cover the sourced libraries, not the top-level script). Sharing the list is what closes that gap: the loop can no longer disagree with the tested list.
- Each of the three helper products builds through the exact invocation the loop now runs, `scripts/swift-safe build --product <name>`, with exit 0 in this worktree.
- Not live-verified on this branch: the machine runs a live fleet, so no restart was performed here. The before/after capability flip described above was observed by hand with the same product build on the same machine.

https://claude.ai/code/session_01Mc559dNKWt4WMua2C1vEmv
[holder-restart-build](https://cheapsteak.github.io/tbd/open/?worktree=C9AFC30F-A8BF-4C54-9423-0E9AB805F2A1)

## Summary by CodeRabbit

- **New Features**
  - Updates now refresh themselves when supporting update scripts change, with safeguards against repeated handoffs.
  - Restart and update operations now build all required runtime products, including holder and peer-helper components.
  - Build metadata documentation now covers the source path and uncommitted changes.

- **Bug Fixes**
  - Restart operations stop safely when any required product fails to build.

- **Tests**
  - Added coverage ensuring all required runtime products are executable targets and sourced-script changes trigger updates correctly.

### DIFF
--- a/docs/specs/2026-09-04-automatic-version-updates-design.md
+++ b/docs/specs/2026-09-04-automatic-version-updates-design.md
@@ -197,14 +197,17 @@ arguments, and inherits the user's environment. The script is the procedure:
    must be safe to `checkout --detach` at any time. Its `origin` is the
    upstream URL resolved as in §5.
 2. **Fetch and check out** `origin/main` detached. If the fetched
-   `scripts/update.sh` differs from the running copy, re-exec the fetched one
-   with the same arguments (guarded by an environment marker so it happens
-   once). This is how the procedure itself updates.
+   `scripts/update.sh`, or either library it sources, differs from the running
+   copy, re-exec the fetched script with the same arguments (guarded by an
+   environment marker so it happens once). This is how the procedure itself
+   updates.
 3. **Stamp** `TBDBuildIdentity.json` into `.build/<config>/` (commit, branch,
-   time, worktree, dirty) and **build** `TBDDaemon`, `TBDApp`, and `tbd`
-   with `scripts/swift-safe build -c release --product …`, the same module
-   cache flags as `restart.sh`. A failed build stops here; the running
-   installation is untouched.
+   time, worktree, dirty) and **build** every product in `RUNTIME_PRODUCTS`
+   (`scripts/restart-bundle-lib.sh`, the list `restart.sh` builds from too:
+   the daemon, the app, the CLI, and the helpers the daemon looks for beside
+   its own binary) with `scripts/swift-safe build -c release --product …`,
+   the same module cache flags as `restart.sh`. A failed build stops here;
+   the running installation is untouched.
 4. **Assemble, sign, and install** the app bundle to `/Applications/TBD.app`
    exactly as `restart.sh` does today. That code moves out of `restart.sh`
    into `scripts/restart-bundle-lib.sh` so the two scripts share it; the
--- a/docs/updating.md
+++ b/docs/updating.md
@@ -36,16 +36,18 @@ these steps in order.
   source worktree, else that worktree's `origin`, else whatever `--remote` you
   pass.
 - **Fetches and detaches onto `origin/main`.** If the fetched copy of
-  `scripts/update.sh` differs from the one running, the fetched one takes over
-  with the same arguments. That is how the procedure updates itself, and an
-  environment marker holds it to a single hop.
+  `scripts/update.sh`, or of either library it sources, differs from the one
+  running, the fetched script takes over with the same arguments. That is how
+  the procedure updates itself, and an environment marker holds it to a single
+  hop.
 - **Stamps the build identity.** `TBDBuildIdentity.json` lands in the build
   directory before the compiler runs, recording the commit, the branch, the
   build time, the clone's path, and whether the tree was dirty. Every binary
   reads the sidecar beside it, which is how `tbd version` knows what is
   running.
-- **Builds** `TBDDaemon`, `TBDApp` and `TBDCLI` through `scripts/swift-safe`,
-  with the same shared module cache `scripts/restart.sh` uses. A failed build
+- **Builds** the same products `scripts/restart.sh` builds — `TBDDaemon`,
+  `TBDApp`, `TBDCLI`, `TBDHolder` and `TBDPeerHelper` — through
+  `scripts/swift-safe`, with the same shared module cache. A failed build
   stops here and the running installation is untouched.
 - **Assembles, signs and installs** `/Applications/TBD.app`, using the same
   code `scripts/restart.sh` uses (`scripts/restart-bundle-lib.sh`). The bundle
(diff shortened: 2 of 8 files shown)


cheapsteak:
Review finding (MEDIUM, all-or-nothing build gate now covering the helpers): acknowledged as a deliberate trade-off rather than fixed. The rationale is now in the `RUNTIME_PRODUCTS` comment in `scripts/restart-bundle-lib.sh` (commit 901be87e) and in a new Assumptions section of the description. Short form: each helper degrades gracefully at runtime when it is absent, which is exactly how a never-built helper went unnoticed, so building one as a warning would reintroduce the same silent degradation one scroll above "Daemon ready". A helper that fails to compile while the daemon builds means the tree is broken, and a broken tree is what a restart should refuse.

The first CI run's test job failed on wait-timeouts in `ControlModeInputRouterTests` and `ServerShutdownLatchTests`, neither touched by this diff (no Swift changes at all), while main is green at the same base commit. The push above starts a fresh run.

cheapsteak:
CI note for the record: the test job failed three times on this head with three disjoint sets of wait-timeout failures (the shutdown latch twice, then three unrelated 60 s time limits in the second parallel pass), each while other runs were active on the repo. The rerun with the queue empty went green with no change to the branch. No Swift in this diff; the only harness the test job runs from the changed files is the bundle-library one, green every time.

PR #813, opened 2026-09-05
Fix: Agents on the pty-holder transport are invisible to every other agent

cheapsteak:
## What's broken

Agents on the pty-holder transport silently vanish from each other. Every Claude Code session TBD spawns with `pty_holder_enabled` on is missing from Claude Code's cross-session peer registry, so `ListAgents` never lists it and a `SendMessage` to it returns "No agent named 'X' is reachable" — which reads like a dead peer. The session itself runs fine, so nobody notices until a coordination handoff goes nowhere.

The same mechanism has a second, worse symptom: those sessions write no transcript. There is no `projects/<slug>/<session-id>.jsonl` under the profile's config dir, so `--resume` after a restart has nothing to resume, and anything TBD reads from the transcript (Session History, archive/restore of a lane's conversation) sees an empty session.

Measured on a live fleet on 2026-09-05, classifying every `claude` process by whether a `TBDHolder` is its ancestor:

| transport | registered | not registered | transcripts on disk |
| --- | --- | --- | --- |
| holder-backed | 0 | 8 | 0 of 8 |
| tmux-backed | 69 | 1 | 1 of 1 checked |

The oldest unregistered holder session had been up for hours, so this is not a startup window.

## Why it happens

A daemon's environment describes the program that launched it as much as the machine it runs on. `scripts/restart.sh` is almost always run from inside a Claude Code session's Bash tool, so the new daemon (and the app) inherit that session's exported identity: `CLAUDECODE=1`, `CLAUDE_CODE_CHILD_SESSION=1`, `CLAUDE_CODE_ENTRYPOINT=cli`, `CLAUDE_CODE_SESSION_ID`, `CLAUDE_CODE_MESSAGING_SOCKET`/`TOKEN`, `CLAUDE_CODE_BRIDGE_SESSION_ID`, `CLAUDE_PID`, plus the pane's `TMUX` and `TMUX_PANE`, and TBD's own exports for the terminal the session ran in (`TBD_TERMINAL_INCARNATION_ID`, `TBD_CLI_PATH`, `TBD_TERMINAL_ID`, `TBD_WORKTREE_ID`, `CLAUDE_CONFIG_DIR` for its profile). The live daemon on the measured machine carries all of these, pointing at the session that restarted it; the `TBDHolder` processes it spawned carry them too, and so does every job those holders forked.

The holder path built the job's environment as the daemon's environment verbatim, with the spawn's own `sensitiveEnv` merged on top. So each holder-backed `claude` starts with `CLAUDE_CODE_CHILD_SESSION=1` in its environment.

Claude Code treats that marker as evidence it is a nested child of another session. In the shipped binary (2.1.261) the decision is: marker present, session interactive, not a team agent, and the marker is not "ambient in tmux". A nested child skips the peer-registry write and disables session persistence and prompt history. That is both symptoms from one cause.

The tmux transport is immune to the Claude Code half, which is why this never showed up before. "Ambient in tmux" is a probe Claude Code runs when `$TMUX` is set: `tmux show-environment -g CLAUDE_CODE_CHILD_SESSION`. If the tmux server's global environment carries the marker, the marker is judged inherited scenery rather than a real parent, and ignored. A tmux server started by a marked daemon carries the marker globally, so the probe says ambient. A long-lived tmux server started from a clean launchd environment (the two main ones on the measured machine date from August and hold 24 variables) has no marker at all. Either way a tmux pane registers. The single marked tmux-backed session in the measurement, running under a server a marked daemon started, is registered, which confirms the probe is what protects that path. The holder transport has no tmux server for the probe to ask, so the inherited marker reads as real.

TBD's own markers have no such probe on either transport. `TBD_TERMINAL_INCARNATION_ID` is exported on every woken or restarted agent; nothing on a fresh spawn overwrites it, and a fresh terminal row has no incarnation, so the SessionStart guard in the terminal store rejects the job's SessionStart, SessionEnd and activity hooks: no session id, no transcript path, activity never updates. A holder job inherits it from the daemon; a tmux pane inherits it from a server a marked daemon created. Only servers that predate the marked daemon are clean, which is why the tmux fleet on the measured machine looked healthy.

Two things ruled out along the way, so nobody repeats them: it is not per-profile registry fragmentation (the `sessions/` slot is a symlink to one host store for every profile, verified intact), and it is not `CLAUDE_CONFIG_DIR` (identical shape on both transports; `ps -Eww` truncates before reaching it, which is where that red herring came from — the full environments here were read with `sysctl KERN_PROCARGS2`).

## When it broke

With the holder transport's first spawn path, #777, which composed the job's base environment from the daemon's without accounting for what the daemon had itself inherited. It slipped through because the flaw is invisible on a daemon launched from a clean shell, and the tmux path had been quietly protected by Claude Code's probe all along, so there was no precedent to compare against. The TBD-side leg on tmux servers predates that and is latent until a marked daemon creates a new server.

## What this PR does

Adds `SpawnBaseEnvironment` (Sources/TBDDaemon/Process/), one named list of the variables that identify the process that launched the daemon, applied in two layers:

- **At daemon startup**, the existing `Daemon.scrubInheritedTBDEnv` (which already unset a handful of TBD and Codex names from the daemon's own environment for exactly this reason) iterates the shared list instead of its own, so every child the daemon spawns through plain inheritance (git, hooks, one-shot tmux clients) starts clean.
- **At the two long-lived spawn seams**, explicitly: `WorktreeLifecycle.holderLaunch` builds the holder job's environment as `SpawnBaseEnvironment.inheriting(daemonEnvironment).merging(sensitiveEnv)`, and `TmuxManager.ensureServer` starts a new tmux server under the same derived environment, since the server's environment is what every pane inherits. Explicit at these seams so they are right even from an injected environment, and testable without touching the process environment.

Existing tmux servers outlive daemon restarts and carry the markers their creating daemon had, so `ensureServer` also repairs a server it finds already running: one tmux invocation that unsets TBD's and Codex's markers from the server's global environment and lists what remains, plus a second call only when the listing shows a TBD-minted `CLAUDE_CONFIG_DIR`. Claude Code's own markers are deliberately left on an old server: panes that already exist hold their own copy, and Claude Code's probe reads the server-global copy as the reason to ignore it, so removing only the global copy would make a `claude` started later in an old shell pane read as nested. Those clear when the server is recycled. `TMUX`/`TMUX_PANE` are the server's own. A failed repair is logged, not swallowed, because tmux stops a command list at its first failure.

The claude-cloud remote provider's spawner builds its environment from the same derived base, since the vendor CLI it runs is a spawned Claude Code like any other.

The trust seeder resolves the `.claude.json` it seeds from the same derived base, so a profile-less spawn under a daemon restarted from a profile-bound session no longer gets seeded into the launcher's profile dir while reading its own home, which would have re-introduced the blocking folder-trust dialog.

The set has four groups under one criterion, identity of the launcher rather than machine, user, or installation configuration:

- **Claude Code's per-session exports.** The nesting markers above, plus `AI_AGENT`, `CLAUDE_EFFORT` and `TRACEPARENT`, which Claude Code's own child-session env builder injects alongside them. An inherited `CLAUDE_EFFORT` would silently set the new session's effort level; an inherited `TRACEPARENT` parents its telemetry into the launcher's dead trace.
- **TBD's own per-process exports.** `TBD_TERMINAL_INCARNATION_ID` (the load-bearing one, above), `TBD_CLI_PATH` (would point the job's hooks at the launcher's worktree CLI), `TBD_TERMINAL_ID`, `TBD_WORKTREE_ID`, the prompt-layer and hook-event variables, and the handover pid the daemon reads before the scrub runs. Installation-wide config such as `TBD_HOME` is not in the set and keeps flowing.
- **Codex's per-session exports**, `CODEX_CI` and `CODEX_THREAD_ID`, carried over from the startup scrub's existing list.
- **The enclosing tmux pane's coordinates**, `TMUX` and `TMUX_PANE`.

One value-dependent rule sits beside the set: `CLAUDE_CONFIG_DIR` is dropped only when it points under the installation's own `profiles/` directory (or is empty), because only TBD writes values there and it does so per spawn. A daemon restarted from a profile-bound session would otherwise hand that profile's credential and settings to every profile-less job. Any other value is the user's configuration and is kept, as TBD's trust seeder already honours it.

The same helper pins `TERM` to `xterm-256color`, which is what the tmux transport already gives every pane through `default-terminal`. A holder job otherwise inherited whatever launched the daemon: another emulator's terminfo, or no `TERM` at all when the app started the daemon itself.

The scrub applies to the inherited base only; values the spawn itself decided (`sensitiveEnv`: profile env overrides, auth and routing) are merged afterwards and win, so a deliberate override under one of these names, `TERM` and `CLAUDE_CONFIG_DIR` included, still reaches the job.

Alternatives considered:

- **Start children from a clean login environment** instead of the daemon's. That would also drop the user's own login exports (PATH additions, API keys) that the tmux path forwards today, a behaviour change well beyond the bug.
- **Set `CLAUDE_CODE_FORCE_SESSION_PERSISTENCE`** on holder jobs. It is Claude Code's internal escape hatch, restores persistence and registration by fiat, and leaves a stale parent identity in the job's environment. The leak is the cause; the flag would paper over it.
- **Only the daemon-level startup scrub**, with no explicit derivation at the spawn seams. It cannot express the `TERM` pin, it depends on the daemon constructing its holder registry after the scrub runs (true today, asserted nowhere), and it leaves the seams untestable except by mutating the test process's environment. Both layers share one list, so they cannot drift.
- **Also pin `TBD_CLI_PATH`** to the daemon's own sibling CLI for fresh spawns. Fresh spawns on both transports resolve `tbd` from PATH today (a replacement process is the only spawn that gets an explicit path), and pinning would change which CLI every fresh session's hooks use, on the tmux path too. Dropping the launcher's stale value restores the existing baseline; the pin is a separate decision and is flagged as a follow-up.

No flag: this is a bug fix restoring the two transports to one behaviour. No new durable resource is created. Doc comments on `holderLaunch` and `HolderRegistry.environment` that claimed the job inherits the daemon environment verbatim are corrected, and the pty-holder design spec's "Environment and launch parameters" bullet states the base and the reason.

Left out deliberately, and worth a separate decision each:

- **The launcher's credential and routing variables.** An API-key or token profile injects `ANTHROPIC_API_KEY` / `CLAUDE_CODE_OAUTH_TOKEN` and inline-exports `ANTHROPIC_BASE_URL`, `ANTHROPIC_MODEL`, `CLAUDE_CODE_USE_BEDROCK`, `AWS_REGION`, `AWS_PROFILE`; a daemon restarted from such a pane inherits them, and this PR does not strip them. It cannot by name: the same names are how a user without profiles authenticates from their login environment, and the tmux path forwards them today. A profile-less spawn under such a daemon authenticated as the launcher's profile before this PR too. The durable fix is at the launcher, having `scripts/restart.sh` and `tbd update` start the daemon without what the enclosing session exported. OAuth profiles, the common case, inject only `CLAUDE_CONFIG_DIR`, which this PR does handle.
- **Pinning `TBD_CLI_PATH`** for fresh spawns, above.

## Assumptions

- Claude Code keeps reading `CLAUDE_CODE_CHILD_SESSION` as the nesting marker and keeps its tmux ambient-marker probe. If it starts reading a different variable, that name must join the set; the pinned-set test makes that a visible edit.
- The user has not deliberately set one of the marker names in their login environment and expected it to reach a spawned session. They are all per-process identity that only a Claude Code session, a TBD terminal, or a tmux pane sets, so a deliberate login export of one of them is not a supported configuration.
- A `CLAUDE_CONFIG_DIR` under the installation's `profiles/` directory was written by TBD for a specific spawn. If a user points their own login `CLAUDE_CONFIG_DIR` inside that directory, it is dropped for profile-less sessions.

## Evidence & verification

- **Repro:** on a machine whose daemon was restarted from a Claude Code session, create a worktree with `pty_holder_enabled` on and check `$CLAUDE_CONFIG_DIR/sessions/<pid>.json` and `projects/<slug>/<session-id>.jsonl` for the new `claude` pid. Both are absent; the same spawn on tmux produces both. The measurement table above is this repro across the live fleet.
- **Discriminating test:** `SpawnBaseEnvironmentTests` (Tests/TBDDaemonTests/Process/) drives `WorktreeLifecycle.holderLaunch` with a daemon environment carrying every marker, a foreign `TERM`, and the variables a job must keep (including `TBD_HOME` and a user-set `CLAUDE_CONFIG_DIR`), and asserts the composed environment equals the kept set plus the pinned `TERM` as a whole dictionary. Without the fix the composed environment is the full daemon environment, so the equality fails. Further tests cover a `sensitiveEnv` value under a marker's name, an unmarked environment passing through unchanged, a daemon with no `TERM`, a spawn choosing its own `TERM`, a TBD-minted `CLAUDE_CONFIG_DIR` being dropped while a user's is kept and the profiles root following the base's own `TBD_HOME`, a profile-bound spawn keeping its own config dir, and the marker set pinned by name.
- **tmux repair tests:** `TmuxServerEnvironmentRepairTests` (in Tests/TBDDaemonTests/TmuxManagerTests.swift) pins the exact argv of the in-place repair, asserts no Claude Code or tmux marker appears in it while TBD's and Codex's do, and checks the config-dir decision against a full `show-environment -g` listing for a TBD-minted value, a user value, tmux's "unset" line, and no line. Group tests assert the four groups partition the set and that the repairable subset spares Claude Code's and tmux's. The existing env-leak test that drives `scrubInheritedTBDEnv` still holds, since every name it asserts is in the shared list; it mutates the process environment at top level, which predates this branch and is noted rather than fixed here.
- **Live tmux test:** `TmuxServerEnvironmentRepairLiveTests` (Tests/TBDDaemonLiveTests/) creates a throwaway server through `ensureServer`, plants a TBD marker, a Codex marker, Claude Code's marker and a TBD-minted `CLAUDE_CONFIG_DIR` in its global environment, calls `ensureServer` again, and asserts the TBD and Codex markers and the minted config dir are gone while Claude Code's marker survives. It also asserts a user's own config dir survives the repair, and that a freshly created server's environment carries the pinned `TERM` and no `TMUX`. tmux reports a removed name as "unknown variable" with a non-zero exit, and the test treats that as removed.
- **Startup scrub test:** the existing test that drives `Daemon.scrubInheritedTBDEnv` sets every marker except `TMUX`/`TMUX_PANE` (a transient `TMUX` in the test process would make concurrently spawned tmux clients refuse to nest) and asserts each is cleared. `CLAUDE_CONFIG_DIR` is not set in-process; its by-value rule is covered by the pure tests above.
- **Local run** of the affected unit suites: `scripts/test.sh --filter SpawnBaseEnvironmentTests --filter TmuxManagerTests --filter TmuxServerEnvironmentRepairTests --filter TBDWorktreeIDLeakTests --filter ClaudeTrustSeederTests` passed, exit 0, plus the live suite above. `swiftlint --strict` on the changed files: 0 violations. Not asserted anywhere: that one-shot tmux client calls still pass `nil` (inherit); `TmuxManager` has no runner seam exposing the child environment.
- **Live verification** requires restarting the daemon, which was not done: a live fleet and a human's session are running on the measured machine. After the next restart from this branch, the check is the repro above: a new holder-backed session's pid gains a `sessions/<pid>.json` row and a transcript, and `ListAgents` from a sibling session lists it.

[20260905-interesting-camel](https://cheapsteak.github.io/tbd/open/?worktree=AE1EF3D9-6AC8-45F4-9284-77A15D4F5269)

https://claude.ai/code/session_01DXvA5KeAgJ7aUzxrvkiy5a

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved environment isolation for spawned jobs, holder sessions, cloud processes, and tmux servers.
  * Prevented inherited session and identity markers from leaking into new worktree sessions.
  * Preserved user-owned configuration while cleaning up stale or tool-managed configuration paths.
  * Standardized terminal settings for spawned sessions.
  * Improved re-adoption of existing tmux servers through automatic environment repair.

* **Documentation**
  * Clarified how holder, job, and tmux environments are derived and maintained.


cheapsteak:
Pushed 88154ef7, addressing the gate's last round:

- **Existing-server repair, end to end:** `TmuxServerEnvironmentRepairLiveTests` creates a real server through `ensureServer`, plants a TBD marker, a Codex marker, Claude Code's marker and a TBD-minted `CLAUDE_CONFIG_DIR` in its global environment, calls `ensureServer` again, and asserts the first three of those are removed while Claude Code's marker survives; a user's own config dir also survives; a fresh server carries the pinned `TERM` and no `TMUX`. tmux 3.6a reports a removed name as "unknown variable" with exit 1, and the test treats that as removed.
- **Startup scrub:** `testScrubInheritedTBDEnv` iterates the shared list minus `TMUX`/`TMUX_PANE` (which an in-process test must not set while sibling suites spawn tmux clients) and asserts every name is cleared.
- **MINOR confirmed and fixed:** the claude-cloud spawner now builds from `SpawnBaseEnvironment.inheriting`, with a test that runs `/usr/bin/env` through `spawn` and asserts no marker reaches the child.

Also since the previous gate run: `TBD_PROMPT_SCRATCH` joined the set (four prompt layers, not three), and an empty `CLAUDE_CONFIG_DIR` is treated as needing removal at all three judgment sites. The description is rewritten to match the current shape.

cheapsteak:
On the gate's latest round:

**`tbd profile login` and the scrub (MEDIUM): not applied, on purpose.** The scrub derives a *spawned child's* base from the *daemon's* environment, where every `TBD_*` name describes the terminal that launched the daemon and is therefore stale. `tbd profile login` runs in a pane's own environment, where the same names describe *that pane*: TBD sets `TBD_TERMINAL_ID` on it deliberately, and the login session's hooks report it back to associate the session with its terminal row. Stripping them there would break that association, and pinning `TERM` there would override the user's real terminal. The Claude Code markers can only be present in that environment by two routes, and both are already handled: a pane under a tmux server carries the server-global copy, which Claude Code's probe reads as ambient; a pane spawned by a daemon on this branch has none. The nested-child decision also does not affect what `/login` exists to do (it gates the peer-registry row, transcript persistence and prompt history, not the credential write), so even the residual case, a login pane spawned by a pre-fix daemon, completes its job. If the login path ever needs the Claude-side strip, `SpawnBaseEnvironment` would have to move to `TBDShared` (the CLI depends only on that target) with a variant that leaves `TBD_*` and `TERM` alone; that is a different change from this one.

**Doc comment on `TBDConstants.configDir` (minor): fixed** in the follow-up commit. It reads `TBD_HOME`, not `CLAUDE_CONFIG_DIR`; the sentence now names the two actual readers and says `TBDConstants.configDir` applies the same empty-means-unset convention to `TBD_HOME`.

**No test forces the repair to fail (minor): acknowledged, not added.** There is no deterministic way to make `set-environment -gu` fail against a live server without a runner seam that `TmuxManager` does not have, and the containment is structural: the repair sits in its own `do`/`catch` inside the existing-session branch, so a throw cannot reach the outer `catch` that creates a session. The unit and live suites cover the success path end to end; the failure path is by inspection, and the PR description says so.

PR #818, opened 2026-09-06
Feat: Teach hibernation the pty-holder transport so idle holder sessions can park and wake

cheapsteak:
## Summary

Sessions on the pty-holder transport can now be hibernated and woken. Until now every park path (idle sweep, merge-park, manual "Hibernate now", the retired suspend RPCs), the wake path, and the usage-limit auto-resume refused a holder-backed row outright, and the reconcile arm deleted a finished holder Claude session rather than parking it. For anyone running with `pty_holder_enabled` on, that meant idle sessions stayed resident forever, the tab menu never offered Hibernate, and a session that ended while nobody was looking was gone rather than resumable.

That refusal was the unbuilt half of a design that is already committed (`docs/specs/2026-08-30-pty-holder-session-transport-design.md`, "Hibernation and revive"): hibernate ends the holder's child and the holder winds down; revive spawns a fresh holder; the input-veto and queued-prompt flags keep their meaning against daemon writes to the pty. This PR builds that half.

## Why this shape

No brainstorming round: the design is the transport spec's, and `Terminal.isManuallyHibernatable` said in as many words that the refusal stood only "until parking learns the holder transport". The decisions this PR had to make on top of it:

- **The row never finalizes parked while the child still runs.** Park writes its intent first (pending incarnation and `hibernatedAt`, so hooks from the dying agent are fenced and a crash in this window is healed by startup reconciliation rather than left ambiguous), then ends the child on a ladder that mirrors the tmux one: `/exit` written to the pty the daemon is reading, then SIGTERM, then the registry's teardown (forget, SIGKILL the job's group, reap the holder), each rung with its own poll on the injected clock, and each signal sent only to a child whose identity verifies against the row's recorded start time and executable. Only then does the park finalize and clear the row's pids. A child that survives the ladder, or whose identity cannot be verified, rolls the intent back and the session is left awake with a named reason. A parked row over a live process is the one state nothing reclaims and a wake would double.
- **The pending-input rail fails closed when the daemon is not the reader**, exactly as the spec's "Two stores" section requires. The rail reads the typed screen the transport now exposes and branches on its source: a screen the daemon is draining right now is judged with the existing sanctioned pending-input check; a stale daemon copy (a viewer holds the pty) or a screen that will not project is refused by name. The idle sweep asks the same source question through the cheaper mode reading before it arms a holder row. Consequence during the soak: a holder session still open in a viewer does not auto-park, until the viewer can answer a screen request.
- **No new holder wire verb.** The spec's "instructs the holder to terminate its child" is met with the verbs that exist; nothing under `Sources/TBDDaemon/Holder/` changed, because a sibling PR is restructuring those files.
- **Manual "Hibernate now" is behind the same flag** as the sweep. The park mechanic is one code path whatever triggers it, and the property the soak validates does not depend on who asked. The app's menus read the flag from daemon capabilities, and they also withhold Hibernate from a holder tab while the app's own viewer holds its pty, because the daemon would refuse that park for the fail-closed reason above; the menu and the rail agree. During the soak that means Hibernate is offered on holder tabs the app is not currently displaying, and the tmux tabs are unchanged. The attached case opens up when the viewer can answer a screen request.
- **The flag gates new parks, never the wake of a row already parked.** Turning the flag off is the soak-abort gesture, and it must not strand what the soak parked: a parked holder row wakes whatever the flag says, and startup reconciliation still heals a parked holder row whose child is verifiably alive or gone. `.holderTransport` on a wake now means only an unparked holder row with the flag off.
- **The reconcile arm parks instead of deleting** a finished holder Claude session when the flag is on, matching the tmux arm. With the flag off it deletes as before, since a park nothing can wake is noise.
- **The row records when its current child started.** Every reclaimer that verifies a recorded pid anchors on a start time; after a wake the child is younger than the row, so anchoring on `createdAt` would read every woken session's child as a stranger.

## What this PR does

- New tri-state config column `holder_hibernation_enabled` (no SQL default) with record, model, RPC setter, `daemon.capabilities` field, a Settings toggle under the pty-holder toggle, and `tbd config holder-hibernation on|off`.
- New nullable column `terminal.holder_child_started_at`, written at holder spawn and wake, cleared at park; the reaper's holder leg anchors on it.
- `Terminal.isManuallyHibernatable` and `isAutoHibernationEligible` become methods taking `holderHibernationEnabled:`, so every caller in the daemon and the app passes the flag and the compiler enforces that the five refusal sites move together.
- `HibernationGate.blockingRail`, `decide` and `decideForMerge` take the flag; `.holderTransport` now means "the soak flag is off".
- `HibernationCoordinator+Holder.swift`: the holder park and wake bodies, a startup-reconciliation branch for parked holder rows (child verified alive un-parks; child verified gone clears stale pids), and a transport-aware input-activity key. The input-veto rail keeps its meaning: its fact source is the app's keystroke stream, and on the holder transport those keystrokes go straight to the pty the viewer holds, which is the state the park already refuses, so the daemon's own emulator plus the fail-closed viewer check is the pending-input rail there. Daemon-originated writes (peer sends, queued prompts, the auto-resume) are deliberately not recorded as typed input.
- `wake` on a parked holder row runs the same preamble as tmux (pinned profile via `loadByID`, env overrides, settings overlay, trust seed, transcript mirror, `ClaudeSpawnCommandBuilder` with the queued prompt on argv), then spawns a fresh holder with `TBD_WORKTREE_ID`/`TBD_TERMINAL_ID` and the rotated incarnation, records its pids, and clears the park.
- `LimitResumeActuator` delivers the auto-resume to a holder row through the injection courier when the flag is on: an Escape write, the same 150 ms pause the tmux arm keeps between Escape and its text (an escape byte followed by a printable in one read parses as a meta key and would type "ontinue"), then `continue` plus a carriage return as one 9-byte write, under the 64-byte tokenizer threshold that stalls unwrapped holder writes.
- Corrected the wake comment that claimed `claude --resume` is cwd-scoped (measured false on claude 2.1.261); `TranscriptProjectDirSync` stays, because TBD reads transcripts from the mirrored project dir itself.

Refusal sites changed, all five: `HibernationGate.blockingRail`, `Terminal.isManuallyHibernatable`, `HibernationCoordinator` (`manualBlockReason`, `wake`), `RPCRouter+ManualSuspendHandlers`, `LimitResumeActuator`.

## Flag & rollout

`holder_hibernation_enabled`, default OFF. Enable for the soak with the Settings toggle "Hibernate pty-holder sessions" or:

```
tbd config holder-hibernation on
```

Both `pty_holder_enabled` and this flag must be on for a holder session to park; the flag alone changes nothing on tmux rows. Graduation: flip `Config.holderHibernationEnabledDefault` after a soak in which no park finalized while its child was still running, no wake left a holder without a row, and the reaper's holder leg never mis-anchored a woken session; an explicit off survives the flip because the column is NULL until touched.

## Assumptions

- A holder row that reaches the park with no recorded child pid (a crash-era or legacy row) can only park through the polite `/exit` window. Once the escalation runs, the registry's own teardown erases the status the verdict would need, and silence is not treated as "gone", so the row is left awake with its holder torn down for reconciliation to judge. Every row written by this daemon records its child pid, so this is the rare shape, and it fails toward leaving a row awake rather than parking over a live process.
- The idle sweep never accrues idle time for a holder row while a viewer holds its pty, so during the soak the auto-park leg exercises only sessions with no open tab. That is the fail-closed rule stated above, and it lifts when the viewer can answer a screen request.

## Orphans

Wake is a new creation path for a holder process, and park is a new teardown path. Who reclaims:

- **A park that half-completes** (daemon dies after the intent, before the child is confirmed gone) leaves a parked row that still names its pids. On the next start `HibernationCoordinator.reconcileOnStartup` un-parks a row whose recorded child is identity-verified alive, and `HolderRegistry.adoptAll` re-adopts its holder. If the holder is gone and the child outlived the hangup, `AgentReaper.sweepHolderChildren` reaps it by the recorded pid with its existing identity check, which the new start-time anchor keeps correct after a wake. The rendezvous files are `HolderRendezvousCollector`'s as before.
- **A wake that half-completes**: the pids are written before the park is cleared. A failed pid write abandons the holder from the failing call itself (the create-time cleanup `WorktreeLifecycle+Create` already does); a failed clear leaves a parked row with live pids, which startup reconciliation un-parks, and which a retried wake adopts rather than respawns: before spawning, the wake checks whether the registry still holds the earlier reader (after a daemon restart, adoption normally re-establishes one for such a row) or the recorded child is identity-verified alive, and if so it only clears the park marker. That is what makes a retry converge instead of stacking a second holder that no row names. The one shape neither leg answers is a live holder that adoption could not take and whose child's identity is uncertain: the wake then proceeds to spawn, the holder's creation lock refuses a second holder, and the row stays parked until that holder dies. Fail-safe, and named here so nobody reads it as a leak.

## Evidence & verification

- The idle sweep skips a holder row whose pty a viewer holds before arming it, so an open tab costs no actuation rows per pass; the holder-specific refusals clear the sweep's idle markers like every other refusal.
- The limit-resume holder arm answers a session the registry has seen end with the same silent cancel the tmux arm gives a dead window.

- Every gate has a test for both branches: the shipped NULL default and an explicit off refuse holder rows exactly as before with the row unmutated and no registry or tmux call; on, the gate elects, manual park reaches the fail-closed screen rail, wake without a spawner leaves the row parked, the reconcile arm parks, startup reconciliation un-parks a live child and clears a dead one. Identical tmux rows pass unchanged under both values.
- A live suite (`HolderHibernationLiveTests`) spawns a real `TBDHolder` with a real job, parks it (the job ignores `/exit`, so the escalation runs), and proves the row is parked with its pids cleared, the child pid answers ESRCH, the holder is gone and the daemon holds no reader; then wakes it and proves a new live holder and child are recorded; and proves the flag-off park leaves the job running. A fourth case drives the rollback itself: with a signaller that reports the child alive whatever the process table says, the park escalates, tears the holder down, and then refuses to finalize, leaving the row awake with its pids intact, its pending incarnation rolled back, and the sweep's idle markers cleared.
- Flag storage tests prove a pre-migration row reads NULL, an explicit false survives a default flip, and NULL follows it.
- CI is the verdict; the local build slot on the development machine is contended. Full suite green on the current head after the rebase onto #819. Two fresh-context review passes, the merge gate's three rounds, and an independent design review all fed fixes into the branch: the SIGTERM rung, identity checks before every signal, the wake-adopt guard and its pid restoration, the polite-write result handling, and the rail-parity matrix. The one item outside this PR's territory is filed as #820.

https://claude.ai/code/session_01MSDmW2YUht9nBzkQaSiouJ

[holder-hibernation](https://cheapsteak.github.io/tbd/open/?worktree=CF66C231-05C1-4E2D-9C7E-41C165A3CAB2)

### DIFF
--- a/Sources/TBDApp/AppState+ModelProfiles.swift
+++ b/Sources/TBDApp/AppState+ModelProfiles.swift
@@ -424,6 +424,41 @@ extension AppState {
         }
     }
 
+    /// Help text for the pty-holder auto-hibernation toggle.
+    ///
+    /// A stored constant rather than a literal in the view, same reasoning as
+    /// `ptyHolderHelp`: it must describe what changes, and it must be honest
+    /// about the one thing it cannot do — check a session open in a viewer for
+    /// unsent input, the way the tmux rail can.
+    static let holderHibernationHelp = """
+        Off by default. When on, idle holder-backed Claude sessions are parked \
+        by the same sweep and rails as tmux sessions, manual Hibernate acts on \
+        them, and wake starts a fresh holder running claude --resume. Sessions \
+        currently open in a viewer are not parked, because the daemon cannot \
+        check them for unsent input.
+        """
+
+    /// Why the holder-hibernation toggle is inert on this daemon — same
+    /// reasoning and the same caption as `ptyHolderUnsupportedCaption`: with no
+    /// `TBDHolder` helper there is no holder-backed session for this gate to
+    /// ever act on.
+    static let holderHibernationUnsupportedCaption =
+        "Requires the TBDHolder helper beside the daemon binary; this daemon could not find it."
+
+    /// Persist the pty-holder auto-hibernation gate, then re-fetch
+    /// capabilities so the Settings toggle reflects the daemon's persisted
+    /// state. Applies on the next hibernation sweep.
+    func setHolderHibernationEnabled(_ enabled: Bool) async {
+        do {
+            try await holderHibernationFlagSetter(enabled)
+            await refreshDaemonCapabilities()
+        } catch {
+            logger.error("Failed to set holder hibernation: \(error, privacy: .public)")
+            showAlert(
+                "Failed to set holder hibernation: \(error.localizedDescription)", isError: true)
+        }
+    }
+
     /// Persist the worktree auto-trust switch, then re-fetch capabilities so
     /// the Settings toggle reflects the daemon's persisted state. Applies to
     /// the next Claude spawn or wake; never un-trusts an already-seeded path.
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -1438,6 +1438,13 @@ final class AppState {
     /// injectable for the same reason as `controlModeSetter`.
     @ObservationIgnored lazy var ptyHolderFlagSetter: @MainActor (Bool) async throws -> Void =
         { [daemonClient] enabled in try await daemonClient.setPtyHolderEnabled(enabled: enabled) }
+    /// How `setHolderHibernationEnabled` persists the pty-holder
+    /// auto-hibernation gate — injectable for the same reason as
+    /// `controlModeSetter`.
+    @ObservationIgnored lazy var holderHibernationFlagSetter: @MainActor (Bool) async throws -> Void =
+        { [daemonClient] enabled in
+            try await daemonClient.setHolderHibernationEnabled(enabled: enabled)
+        }
     /// How `setClaudeCloudEnabled` persists the Claude cloud gate — injectable
     /// for the same reason as `controlModeSetter`, so the Settings toggle's
     /// success and failure branches are testable without a real daemon.
--- a/Sources/TBDApp/DaemonClient.swift
+++ b/Sources/TBDApp/DaemonClient.swift
@@ -1191,6 +1191,17 @@ actor DaemonClient {
         )
     }
 
+    /// Persist the pty-holder auto-hibernation gate (default OFF). Applies on
+    /// the next hibernation sweep — no daemon restart needed. Sending either
+    /// value is an explicit gesture that survives a later change to the
+    /// shipped default.
+    func setHolderHibernationEnabled(enabled: Bool) async throws {
+        try await callVoidAsync(
+            method: RPCMethod.configSetHolderHibernationEnabled,
+            params: ConfigSetHolderHibernationEnabledParams(enabled: enabled)
+        )
+    }
+
     /// Persist the Claude cloud sessions gate (default OFF). The daemon builds
     /// its provider manager only at boot, so this takes effect on the next
     /// daemon restart rather than the next gesture.
--- a/Sources/TBDApp/Settings/SettingsView.swift
+++ b/Sources/TBDApp/Settings/SettingsView.swift
@@ -269,6 +269,7 @@ struct GeneralSettingsTab: View {
                     .help("Experimental: best-effort exit idle Claude instances when the machine is about to sleep, so a tmux server that dies during a long sleep has less to recover. Off by default — may interrupt long-running work.")
                 controlModeToggle
                 ptyHolderToggle
+                holderHibernationToggle
                 hibernateInputVetoToggle
                 autoCloseSetupToggle
                 queuedPromptToggle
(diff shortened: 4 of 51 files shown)


cheapsteak:
Addressed the Medium finding in 43fcdb94: `HolderHibernationLiveTests.parkRollsBackWhenTheChildOutlivesTheEscalation` drives the rollback branch against a real holder and job, with the coordinator's injected signaller reporting the child alive regardless of the process table (a real job cannot survive SIGKILL, and `abandon` nils the registry's status, so the signaller is the only remaining "gone" source). It asserts the refusal names the pid, the row is not parked on either column, the pending incarnation rolled back and the incarnation id is unchanged, the pids stay recorded, the sweep's idle markers are cleared after a real sweep armed them, and a second park returns a refusal rather than `alreadyHibernated`. The scripted suite cannot reach this branch because every scripted holder park stops at the reader rail, which is why it is a live test.

cheapsteak:
Addressed the wake-retry finding in 37128cce: `wakeHolderSection` now checks, before spawning, whether the registry still holds the reader an earlier spawn adopted or whether the row's recorded child is identity-verified alive (start time within the window of `holder_child_started_at`, acceptable executable). Either fact means the earlier wake's holder is live, so the retry only clears the park marker and broadcasts the un-park; every other verdict proceeds to the spawn as before. Pids are still written before the park clears, for the reason in the doc comment. Tests: `wakeAdoptsAHolderThatIsAlreadyRunning` (registry that cannot spawn, so a spawn attempt would be `.respawnFailed`), `wakeWithADeadChildStillSpawns`, and `parkedTmuxRowIsUnaffectedByTheAdoptGuard`.

The test-job failure on 43fcdb94 was `HolderInvocationTests.theLaunchDescriptorIsClosedOnceItIsRead` in the untouched TBDHolder target; the same test failed on a recent main run (34005101602) and passed on this branch's previous head, so it is being treated as a flake and re-run by this push.

PR #821, opened 2026-09-06
Docs: A message composer in the live transcript pane, designed

cheapsteak:
## Summary

Somebody watching a live Claude Code session in the app's transcript pane can read what the agent says but cannot answer it there; answering means switching to the terminal pane, which lags when the terminal is slow. This spec designs a message composer under the live transcript: a text field that sends a message into the running session, completes slash commands, skills and subagents from the session's own Claude Code binary, and accepts pasted or dropped images placed inline in the text. It is the design the implementation PRs argue from; nothing here changes behavior.

## Why this shape

The document is the design rationale. The decisions that most shape it: messages are injected immediately and verbatim, with the daemon's dispatch envelope suppressed only for the app acting for a human; the completion list comes from the installed binary through its control-protocol `initialize` request, never from a table inside TBD, so new commands and skills appear without a release; images are inline tokens whose presence in the text decides what is sent, mirrored by a strip; a hibernated or exited session is woken with the message as its first prompt through the wake RPC's existing prompt argument; one default-off config flag with a Settings toggle gates the whole feature. Rejected alternatives, each with its measured reason, include the rendezvous-socket reply frame, a compiled-in list of built-ins, and disabling the composer for a parked session.

Two measurements in the spec matter beyond it: Claude Code's stdin tokenizer swallows the carriage return of any single unwrapped write of 64 bytes or more, and a pasted path becomes an image attachment only when the paste is the bare quoted path alone. The holder-arm bracketed-paste fix is a dependency on the child-as-contract-party design (#816), not part of this work.

## What this PR does

- Adds `docs/specs/2026-09-05-transcript-composer-design.md`.
- No code. The implementation lands as four PRs the spec's sections map to: the not-running send guard and wake prompt, the flag and completions RPC, the daemon send-path and attachment reconcilers, and the app composer.

## Assumptions

- Claude Code 2.1.261's control protocol and paste handling, as measured on one machine with the zero-token stub; the decoder is written to tolerate added keys, and a renamed key degrades to the filesystem fallback.
- Holder-backed terminals gain park and wake in the hibernation subsystem's own work; the spec lists that as a dependency and does not rely on the gate's current holder exemption.

## Evidence & verification

Spec-only change. The design was brainstormed with a human answering every question, and every measured claim in it names what was run: the control-protocol probe timings and side effects, the 64-byte tokenizer threshold under a real pty, the image-attachment paste rule, and the resume-with-argument behavior.

[🔗 open in tbd](https://cheapsteak.github.io/tbd/open/?worktree=B6F8AB9D-8BD6-4902-98EF-16F8DCA04FE5)

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive design specification for a message composer in the live transcript pane.
  * Documented composer states, keyboard interactions, terminal-specific drafts, image tokens and storage, send and wake flows, transport constraints, input gates, autocomplete ranking, fallback handling, feature flags, cleanup, testing, rejected alternatives, dependencies, and follow-up work.
  * Scope is limited to local Claude Code sessions and excludes archived, remote, Codex, and shell sessions.


cheapsteak:
Addressed all three findings in `docs/specs/2026-09-05-transcript-composer-design.md` (014015bd).

- **HIGH — suppression trusted a self-declared actor.** The "Envelope" section no longer mentions the declared actor as the gate. Suppression now turns on an authenticated *connection*: the daemon reads `LOCAL_PEERPID` at accept through NIO's socket-option provider (a property of the socket, not of the wire) and provisionally matches it against the pid the FD-vending sidecar recorded for its one client, the app; a request that actually asks for suppression then pays for the rest of the identity — start time and command line through the same pid-reuse check `AppLivenessArbiter`/`AgentReaper` already use — and only a `.same` verdict authenticates. Everything else (CLI, agents, any local process) gets the envelope however it stamps `actor`, and the check fails closed: unreadable peer pid, no recorded sidecar client, pid mismatch, or any other verdict attaches the envelope. A closing paragraph reconciles this with `DispatchEnvelopeDisposition`: it stays daemon-internal in the sense that matters — the field can ask, only the connection can authorize. Rejected alternatives gained four entries: "Trust the declared actor field", "Check the peer's executable name", "A nonce the app registers at startup", and "A second, app-only RPC socket" (rejected on cost, not on strength).

- **MEDIUM — the not-running hold waited on any SessionStart.** "Not-running delivery" now scopes the hold to the composer's own wake. `terminal.wake` gains one additive optional field on its result, the session incarnation id it minted, populated only on the `woken: true` path; the app releases on the SessionStart carrying that id and no other, so a competing wake (deduped by `wakesInFlight`), a post-`--fork-session` recapture, and a hand-typed `claude --resume` do not release it. The id already exists and already reaches hooks as `TBD_TERMINAL_INCARNATION_ID`, so nothing new is minted. The one case that does release — a manual resume inside the pane the wake created, inheriting that spawn's env — is named and is correct.

- **MEDIUM — frecency constants had no rationale.** "Filtering and ranking" now says they are Claude Code's own, read from its binary during this design's research (usage count times `max(0.5 ^ (days-since-last-use / 7), 0.1)`), so ranking behaves the same in the transcript composer as in the terminal and muscle memory carries across; the floor keeps a once-favored command from decaying out of the list.

Testing was updated for both mechanisms: peer-identity rows over a `socketpair` with the sidecar identity injected (authenticated connection suppresses, foreign pid does not even when it declares `{"kind":"app"}`, unreadable/absent/mismatched identity keeps the envelope), and hold rows (matching incarnation releases, foreign and absent ids do not, timeout restores the text).

cheapsteak:
Addressed the second review:

- **Incarnation id claim (Medium):** the sentence now states the fact precisely — every respawn that replaces a terminal's process mints the id (the wake path among them); a worktree's first spawn and an archive restore plant none, which is why a SessionStart without one never releases the hold.
- **Wake-prompt wiring vs the flag doctrine (Minor):** the Flag section now states the wiring is inert until the composer exists: every added parameter defaults to nil, no existing call site passes a prompt, and a nil prompt encodes no field.
- **2000 px cap (Minor):** sourced — both the 2000 px and 5 MiB limits are Claude Code's own client-side caps, read from its binary.

cheapsteak:
Addressed the third review:

- **Fuzzy-score constants (Medium):** sourced. The per-character, adjacency, gap, boundary and case-transition numbers are Claude Code's own path scorer, read from its binary, applied to command names; the field weights are the weights its command menu gives the same fields. The one TBD-authored rule, the colon-split segment match, is now marked as such with its reason.
- **5-second probe deadline (Minor):** stated as ten times the measured cold start and a soak knob, matching the treatment the 14-day retention gets.

cheapsteak:
Addressed the fourth review:

- **Wake-prompt premise (High):** the Flag section no longer claims the wiring is unused. It now says the prompt is an existing, unflagged parameter that the CLI's wake command and the nightwatch skill already pass, which is why adding the app's call site needs no flag, and that the app's side alone stays inert until the composer exists. This reconciles with "Not-running delivery", which already said only the app had never passed it.

PR #822, opened 2026-09-06
Fix: A message sent to a terminal whose Claude has exited is refused, not run as a shell command

cheapsteak:
## What's broken

Send a message to a TBD terminal whose Claude has gone away, and the message is
executed as a shell command.

The pathway is ordinary. A terminal tab is running Claude; the agent exits —
because someone typed `/exit` or `Ctrl-D`, because the session was deliberately
hibernated, or because the process ended on its own — and the tmux window stays
alive with a shell prompt sitting in it. Any later `tbd terminal send` (or the
same call from the app, or from another agent coordinating across worktrees)
finds a live pane whose id matches the row, pastes the text into it, presses
Enter, and reports success. What the user sees is their prose typed at a shell:
`what did we just do?` becomes `zsh: command not found: what`. What they
expected is a refusal, or a resumed session.

The blast radius is every caller of `terminal.send` — the CLI, the app, and
agent-to-agent sends — and the failure is silent at the API boundary, because
the daemon returns success.

## Why it happens

Two gaps, one on each side of the send.

**TBD had no machine-readable "Claude is not running here" fact for the exit
case.** Claude Code fires a `SessionEnd` hook when a session ends, and TBD's
`tbd session-end` bridge forwarded the terminal identity — but it never read the
hook payload from stdin at all, so the `reason` that says *why* the session
ended never reached the daemon. Nothing was recorded, and the row went on
looking exactly like a row with a live agent in it.

**The send path's only pre-typing check asked the wrong question.** Before
typing, it asked tmux whether the pane was alive and whether it belonged to this
terminal. A shell prompt answers both affirmatively. Liveness of the *pane* is
not liveness of the *agent*, and nothing in the path distinguished them.

## What this PR does

Three changes, two rails and a small app fix.

**The exit stamp.** `tbd session-end` now reads the hook payload and forwards
its `reason`. The daemon parks the row with a new `HibernateReason.exited` for
the reasons that mean the process is leaving, and `SessionStart` retracts that
stamp — scoped to `.exited`, so a deliberate `.manual` park is not cleared by a
`/clear` in some other session. Both the stamp and its retraction broadcast the
same hibernation delta the hibernation coordinator already sends, so the app
reflects the change immediately rather than on its next poll. An exit-stamped
row is excluded from `AppState`'s focus- and tab-activation auto-wake, the same
way a `.manual` park already is, because `stampSessionExited` never replaces
the pane with an inert placeholder the way every other park does — an
automatic wake there would `respawn-window -k` the live shell Claude's exit
left behind. It wakes only via the explicit affordances (parked-pane banner
click, Wake menu, CLI/composer send, `terminal.wake`).

The stamp is a hibernation rather than a new state because the situation has a
hibernation's exact shape: process gone, terminal alive, session id known. One
state gives one wake path and one UI. `HibernateReason`'s decoder already
resolves an unknown reason to `.auto`, so an older app binary reading `"exited"`
behaves as it does for an idle-sweep park. `Sources/TBDApp/Terminal/TerminalContainerView.swift`
gained an `.exited` banner arm because the new enum case made an existing
exhaustive switch non-exhaustive — compile-forced, and covered by a test.

**The foreground rail.** A text send is refused when the row carries the stamp,
and — for a tmux row with a readable pane pid — when the process table says
Claude does not own the pane's foreground process group. It uses the same
`PaneProcessInspecting` seam the limit-resume path has always used. `--keys`
sends are deliberately unaffected: a key sequence is how you interact with a
shell, and refusing those would break the escape hatch.

The rail is gated on the terminal's KIND, which supplies the process name to
look for: `claude` for a Claude row, `codex` for a Codex one, and nothing for a
shell — a shell's pane pid *is* the shell, with no agent under it, so asking the
inspector about one would refuse every healthy shell send there is. A row whose
`kind` column is NULL is read exactly as migration `v22_terminal_kind`'s
backfill and `Terminal.isClaudeResumable` already read one: Claude when it
carries a Claude session id, shell otherwise. Reading every kindless row as
Claude regardless of session id ran this rail against a plain shell pane that
never had a "claude" process to find — the CI regression fixed below — so the
session id is now load-bearing in the mapping, not just the kind.
The mapping is one pure function, `RPCRouter.foregroundAgentName(kind:claudeSessionID:)`,
with its own test.
`PaneProcessInspecting` gained the kind-aware `foregroundAgentPID(panePID:matching:)`
— the same lowercased-substring match on the command line the Claude question
has always used, and the same lazy grandchild walk — and `foregroundClaudePID`
survives as a thin wrapper over it, so the limit-resume actuator's behavior and
its 57 tests are unchanged.

**The two rails split on a bare Enter.** An empty-text `--submit` carries no
message, and the rails treat it differently on purpose. The **park** rail takes
it: a parked row's pane holds a shell and nothing else, so an Enter there has no
message to lose and no purpose to serve, and the refusal already names wake as
the remedy. The **foreground** rail does not: a bare Enter is how a caller
answers a prompt the agent is already showing, and the inspector is the fallible
fact of the two — a pane it cannot read must not cost a live row its Enter. Both
directions are asserted.

**Codex panes are covered by the foreground rail, and by that rail only.** The
exit stamp does not reach a Codex row and deliberately never will here: the
Codex hook overlay wires `SessionStart`, `Stop`, `UserPromptSubmit` and
`PermissionRequest` and no `SessionEnd`, so `tbd session-end` is never invoked
for one — and stamping one anyway would be worse than not, because
`isClaudeResumable` is unconditionally false for a Codex row, so the wake path
would refuse it and the stamp would be a park nothing can wake. The process
table is therefore the only fact available about a Codex session's liveness, and
it is the fact this rail reads. The bug described at the top of this PR —
message typed at a shell, run, success reported — is closed for a Codex terminal
whose agent exits, in the same shape as for Claude: refused, nothing typed, with
the refusal naming the agent it looked for. Giving Codex a hook-driven stamp (and
a wake path that would make one meaningful) remains separate work.

**The wake prompt.** The app now passes `terminal.wake`'s `prompt`, which the
CLI, the RPC params, the handler, the hibernation coordinator and the spawn
builder all already carried. It is dormant on this branch — no production call
site passes a non-nil prompt until the composer lands — but the plumbing is now
complete end to end.

## Assumptions

- **The `SessionEnd` reason vocabulary is Claude Code's.** `logout`, `exit`,
  `prompt_input_exit` and `other` park the terminal; `clear`, `compact`,
  `resume`, an absent reason, and any unrecognized value do not. It is a
  whitelist, so an unknown value fails toward *not* parking and the foreground
  rail is what catches that exit. `other` is in the parking set because Claude
  Code documents it as "other exit reasons" — a same-process transition reports
  `clear`.
- **A holder-backed row is deliberately not stamped.** The hibernation
  coordinator's wake cannot respawn a pty holder, so a stamped holder row would
  be a park that nothing can wake and no reconciler can reclaim. That guard is a
  limitation, not a design preference; it lifts when holder park/wake lands
  (PR #818's area). Until then an exited holder session is not marked
  not-running, and a send falls through to the holder arm's own liveness
  handling.
- **Foreground-process inspection can be fooled.** Inherited unchanged from the
  limit-resume inspector: a `tcsetpgrp`-style foreground handoff, or a wrapper
  layer deeper than grandchildren, can produce a false refusal. The rail forks
  `ps` — one full-table `ps -axo pid=,ppid=` plus two `ps` per candidate, per
  text send to a Claude tmux terminal — and walks grandchildren only after the
  pane pid and its direct children lose. Collapsing that to a single snapshot is
  a named follow-up, not this PR.
- **The stamp records no suspended-screen snapshot**, so the parked overlay
  shows a blank pane. The `.recovery` park it displaces behaved the same way.
- **Both hook bridges read stdin without an interactive-terminal guard.** That
  is the pre-existing pattern across the hook commands, so running one by hand
  in a terminal hangs waiting for input. Follow-up, not this PR.

## Evidence & verification

**The repro.** Start Claude in a terminal tab, confirm `tbd terminal send
--terminal <id> --text "hello" --submit` lands in the session, exit the agent,
and re-run the same send: before this change the text is typed at the shell
prompt and run, and the call returns success.

**Discriminating tests.** The evidence for the send rails is
`TerminalSendNotRunningTests` — 12 cases, including `aLiveRowWithClaudeForegroundIsNotRefused`
and `aCodexTerminalWithCodexForegroundIsNotRefused` as the positive controls that
fail if the rails refuse everything, `aCodexTerminalWithNoForegroundCodexIsRefused`
which fails if the rail skips Codex panes, and the `--keys` and shell cases that
fail if the rail is applied too broadly. `aKindlessTerminalWithNoForegroundClaudeIsRefused` fails if a NULL-kind row
carrying a Claude session id skips the rail, and its sibling
`aKindlessTerminalWithNoSessionIsNotSubjectToTheForegroundRail` fails if a
NULL-kind row with no session id is wrongly read as Claude — the CI regression
below. The pair `anEmptyTextSubmitToAHibernatedRowIsRefused` /
`anEmptyTextSubmitToALiveRowSkipsTheForegroundRail` pins the bare-Enter split in
both directions. `theForegroundAgentNameFollowsTheKind`
pins the `(kind, claudeSessionID) → agent-name` mapping in every combination,
both NULL together included.
The evidence for the stamp is `SessionExitStampHandlerTests`, whose two
positives were written red first, and which asserts the delta broadcasts;
`aClearDoesNotStamp` and `sessionStartLeavesADeliberateParkAlone` are the
negatives that fail if the stamp is applied bluntly. `SessionEndReasonTests`
pins the reason whitelist, and `HibernateReasonExitedTests` pins the lenient
decode. `SessionEndHookReasonTests` covers the payload parse the stamp is built
on — reason present, reason absent, empty stdin, a payload refused on its byte
count alone at the 1 MiB cap, and malformed JSON — since every one of those must
answer nil rather than throw inside a hook.

`TerminalSendDispatchTests` passes on this branch but is **not** evidence for
the rails: its dry-run pane pid is `"0"`, so it bypasses the foreground rail
entirely and evidences send ordering only.

**CI regression fix, sha `b1249269`.** The quiet pass failed three
`TerminalSendTargetCheckLiveTests` cases ("a dead remain-on-exit pane errors…",
"a live pane belonging to another terminal is refused, naming both ids", "a
live pane carrying no identity still sends") because their fixtures call
`db.terminals.create(worktreeID:tmuxWindowID:tmuxPaneID:)` with neither `kind`
nor `claudeSessionID`, against a real tmux pane running a plain shell. The old
`foregroundAgentName(for:)` read that as `kind ?? .claude`, so the foreground
rail asked the process table for a "claude" process before the target check
under test ever ran, and refused every one of them. `foregroundAgentName` now
takes `claudeSessionID` alongside `kind` and reads a NULL kind exactly as
migration `v22_terminal_kind`'s backfill and `Terminal.isClaudeResumable`
already do — Claude when the row carries a Claude session id, shell otherwise
— so a plain-shell fixture is no longer subject to the rail.
`Tests/TBDDaemonTests/TerminalSendNotRunningTests.swift` gained the
NULL-kind/no-session case and its unit-level mapping assertion; the live suite
itself is unchanged. Dispatched CI run:
https://github.com/cheapsteak/tbd/actions/runs/34022464777.

**Live verification is deferred to the human.** This machine runs a live agent
fleet, so restarting the app and daemon to drive the refusal by hand is not
something this branch's work could do. The discriminating unit tests above and
the CI run are the evidence here; the by-hand run — send, exit, refused send,
`tbd terminal wake --prompt`, send succeeds again — is worth doing before merge.

## No flag

These ship unflagged, as bug fixes. The transcript composer design's Flag
section (`docs/specs/2026-09-05-transcript-composer-design.md`) now carries
both rationales by name, each in its own paragraph: the wake-prompt paragraph
this carve-out was originally named after, and a paragraph for the exit stamp
— it records a fact the agent's own `SessionEnd` hook already reported rather
than deciding anything, unlike the two flagged auto-park causes in this
codebase (`auto_hibernate_enabled`, the merge park's gate in
`HibernationGate`); it is scoped by session incarnation and excludes holder
transport; and it is retracted by `SessionStart` or the wake path, so a wrong
stamp costs one banner and one refused send rather than lost input. The
refusal it enables is the fix for a real defect: text typed at a dead Claude
executes in its shell.

This PR is now stacked on #821 (base branch `tbd/20260905-abundant-earwig`), so
the spec these comments cite is reachable from this PR's own HEAD rather than
only from a branch this PR cannot see. GitHub retargets the base back to `main`
automatically once #821 merges.

An exit-stamped row is excluded from focus- and tab-activation auto-wake for
the same no-flag reason as the stamp itself: it is a targeted refusal against a
real defect (an unguarded auto-wake would kill the live shell an exit leaves
behind), not a new autonomous behavior, so it needs no soak.

**That app-side exclusion is the first line, not the only one.** It lives in
the app, and [`docs/updating.md`](docs/updating.md)'s `--no-app` makes daemon-newer-than-app a
supported skew: an app binary older than this PR decodes `.exited` through the
lenient `HibernateReason` decoder as `.auto`, and its focus-wake and
tab-activation still fire `terminal.wake` — which is `respawn-window -k` over
the live shell the stamp exists to preserve. So the daemon defends the wake
itself, for every caller (an old app, the CLI, the composer) rather than
trusting who asked: an exit-stamped tmux row whose pane's foreground process is
not the pane's own shell is refused with `WakeResult.paneBusy`, naming the pid
and the two ways out (finish the process, or wake from the terminal). An idle
shell — the pane's own pid owning the foreground group — wakes exactly as
before, and an unreadable pane pid proceeds for the same reason the send rail's
does. That guard is also unflagged and for the same reason: it is a refusal
against a defect, it mutates nothing, and it leaves the row exit-stamped so a
retry after the process finishes still wakes.

## Reconciler

No new durable external resource. This writes two columns on an existing row and
refuses an existing act; the stamp is retracted by `SessionStart` or by the wake
path. Nothing new can orphan.

## Coordination

Touches `performTerminalSend` only. `performHolderSend`, `HolderInjectionCourier`
and everything under `Sources/TBDDaemon/Holder/` are untouched — PR #816 owns
them.

## Follow-ups

**A narrower TOCTOU remains between the park/foreground check and the paste**
in `performTerminalSend`: the rails read a snapshot of the row and the
foreground process, and the actual paste lands several awaits later. A
`SessionEnd` stamp that arrives in that window is not caught. It is narrower
than the accepted race already named in this same file — the assumption above
that foreground-process inspection can be fooled — and is left for a
follow-up rather than fixed here.

[composer-a](https://cheapsteak.github.io/tbd/open/?worktree=B6F8AB9D-8BD6-4902-98EF-16F8DCA04FE5)

## Summary by CodeRabbit

- **New Features**
  - Claude sessions that end are marked as exited with clearer hibernation banners.
  - Wake requests can include an optional prompt when sessions resume.
  - Session-end reasons help determine whether terminals should be parked.
  - Wake-on-focus and tab activation avoid automatically reopening exited sessions.
  - Wake operations refuse to replace another process occupying an exited terminal.

- **Bug Fixes**
  - Text messages are blocked for inactive or no-longer-running Claude and Codex sessions, while key input remains available.
  - Session state refreshes correctly when sessions start or end, preserving intentional manual parks.

- **Tests**
  - Added coverage for wake prompts, exit handling, banners, guarded message delivery, and busy terminals.

### DIFF
--- a/Sources/TBDApp/AppState+Hibernation.swift
+++ b/Sources/TBDApp/AppState+Hibernation.swift
@@ -311,9 +311,16 @@ extension AppState {
     /// the worktree. They wake only via the explicit affordances (parked-pane
     /// click, Wake menu). nil-reason (legacy), auto, and recovery parks still
     /// focus-wake.
+    ///
+    /// Exit-stamped sessions (`hibernateReason == .exited`) are excluded for a
+    /// different reason: `stampSessionExited` writes only two DB columns and
+    /// never replaces the pane with an inert placeholder the way every other
+    /// park path does, so a focus-wake would run `tmux respawn-window -k`
+    /// against the live shell Claude's exit left behind — killing it. They
+    /// wake only via the same explicit affordances as `.manual`.
     func terminalIDToWakeOnFocus(worktreeID: UUID) -> UUID? {
         let parked = (terminals[worktreeID] ?? []).filter {
-            $0.isParked && $0.isClaudeResumable && $0.hibernateReason != .manual
+            $0.isParked && $0.isClaudeResumable && $0.hibernateReason != .manual && $0.hibernateReason != .exited
         }
         guard !parked.isEmpty else { return nil }
         if let focusedID = terminalIDForAutofocus(worktreeID: worktreeID),
@@ -329,12 +336,21 @@ extension AppState {
     /// when the tab has no terminals, none are parked, or none are resumable.
     /// Called by `setActiveTab` to wake only what the user is about to see,
     /// never a parallel spawn storm.
+    ///
+    /// Excludes `.manual` (an explicit "Hibernate now" must not be undone by
+    /// tab activation) AND `.exited` (`stampSessionExited` never replaces the
+    /// pane with an inert placeholder, so waking it would `respawn-window -k`
+    /// the live shell Claude's exit left behind — killing it). Both wake only
+    /// via explicit affordances.
     func terminalIDsToWakeOnTabActivation(worktreeID: UUID, tabIndex: Int) -> [UUID] {
         guard let worktreeTabs = tabs[worktreeID], worktreeTabs.indices.contains(tabIndex) else { return [] }
         let tab = worktreeTabs[tabIndex]
         let terminalIDsInTab = terminalIDs(in: tab)
         return (terminals[worktreeID] ?? [])
-            .filter { terminalIDsInTab.contains($0.id) && $0.isParked && $0.isClaudeResumable && $0.hibernateReason != .manual }
+            .filter {
+                terminalIDsInTab.contains($0.id) && $0.isParked && $0.isClaudeResumable
+                    && $0.hibernateReason != .manual && $0.hibernateReason != .exited
+            }
             .map { $0.id }
     }
 }
--- a/Sources/TBDApp/AppState+Hibernation.swift
+++ b/Sources/TBDApp/AppState+Hibernation.swift
@@ -59,15 +59,15 @@ extension AppState {
     /// wraps this. `fallbackToDefaultProfile` opts into the ambient default login
     /// when the pinned profile is gone (used by the fallback retry).
     func wakeTerminalOutcome(terminalID: UUID, worktreeID: UUID, userInitiated: Bool,
-                             fallbackToDefaultProfile: Bool) async -> WakeAttemptOutcome {
+                             fallbackToDefaultProfile: Bool,
+                             prompt: String? = nil) async -> WakeAttemptOutcome {
         guard !wakeInFlight.contains(terminalID) else { return .ok }  // in-flight dedupe = benign no-op
         wakeInFlight.insert(terminalID)
         defer { wakeInFlight.remove(terminalID) }
         do {
             let size = mainAreaTerminalSize()
-            try await daemonClient.terminalWake(
-                terminalID: terminalID, cols: size.cols, rows: size.rows,
-                fallbackToDefaultProfile: fallbackToDefaultProfile)
+            try await terminalWakeSender(
+                terminalID, size.cols, size.rows, fallbackToDefaultProfile, prompt)
             await refreshTerminals(worktreeID: worktreeID)
             return .ok
         } catch {
@@ -103,9 +103,11 @@ extension AppState {
     /// windows) fired a modal per parked terminal. `userInitiated` drives the
     /// failure log level (explicit action → error; background → warning).
     @discardableResult
-    func wakeTerminal(terminalID: UUID, worktreeID: UUID, userInitiated: Bool) async -> String? {
+    func wakeTerminal(terminalID: UUID, worktreeID: UUID, userInitiated: Bool,
+                      prompt: String? = nil) async -> String? {
         switch await wakeTerminalOutcome(terminalID: terminalID, worktreeID: worktreeID,
-                                         userInitiated: userInitiated, fallbackToDefaultProfile: false) {
+                                         userInitiated: userInitiated,
+                                         fallbackToDefaultProfile: false, prompt: prompt) {
         case .ok: return nil
         case .failed(let m): return m
         case .profileMissing(_, _, let m): return m
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -1445,6 +1445,17 @@ final class AppState {
         { [daemonClient] enabled in
             try await daemonClient.setHolderHibernationEnabled(enabled: enabled)
         }
+    /// How `wakeTerminalOutcome` reaches the daemon — injectable for the same
+    /// reason as `controlModeSetter`, so the wake's prompt plumbing and its
+    /// failure branches are testable without a running daemon.
+    @ObservationIgnored
+    lazy var terminalWakeSender:
+        @MainActor (UUID, Int?, Int?, Bool, String?) async throws -> Void =
+        { [daemonClient] terminalID, cols, rows, fallback, prompt in
+            try await daemonClient.terminalWake(
+                terminalID: terminalID, cols: cols, rows: rows,
+                fallbackToDefaultProfile: fallback, prompt: prompt)
+        }
     /// How `setClaudeCloudEnabled` persists the Claude cloud gate — injectable
     /// for the same reason as `controlModeSetter`, so the Settings toggle's
     /// success and failure branches are testable without a real daemon.
--- a/Sources/TBDApp/DaemonClient.swift
+++ b/Sources/TBDApp/DaemonClient.swift
@@ -1632,10 +1632,21 @@ actor DaemonClient {
 
     /// Wake a hibernated terminal: respawn `claude --resume` in its window.
     /// Idempotent — a double-call collapses to one respawn daemon-side.
-    func terminalWake(terminalID: UUID, cols: Int? = nil, rows: Int? = nil, fallbackToDefaultProfile: Bool = false) async throws {
+    ///
+    /// `prompt` is delivered as a trailing argv on `claude --resume`, atomic with
+    /// the respawn, so it reaches ONLY a session this call actually woke. On the
+    /// idempotent no-op paths (already awake, wake already in flight) the daemon
+    /// returns `woken: false` and delivers nothing — which is what makes the
+    /// parameter safe to pass from a UI that may be racing a background wake.
+    func terminalWake(
+        terminalID: UUID, cols: Int? = nil, rows: Int? = nil,
+        fallbackToDefaultProfile: Bool = false, prompt: String? = nil
+    ) async throws {
         try await callVoidAsync(
             method: RPCMethod.terminalWake,
-            params: TerminalWakeParams(terminalID: terminalID, cols: cols, rows: rows, fallbackToDefaultProfile: fallbackToDefaultProfile)
+            params: TerminalWakeParams(
+                terminalID: terminalID, cols: cols, rows: rows,
+                fallbackToDefaultProfile: fallbackToDefaultProfile, prompt: prompt)
         )
     }
 
--- a/Sources/TBDApp/Terminal/TerminalContainerView.swift
+++ b/Sources/TBDApp/Terminal/TerminalContainerView.swift
@@ -407,6 +407,8 @@ enum HibernatedBannerModel {
             return "Parked after a restart — click anywhere in the pane to resume"
         case .merged:
             return "Hibernated after the PR merged — click anywhere in the pane to resume"
+        case .exited:
+            return "Claude exited — click anywhere in the pane to resume"
         case .auto, nil:
             return "Hibernated while idle — click anywhere in the pane to resume"
         }
--- a/Sources/TBDCLI/Commands/SessionEndCommand.swift
+++ b/Sources/TBDCLI/Commands/SessionEndCommand.swift
@@ -42,16 +42,18 @@ struct SessionEndCommand: AsyncParsableCommand {
 
     mutating func run() async throws {
         let environment = ProcessInfo.processInfo.environment
+
+        // Read stdin BEFORE any early return, including the terminal-ID guard
+        // below: Claude Code pipes the payload and waits on the write, so a
+        // command that exits without draining it can leave the hook's writer
+        // blocked on a full pipe until its timeout.
+        let reason = Self.hookReason(from: FileHandle.standardInput.readDataToEndOfFile())
+
         guard let terminalIDString = environment["TBD_TERMINAL_ID"],
               let terminalID = UUID(uuidString: terminalIDString) else {
             sessionEndLogger.debug("suppressed reason=noTerminalID")
             return
         }
-
-        // Read stdin BEFORE the daemon check: Claude Code pipes the payload and
-        // waits on the write, so a command that exits without draining it can
-        // leave the hook's writer blocked on a full pipe until its timeout.
-        let reason = Self.hookReason(from: FileHandle.standardInput.readDataToEndOfFile())
 
         let client = SocketClient()
         guard client.isDaemonRunning else {
--- a/Sources/TBDCLI/Commands/SessionEndCommand.swift
+++ b/Sources/TBDCLI/Commands/SessionEndCommand.swift
@@ -6,7 +6,8 @@ import TBDShared
 private let sessionEndLogger = Logger(subsystem: "com.tbd.cli", category: "sessionEnd")
 
 /// Bridges Claude Code's `SessionEnd` hook into TBD so a session that exits
-/// while background agents are live cannot leave a standing delegation claim.
+/// while background agents are live cannot leave a standing delegation claim,
+/// and so the daemon knows the session's process is gone.
 ///
 /// Every failure is silent, like every other hook bridge: a hook must never
 /// wedge or redden the agent it observes.
@@ -17,8 +18,26 @@ struct SessionEndCommand: AsyncParsableCommand {
         shouldDisplay: false
     )
 
+    /// The fields this command reads out of Claude Code's `SessionEnd` payload.
+    /// Everything optional, everything else ignored — payload extensions are
+    /// additive and must not break the bridge.
+    private struct HookPayload: Decodable {
+        let reason: String?
+    }
+
     static func sessionIncarnationID(environment: [String: String]) -> UUID? {
         environment["TBD_TERMINAL_INCARNATION_ID"].flatMap(UUID.init(uuidString:))
+    }
+
+    /// Lift `reason` out of a hook payload. Static and pure so the parse is
+    /// testable without a pipe. The 1 MiB cap bounds *processing*, matching
+    /// `SessionEventCommand`; a payload that is empty, oversized, or not the
+    /// object this expects yields nil, which the daemon reads as "we do not know".
+    static func hookReason(from data: Data) -> String? {
+        guard !data.isEmpty, data.count <= 1 << 20,
+              let payload = try? JSONDecoder().decode(HookPayload.self, from: data)
+        else { return nil }
+        return payload.reason
     }
 
     mutating func run() async throws {
@@ -28,6 +47,11 @@ struct SessionEndCommand: AsyncParsableCommand {
             sessionEndLogger.debug("suppressed reason=noTerminalID")
             return
         }
+
+        // Read stdin BEFORE the daemon check: Claude Code pipes the payload and
+        // waits on the write, so a command that exits without draining it can
+        // leave the hook's writer blocked on a full pipe until its timeout.
+        let reason = Self.hookReason(from: FileHandle.standardInput.readDataToEndOfFile())
 
         let client = SocketClient()
         guard client.isDaemonRunning else {
@@ -40,7 +64,8 @@ struct SessionEndCommand: AsyncParsableCommand {
                 method: RPCMethod.terminalSessionEnded,
                 params: TerminalSessionEndedParams(
                     terminalID: terminalID,
-                    sessionIncarnationID: Self.sessionIncarnationID(environment: environment))
+                    sessionIncarnationID: Self.sessionIncarnationID(environment: environment),
+                    reason: reason)
             )
         } catch {
             sessionEndLogger.debug(
--- a/Sources/TBDDaemon/Actuation/ActuationOutcome+Park.swift
+++ b/Sources/TBDDaemon/Actuation/ActuationOutcome+Park.swift
@@ -57,6 +57,9 @@ extension ActuationOutcome {
         // The row's transport has a park/wake mechanic, but this install has
         // not armed it: `holder_hibernation_enabled` is off.
         case .holderTransport: return .refused(.notEligible)
+        // The row is exit-stamped and its pane is busy: a well-formed wake the
+        // daemon declines, and nothing was respawned.
+        case .paneBusy: return .refused(.notEligible)
         }
     }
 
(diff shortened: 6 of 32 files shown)


cheapsteak:
Review findings addressed in 5f5b1c16.

- **HIGH — a Codex pane whose agent exits was caught by neither rail: closed.** The foreground rail is now kind-aware rather than Claude-only. `PaneProcessInspecting` gained `foregroundAgentPID(panePID:matching:)` — the same lowercased command-line substring match and the same lazy grandchild walk, with the name supplied by the caller — and `foregroundClaudePID` remains a thin wrapper over it, so `LimitResumeActuator`'s behavior and its 57 tests are unchanged. `performTerminalSend` picks the name from one pure function, `RPCRouter.foregroundAgentName(for:)` (`.claude` → `claude`, `.codex` → `codex`, `.shell` → nil), and the refusal names the agent it looked for. Codex rows are deliberately still not exit-stamped: `isClaudeResumable` is false for one, so a stamp would be a park the wake path refuses — the process table is the only liveness fact a Codex session offers, and it is now the one this rail reads. `TerminalSendNotRunningTests` goes 7 → 9: the old "codex is not subject to the rail" case is replaced by two discriminating ones (no codex foreground → refused, naming `(codex)`; codex foreground → proceeds), plus a test pinning the kind → name mapping in both directions. The shell case stays as the never-subject control. `--filter LimitResume` (57), `--filter TerminalSendNotRunning` (9) and `--filter TerminalSendDispatchTests` (33) all green — 99 tests in one run.
- **MEDIUM — the spec path.** The design write-up for this rail lives in #821, which lands first; this PR is the implementation half and does not duplicate it.

cheapsteak:
Second gate round addressed in a95afff0 — all four applied.

- **MEDIUM, `foregroundAgentName(for:)` skipped a kindless row.** Fixed: `kind ?? .claude`, not `?? .shell`. A NULL `kind` predates the column and is a legacy Claude terminal — the reading `Terminal.isClaudeResumable` and the continue-in-Codex path already give one — and defaulting it to `shell` skipped the rail for exactly the oldest rows on the machine, the likeliest to have outlived their agent. The doc comment now says so, `theForegroundAgentNameFollowsTheKind` asserts `nil → "claude"`, and a new handler-level case, `aKindlessTerminalWithNoForegroundClaudeIsRefused`, drives a `kind: nil` row with no foreground claude to a refusal that names `(claude)`.
- **MEDIUM, the two comments cite a spec that lands in #821.** Kept the path — it is the durable reference — and appended ", landing in PR #821" at both sites (`Models.swift` `HibernateReason.exited`, and `handleTerminalSessionEnded`), so a reader on `main` before that merge knows where to look. The inlined rationale stays, so neither comment depends on the file being there.
- **MINOR, an empty-text `--submit` bypassed both rails.** The **park** rail now covers every text payload, empty included: a parked row's pane holds a shell, so a bare Enter has no message to lose and no purpose to serve, and the refusal already names wake as the remedy. The **foreground** rail deliberately keeps `!body.isEmpty` — a bare Enter is how a caller answers a prompt the agent is already showing, and the inspector is the fallible fact of the two. The split and its reasoning are in the comment above the rails, and both directions are asserted: `anEmptyTextSubmitToAHibernatedRowIsRefused` and `anEmptyTextSubmitToALiveRowSkipsTheForegroundRail`.
- **MINOR, `hookReason(from:)` had no test.** `TBDCLITests` already exists, so the new `Tests/TBDCLITests/SessionEndHookReasonTests.swift` sits beside `SessionEventCommandTests`. Five cases: reason present (with unknown fields ignored), reason absent (`{}`, no key, explicit null), empty stdin, oversized — asserted on the byte count alone, since the cap bounds processing: 1 MiB exactly parses to `"clear"`, 1 MiB + 1 answers nil — and malformed JSON plus JSON that is not the expected object. All nil rather than throwing, which is what a hook bridge owes the agent it observes.

Verification: `scripts/test.sh --filter "TerminalSendNotRunningTests|SessionEndHookReason"` went 5 issues across 3 tests red → 17 tests in 2 suites green; `scripts/test.sh --filter TerminalSendDispatchTests` 33/33 green. `swiftlint --strict` clean on all four touched files. PR description updated for the kindless-row reading, the bare-Enter rule, and the spec's landing order.

cheapsteak:
Addressed the four round-3 review-gate findings:

1. **MEDIUM — dangling spec citation.** `docs/specs/2026-09-05-transcript-composer-design.md` lived only on the unmerged spec branch (`tbd/20260905-abundant-earwig`, PR #821, now green). Rebased this branch onto `origin/tbd/20260905-abundant-earwig` (clean, no conflicts — that branch only adds the spec file) and retargeted this PR's base to `tbd/20260905-abundant-earwig`. The spec file is now present in this PR's tree, and `git log origin/tbd/20260905-abundant-earwig..HEAD` lists exactly this PR's 16 commits. The two code comments citing the spec are left as-is — they stay true. GitHub will retarget the base to `main` automatically once #821 merges.

2. **MEDIUM — unflagged exit stamp.** No flag added. Rewrote the "## No flag" section to explain why: the stamp records a fact the agent's own `SessionEnd` hook already reported (the process has ended), so it kills nothing and sends nothing to a session — unlike `auto_hibernate_enabled` and the merge-hibernate gate, which each initiate a kill. It's scoped by the session-incarnation check and the holder-transport exclusion, and retracted by `SessionStart` or the wake path, so a wrong stamp costs one banner and one refused send rather than lost input. The refusal it enables fixes a real defect: text typed at a dead Claude executes in its shell. The section also now notes the PR is stacked on #821.

3. **MINOR — stdin left undrained on the missing-`TBD_TERMINAL_ID` early return.** Fixed in `Sources/TBDCLI/Commands/SessionEndCommand.swift`: the stdin read now happens unconditionally at the top of `run()`, before the terminal-ID guard, so every invocation drains the hook's payload regardless of which early return fires. `hookReason(from:)` parsing is unchanged in place; only the read call and its comment moved. `scripts/test.sh --filter "SessionEndHookReason|TerminalSendNotRunningTests"` — 17 tests, 0 failures.

4. **MINOR — TOCTOU between the park/foreground check and the paste.** No code change, per ruling. Added a "## Follow-ups" section to the PR body naming the race: the rails read a snapshot and the paste lands several awaits later, so a `SessionEnd` stamp landing in that window isn't caught. It's narrower than the already-accepted foreground-inspection race documented in the same file, and is left for a follow-up.

Claude-Session: https://claude.ai/code/session_01E9SPoxoVoApcVmduLiBWik

cheapsteak:
Round-4 review-gate fixes:

- **Finding 1** (MEDIUM, correctness — `stampSessionExited` skipped the incarnation guard entirely when `reportedIncarnationID` was nil) → 61d64dee. Switched to exact equality (`record.sessionIncarnationID == reportedIncarnationID?.uuidString`), matching the three sibling writers. Added `stampDeclinesANilReportAgainstANamedIncarnation` (RED before, GREEN after) alongside the existing nil-report/nil-record positive case. Checked `clearSessionExitStamp` for the same asymmetry: it takes no `reportedIncarnationID` at all and only runs after `applySessionStart` has already verified an exact-equality incarnation match in the same request, so the defect can't arise there — no change needed.
- **Finding 2** (MEDIUM, convention — the spec's "ships unflagged" paragraph carried only the wake-prompt's rationale) → d612ffc2. Added the exit stamp's own paragraph to `docs/specs/2026-09-05-transcript-composer-design.md`'s Flag section, as timeless design prose. Also updated the PR body's "No flag" section to point at that paragraph instead of restating the rationale independently.

`scripts/test.sh --filter "TerminalExitStampStoreTests|TerminalSendNotRunningTests"`: 19/19 passing after the fix (2 issues, both in the new test, before it).

Claude-Session: https://claude.ai/code/session_01E9SPoxoVoApcVmduLiBWik

cheapsteak:
## Review-gate round-6 fix (HIGH)

**Finding:** An exit-stamped row (`hibernateReason == .exited`, introduced by this PR) is `isParked`, so `AppState.terminalIDToWakeOnFocus(worktreeID:)` (`Sources/TBDApp/AppState+Hibernation.swift`) and `terminalIDsToWakeOnTabActivation` auto-woke it on ordinary focus or tab activation, exactly like `.auto`/`.recovery`/nil. Waking runs `tmux respawn-window -k`, which is safe for every other park because the park path had already replaced the pane with an inert placeholder — but `stampSessionExited` writes only two DB columns and leaves the pane holding the live shell Claude's exit left behind. So a tab switch could kill that shell.

**Ruling:** Exclude `.exited` from both auto-wake decisions, the same way `.manual` is excluded — an exit-stamped row now wakes only on an explicit gesture (parked-pane banner click, Wake menu, CLI/composer send, `terminal.wake`).

**Fix:** commit `1f94e598`. Tests-first in `Tests/TBDAppTests/WakeOnFocusDecisionTests.swift` and `Tests/TBDAppTests/WakeOnTabActivationTests.swift` (RED confirmed before the fix, 2/49 failing; GREEN after, 49/49 passing). `swiftlint lint --strict` clean on touched files.

**Dispatched CI run:** https://github.com/cheapsteak/tbd/actions/runs/34023308166

cheapsteak:
CI regression fix for the three `TerminalSendTargetCheckLiveTests` quiet-pass failures, sha `b1249269`.

**Root cause.** Those fixtures call `db.terminals.create(worktreeID:tmuxWindowID:tmuxPaneID:)` with neither `kind` nor `claudeSessionID`, against a real tmux pane running a plain shell. `RPCRouter.foregroundAgentName(for:)` read a `nil` kind as `kind ?? .claude`, so the foreground rail asked the process table for a "claude" process before the target check under test ever ran, and refused every send.

**Fix.** A NULL `kind` is now read exactly as migration `v22_terminal_kind`'s backfill and `Terminal.isClaudeResumable` already read one: Claude when the row carries a `claudeSessionID`, shell otherwise. `foregroundAgentName` takes `claudeSessionID` alongside `kind`; the call site in `performTerminalSend` and both doc comments describing the mapping are updated to match.

**Tests.** `Tests/TBDDaemonTests/TerminalSendNotRunningTests.swift` gained `aKindlessTerminalWithNoSessionIsNotSubjectToTheForegroundRail` (the exact NULL/NULL shape the live fixtures use) and extended `theForegroundAgentNameFollowsTheKind` with the `(nil, nil) → nil` case. 51 tests in the touched suites pass locally; `swiftlint --strict` is clean on both files.

**CI verdict.** Dispatched run https://github.com/cheapsteak/tbd/actions/runs/34022464777 was pre-empted by a sibling commit's dispatch on this same branch; the superseding run (stacked on top of this fix), https://github.com/cheapsteak/tbd/actions/runs/34023308166, finished **green** — fast pass 1/2, fast pass 2/2, and the quiet pass (which runs `TerminalSendTargetCheckLiveTests`) all succeeded.

PR body's Evidence & verification section updated accordingly.

Claude-Session: https://claude.ai/code/session_01E9SPoxoVoApcVmduLiBWik

cheapsteak:
Restacked onto `tbd/20260905-abundant-earwig`, which was brought up to today's `main` (now carrying #819, the holder send path). The 20 commits replayed with no conflicts; new head `d0e164f6`.

Full test workflow: https://github.com/cheapsteak/tbd/actions/runs/34041663565

cheapsteak:
restacked onto main @ 406a9800 (#818, #827); conflicts resolved in Sources/TBDApp/AppState.swift; run https://github.com/cheapsteak/tbd/actions/runs/34054417834

cheapsteak:
Review-gate fix round — both MEDIUM findings addressed in `5ac61831`.

**1. The foreground rail's fail-open was untested** (`RPCRouter+TerminalHandlers.swift`). The design and its comment stand; the three swallowed conditions are now pinned in `Tests/TBDDaemonTests/TerminalSendNotRunningTests.swift`:

- `anUnaskablePanePIDLetsTheSendThrough` — the `tmux.panePID` call throws and the `try?` swallows it.
- `aNonNumericPanePIDLetsTheSendThrough` — tmux answered `abc`.
- `aZeroPanePIDLetsTheSendThrough` — tmux's own "I don't know", the literal `0`.

Each runs on the same fixture whose *readable* pid is refused (`aLiveRowWithNoForegroundClaudeIsRefused`, same stub inspector answering "no claude here"), so a send that goes through went through because the pid was unreadable and for no other reason. Each also asserts the paste actually reached the pane, not merely that the response said success. `TmuxManager.dryRunPanePID` became throwing — matching `dryRunPaneSendTarget` — so "tmux could not answer" is expressible at all.

**2. An older app binary decodes `.exited` as `.auto` and still auto-wakes** (`Models.swift`). The app-side exclusion stays the first line; the daemon now defends the wake itself, so the answer no longer depends on who asked.

In `HibernationCoordinator.wake`, before the tmux mutation section: an exit-stamped tmux row whose pane's **foreground process is not the pane's own shell** returns the new `WakeResult.paneBusy(pid:)` and respawns nothing. The row stays exit-stamped, so a retry once that process finishes still wakes. The refusal text (`HibernationCoordinator.paneBusyRefusal`) names the pid and the two ways out — finish or stop the process, or wake from the terminal itself — and is shared by `terminal.wake`, the manual-resume handler and the actuation record.

The question is asked of the process table, never the screen: `PaneProcessInspecting` gains `paneForegroundPID(panePID:)` beside the name-matching `foregroundAgentPID`, and the coordinator takes the inspector as an injected seam. An unreadable pane pid proceeds, for the same reason it does on the send rail.

Tests in `HibernationCoordinatorTests`:

- `exitStampedWakeIsRefusedWhenThePaneIsBusy` — foreground pid 9999 ≠ pane pid 4242 → `.paneBusy(pid: 9999)`, no `respawn-window` recorded, row still exit-stamped. Red without the guard, where the case does not exist and the wake returns `.ok` after respawning.
- `exitStampedWakeProceedsWhenThePaneIsAnIdleShell` — foreground pid == pane pid → `.ok`, and the ordinary `respawn-window … claude --resume sess-1` is recorded.
- `aDeliberatelyParkedRowIsUnaffectedByTheBusyPaneGuard` — the same busy inspector on a `.manual` park still wakes, which is what makes the first case's refusal attributable to the exit stamp rather than to the fixture.

**Verification.** `scripts/test.sh --filter "TerminalSendNotRunningTests|TerminalExitStampStoreTests|HibernationCoordinatorTests|WakeOnFocusDecisionTests"` → 135 tests in 6 suites, 0 failures. `swiftlint --strict` clean on all nine touched files. The PR body's "No flag" section now carries the version-skew scenario and this guard.

PR #823, opened 2026-09-06
Feat: A default-off transcript-composer flag, and terminal.completions from the session's own Claude Code

cheapsteak:
## Summary

Somebody watching a live Claude transcript is about to get a message box underneath it. This PR lands the two things that box needs before it can exist: the flag that gates the whole feature, and a way for the daemon to tell the app which slash commands, skills and subagents *this particular session* can complete.

Nothing user-visible changes yet. The one thing a person can see is a Settings toggle, and it is off.

## Why this shape

The design spec is `docs/specs/2026-09-05-transcript-composer-design.md`, and this PR implements its Part B. **This PR is stacked on #822** (base `tbd/composer-a-not-running-guard`), which is itself stacked on #821, the PR that adds that spec. The spec is therefore reachable from this HEAD — a reviewer can read the design in this checkout. GitHub retargets each base automatically as the PRs below merge.

Four decisions carry the shape:

- **The list comes from the installed binary, never from a table inside TBD.** The daemon asks Claude Code itself, over the control protocol's `initialize` request, and gets back the commands and agents that binary actually has. A command Claude Code ships next week, or a skill the person installs this afternoon, appears in the composer with no TBD release in between.
- **The probe asks the executable the session is *running*, not the one on `PATH`.** Five Claude Code versions sit side by side on a working machine; during development two of them were running at once and reported different command counts. Completing against the wrong one is a wrong answer that looks right. The probe resolves the executable from the session's own child process, and deliberately declines when that process is still a login shell that has not exec'd yet — better to fall through to the scan than to pin `/bin/zsh` as the binary.
- **Every flag on the probe's command line closes a side effect that was measured, not guessed.** That is what lets `arguments(mcpConfigPath:)` be a pure function with its exact argument vector asserted in a test: the argument list *is* the safety property, so it is pinned rather than described.
- **The probe runs behind the same per-config-dir lane as the folder-trust seeder**, because both rewrite `.claude.json` and a read-modify-write race there loses somebody's trust entry.

Two alternatives were considered and rejected. **A compiled-in list of built-in commands** is stale the day Claude Code ships a new one, and TBD would have to cut a release to catch up — the exact coupling the first decision above exists to avoid. **Taking the built-ins from the headless `init` frame** looked cheaper, but that frame arrives only after a real, billed message, it omits roughly twenty interactive-only commands, and it carries no descriptions — so the composer would show a shorter list, with blank help text, at the cost of a paid turn.

## What this PR does

- **The flag chain.** A migration adding `transcript_composer_enabled` to `config`; the GRDB record; the `TBDShared` model; `ConfigStore` read/write; a `config.setTranscriptComposerEnabled` RPC; the flag in `daemon.capabilities`.
- **`terminal.completions`.** A new RPC, gated on the flag, served by `CompletionInventoryService`: it probes the session's own binary, caches the answer keyed on a fingerprint of the config directory, and falls back to a filesystem scan when the probe cannot answer (no live child yet, a shell that has not exec'd, a timeout). The result carries its own provenance — `source` (probe or scan) and `freshness` (fresh, stale, fallback) — so a caller can tell a real inventory from a best guess.
- **`ClaudeCompletionProbe`.** Spawns the pinned executable with the control protocol on stdio, sends `initialize`, decodes the response, and races a five-second deadline.
- **`ClaudeCompletionScan`.** The fallback: walks `commands/`, `skills/`, `agents/` and enabled plugin trees under the session's config directory and the worktree, reading frontmatter for descriptions.
- **`ClaudeConfigDirSerializer`.** A per-config-directory lane, now taken by all four writers of `.claude.json`: the trust seeder, the probe, and both `ClaudeProfileConfigDirManager.ensure*Dir` entry points, whose read-modify-write was previously unlocked and could lose a folder-trust key to a probe running concurrently.
- **A deadline fix in `BoundedProcessRunner`**, found by this branch's own test — see below.
- **`ClaudeSpawnCommandBuilder.routingEnv`.** The non-secret routing environment, extracted so the spawn and the probe compose it from one place instead of two.
- **The Settings toggle**, beside the live-transcript toggle it belongs to.
- **`tbd terminal completions --terminal <id> [--json]`**, for reading what the probe returned without a UI.

## A bounded-subprocess deadline bug this branch surfaced

The probe is the first caller of `runBoundedProcess` whose child can hold its stdout open past the deadline, and doing so exposed a real bug in that shared runner.

The deadline action did its whole job on the single `SubprocessWatchdog` thread — the one thread that fires every bounded deadline in the daemon — and part of that job could wait. It snapshotted the child's output pipes, and `snapshot()` must acquire the accumulator lock, which a drain worker can hold across a blocking `FileHandle.availableData` on a pipe read end whose write end the child, or a grandchild that inherited it, still holds.

A wait there is self-sustaining and fleet-wide. The SIGKILL escalation is queued on the very thread the action is parked on, so the one event that would close the write end can never run; and `GitManager`, `TmuxManager`, `ProviderRunner` and every completion probe lose their deadlines behind it. Signalling before snapshotting narrows the window — the kill is what makes a parked read return — but it does not remove the wait from that thread, and the escalation that would deliver the kill is queued behind it.

**The final shape is: signal first, snapshot off the watchdog thread.** The action now does only bookkeeping — SIGTERM, queue the SIGKILL escalation — and hands the snapshot to the task that is already awaiting the call, which runs it on the way out. A `.timedOut` outcome carries no output, so nothing about the result waits on the snapshot, and a wedged drain can delay only the call that owns it, never the shared mechanism. The `.pseudoTerminal` path is untouched; it was never exposed, because it sets `O_NONBLOCK` on the primary at creation.

To be precise about the novelty, because the runner's own comments contradict the easy version of this claim: silent children were *known* — `GitManager` documents a 95-minute `git fetch` as the reason its deadline exists. What was missing was a test that holds a child silent past its deadline, and the probe's is the first in tree. **The fix lands in the shared runner, so it covers every existing caller**, that `git fetch` included — not just the probe.

Two new tests in `BoundedProcessRunnerTests` pin the guarantee against a child that ignores SIGTERM (`trap '' TERM` before `exec`, so `SIG_IGN` survives the exec and the sleeper itself is TERM-proof): the call resolves at its deadline and the child is gone, which only the escalation can achieve; and a second, concurrently-armed call's deadline *and* its escalation are both served while the first call's timeout path is resolving.

**Verified by mutation, twice, because the hazard cannot be provoked from a child process** — Darwin's `readabilityHandler` never fires without data, so the parked drain worker is a latent lock-ordering hazard rather than a reproducible one. Blocking the deadline action for 30 s where the snapshot used to run turns both tests red on their wall-clock bounds (30.6 s / 30.9 s against 1.2 s / 1.4 s green). Blocking it *only* on the watchdog thread leaves every deadline firing on time — the clock armer serves them from the cooperative executor — and the second test still goes red, on both TERM-ignoring children still alive 8 s after their calls returned. Nothing but the watchdog can deliver those SIGKILLs.

## Flag & rollout

The flag is **`transcript_composer_enabled`**: a `config` column added with **no SQL `DEFAULT`**, resolved through `Config.transcriptComposerEnabledDefault = false`.

Omitting the `DEFAULT` is the load-bearing part. NULL means *nobody has chosen*; `0` and `1` mean somebody did. The shipped default therefore lives in exactly one place, and graduation is a one-line change to that constant — it reaches everyone who never touched the toggle while preserving every explicit opt-out. (`auto_hibernate_enabled` is the cautionary precedent: its `DEFAULT` backfilled every existing row, so flipping it later needed a forcing `UPDATE` that could not tell a deliberate opt-in from a backfilled value, and reset both.)

To enable it for the soak: **Settings → General → Claude → "Message composer in the transcript pane"**, or the `config.setTranscriptComposerEnabled` RPC.

Both branches are tested, and the column's three states are proved distinguishable by name: `nullBeforeAnyGesture`, `rowWrittenBeforeTheMigrationStillReadsNull`, and `explicitFalseSurvivesADefaultFlipWhileNullFollowsIt`.

Graduation waits on a soak in which no probe leaves a process or a `session-env/` directory behind and no probe races a trust seed.

## Assumptions

- **The `initialize` response shape is Claude Code's control protocol.** The decoder tolerates unknown keys at both levels, so an *added* field cannot break it. A *renamed* `commands` key would silently yield an empty list — which degrades to the scan's answer rather than to a crash, but degrades quietly.
- **The plugin layout on disk.** This is the assumption to check first if plugin items stop appearing. As measured: `installed_plugins.json` is `{"version": 2, "plugins": {"<plugin>@<marketplace>": [{scope, installPath, version, …}]}}`; plugin trees live at `plugins/cache/<marketplace>/<plugin>/<version>/`; enablement keys on the full `name@marketplace` string in `settings.json`'s `enabledPlugins`; the tree to read comes from the newest record's `installPath`; and the namespace prefix is the half of the key before the `@`.
- **Probe environment carries routing, not credentials.** The probe reuses `routingEnv` — `CLAUDE_CODE_USE_BEDROCK`, `AWS_REGION`, `AWS_PROFILE`, `ANTHROPIC_MODEL`, `ANTHROPIC_BASE_URL`, plus `CLAUDE_CONFIG_DIR` and the merged free-form overrides in the spawn's own precedence. Secrets are deliberately *not* reproduced: injecting an API key into a session running on an OAuth profile changes what that binary reports, losing three subscription-gated commands. The probe reads the same credentials the session would, from the same config directory.

## Side effects, and what suppresses each

Each flag exists because something was measured happening without it.

- **`--settings {"disableAllHooks": true, "disableClaudeAiConnectors": true}`** — without `disableAllHooks`, the probe's startup ran the session's hooks and left a `session-env/<uuid>/` directory behind on every call. Without `disableClaudeAiConnectors`, startup made an authenticated connectors request to the Anthropic API host — a network call, on someone's account, for a command list.
- **`--strict-mcp-config --mcp-config <empty file>`** — without it, the probe started the session's real MCP servers: extra processes to reclaim and log files written into `Library`.

Two silent Keychain reads per probe remain, about 10 ms, with no prompt. They are how the binary loads its own credentials; suppressing them would change the answer.

## Reconciler

Per the named-reconciler doctrine: the probe creates two things, and the claim is scoped to the **direct child only**.

- **The process** is reclaimed by pid on every exit path — normal exit, decode failure, cancellation, and the watchdog's SIGTERM-then-SIGKILL on the five-second deadline.
- **The temp `--mcp-config` file** is removed by a `defer`.

There are no grandchildren to escape the claim, because hooks and MCP servers are exactly what the suppressors above disable. Nothing durable is created, so no sweep covers this.

One window is not covered, and is named rather than papered over: the watchdog is in-process, so a daemon that **crashes or is killed while a probe child is running** takes the watchdog with it. `OrphanProcessCollector` will not reclaim that child either, because it deliberately never signals a process whose cwd is under a still-live worktree — which is exactly where the probe runs. The exposure is one process bounded to about five seconds of the daemon's own lifetime, and it is not worth a new sweep; if a soak ever shows a surviving probe, the fix is to widen the collector, not to add a reconciler.

(The composer's one genuinely durable resource — attachment files — arrives in PR C with its own named reconciler.)

## Known rough edges, carried deliberately

Named here so a soak observation is recognized rather than rediscovered:

- `loadByID` touches `last_used_at` on every completions request, so a warm-on-focus caller will churn that column.
- The cache is unbounded. It is bounded in practice by fleet size (one entry per config directory), and it is not keyed on environment overrides — changing an override does not invalidate it.
- **The cache key omits profile identity.** A profile with no isolated config directory (a Bedrock profile, say) falls back to the shared ambient one, so two such profiles probed against the same worktree and the same binary collapse onto one key, and one can serve the other's list until an unrelated filesystem change moves the fingerprint. It fails open — a stale suggestion in a completion menu, never a crash — and the fix is to fold the profile id into the key.
- A file added inside an *existing* `commands/` subdirectory does not invalidate the fingerprint; a new subdirectory does.
- There is no in-flight de-duplication. Two concurrent requests for the same key are serialized by the lane rather than coalesced, so one probe is wasted.

## Evidence & verification

- **A test per branch of the flag**, plus the three-state proof named above: a pre-migration row reads NULL rather than `0`, and an explicit `false` survives a change to the default constant while NULL follows it.
- **The serializer lane is proved, not assumed.** `probeRunsInsideTheConfigDirLane` fails when `serializer.run` is removed from the probe path — the test discriminates rather than passing for a coincidental reason, and it uses a start signal rather than a sleep window.
- **The probe's argument vector is pinned exactly**, which is what makes the suppressor list above a checked property rather than a comment.
- **Two symlink bugs were caught by tests that were red before their fixes.** Profile config directories mirror the host store by symlink, and two Foundation behaviors bite there: directory enumeration of a symlinked root returns *empty* rather than an error, so the scan silently found nothing; and `attributesOfItem` is `lstat`, so the cache fingerprint stat'd the link and never invalidated. Both paths now resolve symlinks first.

**Live verification is deferred to the human.** This machine runs a live agent fleet and the standing rule forbids restarting the daemon, so the running app was not exercised and the side-effect check was not re-run here. The evidence for the suppressors is the 2026-09-05 spike against Claude Code 2.1.261, measured directly: about 0.5 s wall per probe, 209 commands returned, zero tokens billed, no credentials written, no `session-env/` directory created, no hook frames in the transcript, and no process left behind. The re-check on a restarted daemon — a probe against a live session, `session-env/` counted before and after, and `pgrep` for a surviving probe — belongs to whoever restarts next.

[composer-b](https://cheapsteak.github.io/tbd/open/?worktree=B6F8AB9D-8BD6-4902-98EF-16F8DCA04FE5)

## Summary by CodeRabbit

- **New Features**
  - Added an optional message composer to the transcript pane, configurable from General settings.
  - Added terminal completion support for Claude commands and agents, including freshness indicators and JSON output.
  - Added automatic completion discovery with fallback scanning when live information is unavailable.
- **Bug Fixes**
  - Improved reliability when updating Claude configuration and handling process timeouts.
  - Configuration changes now persist correctly and refresh capabilities immediately, with errors surfaced to users.


cheapsteak:
Addressed all three findings from the review at patch `8aefeff0`.

**HIGH — `ensure*Dir` rewrote `.claude.json` outside the lane** → commit `e821f133`

The finding is right, and the framing that fixes it is that the lane's contract is *every writer of `<configDir>/.claude.json`* — there are four, not two, and a lane half the writers ignore orders nothing. `ensureAPIKeyDir` and `ensureOAuthDir` now take the same directory-keyed lane around their `.claude.json` section; directory creation and the host mirror symlinks touch other paths and stay outside, keeping the critical section to the file it protects.

`ensureOAuthDir`'s exists-check went *inside* the lane with its write, deliberately. Leaving it outside would swap one race for another rather than shrink it: a probe that creates the file between the check and the write turns "it did not exist" into an overwrite of what the probe just wrote.

Reentrancy was checked before making the change, since the lane is not reentrant. The only two lane bodies — `ClaudeTrustSeeder.swift:185` and `CompletionInventoryService.swift:215` — call neither `resolveConfigDir` nor `ensure*Dir`, and the completions handler resolves the config dir *before* the service enters the lane. That is now stated in the serializer's doc comment, which also names all four writers.

Both entry points and `resolveConfigDir` became `async` as a result; all twelve call sites were already in async contexts.

Tests, as asked:
- `ensureAPIKeyDirWaitsForTheLaneOnItsConfigDir` — an ensure issued while the lane is held writes nothing until the lane is released. Verified discriminating: routing the body around the serializer turns it red (`Expectation failed: !(wroteWhileHeld)`).
- `anInterleavedEnsureLeavesTheTrustEntryIntact` — a seeded trust entry and the api-key approval both survive interleaved ensures.

141/141 across the serializer, config-dir-manager, trust-seeder, ModelProfileRPC and ClaudeStateDetector suites; `swiftlint --strict` clean.

**MEDIUM — the spec was not reachable** → base retargeted

This PR is now stacked on #821: rebased onto `tbd/20260905-abundant-earwig` and the base changed to it, so `docs/specs/2026-09-05-transcript-composer-design.md` is reachable from this HEAD and reviewable in the checkout. GitHub retargets the base to `main` when #821 merges.

**MEDIUM — "no caller had a child that stayed silent past its deadline"** → PR body corrected

That claim was wrong, and the runner's own comments say so: `GitManager` documents a 95-minute `git fetch` as exactly such a child. The body now states the accurate version — the pipe-backed path of `BoundedProcessRunner` could park its watchdog behind a silent child; the probe's test is the first in tree to *hold* a child silent past its deadline and so the first to observe it; and because the fix lands in the shared runner it covers every existing caller, that `git fetch` included.

cheapsteak:
Review gate — both `BoundedProcessRunner.swift` findings addressed in **b6ec7c79**.

**High — the deadline action could park the shared watchdog** (`~:511`). Signalling before snapshotting narrowed the window but left the wait on the one thread that fires every bounded deadline in the daemon, with the SIGKILL escalation queued behind it. The action now does only bookkeeping — SIGTERM, queue the escalation — and hands the snapshot to the task already awaiting the call (`DeferredWork`), which runs it on the way out. A `.timedOut` outcome carries no output, so nothing about the result waits on the snapshot, and a wedged drain can delay only the call that owns it. The deposit is ordered before the resume, or the resumed task can find an empty slot and leave the handlers attached. `.pseudoTerminal` is untouched.

**Minor — the SCHEDULING doc comment described the pre-fix ordering** (`~:315`). Rewritten to state the new order and, as timeless design prose, why the snapshot is the one step that cannot run there.

**Discrimination, and an honest caveat.** The wedge cannot be provoked from a child process: Darwin's `readabilityHandler` never fires without data, so a drain worker never parks inside `availableData` — measured across silent, chatty, grandchild-holding and pool-saturated children. It is a latent lock-ordering hazard, so the two new tests pin the *guarantees* and are proved by mutation instead:

- Blocking the deadline action for 30 s where the snapshot used to run turns both tests red on their wall-clock bounds (30.6 s / 30.9 s against 1.2 s / 1.4 s green).
- Blocking it **only** on the watchdog thread leaves every deadline firing on time — the clock armer serves them from the cooperative executor — and `aSecondDeadlineIsServedWhileAnothersTimeoutPathResolves` still goes red, on both TERM-ignoring children alive 8 s after their calls returned. Nothing but the watchdog can deliver those SIGKILLs.

`scripts/test.sh --filter "BoundedProcessRunner|ClaudeCompletionProbeTests"`: 27 tests, 2 suites, 0 issues. Full suite dispatched on CI as run `34026955993`.

cheapsteak:
### CI failure + review MEDIUM addressed in `96c043cd`

**CI (run 34026955993, fast parallel pass 1) — `BoundedProcessRunnerTests`.** Both failures were wall-clock ceilings measuring the runner, not the code: `aTermIgnoringChildIsKilledByTheEscalationAtItsDeadline` at 31.3 s and `aSecondDeadlineIsServedWhileAnothersTimeoutPathResolves` at 40.6 s, against 20 s bounds, on green code. Fast pass 1 runs ~1800 tests in parallel on a loaded macOS runner whose per-test latency has a p50 near 65 s.

- Test 1 keeps its shape; both bounds become `TestDeadlines.saturatedPass`. Its discriminator is the child, not the clock, so the TERM-ignoring sleeper now outlives that bound several times over — with the old 30 s sleeper, "natural exit" and "the escalation ran" produced the same observation inside a 90 s wait.
- Test 2 is now deterministic and renamed `aSecondDeadlineIsServedWhileAnotherCallsSnapshotIsWedged`. It compares no elapsed time. `runBoundedProcess` gains a defaulted, test-only `beforeDeadlineSnapshot` seam that runs immediately before the deadline path's snapshot *wherever that snapshot runs*, so holding it models the hazard exactly. The first call is given an `EventDrivenTestClock` nobody advances, so armer 2 cannot fire and the action is reached from the watchdog thread — otherwise which armer wins is a coin flip and the hold lands somewhere harmless. With the hold in place, the second call must still get its deadline **and** its SIGKILL, which only the watchdog can deliver.

Mutation for test 2 (`beforeDeadlineSnapshot?(); _ = snapshot()` inline in the deadline action, i.e. the pre-fix structure): **RED** at 90.7 s — `timed out waiting for the second call's TERM-ignoring child to be reaped while a snapshot is wedged — observed pid 66070 still alive`. Restored: green in 1.3 s. The mutant was not committed.

**Review MEDIUM — `ClaudeConfigDirSerializerTests.anInterleavedEnsureLeavesTheTrustEntryIntact`.** Correct finding: the trust entry was seeded synchronously before the concurrent ensures started, so every ensure read a file that already carried it and no interleaving could lose it.

The ensure's merge is straight-line synchronous code inside the lane body, so the read-then-write window cannot be straddled from outside. `ensureAPIKeyDir` therefore takes `afterReadForTesting`, a defaulted test-only hook awaited between its read and its write, and the test now uses the same handshake as `ensureAPIKeyDirWaitsForTheLaneOnItsConfigDir` next door: a holder takes the lane, waits until the ensure is issued, gives a lane-ignoring ensure 50 ms to reach its read and park in the hook, writes the trust entry, then releases. With the lane the ensure has not read yet and reads the trust-bearing file; without it, it read a file that had none and rewrites it away.

Mutation (bypass `ClaudeConfigDirSerializer.shared.run` in `ensureAPIKeyDir`): **RED** on the missing `hasTrustDialogAccepted`. Restored: green.

**Verification.** `scripts/test.sh --filter "BoundedProcessRunner|ClaudeCompletionProbeTests|ClaudeConfigDirSerializer"` — 33 tests in 3 suites, all passing. `swiftlint --strict` clean on the touched files. Full suite dispatched as run 34028751254.

cheapsteak:
Restacked onto the rebased `tbd/composer-a-not-running-guard`, which now sits on `tbd/20260905-abundant-earwig` brought up to today's `main` (now carrying #819, the holder send path). The 21 commits replayed with no conflicts; new head `8837ef5c`.

Full test workflow: https://github.com/cheapsteak/tbd/actions/runs/34041915874

cheapsteak:
restacked onto main @ 406a9800 (#818, #827); conflicts resolved in Sources/TBDDaemon/Database/ConfigStore.swift, Sources/TBDShared/Models.swift, Tests/TBDDaemonTests/SQLMigrationLoaderTests.swift, Sources/TBDApp/AppState+ModelProfiles.swift, Sources/TBDApp/DaemonClient.swift, Sources/TBDApp/Settings/SettingsView.swift, Sources/TBDDaemon/Server/RPCRouter.swift, Sources/TBDShared/RPCProtocol.swift; run https://github.com/cheapsteak/tbd/actions/runs/34054755690

cheapsteak:
Restacked onto A @ `5ac61831` (review-gate fix round on #822) — no conflicts, 21 own commits, new head `99a4f2a1`; `scripts/test.sh --filter "TranscriptComposerSettingsTests|ClaudeCompletionProbeTests"` → 16 tests, 0 failures. CI: run 34057456465.

PR #824, opened 2026-09-06
Feat: terminal.send delivers a composed message, and composer attachments get their reconcilers

cheapsteak:
## Summary

TBD's transcript pane is getting a message composer: a place to type to the agent
running in a session — with images dropped in among the words — without leaving the
transcript for the terminal. This PR is the daemon half of that composer's send path,
plus the reclaim story for the files it will stage. Nothing in the app calls it yet;
the composer itself is the next PR in the stack.

What the daemon can do after this PR that it could not before:

- **Deliver a composed message, not a string.** `terminal.send` accepts an ordered
  list of parts — runs of text with image paths between them — and delivers them so
  the session shows *one* user turn reading `look at [Image #1] and tell me what it
  is`, image attached inline, rather than three separate messages.
- **Send in the person's own voice.** A send that TBD's own app makes on a human's
  behalf can ask for the `<tbd-dispatch/>` framing line to be left off, so what lands
  in the session is the message the person wrote and nothing else. That request is
  honoured only for the app's own connection, and cannot be obtained by asking.
- **Refuse rather than type into a dialog.** A caller can opt in to a gate that turns
  a send into a refusal — carrying the dialog's own message — when the session is
  sitting on a permission prompt, instead of having the keystrokes eaten by it.
- **Reclaim what the composer stages.** Images a composer prepares live under
  `~/tbd/attachments/<worktreeID>/`, and this PR gives that new durable resource its
  reconcilers before the code that creates it ships.

## Why this shape

Spec: [`docs/specs/2026-09-05-transcript-composer-design.md`](docs/specs/2026-09-05-transcript-composer-design.md)
— principally "Send pipeline" (Payload, Envelope, Refusals and the gate), "Transport
behavior and the holder dependency", "Attachments → Reclaim", and "Rejected
alternatives".

**Parts exist because of how the agent's paste handler works.** Claude Code turns a
paste into an attached image only when the *whole* paste is one quoted path with an
image extension (measured on 2.1.261). A single paste of `look at '/tmp/a.png' and
tell me` therefore arrives as literal text with a path in it. So the message is split
at its image tokens and each piece is pasted separately into the same composer line —
which reproduces byte for byte what a person dragging a file into the terminal
produces. Text pieces submit as one turn; image pieces attach.

**Envelope suppression is authenticated on the connection, never on the request.**
The kernel names the peer of an `AF_UNIX` socket, so a pid read with `getsockopt` at
accept time is a fact about who is on the other end and not a claim the caller can
make. A request that asks for suppression additionally re-verifies, through
`ProcessIdentityCheck`, the identity the FD-vending sidecar recorded for its current
client; only `.same` authenticates, and every other answer attaches the envelope. The
spec names the rejected alternatives: trusting the actor the request declares, letting
any caller suppress, matching the peer's executable *name* (spoofable by copying a
binary), and a startup nonce (a secret in every log and crash report). A dedicated
second socket for app traffic was rejected on cost alone, and stays the clean answer
if this ever needs to be more than a dev-tool boundary.

`ActuationActor` — the `{"actor":{"kind":"app"}}` field the RPC already carries —
gains no authority here. It is an ambient self-declaration that any local process can
stamp on a socket the CLI, the app, and every agent session share; it stays what it
was, a provenance label for the record.

**The awaiting-input gate is opt-in, for the mirror reason.** Agents answer permission
dialogs with this exact verb, on purpose. A daemon-wide gate would refuse precisely
those sends. So the gate is a field the caller sets, and the composer sets it because
a human typing prose into a transcript never means "press 1".

**The holder arm refuses what its single write cannot frame.** A holder-backed session
receives body and carriage return in one unwrapped write, and past a threshold in the
agent's own tokenizer the carriage return is swallowed rather than submitting — a
single unwrapped write of 63 bytes submits, 64 does not (same measurement,
2.1.261). Multi-line and multi-part payloads are therefore refused on that transport,
with a message that says why and names the fix. That fix is bracketed-paste wrapping,
which belongs to the holder work already in flight (#816); when it lands, the refusal
lifts. Note what the refusal is *not* keyed on: length. Length is #816's, and this PR
adds no byte-count leg.

## What this PR does

**`terminal.send` gains three optional params** (`Sources/TBDShared/RPCProtocol.swift`),
all of which decode absent as today's behaviour:

- `parts` — an ordered array of `{kind: "text" | "imagePath", value: …}`. Shape is
  validated before any session row is looked up, so a malformed payload fails the same
  way whatever the target.
- `envelope` — `"attached"` (default) or `"suppressed"`, a *request* that the
  connection must earn.
- `gateOnAwaitingInput` — the opt-in refusal above.

**Delivery.** On the tmux arm each part becomes its own bracketed paste in order, then
one Enter if submitting. When the envelope is attached and no part carries non-empty
text — an image-only message — the envelope goes out as its own leading paste so the
image paste stays bare and alone and still attaches, rather than being merged into a
line the agent would read as text.

**Authentication.** A new `RPCConnectionContext` carries the peer pid read once per
accepted connection (`Sources/TBDDaemon/Server/SocketServer.swift`), and the send
handler re-verifies the sidecar's recorded client identity before honouring a
suppression request.

**The gate.** `AwaitingInputSendGate` reuses the existing `AwaitingInputSupersession`
record, so it refuses on the same standing-dialog signal the rest of the daemon
already trusts, and carries that dialog's own message into the refusal. It sits in one
block that both transports fall into — not two call sites — because the bug that
motivated the restructure was exactly a rail added to one arm and forgotten in the
other.

**`terminal.wake` reports its incarnation.** `TerminalWakeResult` gains an optional
`sessionIncarnationID`, so a caller that wakes a hibernated session and then sends to
it can scope its wait to the session the wake actually minted. The no-op path (nothing
was hibernated) names no incarnation, which is what makes the field meaningful.

**Attachments paths and reclaim.** `TBDConstants.attachmentsDir` / `attachmentPath`
(TBD_HOME-honouring, like every other TBD-owned path), an unlink on the archive path,
and a new `reclaimAttachments` leg in `OrphanGC` — see Reconciler below.

**Builds on the two PRs below it.** PR A's two not-running rails, `paneProcessInspector`,
`HibernateReason.exited` and the app's wake prompt are already in `performTerminalSend`
and are neither re-added nor moved here — a reviewer reading that function top to
bottom is reading three PRs' rails at once. PR B's `transcript_composer_enabled` is the
column this PR's GC leg reads; it is not added here. One inherited behaviour worth
restating because the send path depends on it: a session row with a NULL `kind` reads
as Claude only when it carries a Claude session id, and as a shell otherwise — the same
rule the `v22` backfill used.

## Reconciler

**This PR introduces a new kind of durable resource — files under
`~/tbd/attachments/<worktreeID>/` — and names two reconcilers and three reclaim rules
for it.**

- **`OrphanGC.removedWorktreeCleanup`** (`Sources/TBDDaemon/GC/OrphanGC.swift`) unlinks
  a worktree's attachments directory on the existing worktree-removed callback that the
  archive path and `scratch.delete` already fire — the **archive-time unlink**. Best
  effort, under the `gcEnabled` master switch, and derived from the injected attachments
  base rather than a re-resolved global constant. `forgetWorktree` does not fire that
  callback; the hourly leg below covers what it leaves.
- **`OrphanGC.reclaimAttachments`** runs on the existing hourly cadence and is the
  standing guarantee. It carries the other two rules:
  - **An orphan directory past the floor** — no worktree row of that id remains, and
    nothing inside is younger than fourteen days — goes whole. Keep-biased and
    all-or-nothing: it keeps an entry that is not a directory, whose name is not a
    UUID, whose contents it cannot read, or that holds any file the floor has not
    reached; an empty one is measured on its own mtime against the same floor.
  - **A file past the floor inside a live worktree's directory** is reclaimed on its
    own, while the directory stays — never removed while the row lives, even once it
    holds nothing. This is safe because a staged image is read at paste time, or at
    resume time on the wake path: the file is needed for minutes, never days, so one
    that has sat untouched for two weeks belongs to a message nobody is going to send.
    Only regular files age; an unreadable file age keeps, as does an unreadable
    directory.

  It skips the whole leg on a failed database read, because "no worktrees" and "could
  not ask" must not look alike. Fourteen days is a soak knob, not a load-bearing number.

No `ReapRecord` is written for an attachments reap — matching the holder-rendezvous and
rowless-holder legs — because there is nothing a `tbd gc restore` could put back: the
image was staged for a message in a worktree that no longer exists. Consequently these
reaps do not appear in the History UI, and `.reapRecordsChanged` fires only via the
shared `reaped > 0` counter.

## Coordination

`performHolderSend`, `HolderInjectionCourier` and everything under
`Sources/TBDDaemon/Holder/` are untouched. The holder refusal is placed in
`performTerminalSend`, textually ahead of the holder branch, so that arm sees only
payloads it can already deliver. PR #816 owns the holder transport and lifts both
refusals this PR adds — the multi-line/multi-part one and the unauthenticated
image-only one.

## Flag & rollout

`transcript_composer_enabled` — a `config` column added in the PR below this one,
**default OFF**, no SQL default, so "nobody has chosen" stays a third state distinct
from an explicit opt-out. Enable it for a soak in Settings → General → Experimental.

What the flag gates here is exactly one thing: the `reclaimAttachments` GC leg, which
also requires `gcEnabled`. Both branches are tested, and a `--dry-run` sweep bypasses
the flag deliberately, so an operator can see what enabling it *would* reclaim before
enabling it.

Two things are deliberately **not** flag-gated, and the reasoning is the same for both:
a flag that gates *creation* must never gate *reclaim*, or turning the feature off
strands its files forever. So the archive-time unlink and the `scratch.delete` unlink
run under `gcEnabled` alone.

The send-path fields are ungated because they are inert: no shipped caller sets
`parts`, `envelope` or `gateOnAwaitingInput`, and each decodes absent as exactly
today's behaviour. Graduation is the composer PR's to make — the flag flips when the
composer itself is ready, and the change is the one `?? Config.transcriptComposerEnabledDefault`
constant.

## Assumptions

- **The agent's paste handler keys image detection on a whole-paste quoted path with
  an image extension.** If that changes, images arrive as literal paths and the agent
  reads them with its file-reading tool instead: degraded, not broken.
- **The holder byte threshold is a property of the agent's tokenizer**, not a number
  TBD controls or can negotiate. It was measured, not documented, and it can move.
- **The FD-vending sidecar has exactly one client, and that client is the app**, which
  connects eagerly and unconditionally as soon as the RPC socket answers — the same
  assumption the app-liveness arbiter already rests on. `FDVendingServer.adoptConnection`
  records whatever process connects, last writer wins, and a sidecar client can already
  inject raw bytes; suppression inherits that existing trust boundary rather than
  creating a new one. If the sidecar is not connected, suppression fails closed and a
  human's message carries a dispatch line — nothing is lost.
- **Wire compatibility.** An old caller's `{terminalID, text, submit}` decodes with the
  three new fields nil; an unrecognised `envelope` string decodes as `.attached`, the
  safe direction; `TerminalWakeResult.sessionIncarnationID` is additive.

## Evidence & verification

`scripts/swift-safe build --build-tests` clean; `swiftlint --strict` reports 0
violations across 992 files. The branch's own suites —
`TerminalSendPartsDeliveryTests`, `EnvelopeSuppressionAuthTests`,
`TerminalSendPartsShapeTests`, `TerminalSendDispatchTests`,
`TerminalSendNotRunningTests`, `AwaitingInputSendGateTests`,
`HolderCompositeSendRefusalTests`, `OrphanGCAttachmentsTests`, `OrphanGCTests`,
`AttachmentArchiveCleanupTests`, `TerminalWakeIncarnationTests`,
`AttachmentPathTests` — **145 tests in 15 suites, 0 failures** on the rebased head.

Full-suite CI was dispatched for this branch, since `test.yml` fires automatically only
for PRs based on `main`:
https://github.com/cheapsteak/tbd/actions/runs/34028298661 — **verdict pending at the
time of writing.**

**The discriminating tests**, i.e. the ones that fail if a rule is applied bluntly
rather than precisely:

- `anOptedOutSendIntoTheSamePromptIsNotGated` and `anOptedOutSendToAHolderRowIsNotGated`
  — fail if the awaiting-input gate is made unconditional on either transport.
- `anOptedInSendToAHolderRowIsRefused` — fails if the gate is skipped for one transport,
  the defect that motivated hoisting it above the transport fork. It asserts nothing
  reached the pty, through the harness's real injection courier.
- `aCLIConnectionAskingForSuppressionStillGetsTheEnvelope` and
  `theDeclaredActorChangesNothingInEitherDirection` — fail if `ActuationActor` is ever
  read for authority, in either direction.
- `aReVerifyThatIsNotSameKeepsTheEnvelope` — fails if only the peer pid is checked and
  the sidecar's recorded identity is not re-verified.
- `anUnauthenticatedImageOnlySendPastesTheEnvelopeFirst` /
  `anAuthenticatedImageOnlySendPastesOnlyThePath` — one asserts two pastes, the other
  one; a change that merges the envelope into the image paste fails the first, and a
  change that always emits it fails the second.
- `aTmuxRowStillAcceptsMultiLineAndParts` and PR A's
  `aLiveRowWithClaudeForegroundIsNotRefused` — the positive controls that fail if the
  rails start refusing everything.
- `theFlagOffLeavesEverythingAlone` and `aDryRunPlansWithTheFlagOffAndUnlinksNothing` —
  the flag's two branches, plus the dry-run bypass.
- `anUnreadableOrphanDirectoryIsKept`, `anEmptyOrphanYoungerThanTheFloorIsKept` vs
  `anEmptyOrphanOlderThanTheFloorIsReaped`, `aNonUUIDDirectoryIsKept`,
  `anArchivedWorktreesAttachmentsAreKept` — the keep-biased sweep's rules, each with a
  counterpart that reaps, so "keeps everything" cannot pass.
- `aStaleFileInALiveWorktreeIsReapedWhileTheDirectoryStays` — one 30-day file and one
  1-day file in the same live worktree's directory: fails if the per-file rule reaps
  the directory, spares the stale file, or takes the young one with it.
  `aDryRunPlansALiveWorktreesStaleFileWithoutUnlinkingIt` and
  `aNonRegularEntryInALiveWorktreeIsIgnored` fence the other two branches.

Every one of the above was observed RED before its fix and GREEN after, except the two
parts-payload additions to the auth suite, which are additive coverage of a rule that
already held and passed from the start.

**Live verification is deferred, and here is exactly what that means.** This machine is
running a live fleet of agent sessions, and the standing rule forbids restarting the
daemon or app, sending raw frames to the RPC socket, or running a real GC sweep against
the real home. So three checks the plan called for were **not** run:

- *Restart and confirm the process pair.* Nothing here changes the launch path.
- *Drive a parts send, an actor-spoofing send, and a gated send through the socket by
  hand.* Standing in for them: `TerminalSendPartsDeliveryTests` (order and paste
  shape end to end through the handler), `EnvelopeSuppressionAuthTests` (both the
  authenticated and the stranger connection, on `text` and on `parts`), and
  `AwaitingInputSendGateTests` (refusal message, opt-out, both transports).
- *A real `tbd gc sweep-now` and a staged, backdated orphan directory.*  Standing in
  for it: `OrphanGCAttachmentsTests`, which stages exactly that tree under a fenced
  `TBD_HOME` and an injected attachments base, backdates it through the injected `now`
  seam, and asserts both the dry-run plan and the real reap.

The one thing no test replaces is watching a real session render `look at [Image #1]`
as a single attributed turn. That needs the composer to exist to click, so it is the
first thing to check when the composer PR lands.

## Follow-ups

- The sidecar's trust anchor is last-writer-wins. First-client-wins, or a
  daemon-minted token handed to the app at launch, belongs to the sidecar rather than
  to this send path.
- `tbd terminal wake --json` does not yet print `sessionIncarnationID`.
- A `logger.debug` when the RPC channel is not a `SocketOptionProvider`, so
  suppression failing closed is visible rather than silent.
- The TOCTOU window between the not-running rails and the paste (noted in the PR
  below) applies to a parts send too — it is wider, since a parts send makes several
  pastes.
- #816 lifts both holder refusals.
- The `.reapRecordsChanged`-with-no-record edge is house style across several GC legs,
  not something this PR changes.

### Live run, 2026-09-06 — daemon and app built from `79d9e85d`, against a zero-token stub session

Run against the live fleet with `scripts/restart.sh` from `composer-d`; daemon
`daemon status` reported `Build: 79d9e85d (tbd/composer-d-app-composer)`. Every model
request went to `scripts/claude-stub.py` on loopback (both stub servers printed
`no model request left this machine`), so the session cost nothing. Full record:
`.superpowers/sdd/2026-09-05-composer-d-app-composer/live-verification-report.md`.

**The parts payload arrives as one turn with the image inline.** `terminal.send` with
`actor:{"kind":"app"}`, `envelope:"suppressed"`, `gateOnAwaitingInput:true` and three
parts, read back out of the session's transcript JSONL — not the screen:

```
[user] TEXT:'<tbd-dispatch id="965rlv98pe4y" from="app"/>\nlook at [Image #4] and tell me what it is' | IMAGE:image/png 240b
[user] TEXT:'[Image: source: /tmp/a.png]'
```

**The declared actor changes nothing.** The same payload re-sent as
`actor:{"kind":"session",…}` produced the identical shape — one turn, image inline,
envelope attached. Only the envelope's `from=` label follows the declaration.

**Suppression is refused on the connection, and the reason proves the whole chain.**
Each suppression-asking `nc` send logged exactly one line:

```
[com.tbd.daemon:terminalHandlers] send: envelope suppression refused (peer-pid-is-not-the-apps) — attaching the dispatch line
```

That reason is only reachable *after* `connection.peerPID` resolved and
`recordedAppIdentity()` returned a value, so the live run also demonstrates the
`getsockopt` at accept and a connected FD-vending sidecar with the app recorded as its
client — a sidecar-less daemon would have said `no-recorded-sidecar-client`, and an
unestablished peer `peer-pid-unestablished`.

**The awaiting-input gate refuses, and the opt-out passes.** TBD spawns Claude with
`--dangerously-skip-permissions`, so the state was produced through the hook's own RPC,
`terminal.notificationEvent` with `notification_type: "permission_prompt"` (the row took
it as `classification: "prompt_on_screen"`). Gated:

```
{"success":false,"error":"terminal.send was refused: the session has a prompt on screen — \"Claude needs your permission to run: echo hello-from-stub\". Nothing was typed; a pasted message plus Enter would commit whichever option is highlighted. Answer it in the terminal."}
```

Nothing was typed. The same payload with the field omitted, against the same standing
prompt, went through and the text appeared in the composer.

**The wake names its spawn, once.**

```
wake #1: {"woken":true,"sessionIncarnationID":"1FC3EB77-1DAD-4C1E-BF8A-8759A1E2BAE9"}
wake #2: {"woken":false}
```

and the woken pane's own `TBD_TERMINAL_INCARNATION_ID` read
`1FC3EB77-1DAD-4C1E-BF8A-8759A1E2BAE9`, matching the row.

**The holder refusals carry their reason.** Against a holder-backed terminal, a
multi-part payload and a payload whose text contains a newline are each refused with
"…delivers a message in one unwrapped write… Nothing was typed. Send a single-line
message, or move the session to tmux."

**The GC leg, all three branches, on staged fixtures under `~/tbd/attachments/`:**

```
flag ON,  --dry-run : KEEP not-a-worktree-id / KEEP live-worktree / REAP attachments
flag OFF, --dry-run : identical plan (the dry run bypasses the flag)
flag OFF, real sweep: all three directories still present
flag ON,  real sweep: only the backdated orphan gone
```

`tbd terminal completions --json` returned `{agents: 18, commands: 213, freshness:
"stale", source: "probe"}`.

**One defect found, and it is this PR's.** A parts message's `[Image #N]` token can land
at the **end** of the message instead of at its place among the words, and that send can
fail to submit. Observed twice, both times on the first image a Claude Code process had
ever been given:

```
expected: look at [Image #1] and tell me what it is
observed: look at  and tell me what it is[Image #1]      (and submit:true did not submit)
```

Repeats of the same payload in the same session landed correctly (runs 2–4), and a third
session's first parts send was also correct, so it is a race rather than a constant. The
shape points at the `.parts` arm pasting each part back-to-back with no pause while
Claude Code's paste handler inserts the token asynchronously: when the following text
paste wins, the token is appended at the caret's new position, and an Enter arriving
mid-attach is dropped. An image-only paste measured ~1 s to produce its token with
nothing competing. The splitting is right; the ordering needs a wait on the image part's
token before the next paste, and certainly before Enter.

[composer-c](https://cheapsteak.github.io/tbd/open/?worktree=98C2D440-D4F2-4EB3-966F-D67650C149DF)

## Summary by CodeRabbit

- **New Features**
  - Terminal messages support ordered text and image parts, image delivery controls, and submission options.
  - Sends can be gated while an active terminal prompt requires a response.
  - Dispatch-envelope suppression is authenticated for supported sends.
  - Successful terminal wake responses can identify the newly started session.

- **Bug Fixes**
  - Removed worktrees now clean up associated attachments and scratch data.
  - Attachment cleanup preserves active, recent, unknown, or unreadable data while reclaiming eligible stale files.
  - Unsupported multipart sends are prevented through holder terminals.
  - Wake handling now returns a safe error instead of terminating on unexpected results.

### DIFF
--- a/Sources/TBDDaemon/Daemon.swift
+++ b/Sources/TBDDaemon/Daemon.swift
@@ -966,8 +966,9 @@ public final class Daemon: Sendable {
         if mockMode == nil {
             let gc = OrphanGC(db: database, git: git, broadcast: { [subs] delta in subs.broadcast(delta: delta) })
             self.orphanGC = gc
-            lifecycle.onWorktreeRemoved = { [gc] path, repoPath in
-                await gc.scratchpadCleanup(forRemovedWorktreePath: path, repoPath: repoPath)
+            lifecycle.onWorktreeRemoved = { [gc] worktreeID, path, repoPath in
+                await gc.removedWorktreeCleanup(
+                    worktreeID: worktreeID, worktreePath: path, repoPath: repoPath)
             }
         }
 
--- a/Sources/TBDDaemon/Daemon.swift
+++ b/Sources/TBDDaemon/Daemon.swift
@@ -1057,6 +1057,12 @@ public final class Daemon: Sendable {
             pendingQuestions: pendingQuestions,
             remoteManager: remoteManager,
             claudeCloudLive: claudeCloudLive,
+            // Envelope suppression is authenticated against the sidecar's
+            // recorded client — the app — and against nothing the request says
+            // about itself. See `authenticatesEnvelopeSuppression`.
+            recordedAppIdentity: { [fdVendingServer] in
+                await fdVendingServer.currentClientIdentity()
+            },
             actuationLog: actuationLog
         )
         // Wire the shared input activity tracker to the coordinator
--- a/Sources/TBDDaemon/GC/AttachmentsCollector.swift
+++ b/Sources/TBDDaemon/GC/AttachmentsCollector.swift
@@ -32,10 +32,20 @@ struct AttachmentsCollector: Sendable {
         self.now = now
     }
 
-    /// Every entry directly under the attachments base.
+    /// Every directory directly under the attachments base.
+    ///
+    /// A non-directory entry is not a candidate at all, matching
+    /// `HolderRendezvousCollector.candidates()`: the wrong node type is filtered
+    /// out here rather than classified by `decide`, which never considered it.
+    /// Without this, a bare-UUID *file* at the top level would read as an
+    /// orphan directory whose `contentsOfDirectory` comes back empty — reaped
+    /// on the very first sweep past the floor.
     func candidates() -> [URL] {
-        (try? FileManager.default.contentsOfDirectory(
+        let entries = (try? FileManager.default.contentsOfDirectory(
             at: base, includingPropertiesForKeys: [.isDirectoryKey])) ?? []
+        return entries.filter {
+            (try? $0.resourceValues(forKeys: [.isDirectoryKey]))?.isDirectory == true
+        }
     }
 
     /// **Every branch fails toward keeping.**
--- a/Sources/TBDDaemon/GC/AttachmentsCollector.swift
+++ b/Sources/TBDDaemon/GC/AttachmentsCollector.swift
@@ -9,6 +9,16 @@ struct AttachmentsCollector: Sendable {
     enum Decision: Equatable, Sendable {
         case keep(reason: String)
         case reap
+        /// The directory names a worktree row that still exists. The directory
+        /// itself always stays — even once it holds nothing — and each staged
+        /// file the floor has passed goes on its own.
+        case reapFiles(stale: [URL], kept: [KeptFile])
+    }
+
+    /// One file inside a live worktree's directory that stays, and why.
+    struct KeptFile: Equatable, Sendable {
+        let file: URL
+        let reason: String
     }
 
     /// How long an orphan's files must have sat untouched before they go.
@@ -56,10 +66,10 @@ struct AttachmentsCollector: Sendable {
         guard let id = UUID(uuidString: directory.lastPathComponent) else {
             return .keep(reason: "not-a-worktree-id")
         }
-        guard !liveWorktreeIDs.contains(id) else {
-            return .keep(reason: "live-worktree")
-        }
         let floor = now().addingTimeInterval(-Double(floorDays) * 86_400)
+        if liveWorktreeIDs.contains(id) {
+            return decideLiveWorktree(directory, floor: floor)
+        }
         // **An unreadable directory is not an empty one.** Reading through `try?`
         // and falling back to `[]` made a directory the sweep cannot open — a
         // permissions problem, a volume that answered badly — look like a
@@ -92,10 +102,50 @@ struct AttachmentsCollector: Sendable {
         return .reap
     }
 
-    /// Unlink the directory and everything in it. Returns whether it went.
-    func reap(_ directory: URL) -> Bool {
+    /// A live worktree's directory outlives its files.
+    ///
+    /// The row is what makes the directory worth keeping, so the directory
+    /// stays for as long as the row does — but the images inside it do not age
+    /// with the worktree. A staged file is read at paste time, or at resume
+    /// time on the wake path, so it is needed for minutes; one the floor has
+    /// passed is spent, and the same fourteen days that govern an orphan govern
+    /// it. Every branch here still fails toward keeping.
+    private func decideLiveWorktree(_ directory: URL, floor: Date) -> Decision {
+        // An unreadable directory is not an empty one, exactly as in the orphan
+        // branch: it yields no files to classify, so nothing here may act.
+        guard let entries = try? FileManager.default.contentsOfDirectory(
+            at: directory, includingPropertiesForKeys: [.isRegularFileKey]) else {
+            return .keep(reason: "contents-unreadable")
+        }
+        var stale: [URL] = []
+        var kept: [KeptFile] = []
+        for entry in entries {
+            // Only a regular file is a staged image. A subdirectory, a symlink,
+            // or anything else a later feature puts here is left alone rather
+            // than classified by a rule that never considered it.
+            guard (try? entry.resourceValues(forKeys: [.isRegularFileKey]))?
+                .isRegularFile == true else { continue }
+            guard let attrs = try? FileManager.default.attributesOfItem(atPath: entry.path),
+                  let modified = attrs[.modificationDate] as? Date else {
+                // A file whose age cannot be read is a file whose age is
+                // unknown, and unknown is not old.
+                kept.append(KeptFile(file: entry, reason: "age-unreadable"))
+                continue
+            }
+            if modified > floor {
+                kept.append(KeptFile(file: entry, reason: "younger-than-floor"))
+            } else {
+                stale.append(entry)
+            }
+        }
+        return .reapFiles(stale: stale, kept: kept)
+    }
+
+    /// Unlink one node: a whole orphan directory with everything in it, or a
+    /// single staged file inside a live worktree's. Returns whether it went.
+    func reap(_ node: URL) -> Bool {
         do {
-            try FileManager.default.removeItem(at: directory)
+            try FileManager.default.removeItem(at: node)
             return true
         } catch {
             return false
--- a/Sources/TBDDaemon/GC/AttachmentsCollector.swift
+++ b/Sources/TBDDaemon/GC/AttachmentsCollector.swift
@@ -60,14 +60,31 @@ struct AttachmentsCollector: Sendable {
             return .keep(reason: "live-worktree")
         }
         let floor = now().addingTimeInterval(-Double(floorDays) * 86_400)
-        let files = (try? FileManager.default.contentsOfDirectory(
-            at: directory, includingPropertiesForKeys: [.contentModificationDateKey])) ?? []
-        // An empty orphan directory is reapable: there is nothing left to protect.
+        // **An unreadable directory is not an empty one.** Reading through `try?`
+        // and falling back to `[]` made a directory the sweep cannot open — a
+        // permissions problem, a volume that answered badly — look like a
+        // directory with nothing in it, and an empty orphan is reapable. The
+        // failure would have destroyed exactly the files it could not see.
+        guard let files = try? FileManager.default.contentsOfDirectory(
+            at: directory, includingPropertiesForKeys: [.contentModificationDateKey]) else {
+            return .keep(reason: "contents-unreadable")
+        }
         for file in files {
             guard let attrs = try? FileManager.default.attributesOfItem(atPath: file.path),
                   let modified = attrs[.modificationDate] as? Date else {
                 // A file whose age cannot be read is a file whose age is unknown,
                 // and unknown is not old.
+                return .keep(reason: "age-unreadable")
+            }
+            if modified > floor { return .keep(reason: "younger-than-floor") }
+        }
+        // An EMPTY orphan directory gets the floor too, measured on its own
+        // mtime: it has no file to age, and a directory minted moments ago is a
+        // composer somebody has open whose first image has not landed yet. Only
+        // past the floor is there nothing left to protect.
+        if files.isEmpty {
+            guard let attrs = try? FileManager.default.attributesOfItem(atPath: directory.path),
+                  let modified = attrs[.modificationDate] as? Date else {
                 return .keep(reason: "age-unreadable")
             }
             if modified > floor { return .keep(reason: "younger-than-floor") }
--- a/Sources/TBDDaemon/GC/AttachmentsCollector.swift
+++ b/Sources/TBDDaemon/GC/AttachmentsCollector.swift
@@ -1,0 +1,77 @@
+import Foundation
+import TBDShared
+
+/// Decides what to do with one directory under `~/tbd/attachments`.
+///
+/// Pure and injectable, so every case is stated in one line of a test without a
+/// filesystem walk or a database.
+struct AttachmentsCollector: Sendable {
+    enum Decision: Equatable, Sendable {
+        case keep(reason: String)
+        case reap
+    }
+
+    /// How long an orphan's files must have sat untouched before they go.
+    ///
+    /// A soak knob, not a load-bearing number. It exists because the failure it
+    /// guards against is asymmetric: a leaked image is a few hundred kilobytes a
+    /// later sweep still finds, while a wrong reap destroys something a person
+    /// staged for a message they have not sent yet — and a draft can sit unsent
+    /// across a weekend.
+    static let defaultFloorDays = 14
+
+    let base: URL
+    /// The date seam. Ages here are persisted file timestamps compared against a
+    /// time, which is data rather than behavior, so this is a `Date` provider
+    /// and not a `Clock` — and it is the sweep's single reading of now, handed
+    /// down so a test can state a fixture's age exactly.
+    let now: @Sendable () -> Date
+
+    init(base: URL, now: @escaping @Sendable () -> Date) {
+        self.base = base
+        self.now = now
+    }
+
+    /// Every entry directly under the attachments base.
+    func candidates() -> [URL] {
+        (try? FileManager.default.contentsOfDirectory(
+            at: base, includingPropertiesForKeys: [.isDirectoryKey])) ?? []
+    }
+
+    /// **Every branch fails toward keeping.**
+    func decide(_ directory: URL, liveWorktreeIDs: Set<UUID>, floorDays: Int) -> Decision {
+        // A whitelist: only a directory whose name IS a worktree UUID is a
+        // candidate at all. Anything else a human or a future feature put here is
+        // left alone rather than classified by a rule that never considered it.
+        guard let id = UUID(uuidString: directory.lastPathComponent) else {
+            return .keep(reason: "not-a-worktree-id")
+        }
+        guard !liveWorktreeIDs.contains(id) else {
+            return .keep(reason: "live-worktree")
+        }
+        let floor = now().addingTimeInterval(-Double(floorDays) * 86_400)
+        let files = (try? FileManager.default.contentsOfDirectory(
+            at: directory, includingPropertiesForKeys: [.contentModificationDateKey])) ?? []
+        // An empty orphan directory is reapable: there is nothing left to protect.
+        for file in files {
+            guard let attrs = try? FileManager.default.attributesOfItem(atPath: file.path),
+                  let modified = attrs[.modificationDate] as? Date else {
+                // A file whose age cannot be read is a file whose age is unknown,
+                // and unknown is not old.
+                return .keep(reason: "age-unreadable")
+            }
+            if modified > floor { return .keep(reason: "younger-than-floor") }
+        }
+        return .reap
+    }
+
+    /// Unlink the directory and everything in it. Returns whether it went.
+    func reap(_ directory: URL) -> Bool {
+        do {
+            try FileManager.default.removeItem(at: directory)
+            return true
+        } catch {
+            return false
+        }
+    }
+}
--- a/Sources/TBDDaemon/GC/OrphanGC.swift
+++ b/Sources/TBDDaemon/GC/OrphanGC.swift
@@ -961,10 +961,16 @@ public actor OrphanGC {
     /// `.active` here would let this leg reap out from under a worktree somebody
     /// can still revive.
     ///
+    /// A directory whose row still exists is never removed, but the files in it
+    /// age individually: a staged image is read at paste time, or at resume time
+    /// on the wake path, so one the floor has passed is spent and goes on its
+    /// own.
+    ///
     /// No `ReapRecord` is written. The record type is keyed by the worktree a
     /// reap removed and carries a restorability pointer; there is nothing a
     /// `tbd gc restore` could put back here, because the image was staged for a
-    /// message in a worktree that no longer exists.
+    /// message in a worktree that no longer exists — or, for a per-file reap,
+    /// for one nobody sent in two weeks.
     private func reclaimAttachments(
         config: Config, dryRun: Bool, planned: inout [String], reaped: inout Int
     ) async {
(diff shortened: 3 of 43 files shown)


cheapsteak:
Restacked onto PR B's latest head `96c043cd` so this run carries PR B's test-only fix for the `BoundedProcessRunnerTests` timing tests. Clean rebase, no conflicts. New head: `f8db3274` (21 commits above PR B).

Full CI run: https://github.com/cheapsteak/tbd/actions/runs/34037630303

cheapsteak:
**Live verification run — 2026-09-06, daemon and app built from `79d9e85d`.**

The fleet's daemon and app were replaced by this branch's (with the user's
authorization) and a Claude session was driven against `scripts/claude-stub.py` on
loopback. Zero tokens: both stub servers printed `no model request left this machine`.
The Evidence section now carries the excerpts; the full record is in
`.superpowers/sdd/2026-09-05-composer-d-app-composer/live-verification-report.md`.

**PASS, observed live and read from machine interfaces, never the screen:**

- Plan Task 9 Steps 2.4 and 2.5 — a three-part payload arrives as **one** user turn with
  the image attached inline, and the declared actor changes nothing in either direction.
- Step 2.6 — every suppression-asking `nc` send logged
  `envelope suppression refused (peer-pid-is-not-the-apps)`. Worth naming: that reason is
  only reachable *after* `connection.peerPID` resolved and `recordedAppIdentity()`
  returned, so the run also demonstrates the accept-time `getsockopt` and a connected
  FD-vending sidecar with the app recorded as its client.
- Step 2.7 — the gate refuses with the prompt's own message and types nothing; the same
  payload without `gateOnAwaitingInput` passes the same standing prompt. The prompt state
  was produced through the hook's own RPC (`terminal.notificationEvent`,
  `notification_type: "permission_prompt"`), because TBD spawns Claude with
  `--dangerously-skip-permissions` and no real dialog can be raised in a TBD session.
- Step 2.8 — `{"woken":true,"sessionIncarnationID":"1FC3EB77-…"}` then `{"woken":false}`,
  with the woken pane's own `TBD_TERMINAL_INCARNATION_ID` matching the row.
- The holder refusals, both shapes, each carrying its reason and typing nothing.
- Step 3 — the GC leg on all three branches: the dry run bypasses the flag, a real sweep
  with the flag off keeps everything, a real sweep with it on reaps only the backdated
  orphan and keeps both the non-UUID and the live-worktree directory.
- `tbd terminal completions --json` — 18 agents, 213 commands, `source: "probe"`.

**One defect, and it belongs to this PR.** A parts message's `[Image #N]` token can land
at the **end** of the message rather than among the words, and that send can fail to
submit:

```
expected: look at [Image #1] and tell me what it is
observed: look at  and tell me what it is[Image #1]      (submit:true did not submit)
```

Observed twice, both times on the first image a Claude Code process had ever been given;
correct on every repeat within a warmed session, and correct on a third session's first
send — so a race, not a constant. The `.parts` arm pastes each part back-to-back with no
pause while Claude Code's paste handler inserts the token asynchronously; when the
following text paste wins, the token lands at the caret's new position, and an Enter
arriving mid-attach is dropped. An image-only paste measured ~1 s to produce its token
with nothing competing. The splitting is right — the ordering needs to wait on the image
part's token before the next paste, and certainly before Enter.

cheapsteak:
Restacked onto the new PR B head (`8837ef5c`), which now carries `main` including #819's holder send path. C's 22 commits replayed on top; new head `364acfdf`.

Four conflicts, all in `Sources/TBDDaemon/Server/RPCRouter+TerminalHandlers.swift`, resolved keeping #819's holder composition **and** C's rails/gate/parts/settle:

- #819's mode reading (`holderModeReading` → `modeSource`/`modeAge`) moved into C's extracted `deliverHolderText`, so the single-part `.parts` arm composes against the same reading the `.text` arm does.
- The body composition keeps #819's `HolderSendComposition.compose(body:submit:bracketedPaste:)` and gains C's `envelopeEligible` gate, so a lone image part is still never prefixed.
- The two `performHolderSend` doc-comment conflicts merged into one list: #819's verify/keys bullets plus C's composite and image-only bullets, with #819's separate no-courier paragraph left standing.

The D1 paste-race fix (`abf5c37b`, a 2 s settle per image part, 0/5 → 5/5 in the harness) is on the branch as its tip commit.

Local verification on the restacked head: `scripts/swift-safe build --build-tests` clean; `scripts/test.sh --filter "TerminalSendPartsDeliveryTests|EnvelopeSuppressionAuth|HolderCompositeSendRefusalTests|TerminalSendNotRunningTests|OrphanGCAttachmentsTests|TerminalWakeIncarnationTests"` → 65 tests in 8 suites, 0 failures.

Full CI: https://github.com/cheapsteak/tbd/actions/runs/34042727928

cheapsteak:
restacked onto main @ 406a9800 (#818, #827); no textual conflicts, but a semantic one: #818's holder wake path returns `WakeResult.ok`, which this branch widened to carry `sessionIncarnationID`. Resolved in Sources/TBDDaemon/Lifecycle/HibernationCoordinator+Holder.swift (spawn arm reports the minted incarnation, adopt arm reports nil), Tests/TBDDaemonLiveTests/Holder/HolderHibernationLiveTests.swift and Tests/TBDDaemonTests/Holder/HolderTmuxAssumptionGateTests.swift (`.isOk`); run https://github.com/cheapsteak/tbd/actions/runs/34055228479

cheapsteak:
Restacked onto B @ `99a4f2a1` (which is itself restacked onto A's fix round, #822 @ `5ac61831`) — 22 own commits, new head `e7b293a9`. Two rebase adaptations to A's new `WakeResult.paneBusy` case, each folded into the commit that owns the code it touches: `fix(wake): exhaust WakeResult in the payload switch` gains `.paneBusy` in `wakeResultPayload`'s exhaustive list, and `feat(send): deliver a parts payload in order on the tmux arm` gains `paneForegroundPID` on `SendHarness.StubInspector` (A added it to `PaneProcessInspecting`). `scripts/test.sh --filter "TerminalSendPartsDeliveryTests|EnvelopeSuppressionAuth|TerminalWakeIncarnationTests"` → 29 tests, 0 failures. CI: run 34057860313.

cheapsteak:
Review findings, one closed and one held.

**Finding 2 — no end-to-end test that `channelActive` populates `RPCConnectionContext.peerPID` (MEDIUM): fixed in `aa83e090`.**

`Tests/TBDDaemonTests/SocketServerPeerPIDTests.swift` drives the whole chain from a real `connect(2)` in the test process: a `SocketServer` on a scratch `/tmp` socket over a `SendHarness` router, a plain `AF_UNIX` client, one `terminal.send` asking for `envelope: .suppressed` with `{"kind":"app"}`. The recorded sidecar identity is this process, so the suppression happens if and only if the pid the router received off the socket is `getpid()` — on the first and only line the connection ever carries, which is what makes it a statement about ordering rather than just about plumbing. The negative row moves the recorded pid by one and the envelope comes back, so the comparison is against a specific number the kernel supplied rather than against anything at all. The other direction the design rests on — a context with no peer pid keeps the envelope — is already pinned by `EnvelopeSuppressionAuthTests.anUnreadablePeerPIDKeepsTheEnvelope` and `noConnectionContextKeepsTheEnvelope`, so this suite references those rather than restating them.

No production change: the seam is the router the server is handed, and the observable is the send path's own outcome.

`scripts/test.sh --filter "SocketServerPeerPIDTests|EnvelopeSuppressionAuthTests|SocketServer"` → 23 tests in 5 suites, 0 failures (`SocketServerSocketOwnershipTests` green locally; it has flaked on CI's loaded runner against its 90 s gate). `swiftlint --strict` clean on the new file.

**Finding 1 — an abandoned attachment in a still-live worktree has no reconciler (MEDIUM): held.**

That is a spec decision rather than an oversight to patch here — what "abandoned" means while the worktree is alive, and whether a sweep may reclaim under a live composer at all, is the maintainer's call. Pending their answer.

cheapsteak:
**Finding 1 — an abandoned attachment in a still-live worktree has no reconciler (MEDIUM): closed in `8b03cb99` (code + tests) and `4b570b47` (spec).**

The maintainer's call: a file inside a live worktree's directory is reclaimed **per file** once the existing fourteen-day floor has passed it. The directory itself is still never removed while its row lives, even once it holds nothing — the row is what makes the directory worth keeping, but the images inside it do not age with the worktree.

Why fourteen days is not "under a live composer": a staged image is read at paste time, or at resume time on the wake path, so the file is needed for minutes, never days. One that has sat untouched for two weeks belongs to a message nobody is going to send.

`8b03cb99` — `AttachmentsCollector.decide` now routes a live-row directory to `decideLiveWorktree`, which classifies each entry instead of keeping the directory unconditionally. Every branch still fails toward keeping: an unreadable directory keeps (`contents-unreadable`), an unreadable file age keeps that file (`age-unreadable`), a non-regular entry is ignored rather than classified, and any `try?` failure keeps. The leg emits `KEEP live-worktree <dir>`, a `KEEP <reason> <file>` line per kept file, and `REAP attachments <file> live-worktree-file-past-floor` per reclaimed one; the dry run plans those reaps and unlinks nothing. The orphan-directory branch and both flag branches are unchanged.

`4b570b47` — the spec's "Attachments → Reclaim" section now states the per-file rule and its safety argument.

Discriminating tests: `aStaleFileInALiveWorktreeIsReapedWhileTheDirectoryStays` stages one 30-day and one 1-day file in the same live worktree's directory and fails if the rule reaps the directory, spares the stale file, or takes the young one with it; `aDryRunPlansALiveWorktreesStaleFileWithoutUnlinkingIt` and `aNonRegularEntryInALiveWorktreeIsIgnored` fence the other two branches. Both new reap tests were observed RED (6 issues) before the collector change and GREEN after.

`scripts/test.sh --filter "OrphanGCAttachmentsTests|OrphanGCTests|AttachmentArchiveCleanupTests"` → **41 tests in 4 suites, 0 failures**. `swiftlint --strict` clean on the three touched files.

PR #825, opened 2026-09-06
Feat: A message composer in the live transcript pane

cheapsteak:
## Summary

Somebody watching a live Claude session in the transcript pane can read what the
agent says but cannot answer it there. Answering means switching to the terminal
pane, finding the input box, and typing into it — and when the terminal is busy
repainting, the reply lags behind the conversation it belongs to.

This adds a composer under the live transcript: a multi-line text field that
sends into the running session, completes Claude Code's own slash commands,
skills and subagents as you type, and accepts pasted or dropped images placed
inline among the words. Escape returns focus to the transcript; ⌘/ brings it
back. If the session has parked or exited, the composer wakes it and hands the
message to the spawn it just started, instead of pasting into a shell prompt.

It is the last of a four-PR stack and the only one a user can see.

## Why this shape

Design spec: [`docs/specs/2026-09-05-transcript-composer-design.md`](docs/specs/2026-09-05-transcript-composer-design.md).
The decisions worth naming for a reviewer:

- **The command list comes from the session's own Claude Code binary**, through
  `terminal.completions` (landed in PR B). A command that Claude Code ships, or
  a skill somebody installs into their own project, appears in the menu with no
  TBD release in between. TBD ships no command list of its own.
- **The composer suppresses the dispatch envelope**, because the person is
  speaking in their own voice and not relaying an agent's instruction — and the
  daemon honours that suppression only on a connection it has authenticated as
  this app's: the kernel-reported peer pid of the socket, matched against the
  identity the FD-vending sidecar recorded and re-verified through the daemon's
  one start-time and command-line check. Never on the strength of a declared
  actor field in the request.
- **The message arrives through the existing send path.** Nothing new reaches
  the agent; the payload is the same shape a `tbd terminal send` produces.

Rejected alternatives, from the spec's "Rejected alternatives" section:

- **A SwiftUI-only composer.** No SwiftUI text API exposes marked-text state,
  and marked text is the single fact the Enter-versus-IME decision depends on.
  The multi-line case also takes Up and Down away from the move-command
  modifier. Hence an `NSTextView` behind a representable, with a pure key router
  in front of it.
- **Images as inline rich-text attachments.** Enabling rich text reopens the
  substitution and paste-formatting hazards the submitting editor closed, and a
  plain-text `[Image #N]` token with temporary attributes gives the same visual
  anchor for none of that cost.
- **A chip strip with no inline anchors.** There would be no way to refer to a
  particular image at a point in the sentence.
- **Holding messages until the session is idle.** Claude Code queues typed input
  itself and shows it pending, so immediate injection mirrors what the terminal
  already does.

## What this PR does

At survey altitude, under `Sources/TBDApp/Panes/Transcript/Composer/`:

- **`ComposerKeyRouter`** — a pure function from a key, the modifiers, and
  whether a menu is open, to an action. Enter, Tab, Escape, the arrows, ⌃N/⌃P,
  ⌘↩. Marked text passes everything through untouched.
- **`CompletionTrigger`** — decides from the text and the caret whether a
  completion query is live, and refuses to open where a trigger would be a trap
  (mid-word, inside a path, after a token that is not at a word boundary).
- **`CommandRanker`** — prefix, then subsequence, then a frecency tiebreak; and
  `FrecencyStore`, one global `@MainActor` store over `UserDefaults` with a
  seven-day half-life.
- **`CompletionController`** — the query/loading/rows/no-match state machine over
  the inventory the daemon returns, memoized per query.
- **`MessageComposerTextView`** — the `NSTextView` representable, driven by
  one-shot commands (`replaceRange`, `insertAtCaret`, `restore`, `clear`) so a
  SwiftUI update never writes into AppKit mid-pass.
- **`CompletionOverlayView`** — the menu, opening upward above the field.
- **`AttachmentStrip`** and **`ComposerImagePreparer`** — the thumbnail row, and
  the ImageIO path that downscales, bakes EXIF orientation into the pixels, and
  writes a PNG atomically.
- **`ComposerTokens`**, **`ComposerDraft`**, **`ComposerArgumentHint`** — the
  `[Image #N]` token scanner, the per-terminal draft (text, caret, attachments),
  and the hint shown under the field for a command that takes arguments.
- **`ComposerState`** — hidden / disabled / enabled, resolved from the terminal.
- **`ComposerSendCoordinator`** — the send itself, including the wake path: it
  wakes a parked or exited session, then holds the message until the incarnation
  *that wake minted* reports its own `SessionStart`.
- **`MessageComposerView`** — the assembly, and the mount under the transcript
  table in `TableTranscriptPaneView`, outside the pane's session-identity `.id`
  so a session rollover does not tear the composer down under a half-typed
  message.

Around it: `AppState+Composer` (drafts), `AppState+ComposerFocus` (the composer
and transcript focus registries, and ⌘/ ⌘⌥/), `AppState+SessionStartWaiters`
(the waiter registry the hold parks on, with a latch so a `SessionStart` that
lands before the waiter registers is not missed), two `DaemonClient` methods, and
a draft discard on tab close.

## Not purely app-side

Two lines live outside `Sources/TBDApp/`:

- One additive optional field, `TerminalSessionDelta.sessionIncarnationID`, in
  `Sources/TBDShared/StateDelta.swift`.
- The one argument that populates it, in the daemon's
  `handleTerminalSessionEvent` broadcast
  (`Sources/TBDDaemon/Server/RPCRouter+TerminalHandlers.swift`).

Both are wire-compatible in each direction: the field is optional with a `nil`
default, old JSON decodes, and a new daemon's JSON decodes into an old app. The
in-place profile-swap broadcast deliberately leaves it nil, because a profile
swap is not a `SessionStart` and a consumer scoping a wait to one spawn must
read nil as "not this one", never as "close enough".

The field exists because the `SessionStart` hook is the only fact that says
*this particular spawn* reported in. Polling the terminal row cannot distinguish
the spawn a wake minted from one that started for any other reason, which is
exactly the distinction the hold depends on.

Both files are also touched by PR C. The stack is linear, so there is no
conflict — but a reviewer reading the two diffs side by side should expect to
see the same files.

## Flag & rollout

`transcript_composer_enabled` — the `config` column landed in PR B, with no SQL
`DEFAULT`, so NULL still means "nobody chose". The shipped default lives in one
place, `Config.transcriptComposerEnabledDefault`, and is **false**.

Enable it for the soak in Settings → General, in the Claude section: "Message
composer in the transcript pane".

The gating is **transitive, by design**. Only the `composerMount` decision reads
the flag. `fetchCompletions` and the attachment writer carry no check of their
own, because neither is reachable except from a mounted composer.
`ComposerMountingTests` pins both branches — `theFlagOffMountsNoComposer` is the
negative, `aLiveClaudeTerminalOnALocalWorktreeMounts` the positive.

Graduation: after a soak in which no message reaches a session that is not
running and no draft is lost across a session rollover, flip
`Config.transcriptComposerEnabledDefault` to true — a one-line change that
reaches everybody who never touched the toggle and preserves every explicit
opt-out.

## Decisions

Rulings taken while this branch was built, recorded so a reviewer does not have
to re-derive them:

- **Enter and Tab.** Enter with a row highlighted accepts it, and also *sends*
  when the completed token is the whole message — so `/comp` plus Enter runs
  compact, exactly as it does in the terminal. Enter with nothing highlighted
  sends the text as typed. Tab is the only key that takes the accept target and
  it never sends. The IME guard is live: `hasMarkedText` flows out of the text
  view into the router, and marked text passes every key through.
- **Cmd+Return is routed by `ComposerKeyRouter` only.** The send button carries
  no SwiftUI `.keyboardShortcut`, because a key equivalent is live for the whole
  window: in a split, ⌘↩ typed with focus in the sibling terminal pane fired the
  composer's send. `commandReturnAlwaysSends` pins the surviving site.
- **Two deferred main-actor hops, on purpose.** The composer registers its view
  one turn after `makeNSView`, and the text view reports its state one turn after
  a one-shot command is applied. Both avoid `@Observable` writes during a SwiftUI
  update; both are asserted (`anInsertionReportsOnceOnALaterTurn` asserts nothing
  is reported synchronously and exactly one report lands after the turn).
- **Zero matching rows closes the menu** rather than showing an empty one. One
  character with nothing matched shows no menu at all; "No commands match"
  appears from two characters up.
- **The empty-query list leads with frecency and its tail is alphabetical.** The
  completions wire type carries no source field, so built-ins, user, project and
  plugin commands cannot be grouped yet — see the follow-ups.
- **`FrecencyStore` is `@MainActor`**, matching every other `UserDefaults` holder
  in the tree, with `@Sendable` closures at the seams.
- **`VisualEffectView` was promoted** from a private struct inside
  `JumpMenu/JumpMenuView.swift` to an internal helper in `Helpers/`. The jump
  menu's call site is unchanged; that file is touched by the move and nothing
  else.
- **Two `.swiftlint.yml` exclusions** from the transcript `ScrollView` ban, for
  `CompletionOverlayView` and `AttachmentStrip`, following that rule's own
  self-exemption convention. Both are pane-level chrome mounted beside the
  transcript table, never inside a transcript row card, so neither sits in the
  row-measurement path the rule exists to protect.
- **The hold's timeout says delivery "could not be confirmed"**, not that the
  message was not delivered. The prompt travelled on the respawn's own argv and
  may well have landed; silence means TBD heard nothing *from* the session. The
  message goes back in the box either way.
- **The hold is scoped to one spawn.** It releases only on the terminal's own
  `SessionStart` for the incarnation the wake minted. A no-op wake (the session
  was already running) never enters the hold; a wake that reports no incarnation
  does not hold either.
- **A dropped file becomes an attachment only if it is an image**, by declared
  `UTType` or, for a file whose type says nothing, an ImageIO decode of the first
  8 KB — type only, never a frame count a truncated buffer cannot answer. A text
  drop or paste falls through to AppKit unchanged.

## Dependencies

Merge bottom-up; GitHub retargets each base as the one below it merges.

- **[#821](https://github.com/cheapsteak/tbd/pull/821)** — the design spec.
- **[#822](https://github.com/cheapsteak/tbd/pull/822)** — the not-running
  refusal, the exit stamp (`HibernateReason.exited`, `Terminal.isExitStamped`),
  and the wake prompt the app now passes.
- **[#823](https://github.com/cheapsteak/tbd/pull/823)** — the
  `transcript_composer_enabled` flag and `terminal.completions`.
- **[#824](https://github.com/cheapsteak/tbd/pull/824)** — `parts`,
  peer-authenticated envelope suppression, the awaiting-input gate,
  `TerminalWakeResult.sessionIncarnationID`, and the attachment reconcilers.
  **This PR's base.**

One standing dependency outside the stack: the composer refuses a multi-line or
multi-part message to a holder-backed session until
[#819](https://github.com/cheapsteak/tbd/pull/819)'s bracketed-paste wrapping
lands. That refusal lifts with it, with no change here.

Follow-ups this PR knowingly leaves open:

- **`CompletionCommand.source` on the wire**, so the menu can group built-ins,
  user, project and plugin commands and weight a display name. Needs a small
  protocol change in a later PR.
- **A per-file age leg for attachments in live worktrees** (see Reconciler
  below). It changes what the spec says gets reclaimed and when, so it needs a
  spec amendment and a human decision, not a quiet patch.
- **The sidecar trust anchor is first-client-wins**, noted in PR C. The composer
  inherits it unchanged.
- **`terminal.wake --json` does not print the session incarnation yet**, so the
  CLI cannot do what the app now does. A CLI follow-up.

## Assumptions

Text sitting unsent in the terminal's own input box is invisible to the
composer, and a message sent from here appends to it rather than replacing it.
No signal for that exists outside the rendered screen: Claude Code keeps its
composer in process memory, writes no draft file for a pty-hosted session, and
exposes nothing about it on its peer socket or control protocol. Reading the
screen to find out is exactly what this repo forbids. The design accepts and
documents the limitation rather than warning from a keystroke timestamp that
would be wrong in both directions.

If Claude Code ever does expose its pending input, that warning becomes possible
and this assumption should be revisited.

## Reconciler

The composer writes attachment PNGs under `~/tbd/attachments/<worktreeID>/`.
That resource is not new here, and neither are its reconcilers: PR C landed both
the archive-time unlink and the hourly `reclaimAttachments` leg of `OrphanGC`.
**This PR adds the first writer, not a new kind of resource.**

One gap is worth stating plainly rather than discovering later.
`AttachmentsCollector` reclaims a worktree's attachments directory only once the
worktree row is gone, behind a 14-day floor. A file staged into a long-lived
worktree is therefore never reclaimed while that worktree lives — every pasted
image accumulates until the worktree is archived. For the soak population that
is bounded by hand and by the flag being off; as a permanent answer it is not
good enough, which is why "a per-file age leg for live worktrees" is listed
above as a follow-up needing a spec amendment.

## Evidence & verification

**Live verification is deferred, and this section leads with that.** This
machine is running a live fleet of agent sessions, and the standing rule forbids
restarting the app or the daemon underneath them. Task 14's live pass — restart
the app, drive a stub-backed session, paste an image, watch the wake hold — was
**not run**. Nothing below was exercised against a running app.

That matters more here than it usually would: transcript work in this repo has
greened headless while being visibly broken live, more than once. So each
deferred check is named with the standing test that stands in for it, and none
of them is claimed as a substitute:

- *Deferred: type into the composer and watch the menu open, filter and accept.*
  Stands on `CompletionTriggerTests` (where a trigger opens and where it is a
  trap), `CommandRankerTests` (prefix, subsequence, and the frecency tiebreak
  fixture), and `CompletionControllerTests` —
  `escapeWhileLoadingStaysClosedWhenInventoryArrives` (a menu dismissed while
  the inventory is in flight stays dismissed when it lands) and
  `repeatedUpdatesMemoizeByQuery` (the same query is not re-fetched).
- *Deferred: press each key with the menu open and closed and see what happens.*
  Stands on `ComposerKeyRouterTests`, a table over the router's whole input
  space — menu closed, menu open, marked text, ⇧↩, ⌘↩, ⌃N/⌃P — including
  `commandReturnAlwaysSends` and `returnAndTabAreDistinguishableWithTheMenuOpen`.
- *Deferred: paste and drop images, and see the thumbnails and tokens.* Stands
  on `ComposerImagePreparerTests`, including `orientationIsBakedIntoThePixels`,
  which uses a JPEG fixture carrying a real EXIF orientation of 6 and asserts
  the output pixels are rotated — a fixture built to discriminate, since an
  untagged fixture would only have read the default back.
- *Deferred: park a session, send from the composer, and watch the hold.* Stands
  on two suites. `ComposerSendCoordinatorTests` covers the coordinator's own
  decisions and was mutation-proved (see below):
  `aSessionStartFromAnotherIncarnationDoesNotRelease`,
  `aNoOpWakeSurfacesWithoutEnteringTheHold`, and
  `aWokenWakeWithNoIncarnationDoesNotHold`. `ComposerMountingTests` covers the
  waiter registry underneath it: `anotherIncarnationDoesNotRelease`,
  `anIncarnationlessSessionStartReleasesNobody`, and the latch pair —
  `aSessionStartBeforeTheWaiterStillReleases` (a start that lands before the
  waiter registers is not missed, so the send cannot wait to its timeout) and
  `aLatchedDifferentIncarnationDoesNotReleaseALaterWaiter` (the latch does not
  become a way for the wrong spawn to release a later hold).
- *Deferred: toggle the flag and see the composer appear and vanish.* Stands on
  `ComposerMountingTests`' both-branch pair, plus the remote-worktree,
  non-Claude-terminal, missing-row and hibernated-terminal cases.

**The mutant.** The hold tests were not written and assumed to discriminate;
they were checked against a deliberately wrong implementation in which the hold
is not scoped to the wake's own spawn (`.noOp` and a wake with no incarnation
both mint a substitute and enter the hold, and the hold's answer is discarded so
any start releases it). Against that mutant,
`scripts/test.sh --filter ComposerSendCoordinatorTests` gave **13 tests, 5
failures**: the three coordinator tests named above, plus
`aWakeWhoseSessionNeverStartsFails` and `theHoldTimesOutOnTheInjectedClock`.
`aSessionStartFromAnotherIncarnationDoesNotRelease`, and its registry-level twin
`anotherIncarnationDoesNotRelease`, are the ones a reviewer cannot judge from the
diff: a hold releasing on somebody *else's* `SessionStart` looks identical in the
source to one that does not.

Other RED/GREEN pairs verified by reverting the source and re-running: the
deferred-report change (three synchronous reports before the fix, zero after),
and the timeout copy (all four assertions red against the old sentence).

**Local runs** (filtered, never the full suite — the machine is shared):

- `scripts/swift-safe build --build-tests` — `Build complete!`, no new warnings,
  on every commit's tree.
- `scripts/test.sh --filter "Composer|Completion"` — **287 tests, 33 suites, 0
  failures**.
- `scripts/test.sh --filter "ComposerMountingTests|ComposerStateTests"` — 41
  tests, 0 failures.
- `swiftlint lint --strict` on the touched files — 0 violations.

**CI.** The full suite was dispatched for this branch (`test.yml` fires
automatically only for PRs based on `main`, and this one is stacked):
https://github.com/cheapsteak/tbd/actions/runs/34036456842 — **verdict pending
at the time of writing.**

### Live run, 2026-09-06 — the daemon half is now verified live; the app half still is not

With the user's authorization the fleet's daemon and app were replaced by this
branch's (`scripts/restart.sh` from `composer-d`, `daemon status` reporting
`Build: 79d9e85d (tbd/composer-d-app-composer)`), and a zero-token Claude session was
driven against `scripts/claude-stub.py` on loopback. Full record:
`.superpowers/sdd/2026-09-05-composer-d-app-composer/live-verification-report.md`.

**What that reached, and it is the half below the composer.** The send path this PR
calls is now confirmed live end to end: a three-part payload arrives as one user turn
with the image inline (read out of the session's transcript JSONL, not the screen);
suppression asked for over a socket that is not the app's is refused on the connection
with the daemon naming `peer-pid-is-not-the-apps` — a reason only reachable after the
peer pid resolved *and* the FD-vending sidecar's recorded client was found, so the
sidecar is demonstrably connected with the app recorded against it; the awaiting-input
gate refuses with the prompt's own message and types nothing, while the same payload
without `gateOnAwaitingInput` passes the same standing prompt; and
`terminal.wake` answers `{"woken":true,"sessionIncarnationID":"1FC3EB77-…"}` on the
first call and `{"woken":false}` on the second, with the woken pane's own
`TBD_TERMINAL_INCARNATION_ID` matching. The composer's completions source answers too:
`tbd terminal completions --json` returned 18 agents and 213 commands, `source: "probe"`.

**Task 14 Steps 2–9 remain NOT DRIVEN, and the reason has changed.** It is no longer
the fleet rule — the restart happened. It is that an automated session on this machine
cannot see or reach the app's window: `screencapture` fails with
`could not create image from display` (no Screen Recording grant), `cua-driver` reports
`screen_recording_capturable: false` and its AX walk of TBD's main window returns 1790
elements that are entirely menu bar with no `AXWindow` and no window content, and
`osascript` keystrokes are disabled on this box. Posting blind keystrokes into a fleet
app risks typing into another agent's live session, so nothing was typed. Driving by RPC
instead did not help: `tbd terminal focus --activate` and the `tbd://open?worktree=…`
deep link both succeeded, but `tbd panel list` answered `No tabs found` for a worktree
the app had not been opened on by hand, so `tbd panel open --transcript` had no tab.

What the running binary does confirm: both menu commands exist and are wired —
`Focus Message Composer` (⌘/) and `Focus Transcript` (⌘⌥/) were read out of the live
app's own accessibility tree. Everything below that line still needs a person, and the
report above ends with a step-by-step list of exactly what to press and what to look
for. **The composer flag is left ON on the running daemon**
(`transcriptComposerEnabled: true`) so that pass can start immediately.

**One defect found, and it is reachable from this composer.** PR #824's parts delivery
can place an `[Image #N]` token at the **end** of a message instead of among the words,
and that send can fail to submit — observed twice, both on the first image a Claude Code
process had ever been given, and correct on every repeat within a warmed session. It
looks like a race between the `.parts` arm's back-to-back pastes and Claude Code's
asynchronous image attach. Step 7 of the report's hand-verification list says to watch
for it specifically, because a paste into this composer takes the same path.

[composer-d](https://cheapsteak.github.io/tbd/open/?worktree=3CFAF90E-C785-4CC5-BC83-E703E10064B2)

## Summary by CodeRabbit

- **New Features**
  - Added a terminal message composer with per-terminal draft restoration and send/resume actions.
  - Added slash-command and at-mention completions with ranking, keyboard navigation, previews, and argument hints.
  - Added image paste/drop support with staged thumbnails, reinsertion, removal, and size validation.
  - Added shortcuts to focus the composer or return focus to the transcript.
  - Added composer availability messaging for unsupported, blocked, or inactive sessions.

- **Bug Fixes**
  - Composer state is now cleared when tabs or terminals are removed.
  - Improved session-start handling for reliable message delivery.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -277,6 +277,13 @@ custom_rules:
       - "Sources/TBDApp/Panes/Transcript/TranscriptOverlayView\\.swift"
       # OverlayFileView is the body of a file frame in the overlay — outside the transcript list, real ScrollView is safe.
       - "Sources/TBDApp/Panes/Transcript/OverlayFileView\\.swift"
+      # CompletionOverlayView is the composer's completion menu, a fixed-height
+      # floating overlay above the composer — never hosted or measured as a
+      # transcript row card.
+      - "Sources/TBDApp/Panes/Transcript/Composer/CompletionOverlayView\\.swift"
+      # AttachmentStrip is the composer's thumbnail strip above the text field —
+      # also outside the transcript list's row-measurement path.
+      - "Sources/TBDApp/Panes/Transcript/Composer/AttachmentStrip\\.swift"
     regex: '\bScrollView\b'
     match_kinds:
       - identifier
--- a/Sources/TBDApp/AppState+Composer.swift
+++ b/Sources/TBDApp/AppState+Composer.swift
@@ -26,6 +26,32 @@ extension AppState {
         composerDrafts[terminalID] = nil
     }
 
+    /// Forget everything the composer keys on this terminal, because the
+    /// terminal itself is gone.
+    ///
+    /// Called from both deaths a terminal has: the tab close that deletes it,
+    /// and `removeDeletedTerminalFromState`, which every other route — a pane
+    /// close, an archive, a daemon-reported removal — funnels through. The tab
+    /// close alone was not enough: it is the rarer of the two, and a draft left
+    /// behind by the common path sits in the map for the app's lifetime, joined
+    /// by a fresh empty one whenever a send finishing after the row is gone asks
+    /// `composerDraft(for:)` again.
+    ///
+    /// Memory only, and safe by construction: the terminal row no longer exists,
+    /// so nothing can send this draft, mount this composer, or spawn into this
+    /// incarnation. The two focus registries hold their views weakly and leak
+    /// nothing, but they do accumulate empty boxes, so they are pruned here too.
+    ///
+    /// Waiters are RESUMED rather than dropped — see
+    /// `releaseSessionStartWaiters`.
+    func forgetComposerState(for terminalID: UUID) {
+        discardComposerDraft(for: terminalID)
+        composerFocusTargets.removeValue(forKey: terminalID)
+        transcriptFocusTargets.removeValue(forKey: terminalID)
+        lastStartedIncarnation.removeValue(forKey: terminalID)
+        releaseSessionStartWaiters(terminalID: terminalID)
+    }
+
     /// Whether the daemon reports the composer as enabled. False until
     /// capabilities have been fetched, which is the conservative reading: a
     /// composer that flashed in and then disappeared would be worse than one that
--- a/Sources/TBDApp/AppState+Composer.swift
+++ b/Sources/TBDApp/AppState+Composer.swift
@@ -1,0 +1,51 @@
+import Foundation
+import os
+import TBDShared
+
+private let logger = Logger(subsystem: "com.tbd.app", category: "composer")
+
+extension AppState {
+
+    /// This terminal's draft, created on first ask and kept for the app's
+    /// lifetime — or until `discardComposerDraft` forgets it.
+    ///
+    /// Kept in an `@ObservationIgnored` dictionary because the DICTIONARY is not
+    /// what anything observes: each `ComposerDraft` is itself `@Observable`, and
+    /// a view that holds one re-renders on its changes. Making the registry
+    /// observable would republish every composer in the app whenever any of them
+    /// gained a draft.
+    func composerDraft(for terminalID: UUID) -> ComposerDraft {
+        if let existing = composerDrafts[terminalID] { return existing }
+        let draft = ComposerDraft()
+        composerDrafts[terminalID] = draft
+        return draft
+    }
+
+    /// Forget a terminal's draft — on a successful send, and when its tab closes.
+    func discardComposerDraft(for terminalID: UUID) {
+        composerDrafts[terminalID] = nil
+    }
+
+    /// Whether the daemon reports the composer as enabled. False until
+    /// capabilities have been fetched, which is the conservative reading: a
+    /// composer that flashed in and then disappeared would be worse than one that
+    /// appeared a moment late.
+    var transcriptComposerEnabled: Bool {
+        daemonCapabilities?.transcriptComposerEnabled ?? false
+    }
+
+    /// Fetch this terminal's completion inventory. nil on any failure — the menu
+    /// shows its loading row and then simply has nothing to offer, which is a
+    /// smaller loss than an error banner over a text field.
+    func fetchCompletions(terminalID: UUID) async -> TerminalCompletionsResult? {
+        do {
+            return try await daemonClient.terminalCompletions(terminalID: terminalID)
+        } catch {
+            logger.debug("""
+            completions unavailable for terminal \
+            \(terminalID.uuidString, privacy: .public): \(error, privacy: .public)
+            """)
+            return nil
+        }
+    }
+}
--- a/Sources/TBDApp/AppState+ComposerFocus.swift
+++ b/Sources/TBDApp/AppState+ComposerFocus.swift
@@ -1,0 +1,102 @@
+import AppKit
+import Foundation
+
+/// A weak handle on one composer's text view, so focus can be moved to it from
+/// a menu command without the registry keeping a torn-down pane alive.
+///
+/// The `TerminalFocusTarget` shape (`AppState+TerminalFocus.swift`), for the
+/// same reason: focus is a fact about a view that exists right now.
+@MainActor
+final class ComposerFocusTarget {
+    weak var view: NSView?
+
+    init(_ view: NSView) { self.view = view }
+}
+
+/// What a composer's wake answered.
+///
+/// Three cases rather than a `Bool` and an error string, because the middle one
+/// is the interesting one: a wake the daemon answered `woken: false` delivered
+/// the prompt nowhere, which is neither a success nor a failure and must not be
+/// reported as either.
+///
+/// Named at file scope rather than nested in `ComposerSendCoordinator` so this
+/// plumbing compiles before the coordinator exists; the coordinator adopts it
+/// as its `WakeReply`.
+enum ComposerWakeReply: Equatable, Sendable {
+    /// A spawn happened. The incarnation is the id that spawn will echo on its
+    /// `SessionStart`, or nil from a daemon that predates the field — in which
+    /// case there is nothing to scope a wait to.
+    case woken(incarnationID: UUID?)
+    /// Idempotent no-op: already awake, or another wake was in flight. The
+    /// prompt reached nobody.
+    case noOp
+    case failed(message: String)
+}
+
+extension AppState {
+    func registerComposerView(_ view: NSView, for terminalID: UUID) {
+        composerFocusTargets[terminalID] = ComposerFocusTarget(view)
+    }
+
+    /// Only the CURRENT view may unregister. A pane rebuilt while the old view is
+    /// still deallocating would otherwise evict its own replacement.
+    func unregisterComposerView(_ view: NSView, for terminalID: UUID) {
+        guard composerFocusTargets[terminalID]?.view === view else { return }
+        composerFocusTargets.removeValue(forKey: terminalID)
+    }
+
+    /// Cmd+/ — put the caret in this terminal's composer. A no-op when no
+    /// composer is mounted, which is the honest answer for a pane that is closed.
+    func focusComposer(terminalID: UUID) {
+        guard let view = composerFocusTargets[terminalID]?.view else { return }
+        view.window?.makeFirstResponder(view)
+    }
+
+    /// Escape from the composer — hand focus back to the transcript, which is
+    /// where the person was reading.
+    func focusTranscript(terminalID: UUID) {
+        guard let view = transcriptFocusTargets[terminalID]?.view else {
+            // No transcript table registered: give up first responder rather
+            // than holding it, so the pane's own key handling resumes.
+            composerFocusTargets[terminalID]?.view?.window?.makeFirstResponder(nil)
+            return
+        }
+        view.window?.makeFirstResponder(view)
+    }
+
+    /// The Reveal Terminal action on the blocked banner. Answering a dialog in
+    /// the terminal is the correct resolution, so the composer's job is to get
+    /// the person there rather than to offer a second way to answer.
+    func revealTerminal(terminalID: UUID) {
+        guard let view = terminalFocusTargets[terminalID]?.view else { return }
+        view.window?.makeFirstResponder(view)
+        view.window?.makeKeyAndOrderFront(nil)
+    }
+
+    /// The Reveal Terminal action's sibling for the wake path: wake this
+    /// terminal and report what the daemon answered, incarnation included.
+    ///
+    /// A sibling of `wakeTerminal` rather than a change to it. That one returns
+    /// an error string, which is all its five call sites want, and threading a
+    /// result through PR A's shared `terminalWakeSender` seam would change every
+    /// one of them to serve one screen. The app's own `wakeInFlight` dedupe is
+    /// deliberately not consulted here either: the daemon has its own
+    /// `wakesInFlight`, and it answers a racing wake with `woken: false`, which
+    /// is exactly the honest reply for a composer — the prompt went nowhere.
+    func wakeTerminalForComposer(
+        terminalID: UUID, worktreeID: UUID, prompt: String
+    ) async -> ComposerWakeReply {
+        do {
+            let size = mainAreaTerminalSize()
+            let result = try await composerWakeSender(
+                terminalID, size.cols, size.rows, prompt)
+            await refreshTerminals(worktreeID: worktreeID)
+            return result.woken
+                ? .woken(incarnationID: result.sessionIncarnationID)
+                : .noOp
+        } catch {
+            return .failed(message: error.localizedDescription)
+        }
+    }
+}
--- a/Sources/TBDApp/AppState+ComposerFocus.swift
+++ b/Sources/TBDApp/AppState+ComposerFocus.swift
@@ -134,27 +134,50 @@ extension AppState {
 }
 
 extension AppState {
-    /// The terminal the composer menu commands act on.
+    /// The terminal the composer menu commands act on: the first one, among the
+    /// tabs that could plausibly be on screen, that actually has a composer
+    /// mounted.
     ///
-    /// `resolvedFocusedTabCloseContext()` is the accessor the Terminal menu's
-    /// other items already use to name what has focus. It answers nil while
-    /// focus sits in something that is not a terminal view — the transcript
-    /// table and the composer itself both qualify, and both are exactly where
-    /// somebody pressing these shortcuts is — so the last-focused tab is the
-    /// fallback, which is the same value that method falls back to when nothing
-    /// has registered at all.
+    /// **The registration is the answer, not a tie-breaker.** ⌘/ is only ever
+    /// meaningful for a terminal whose composer exists; `focusComposer` on any
+    /// other terminal is a no-op. Naming the registered one — and nil when there
+    /// is none — is what makes the menu items' `.disabled(…)` honest instead of
+    /// offering an action that quietly does nothing.
+    ///
+    /// **Two candidate tabs, in order.** The focused-terminal context first, so
+    /// a split holding both a terminal and a transcript still answers for the
+    /// half the person is in. Then the selected worktree's active tab, which is
+    /// the only candidate that exists at all while focus sits in the transcript
+    /// table or the composer itself: `resolvedFocusedTabCloseContext()` answers
+    /// nil for anything that is not a `TBDTerminalView`, and nothing writes
+    /// `focusedTabCloseContext` for a `.liveTranscript` tab — so relying on it
+    /// alone made ⌘/ in a transcript pane name the last-focused *terminal* tab,
+    /// which is either a no-op or, worse, a caret moved in a background tab.
+    ///
+    /// The unconditional `resolvedFocusedTabCloseContext()` call also keeps the
+    /// observable dependency it documents: it reads `focusedTabCloseContext`,
+    /// the one observable property that moves when focus does, so `.disabled(…)`
+    /// re-evaluates. `composerFocusTargets` is `@ObservationIgnored`, so a
+    /// composer mounting or going away still does not re-evaluate the menu on
+    /// its own — the same staleness `canCloseFocusedTab` documents, and the
+    /// consequence is a stale menu state rather than a wrong action.
     ///
     /// `terminalIDs(in:)` rather than the layout's own enumeration, because a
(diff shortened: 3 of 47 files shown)


cheapsteak:
Restacked onto PR C's new head, which now sits on PR B's latest head `96c043cd` — so this run carries PR B's test-only fix for the `BoundedProcessRunnerTests` timing tests that reddened the previous run. New head: `2cedac31` (31 commits above PR C).

Full CI run: https://github.com/cheapsteak/tbd/actions/runs/34037848435

cheapsteak:
**Fixed the fast-pass failure in `theHoldTimesOutOnTheInjectedClock`** — `79d9e85d`.

**The failure** (run [34037848435](https://github.com/cheapsteak/tbd/actions/runs/34037848435), fast parallel pass 2/2): `ComposerSendCoordinatorTests.swift:200` — `EventDrivenTestClock: no task suspended on the clock within 45.0 seconds — observed zero registered sleepers, virtual now = 0.0 seconds.` Green locally, every time.

**The cause is the runner, not the coordinator.** Reaching the timeout child's `sleep` takes: the `async let` child getting a thread, then a main-actor stretch through the `wake` double and into `holdUntilSessionStart`'s task group — and the main actor is a process-wide serialization point every other `@MainActor` test in the pass queues on — then a thread for the group's own timeout child. The clock's 45 s hang-guard default sits *below* fast pass 2's measured p90 per-test latency (51.4 s, max 55.3 s), so a wall-clock 45 s bound on that chain measures scheduling. Nothing in the test's doubles can hold the group back: the `awaitSessionStart` double parks on its `AsyncStream` at its first suspension, releasing the main actor, and the timeout child never touches it. No production defect found — `ComposerSendCoordinator` arms the sleep as directly as the shape allows.

**The change is in the test.** `requireAdvanceWhenArmed` is split into its two halves so the arming is proven before virtual time moves, and the arming wait is bounded by `TestDeadlines.saturatedPass` (90 s) instead of the 45 s default. The test also takes `.clockDriven`, so a genuinely un-armed timer still lands the clock's named diagnostic rather than tripping an outer limit first. `EventDrivenTestClock`'s docs now say when a call site should raise that default rather than claiming every one takes it.

**The other two `EventDrivenTestClock` suites in `Tests/TBDAppTests`** (`SearchQueryDebouncerTests`, the appearance debounce suite) do not carry this exposure: both arm one unstructured-task hop from a synchronous call in the test body, with no main-actor round trip and no task group in between, and both already run under `.clockDriven`. Left alone.

**Verification:** `scripts/test.sh --filter "ComposerSendCoordinatorTests|ComposerMountingTests"` — 50 tests in 2 suites, 0 failures. Full CI: https://github.com/cheapsteak/tbd/actions/runs/34039144349

cheapsteak:
**Live verification run — 2026-09-06, daemon and app built from `79d9e85d`.** The
Evidence section is updated; the full record is in
`.superpowers/sdd/2026-09-05-composer-d-app-composer/live-verification-report.md`.

**The half below the composer is now verified live**, against a zero-token session
(`scripts/claude-stub.py` on loopback; both stub servers printed
`no model request left this machine`). Everything this PR calls into works end to end:
a three-part payload arrives as one user turn with the image inline; suppression asked
for from a socket that is not the app's is refused on the connection with the daemon
naming `peer-pid-is-not-the-apps` — a reason only reachable after the peer pid resolved
*and* the FD-vending sidecar's recorded client was found, so the sidecar is demonstrably
connected with the app recorded against it; the awaiting-input gate refuses with the
prompt's own message and types nothing while the opt-out passes the same prompt;
`terminal.wake` names its spawn once and never again, matching the woken pane's
`TBD_TERMINAL_INCARNATION_ID`; and `tbd terminal completions --json` answers with 18
agents and 213 commands.

**Task 14 Steps 2–9 are still NOT DRIVEN, and the reason has changed.** It is no longer
the fleet rule — the restart happened. An automated session on this machine simply cannot
see or reach the app's window: `screencapture` fails with
`could not create image from display`, `cua-driver` reports
`screen_recording_capturable: false` and its AX walk of TBD's main window returns 1790
elements that are entirely menu bar (no `AXWindow`, no window content), and `osascript`
keystrokes are disabled here. Typing blind into a fleet app could inject text into another
agent's live session, so nothing was typed. Driving by RPC did not help either:
`tbd terminal focus --activate` and the `tbd://open?worktree=…` deep link both succeeded,
but `tbd panel list` answered `No tabs found` for a worktree nobody had opened by hand, so
`tbd panel open --transcript` had no tab to attach to.

What the running binary does confirm from its own accessibility tree: `Focus Message
Composer` (⌘/) and `Focus Transcript` (⌘⌥/) both exist in the live app's menu bar.

**The composer flag is left ON on the running daemon** (`transcriptComposerEnabled:
true`), so a hand pass can start immediately — the report ends with a step-by-step list
of exactly what to press and what to look for, including the one live proof this run
could not reach: a composer Return should produce a user turn with **no
`<tbd-dispatch/>` line**, which is where PR #824's peer authentication is granted rather
than refused.

**One defect to watch for during that pass.** PR #824's parts delivery can put an
`[Image #N]` token at the end of a message instead of among the words, and that send can
fail to submit — observed twice, both on the first image a Claude Code process had ever
been given. A paste into this composer takes the same path, so Step 7 of the list calls
it out specifically.

cheapsteak:
Restacked onto the new PR C head (`364acfdf`). The whole chain now carries `main` including #819's holder send path, plus C's restack fixes. D's 33 commits replayed with **no conflicts** — the two hunks that could have collided (`RPCRouter+TerminalHandlers.swift` and `Sources/TBDShared/StateDelta.swift`) both merged cleanly and are intact on the new head `fd1cf062`.

The chain as it stands: `main` (incl. #819) → #822 → #823 → #824 (`364acfdf`, carries the D1 paste-race fix) → #825 (`fd1cf062`).

Local verification on the restacked head: `scripts/swift-safe build --build-tests` clean; `scripts/test.sh --filter "Composer|Completion"` → 287 tests in 33 suites, 0 failures.

Full CI: https://github.com/cheapsteak/tbd/actions/runs/34043050422

cheapsteak:
restacked onto main @ 406a9800 (#818, #827); conflicts resolved in the branches below (none left to resolve here); run https://github.com/cheapsteak/tbd/actions/runs/34055654550

cheapsteak:
Restacked onto C @ `e7b293a9` (stack rebased onto A's review-gate fix round, #822 @ `5ac61831`) — no conflicts, 33 own commits, new head `56c197f0`; `scripts/test.sh --filter "Composer|Completion"` → 287 tests in 33 suites, 0 failures. `git merge-tree --write-tree origin/main origin/tbd/composer-d-app-composer` is clean. CI: run 34058009281.

PR #829, opened 2026-09-06
Fix: The composer's slash menu opens upward instead of covering the text being typed

cheapsteak:
## What's broken

Type `/` in the transcript composer and the slash-command menu drops **down over
the text field** instead of opening above it. The list covers the words being
typed — you cannot see the command you are narrowing while you narrow it — and
because there is nothing below the composer but the bottom of the pane, most of
the list is cut off: an eight-row menu shows about three rows, and the rest is
unreachable.

It happens on every completion the composer offers, for everyone with the
transcript composer turned on. An offscreen render of the newly landed composer
is what caught it: the menu in the picture sat squarely on top of the message,
with its rows running off the bottom edge.

## Why it happens

Two independent mistakes, both in how the list is attached to the composer, and
each on its own enough to produce a wrong-looking menu.

**The anchor named the wrong edge.** The list is a SwiftUI overlay on the
composer, attached with an alignment plus an alignment guide meant to flip it —
pin the overlay to the composer's bottom, then move the overlay's bottom guide up
to its own top, so what lands on the composer's bottom edge is the list's top and
the list rises from there. That reads plausibly and is wrong twice over: it
describes a list opening *downward* from the composer's bottom, and `.overlay`
aligns plain edges and ignores the guide entirely — so the list simply hung off
the edge the alignment named.

**The height was a ceiling, not a size.** The list bounded itself with
`.frame(maxHeight:)`. An overlay is proposed the size of the view it hangs off,
so the ceiling never came into play: the menu came out the height of the text
field, whatever the row count, which is why only about three of its eight rows
were ever drawn.

## When it broke

Present since the composer landed in #825 — the menu has never opened upward. The
composer's own tests pin the decisions around the menu (which key accepts, what
the ranker returns, what the send button says), all of which are pure functions;
where the list is *placed* is the one thing none of them could see, because it
takes a mounted view in a real window to have a frame at all.

## What this PR does

- Anchors the list with geometry instead of a guide: a zero-height anchor pinned
  to the composer's **top** edge, with the list as that anchor's bottom-aligned
  overlay. Its bottom edge lands on the composer's top and its rows stack upward,
  out over the transcript. Only edge alignment is involved, which is the part of
  `.overlay` that demonstrably works.
- Gives the list a **definite** height — one row per row it has, up to eight —
  rather than a maximum a layout proposal can undercut. The eight-row cap and the
  scroller past it are unchanged; so is the 460 pt width.
- Gives the composer `.zIndex(1)` in the pane's stack, so the list paints over
  the transcript table rather than under it. The table is an AppKit view and the
  list's sibling, so which one wins is stated rather than left to stacking order.
  This lifts the whole composer subtree, so an open menu also paints over the
  table's own bottom-leading jump-to-bottom button when the transcript is
  scrolled up. That is intended and now says so at the modifier: the menu is a
  transient popup and owns the pane while it is open — Escape or a token change
  closes it and the button is back.
- Routes `AppState.fetchCompletions` through an injectable
  `composerCompletionsFetcher`, the same seam shape as the existing
  `composerWakeSender`. Opening the menu requires an inventory, and this is what
  lets a test supply one without a daemon on the other end of the socket.

No flag: this is a bug fix to a surface that already ships behind the composer's
own gate, and it restores the behavior the design already specifies ("the list
opens upward from the composer"). No spec change for the same reason.

## Evidence & verification

**The repro.** An offscreen render of the composer with a short fake transcript
and the `please /comp` menu open — `NSHostingView` + `cacheDisplay`, 2x, 720 pt
wide. Before: the list starts at the composer's top edge and runs *downward*
across the message and off the bottom of the frame, with the typed text hidden
behind it and three rows visible. After: the list sits entirely above the field,
its bottom edge on the composer's top, `please /comp` legible below it, every row
drawn and none clipped.

**The discriminating test.** `Tests/TBDAppTests/CompletionOverlayPlacementTests.swift`
mounts the real `MessageComposerView` in an offscreen borderless window with a
transcript's worth of space above it, opens the menu through the production
draft-restore path, waits for the list to reach full height, and measures both
frames in window coordinates:

- `theListOpensAboveTheTextField` — the list's bottom edge is at or above the
  text field's top edge. Fails against the old placement (list bottom 520, field
  top 380).
- `theWholeListFitsInsideTheWindow` — the list is 460 x 352 (eight rows) and
  inside the pane on both edges. Fails against the old sizing (148 pt tall).
- `theListIsAsTallAsItsRowsUpToEight` — the height arithmetic, without a window.

Red before the fix, green after: `scripts/test.sh --filter
"CompletionOverlayPlacementTests|MessageComposerViewTests"` runs 28 tests, 2
failures on the old code and 0 on this branch. `swiftlint --strict` clean on
every touched file.

**A paint-order guard, and what it does not prove.** Those three assertions are
geometry, and a frame is the same whether the list paints over its neighbour or
under it — the suite stood the transcript table up as `Color.clear`, a SwiftUI
view, which cannot pose the question `.zIndex(1)` answers.
`theListPaintsOverTheAppKitViewBesideIt` mounts a real `NSViewRepresentable`
sibling — the shape of `TableTranscriptView` — that fills the region the list
grows into with solid red, opens the menu the same way, renders the hosting view
with `bitmapImageRepForCachingDisplay` + `cacheDisplay(in:to:)`, and reads the
list's own interior back out. A control pixel beside the list must be the
sibling's red first, so a capture in which nothing painted fails as the harness
problem it is; then no measurable share of the list's interior may match it.

It is a guard on the outcome, not a discriminating test of the modifier. With
`.zIndex(1)` removed the capture comes back pixel-for-pixel identical — 0.94 grey
interior, the selected row's accent tint along the top, 0% of it the sibling's
red — because an offscreen `cacheDisplay` of an `NSHostingView` already orders
the composer's overlay above a representable sibling declared before it. The
mutant is stated here rather than dressed up as a red: the compositing this fix
is about does not reproduce on that route. The modifier stays because the live
pane's sibling is a layer-backed scrolling table rather than a solid-color
`NSView`, and stating the order is cheaper than depending on declaration order
there; the test stays because it fails loudly if the list ever stops painting
over its neighbour at all. Its doc comment carries the same caveat.

To be explicit about what is and is not proven here: of the four tests this PR
adds, three — `theListOpensAboveTheTextField`, `theWholeListFitsInsideTheWindow`,
`theListIsAsTallAsItsRowsUpToEight` — discriminate the geometry fix, each red
against the old placement or sizing. `theListPaintsOverTheAppKitViewBesideIt`
does not discriminate `.zIndex(1)`; it is a guard against the list vanishing
behind its sibling altogether, and its own comment records that the capture is
identical with the modifier removed. The `.zIndex(1)` paint-order change is
therefore defensive rather than test-verified: it guards a live-app case (a
layer-backed `NSTableView` compositing over a SwiftUI sibling) that the
offscreen harness cannot reproduce, and the live pass is what actually verifies
it.

[composer menu upward](https://cheapsteak.github.io/tbd/open/?worktree=B6F8AB9D-8BD6-4902-98EF-16F8DCA04FE5)

## Summary by CodeRabbit

- **New Features**
  - Terminal completion suggestions now open upward from the message composer, keeping the text field visible.
  - The completion list adjusts its height to the number of available suggestions, up to eight rows.
  - Suggestions remain visible above the transcript and surrounding interface elements.

- **Bug Fixes**
  - Improved overlay placement so suggestions stay within the available window area and avoid covering the composer.
  - Improved completion loading reliability.

### DIFF
--- a/Sources/TBDApp/AppState+Composer.swift
+++ b/Sources/TBDApp/AppState+Composer.swift
@@ -65,7 +65,7 @@ extension AppState {
     /// smaller loss than an error banner over a text field.
     func fetchCompletions(terminalID: UUID) async -> TerminalCompletionsResult? {
         do {
-            return try await daemonClient.terminalCompletions(terminalID: terminalID)
+            return try await composerCompletionsFetcher(terminalID)
         } catch {
             logger.debug("""
             completions unavailable for terminal \
--- a/Sources/TBDApp/AppState.swift
+++ b/Sources/TBDApp/AppState.swift
@@ -1491,6 +1491,16 @@ final class AppState {
             try await daemonClient.terminalWakeReporting(
                 terminalID: terminalID, cols: cols, rows: rows, prompt: prompt)
         }
+    /// How the composer's completion inventory is fetched — injectable for the
+    /// same reason as `composerWakeSender`, so a mounted composer can open its
+    /// menu with real rows in a test, with no daemon on the other end of the
+    /// socket.
+    @ObservationIgnored
+    lazy var composerCompletionsFetcher:
+        @MainActor (UUID) async throws -> TerminalCompletionsResult =
+        { [daemonClient] terminalID in
+            try await daemonClient.terminalCompletions(terminalID: terminalID)
+        }
     /// How `setTranscriptComposerEnabled` persists the composer gate —
     /// injectable for the same reason as `controlModeSetter`.
     @ObservationIgnored
--- a/Sources/TBDApp/Panes/Transcript/Composer/CompletionOverlayView.swift
+++ b/Sources/TBDApp/Panes/Transcript/Composer/CompletionOverlayView.swift
@@ -11,9 +11,12 @@ import SwiftUI
 /// the floating-panel code documents that adding a child window to the split-view
 /// window raises an exception.
 ///
-/// It opens **upward** from the composer, and its height is fixed at eight rows
-/// with a scroller, so a change in row count never shifts the text field under
-/// the caret.
+/// It opens **upward** from the composer: its BOTTOM edge is pinned to the top
+/// of the composer and the rows stack up from there, out over the transcript, so
+/// the list never covers the text being typed. Eight rows at most, with a
+/// scroller past that — and because the pinned edge is the bottom one, a change
+/// in row count grows the list away from the text field rather than shifting the
+/// field under the caret.
 struct CompletionOverlayView: View {
     let controller: CompletionController
     let onAccept: (CommandRanker.Row) -> Void
@@ -25,7 +28,22 @@ struct CompletionOverlayView: View {
     @State private var hoveredIndex: Int?
 
     static let rowHeight: CGFloat = 44
-    static let maxHeight: CGFloat = rowHeight * CGFloat(CompletionController.visibleRowCount)
+    /// The tallest the list ever gets: eight rows, and a scroller past that.
+    static var maxHeight: CGFloat {
+        listHeight(rowCount: CompletionController.visibleRowCount)
+    }
+
+    /// How tall the list is for a given number of rows: as many rows as it has,
+    /// up to eight.
+    ///
+    /// **A definite height, not a cap.** This is an overlay on the composer, and
+    /// an overlay is proposed the composer's own size — so a list that only
+    /// bounded itself with `.frame(maxHeight:)` came out the height of the text
+    /// field, three rows of an eight-row menu, however many rows it had. Stating
+    /// the height is what lets it be taller than the view it hangs off.
+    static func listHeight(rowCount: Int) -> CGFloat {
+        rowHeight * CGFloat(min(max(rowCount, 1), CompletionController.visibleRowCount))
+    }
 
     var body: some View {
         // `.closed` renders nothing at all — no chrome, no zero-height frame —
@@ -114,7 +132,7 @@ struct CompletionOverlayView: View {
                     }
                 }
             }
-            .frame(maxHeight: Self.maxHeight)
+            .frame(height: Self.listHeight(rowCount: controller.rows.count))
             .onChange(of: controller.selectedIndex) { _, index in
                 guard let index, controller.rows.indices.contains(index) else { return }
                 proxy.scrollTo(controller.rows[index].id, anchor: .center)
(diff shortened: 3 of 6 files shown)


cheapsteak:
Both review findings addressed in `27f11e1f`.

**1 — `.zIndex(1)` also covers the table's `jumpToBottomButton`** (`TableTranscriptPaneView.swift`). Intended, and now stated at the modifier: the modifier lifts the whole composer subtree, so while the completion menu is open it paints over everything in the pane, the bottom-leading jump-to-bottom button included. The menu is a transient popup that owns the pane for as long as it is open; Escape or a token change dismisses it and the button is back. No code change — a comment at the modifier and a paragraph in the PR body.

**2 — the placement suite could not see the fix** (`CompletionOverlayPlacementTests.swift`). It stood the transcript table up as `Color.clear`, so the geometry assertions passed with or without `.zIndex(1)`. Added `theListPaintsOverTheAppKitViewBesideIt`: a real `NSViewRepresentable` sibling (an `NSView` subclass filling itself solid red) occupies the region the list grows into, the menu is opened through the same draft-restore path, the hosting view is rendered with `bitmapImageRepForCachingDisplay` + `cacheDisplay(in:to:)`, and the list's own interior is read back out. A control pixel beside the list is asserted red first, so an empty capture fails as a harness problem rather than passing as a clean list.

**The mutant did not go red, and that is reported rather than papered over.** With `.zIndex(1)` removed the capture comes back pixel-for-pixel identical — interior 0.94 grey, the selected row's accent tint along the top, 0% of it the sibling's red. Offscreen, `cacheDisplay` of an `NSHostingView` already orders the composer's overlay above a representable sibling declared before it, so this route does not exercise the stacking the modifier states. Per the ruling the test is kept as a guard (it fails loudly if the list ever stops painting over its neighbour), `.zIndex(1)` is kept with its comment, and the caveat is written into both the test's doc comment and the PR body.

`scripts/test.sh --filter "CompletionOverlayPlacementTests|MessageComposerViewTests"` — 28 tests, 0 failures. `swiftlint --strict` clean on both touched files.

cheapsteak:
Addressed the review gate's MEDIUM (no discriminating test for `.zIndex(1)`, Evidence section silent about it) in 834f7d02: the modifier's comment now states plainly that no test discriminates it — `theListPaintsOverTheAppKitViewBesideIt`'s own comment already admits the capture is identical without it — and that it is kept defensively for the live app, where an `NSTableView` sibling may composite over the SwiftUI overlay in a way the offscreen harness cannot reproduce. The PR body's Evidence section now says the same: three tests discriminate the geometry fix, the fourth is a non-discriminating guard, and the paint-order change itself is verified only by the live pass.

Unrelated to this: the previous CI run's test failure was `AppearanceDebounceTests` on rapid scheme changes, a clock-driven flake in a file this PR never touches. This push re-triggers both the review gate and the test workflow.

PR #831, opened 2026-09-07
DX: Give the transcript composer stable accessibility identifiers so GUI drivers and tests can find its controls

cheapsteak:
## Summary

Anyone driving TBD's UI from outside — a GUI driver such as cua-driver, an
XCUITest-shaped tool, or the repo's own offscreen hosting harness — finds a
control by asking the accessibility tree for it by identifier. A
live-verification pass found the transcript composer had none: the text field,
the send button, the blocked and error banners, the completion rows and the
attachment thumbnails were all anonymous. A driver could see shapes but had
nothing stable to address, and a test could only reach a control by AppKit
class or by matching the sentence a person hears — which changes with the
terminal's name, the daemon's message, and the image's number.

This PR gives every composer control a stable identifier. It is the
prerequisite for driving the composer at all, and it is additive metadata: no
behavior a user sees changes.

## Why this shape

**No spec.** This is UI/test-infrastructure metadata — identifiers plus the
test that proves they land — not a revision to any theory of the system. Per
CLAUDE.md, a minor UI change of this kind needs none.

**Identifiers are not labels, and both stay.** A label is the sentence
VoiceOver reads and it is allowed to change with the copy; an identifier is
the handle automation holds and it must not. Every existing
`.accessibilityLabel` is left exactly as it was, and no identifier was folded
into one.

**One place.** They are `static let`s and small functions on a new
`ComposerAccessibility` enum, so a driver and a test that both mean
`composer.send` are reading the same constant rather than each retyping a
literal.

**Rows are named by command, not by index.** The ranker reorders the
completion list as the query changes, so `composer.menu.row.3` would name a
different row from one keystroke to the next; `composer.menu.row.compact` names
the same one every time.

**Containers are stated, not hoped for.** A SwiftUI `VStack` is not an
accessibility element at all, so an identifier on one alone names nothing.
The composer root, the completion list and the attachment strip declare
`.accessibilityElement(children: .contain)` — which keeps every child
individually addressable, where `.combine` would flatten them into one.

## What this PR does

New file `Sources/TBDApp/Panes/Transcript/Composer/ComposerAccessibility.swift`
holds every identifier:

- `composer.root` — the whole composer, as an accessibility container
- `composer.field` — the `NSTextView`, set in `makeNSView`
- `composer.send` — the send/resume button
- `composer.note` — the not-running note
- `composer.blocked.message`, `composer.blocked.reveal` — the blocked banner's
  sentence and its Reveal Terminal button
- `composer.error` — the banner a failed send or a refused image raises
- `composer.menu` — the completion list
- `composer.menu.row.<command>` — one row per command, e.g.
  `composer.menu.row.compact`
- `composer.attachments` — the attachment strip
- `composer.attachments.item.<n>` — one staged image's thumbnail
- `composer.attachments.remove.<n>` — that image's remove button

The argument hint is not in the list because it is not a view: it is a string
the composer's `NSTextView` draws itself, with no element to identify.

One behavioural change came with the strip. A thumbnail was a `.combine`d
accessibility element, which folds the remove button into the picture and
leaves nothing to press separately — for a driver or for VoiceOver. It is now
a container that keeps the same label and carries the thumbnail's own click as
an explicit accessibility action, so both the thumbnail and its x are
addressable.

No flag: this is additive metadata on views that already existed, it acts on
nothing by itself, and it destroys no state.

## Evidence & verification

New `Tests/TBDAppTests/ComposerAccessibilityIdentifierTests.swift` mounts the
real `MessageComposerView` in an offscreen window — the harness shape
`CompletionOverlayPlacementTests` already uses — and asserts on the
**accessibility tree**, not on the source. Asserting on the tree is the point:
an identifier on a view SwiftUI declines to make an element of is a no-op that
reads perfectly well in a diff.

Two facts about that tree had to be measured rather than assumed, and both are
written down in the helper:

- **SwiftUI builds no accessibility tree until a client asks for one.** With
  `AXEnhancedUserInterface` clear, a hosting view's accessibility children come
  back empty however much the views declare. The harness sets that flag, which
  is exactly what a driver's attaching does — so the test measures the tree the
  driver will see.
- **SwiftUI's nodes do not declare the Objective-C accessibility protocol in
  Swift's eyes.** A typed `as? NSAccessibilityProtocol` walk drops every one of
  them and reports an empty tree. The walk reaches them by selector instead.

RED/GREEN, measured both ways. With the identifiers reverted from `Sources/`
and the suite unchanged, 5 of the 6 tests fail; the sixth is the negative
control (`theNotRunningNoteIsAbsentWhileRunning`), which passes either way and
is what keeps the positive assertions honest — it proves the note assertion is
about the state and not about a string that is always present. With the
identifiers in place all 6 pass.

`scripts/test.sh --filter "ComposerAccessibilityIdentifierTests|CompletionOverlayPlacementTests|MessageComposerViewTests|MessageComposerTextViewTests"`
— 45 tests in 4 suites, 0 failures. `swiftlint --strict` clean across the tree
(0 violations in 1032 files).

The AX walk helper (`axIdentifiers(in:)`, plus a tree-dump used in failure
messages) lives in the test file rather than in `Tests/TestSupport/`: that
target is the daemon's support library and links into the daemon suites, so
putting an AppKit-only helper there would drag AppKit into all of them. As a
top-level function in the app test target it is already reusable by every suite
in it.

### The bridge switch is scoped, and its suites are serialized (d557a18b)

The first push of this branch turned the accessibility bridge on and left it
on. `AXEnhancedUserInterface` is set on `NSApplication.shared`, so it is
process-global, and every test target compiles into one process where Swift
Testing runs suites in parallel — the flag was therefore live under every other
suite in the run. `AccessibilityWindowGeometryGuardTests` drives an `NSWindow`
through the same accessibility setters the guard swizzles and then reads
`window.frame` back, and with enhanced UI on AppKit stops applying such a move
synchronously (it is the mode window managers switch off before they move
anything), so two of its assertions failed on CI in run 34074933464. That is a
real AppKit behavior change rather than a weakness in the guard, so the guard is
unchanged: the fix is that the flag must not be left on. The switch is now
flipped only by `withAccessibilityBridge`, which records the prior value and
restores it in a `defer`, and both suites nest under a new `@Suite(.serialized)
enum AccessibilityBridgeSerialized` — the `TBDHomeSerialized` idiom — so one
suite's open scope can never overlap another suite's accessibility read; the
parent carries a comment saying why, and a test asserts the switch is on inside
the scope and back to its prior value after it. Any future suite that touches
the bridge, or reads back behavior the bridge changes, nests there.
`scripts/test.sh --filter
"AccessibilityWindowGeometryGuardTests|ComposerAccessibilityIdentifierTests|CompletionOverlayPlacementTests|AccessibilityBridgeScopeTests"`
— 17 tests in 5 suites, 0 failures, three runs in a row (and the failure
reproduced locally before the fix, with the guard passing when run alone).

[composer a11y ids](https://cheapsteak.github.io/tbd/open/?worktree=54D89E98-AFA6-4B66-A626-6CE0E0692A5E)

## Summary by CodeRabbit

- **Accessibility**
  - Added stable accessibility identifiers for the message composer, text field, send control, notices, blocked states, completion menus, attachments, thumbnails, removal controls, and terminal-reveal actions.
  - Improved assistive technology navigation by exposing attachment and completion actions as distinct, identifiable controls.

- **Tests**
  - Added comprehensive accessibility coverage for composer controls, completion rows, attachments, and composer states.
  - Improved reliability of accessibility testing by safely managing shared accessibility settings and serialized test execution.

### DIFF
--- a/Sources/TBDApp/Panes/Transcript/Composer/AttachmentStrip.swift
+++ b/Sources/TBDApp/Panes/Transcript/Composer/AttachmentStrip.swift
@@ -40,6 +40,8 @@ struct AttachmentStrip: View {
                 .padding(.horizontal, 8)
                 .padding(.vertical, 6)
             }
+            .accessibilityElement(children: .contain)
+            .accessibilityIdentifier(ComposerAccessibility.attachments)
         }
     }
 
@@ -68,6 +70,8 @@ struct AttachmentStrip: View {
                 .buttonStyle(.plain)
                 .padding(2)
                 .accessibilityLabel("Remove image \(attachment.number)")
+                .accessibilityIdentifier(
+                    ComposerAccessibility.attachmentRemove(number: attachment.number))
             }
             Text(detached ? "not in message" : "#\(attachment.number)")
                 .font(.system(size: 9))
@@ -88,11 +92,26 @@ struct AttachmentStrip: View {
                 onFocusToken(attachment.number)
             }
         }
-        .accessibilityElement(children: .combine)
+        // `.contain` rather than `.combine`. Combining folds the remove button
+        // into this one element, and a button that is not its own element is a
+        // button neither a driver nor VoiceOver can press separately from the
+        // thumbnail it sits on. The group keeps the sentence it always had, and
+        // carries the thumbnail's own click as an explicit action instead of
+        // relying on the merge to expose it.
+        .accessibilityElement(children: .contain)
         .accessibilityLabel(
             detached
                 ? "Image \(attachment.number), not in message. Click to re-insert."
                 : "Image \(attachment.number), in message. Click to go to it.")
+        .accessibilityIdentifier(
+            ComposerAccessibility.attachmentItem(number: attachment.number))
+        .accessibilityAction {
+            if detached {
+                onReinsert(attachment.number)
+            } else {
+                onFocusToken(attachment.number)
+            }
+        }
     }
 }
 
--- a/Sources/TBDApp/Panes/Transcript/Composer/CompletionOverlayView.swift
+++ b/Sources/TBDApp/Panes/Transcript/Composer/CompletionOverlayView.swift
@@ -54,6 +54,10 @@ struct CompletionOverlayView: View {
             EmptyView()
         } else {
             content
+                // A container, so every row underneath stays individually
+                // addressable rather than collapsing into one element.
+                .accessibilityElement(children: .contain)
+                .accessibilityIdentifier(ComposerAccessibility.menu)
                 .frame(maxWidth: .infinity, alignment: .leading)
                 .background(VisualEffectView(material: .menu, blendingMode: .withinWindow))
                 .overlay(
@@ -188,6 +192,10 @@ private struct CompletionRowView: View {
         .accessibilityElement(children: .ignore)
         .accessibilityLabel(accessibilityLabel)
         .accessibilityAddTraits(isSelected ? [.isSelected] : [])
+        // Named by the command, not by position: the ranking reorders these
+        // rows as the query changes, so an index would name a different row
+        // from one keystroke to the next.
+        .accessibilityIdentifier(ComposerAccessibility.menuRow(command: row.command.name))
     }
 
     private var accessibilityLabel: String {
--- a/Sources/TBDApp/Panes/Transcript/Composer/ComposerAccessibility.swift
+++ b/Sources/TBDApp/Panes/Transcript/Composer/ComposerAccessibility.swift
@@ -1,0 +1,49 @@
+import Foundation
+
+/// The composer's accessibility identifiers, in one place.
+///
+/// An identifier is **for automation** — a GUI driver, or the offscreen hosting
+/// harness in the test suite, asking the accessibility tree for one control by a
+/// name that does not move when the copy does. That is exactly what a label is
+/// not: a label is the sentence a person hears, and it changes with the wording,
+/// the terminal's name, and the attachment's number. So the two live side by
+/// side here, and nothing reads a label to find a control.
+///
+/// Named constants rather than literals at the call sites for the ordinary
+/// reason: a driver and a test that both spell `composer.send` by hand agree
+/// only until one of them is edited.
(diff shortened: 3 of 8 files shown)


cheapsteak:
**CI fix pushed: d557a18b.**

The Test run on 3a39b14e ([34074933464](https://github.com/cheapsteak/tbd/actions/runs/34074933464), fast pass 2/2) failed in `AccessibilityWindowGeometryGuardTests` — "forwards a finite Accessibility position" and "forwards a finite Accessibility size" — a suite this branch does not touch.

Cause: this branch's new `enableAccessibilityBridge()` set `AXEnhancedUserInterface` on `NSApplication.shared` and deliberately left it on. That flag is process-global, all test targets compile into one process, and Swift Testing runs suites in parallel, so it was live under the geometry guard suite — which drives an `NSWindow` through the same accessibility setters and reads `window.frame` back. With enhanced UI on, AppKit no longer applies an accessibility-driven window move synchronously (it is the mode window managers switch off before moving anything), so the forwarded value is not in the frame when the test looks. Reproduced locally: both suites together fail exactly as CI did; the guard suite alone passes.

The guard itself is not weakened and is unchanged — the flag legitimately changes AppKit's behavior, so the fix is to stop leaving it on. The switch is now flipped only by `withAccessibilityBridge { }`, which records the prior value and restores it in a `defer`, and both suites nest under a new `@Suite(.serialized) enum AccessibilityBridgeSerialized` (the `TBDHomeSerialized` idiom) so one suite's open scope cannot overlap another suite's accessibility read. A new test asserts the switch is on inside the scope and restored after it.

Verification: `scripts/test.sh --filter "AccessibilityWindowGeometryGuardTests|ComposerAccessibilityIdentifierTests|CompletionOverlayPlacementTests|AccessibilityBridgeScopeTests"` — 17 tests in 5 suites, 0 failures, three consecutive runs. `swiftlint --strict` clean on the touched files. New CI run: [34076015063](https://github.com/cheapsteak/tbd/actions/runs/34076015063).

PR #832, opened 2026-09-07
DX: One offscreen GUI harness for TBDApp tests, and an env-gated render suite for the composer

cheapsteak:
Stacked on #831.

## Summary

Somebody writing a test for a SwiftUI view in this app — where did it land, can a driver reach it, what does it actually look like — has to mount that view in a real window and drive it. Three suites had each worked that out separately and kept a private copy: a borderless offscreen `NSWindow`, an `NSHostingView`, a pump that turns both the run loop and the cooperative pool, a bounded settle, a subview walk, a `cacheDisplay` capture. Each copy knew only what its own suite needed, none of them said what the technique proves and what it does not, and the fourth suite to need one would have written a fourth copy.

There is now one, `OffscreenHost`, with its contract written down, plus `ComposerHarness` for the composer's own fixtures. And the render harness that caught the completion list covering the words being typed (#829) — which was thrown away with the branch that produced it — is back, kept, and env-gated.

## Why this shape

No spec: this is test infrastructure. It adds no product behavior, no flag and no database column, and it changes no theory of the system — the three refactored suites make exactly the assertions they made before.

The decisions worth naming:

- **`Tests/TBDAppTests/Support/`, not `Tests/TestSupport/`.** That target depends on `TBDDaemonLib` and links into the daemon suites; putting AppKit and SwiftUI in it would drag them into every one of those for two AppKit-only callers.
- **The contract is half the deliverable.** `OffscreenHost`'s doc comment states what an offscreen mount proves and — the part a copied helper never carries — what it does not: AppKit-versus-SwiftUI compositing does not reproduce offscreen (measured: the placement suite's paint-order capture is pixel-for-pixel identical with the production `.zIndex` removed, so it guards an outcome and does not discriminate a modifier); nothing about events, since the window is never key; nothing about elapsed time, since every wait is bounded by a pump count rather than a clock.
- **Composer state comes from the terminal's own columns.** The harness builds a `Terminal` with `hibernatedAt` or `awaitingInputReason` set and runs `ComposerState.resolve` over it, rather than handing the view a state value directly, so a suite mounting a blocked composer is mounting what the daemon's record would actually produce.
- **The blank-render guard is the render suite's whole discipline.** A view that was never laid out captures as a large, perfectly plausible white rectangle. Every shot's luminance variance has to clear a named threshold before the file is written, and `WorkbenchSnapshotTests` gains the same check beside its existing PNG-size floor — a big PNG can still be a picture of nothing.
- **`TableTranscriptHarness` is left alone.** Its two hosting sites are an oracle that measures a row through `NSHostingController.sizeThatFits` with no window at all, and a ground-truth render pinned to that oracle's height. Neither is a mounted-in-a-window view, which is the only thing the host offers, and its env gate is untouched.

## What this PR does

- **`Tests/TBDAppTests/Support/OffscreenHost.swift`** — generic over `View`. Mounts at a caller-given size on an opaque, appearance-resolved ground; `pump()` / `settle(maxPumps:until:)` / `pumpSynchronously(times:spin:)`; `descendants` and `firstDescendant`; `flippedWindowFrame(_:)` so "above" reads as a smaller `y`; `accessibilityIdentifiers()` and `accessibilityTreeDescription()` over the walk moved out of the identifiers suite (by selector, because SwiftUI's nodes do not declare `NSAccessibilityProtocol`); `capture(scale:)` returning a `Capture` with `pixelFrame(of:)`, `color(x:y:)`, `fraction(in:matching:)`, `luminanceVariance()` and `writePNG(to:)`. `withDrawingAppearance` carries the trap that dynamic colours resolve once, at creation.
- **`Tests/TBDAppTests/Support/ComposerHarness.swift`** — isolated `AppState` on a `UserDefaults(suiteName:)` with teardown, a local worktree, running / parked / blocked terminal builders, a `prepare` hook that stages images into a `fencedScratchRoot` directory the harness removes, `openMenu(typing:)`, the text-view and completion-list lookups, and the two accessibility settles. Plus `TranscriptPreviewStrip`, a deliberately dull fixture conversation that stands in for the transcript above the composer.
- **The three suites move onto it**, assertions unchanged (see below).
- **`Tests/TBDAppTests/ComposerRenderHarness.swift`** — inert unless `TBD_COMPOSER_SHOTS_DIR` is set; writes `1-menu.png`, `2-attachment.png`, `3-not-running.png`, `4-blocked.png` at 2x.
- **`docs/test-doubles.md`** gains a "Harnesses for the app's own views" section: both entries, the env var, and the compositing caveat.

## Evidence & verification

`scripts/test.sh --filter "CompletionOverlayPlacementTests|ComposerAccessibilityIdentifierTests|WorkbenchSnapshotTests|MessageComposerViewTests|ComposerRenderHarness|AccessibilityBridgeScopeTests|AccessibilityWindowGeometryGuardTests"` — **46 tests in 8 suites, zero failures**, both with `TBD_COMPOSER_SHOTS_DIR` set (four PNGs written) and without it (no directory created, the render tests inert). `swiftlint --strict` over the repository: 0 violations in 1036 files.

Assertion counts before and after the refactor, from a grep of `#expect` / `#require` / `Issue.record` / `throw`:

- `ComposerAccessibilityIdentifierTests` — 10 → 10.
- `CompletionOverlayPlacementTests` — 18 → 17 assertions. The one that moved: `try #require(host.bitmapImageRepForCachingDisplay(...))` is now `OffscreenHostError.couldNotMakeBitmap`, thrown inside `capture()`, which names itself on the failure line.
- `WorkbenchSnapshotTests` — 4 → 3 raised failures, and strictly stronger. Its two `RenderError.couldNotMakePNG` throws became one `OffscreenHostError` inside the host; the new `RenderError.blankRender` variance check is added on top of the PNG-size floor it kept.

The four renders were inspected by eye, not merely produced: the completion list opens above the field with `please /comp` visible under it; the staged attachment shows its thumbnail, its `#1` caption and the `[Image #1]` token in the message; the parked session shows "Claude exited in this terminal. Sending will resume the session." over a **Resume claude** button; the blocked session shows the permission sentence with **Reveal Terminal** beside it and a disabled send.

[composer harness](https://cheapsteak.github.io/tbd/open/?worktree=73F947A5-C6BD-4A6A-8120-965909BE8733)

## Summary by CodeRabbit

- **Tests**
  - Improved coverage for composer placement, accessibility, rendering, attachments, blocked sessions, and inactive sessions.
  - Added offscreen rendering checks that capture screenshots and reject blank or flat images.
  - Updated workbench snapshot testing for more reliable layout, appearance, and rendering validation.
  - Standardized completion menu width checks.
  - Added lifecycle coverage for reliable offscreen test cleanup.

- **Documentation**
  - Expanded testing guidance for SwiftUI view harnesses, offscreen hosting, accessibility timing, and screenshot capture.

### DIFF
--- a/Sources/TBDApp/Panes/Transcript/Composer/CompletionOverlayView.swift
+++ b/Sources/TBDApp/Panes/Transcript/Composer/CompletionOverlayView.swift
@@ -27,6 +27,10 @@ struct CompletionOverlayView: View {
     /// so a pointer resting over the list cannot change what Return would take.
     @State private var hoveredIndex: Int?
 
+    /// The list's own width — the composer proposes it, rather than the list
+    /// proposing itself, so this is the one place that width is spelled.
+    static let width: CGFloat = 460
+
     static let rowHeight: CGFloat = 44
     /// The tallest the list ever gets: eight rows, and a scroller past that.
     static var maxHeight: CGFloat {
--- a/Sources/TBDApp/Panes/Transcript/Composer/MessageComposerView.swift
+++ b/Sources/TBDApp/Panes/Transcript/Composer/MessageComposerView.swift
@@ -105,7 +105,7 @@ struct MessageComposerView: View {
             // in the first place.
             if let controller, controller.isOpen {
                 Color.clear
-                    .frame(width: 460, height: 0)
+                    .frame(width: CompletionOverlayView.width, height: 0)
                     .overlay(alignment: .bottomLeading) {
                         CompletionOverlayView(
                             controller: controller,
@@ -113,7 +113,7 @@ struct MessageComposerView: View {
                             // Reached only from an explicit click; hover is
                             // handled inside the overlay and moves nothing.
                             onHighlight: { controller.moveTo(index: $0) })
-                        .frame(width: 460)
+                        .frame(width: CompletionOverlayView.width)
                     }
             }
         }
--- a/Tests/TBDAppTests/CompletionOverlayPlacementTests.swift
+++ b/Tests/TBDAppTests/CompletionOverlayPlacementTests.swift
@@ -10,202 +10,34 @@ import TBDShared
 /// difference between opening upward and opening downward over the text field
 /// is one edge name.
 ///
-/// So this suite mounts the real `MessageComposerView` in a real window, opens
-/// the menu by typing into the real text view, and measures the two frames in
-/// window coordinates. The first shipped guide anchored the overlay's top to the
+/// So this suite mounts the real `MessageComposerView` in a real window through
+/// `ComposerHarness`, opens the menu, and measures the two frames in window
+/// coordinates. The first shipped guide anchored the overlay's top to the
 /// composer's bottom — a list that covered the words being typed and ran off the
 /// bottom of the pane — and every assertion here fails against it.
 @MainActor
 @Suite("completion overlay placement")
 struct CompletionOverlayPlacementTests {
 
-    // MARK: - Fixtures
-
-    /// Enough commands that the list is taller than its eight-row cap, so the
-    /// cap is exercised rather than assumed.
-    private static let inventory = TerminalCompletionsResult(
-        commands: (0..<20).map {
-            CompletionCommand(
-                name: "compact\($0)", description: "Compact the conversation, take \($0)")
-        },
-        agents: [], freshness: .fresh, source: .probe)
-
-    private func makeAppState() -> (AppState, String) {
-        let suiteName = "CompletionOverlayPlacementTests-\(UUID().uuidString)"
-        let state = AppState(userDefaults: UserDefaults(suiteName: suiteName)!)
-        state.composerCompletionsFetcher = { _ in Self.inventory }
-        return (state, suiteName)
-    }
-
-    private func makeWorktree() -> LocalWorktree {
-        LocalWorktree(Worktree(
-            id: UUID(), repoID: UUID(), name: "wt", displayName: "WT", branch: "main",
-            path: "/tmp/wt", status: .active, tmuxServer: "test-server", location: .local))!
-    }
-
-    private func makeTerminal(worktreeID: UUID) -> Terminal {
-        Terminal(
-            id: UUID(), worktreeID: worktreeID, tmuxWindowID: "@1", tmuxPaneID: "%1",
-            kind: .claude)
-    }
-
-    // MARK: - The harness
-
-    /// A mounted composer, pinned to the bottom of an offscreen window with the
-    /// transcript's worth of empty space above it — the shape the pane gives it,
-    /// and the space the list has to be allowed to grow into.
-    private struct Mounted {
-        let window: NSWindow
-        let host: NSHostingView<AnyView>
-        let state: AppState
-        let suiteName: String
-    }
-
-    /// What stands above the composer in the mounted stack.
-    enum Sibling {
-        /// `Color.clear`. Enough to give the list a region to grow into, and all
-        /// the geometry assertions need.
-        case transparent
-        /// A real `NSViewRepresentable` that fills itself solid red — the shape
-        /// of the composer's actual neighbour, `TableTranscriptView`. Ordering a
-        /// SwiftUI overlay against a HOSTED AppKit view is precisely what the
-        /// pane's `.zIndex(1)` states, and a SwiftUI-only sibling cannot pose
-        /// that question: SwiftUI orders its own layers among themselves either
-        /// way.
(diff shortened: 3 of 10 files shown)


cheapsteak:
Review fix round applied on top of 0a367335:

- **40a6691d** (`refactor(composer)`) — finding 4 (`CompletionOverlayPlacementTests.swift` ~82 / `MessageComposerView.swift` ~108,116): hoisted the duplicated `460` into `CompletionOverlayView.width`; both production sites and the test now reference it.
- **e55ab61d** (`fix(tests)`) — findings 1, 2, 3, 5:
  - Finding 1 (`OffscreenHost.swift` ~167): `tearDown()` now actually releases the window. `orderOut` + `close()` alone still left it in `NSApp.windows` — measured directly: with `isReleasedWhenClosed = false`, AppKit only drops a window from that list as part of releasing it, so `close()` is a no-op there no matter how many times it runs. Fix flips `isReleasedWhenClosed = true` immediately before `close()`. Added the `NSApp.windows` assertion in `WorkbenchSnapshotTests` (which mounts 5 hosts/run); confirmed it goes RED against the naive `orderOut`-only teardown (5 issues, one per configuration) and GREEN with the real fix.
  - Finding 2 (`ComposerRenderHarness.swift` ~36,158): the blank-render guard now actually gates the write — throws before `writePNG` runs — matching what the doc always claimed. (Superseded my first pass, which had instead reworded the doc to match the old write-unconditionally behavior; the review gate caught that the doc's claim was the one worth making true.)
  - Finding 3: `minLuminanceVariance` moved to `OffscreenHostDefaults`; both suites reference it.
  - Finding 5 (`ComposerHarness.swift` ~54,96): removed the unexercised `size:`/`appearance:`/`inventory:` init params and `inventory(commandCount:)`'s parameter — no caller ever passed a non-default value. Also fixed a `UserDefaults` domain leak found while touching the init: a throwing `prepare` closure (reachable from the staged-attachment render test) left the domain registered with no harness instance around to tear it down; it's now removed before rethrowing.

Verification: `scripts/test.sh --filter "CompletionOverlayPlacementTests|ComposerAccessibilityIdentifierTests|WorkbenchSnapshotTests|MessageComposerViewTests|ComposerRenderHarness|OffscreenHost"` — 39 tests, 0 failures, both without `TBD_COMPOSER_SHOTS_DIR` and with it set (four PNGs written). `swiftlint --strict` on all touched files: 0 violations.

Also: the PR body's assertion-count line for `CompletionOverlayPlacementTests` should read **18 → 17**, not 17 → 16 (recounted with the same `#expect`/`#require`/`Issue.record`/`throw` grep against 6e20c144 vs. current — my width-hoist commit doesn't change the count, still 17 after).

Claude-Session: https://claude.ai/code/session_01E9SPoxoVoApcVmduLiBWik

cheapsteak:
Fixed the signal 11 that killed the whole-suite run on `e55ab61d` (fast parallel pass 2/2, right after the workbench snapshots wrote their PNGs).

**Cause.** `OffscreenHost.tearDown()` flipped `isReleasedWhenClosed = true` immediately before `close()`. That asks AppKit to send a `release` ARC never balanced, against a window the host still held a strong reference to — an over-release whose use-after-free lands later than the test that caused it, which is why narrow local runs stayed green.

**Fix (597c79aa).** The flag stays `false`, set once at creation. Teardown now does `orderOut` → `contentView = nil` → `close()` and then drops the host's own references to the window, the hosting view and the ground; those three moved to `Optional` storage behind the same non-optional accessors, so no call site changes and a use-after-teardown traps with a sentence instead of returning an empty capture.

**What I measured while fixing it**, because the old doc comment asserted the opposite and it is worth having on the record: an ordered-front `NSWindow` is *never* deallocated in this process — not even a bare one with no SwiftUI in it — so it stays in `NSApp.windows` regardless of `close()`. The flag flip "removed" it only by decrementing a count it did not own. The hosting view and the ground *do* deallocate on teardown, and they are the expensive half of a mount.

Consequently `WorkbenchSnapshotTests` no longer asserts `!NSApp.windows.contains(window)` while holding the window (that assertion could only pass through the over-release). Each render now asks a weak reference whether its hosted tree was released — the resource that actually accumulates, five per run.

**Tests.** New `OffscreenHostLifecycleTests`: the tree is released by teardown itself (host still alive), the window is left for AppKit to release (still in `NSApp.windows`), and teardown is idempotent. All three go red against the old teardown — hosting view and ground still mounted, and `NSApp.windows` empty of the window — and green against the fix. 18 tests across the six affected suites pass locally; CI has the verdict.

PR #834, opened 2026-09-07
Refactor: holder-ness is a transport property, not five soak flags

cheapsteak:
## Summary

Holder-ness is a transport property, not a separate opt-in. Someone who turned on auto-hibernation expects idle sessions to park whether they run under tmux or under a pty holder; someone who left orphan GC on expects dead holders' residue reclaimed alongside every other orphan. Until now each holder leg sat behind a flag of its own, so a fleet with auto-hibernate on and holder sessions in it silently left those sessions running, and a fleet with GC on left dead holders' sockets on disk, unless the operator had found and flipped five more switches. This PR removes the five switches. If a subsystem is on and the session is holder-backed, the holder leg runs.

The five flags, and what each collapses into:

- **`holder_hibernation_enabled`** – deleted. The idle sweep arms a holder row on `auto_hibernate_enabled`, exactly as it arms a tmux row. Manual "Hibernate now" takes no flag on any transport, as it never has on tmux. The merge-triggered park rides the same per-worktree setting as tmux. Wake of an unparked holder row is classified against the process table rather than refused. The limit-resume rail is served on holder and gates on exactly what gates it on tmux.
- **`gc_holder_rendezvous_enabled`** – derived from `gc_enabled`. The rendezvous file sweep runs under the same outer guard as the agent-worktree loop.
- **`gc_rowless_holders_enabled`** – derived from `gc_enabled`, the same way.
- **`reap_holder_children_enabled`** – gone; the reaper's holder leg runs on every pass, like its tmux leg.
- **`holder_row_reconcile_enabled`** – gone; the reconcile holder arm judges on every pass, like its tmux arm. A finished, resumable holder row is now parked rather than deleted, because every transport has a wake path.

## Why this shape

The flags were deliberately separate from `pty_holder_enabled` when they landed: that outer gate must be on for any holder row to exist, so it could not express "transport on, this destructive leg not yet", and the soak protocol introduced one destructive reclaimer at a time. That was a rollout argument, not an architectural one. Every leg has now landed and been exercised, so the separate switches became duplicate state that could disagree with the subsystem switch the user was actually looking at.

Two facts make the collapse safe:

- `Config.ptyHolderDefault` is still `false`. The transport is default-off, so nobody who has not deliberately opted into holder sessions is affected by any of this.
- Every one of these legs has been exercised on a live machine, and the transport itself remains the soak gate.

The decision was the maintainer's, made before implementation began, in these words:

> holder_hibernation_enabled shouldn't be a separate flag — should be the same flag as holder. everything gets auto hibernate if auto hibernate is on. if holder is active, and if auto hibernate is active, then holder should auto hibernate. are there other duplicate state flags that should instead be derived for holder? let's fix all of them please

That ruling is the human answer to the brainstorming question. The written record of it is [`docs/specs/2026-09-07-holder-flag-consolidation-design.md`](docs/specs/2026-09-07-holder-flag-consolidation-design.md): the rule, why the per-leg switches were a rollout device rather than a design property, why retiring them is safe, what remains the soak gate, and the alternative rejected. The transport spec's Rollout section names the rule beside the spawn-only flag and states that flipping that flag's default is the transport's only graduation event. The in-place rewrites of the two existing specs state the current design without revision narrative, per the repo rule.

Rejected: flipping each flag's default to on first, then deleting later. That two-step protects explicit opt-outs, but an opt-out from a holder leg while its subsystem stays on is exactly the disagreement this PR exists to remove, so preserving it would preserve the defect.

## What this PR does

- `Config` loses the five properties, their defaults, decode paths, and memberwise-init parameters. `ConfigRecord` stops reading the five columns and carries a one-sentence note naming them as vestigial. No migration is added or touched. The columns stay in the schema unread, because `DROP COLUMN` support is version-dependent and an unread column is harmless.
- The five `config.set…` RPC methods, their param structs, router cases and handlers, and the `holderHibernationEnabled` capability field are removed.
- `Terminal.isManuallyHibernatable` and `isAutoHibernationEligible` take no transport argument. `HibernationGate.Decision.holderTransport`, `WakeResult.holderTransport`, and the two `holderTransportRefusal` strings are gone.
- `OrphanGC` drops the inner guards on its two holder arms. `AgentReaper.sweepHolderChildren` takes no parameter. The reconcile pass no longer reads config for its holder arm or its disposal rule.
- CLI: `tbd config holder-hibernation` and `tbd gc holders|rowless-holders|holder-children|holder-rows` are removed.
- App: the "Hibernate pty-holder sessions" toggle, its help text, and the client setter are removed. The park menu model no longer takes the flag.
- Docs: `CLAUDE.md`'s reconciler bullets, the pty-holder transport spec, and the single-typist injection spec now state the derived rule.

## Assumptions

- An install that had explicitly written `0` or `1` to one of the five columns loses that as a distinct setting. Nothing reads the column, so an operator who had opted out of, say, the rowless-holder sweep now gets it whenever `gc_enabled` is on. This is intended: the holder leg follows its subsystem.
- The reaper's holder leg and the reconcile holder arm answer to no switch at all, not even `gc_enabled`, matching their tmux counterparts. An operator freezing reclaimers by setting `gc_enabled = 0` stops the two OrphanGC holder arms but not those two.
- An older `TBDApp` talking to this daemon still renders the removed toggle and, if flipped, gets an unknown-method error from `config.setHolderHibernationEnabled`. Daemon-newer-than-app skew is otherwise unaffected: the capability field decodes leniently.
- The reconcile arm parks a finished, resumable holder row even on a daemon that cannot currently start a holder because the `TBDHolder` helper is missing. The wake path already names that state and keeps the row parked, so the session id survives until the helper is back. Deleting the row instead would destroy a resumable session over a transient helper absence, which is the worse outcome.

## Evidence & verification

Tests cover each branch of the derived conditions:

- A holder row with `autoHibernateEnabled` on is auto-hibernation eligible; with it off it is `.featureDisabled`, the same answer a tmux row gets (`HibernationGateTests`, `HibernationGateMergeTests`, `HolderTmuxAssumptionGateTests`).
- Manual "Hibernate now" is permitted on a holder row with auto-hibernate explicitly off, exactly as on tmux (`ModelsTests`, `HolderTmuxAssumptionGateTests`).
- The rendezvous and rowless-holder sweeps run under `gc_enabled` and touch nothing with it off; dry-run still plans (`OrphanGCHolderRendezvousTests`, `OrphanGCRowlessHolderTests`).
- The reaper's holder leg reaps with nothing armed, and the reconcile arm judges and parks with no config gesture (`AgentReaperHolderLegTests`, `HolderReconcileInventoryTests`, `HolderStartupReconcileTests`).
- Unparked holder wake is classified, parked holder wake reaches the holder mechanic, and limit-resume is served on a holder row (`HolderTmuxAssumptionGateTests`).
- Negative registration tests pin that the CLI no longer offers the removed subcommands.

Each new test was verified red with its piece of the change reverted: with the inner GC guards, the reaper guard, the reconcile guard, and a transport refusal in the manual and auto rails all temporarily restored, the filtered run reported 144 tests with 68 issues, naming each derived-condition test; with sources restored, 501 tests in 34 suites passed.

Five test files were deleted because their only subject was the tri-state resolution of a flag that no longer exists: `HolderHibernationFlagTests`, `GCHolderRendezvousFlagTests`, `GCRowlessHoldersFlagTests`, `ReapHolderChildrenFlagTests`, `HolderRowReconcileFlagTests`. One test, the flag-on twin of `autoResumeStillActsOnTmuxRow`, was merged into its sibling.

## Rebase note: the hang-stacks flag is not one of the five

This branch is rebased onto `f125742b` (#760, the hang-stacks reclaimer), which landed while it was open. Five files conflicted — `CLAUDE.md`, `Sources/TBDCLI/Commands/GCCommands.swift`, `Sources/TBDDaemon/Server/RPCRouter.swift`, `Sources/TBDShared/Models.swift`, `Sources/TBDShared/RPCProtocol.swift` — and each was resolved as a union: main's `gcHangStacksEnabled` plumbing is kept in full (the `Config` property, the migration `20260830021944_config_gc_hang_stacks.sql`, `config.setGCHangStacksEnabled`, the `tbd gc hang-stacks` subcommand and its registration and parse tests), while only the five holder switches are dropped.

`gcHangStacksEnabled` is deliberately **not** one of the five, and the distinction is the one this PR turns on. The five hang off `pty_holder_enabled`, whose shipped default is `false`, so retiring them reaches only installs that deliberately opted into the holder transport. `gcHangStacksEnabled` hangs off `gc_enabled`, which resolves `gc_enabled ?? true` — retiring it would ship a file-deleting sweep to every install. `gcProfileDirsEnabled`, `gcOrphanProcessesEnabled` and `gcRetainedTranscriptsEnabled` are untouched for the same reason.

Two fresh-context review rounds at high effort found no logic bugs. Their remaining findings were a test that read the real process table for a fixture pid, a weak negative assertion, and stale prose; all are fixed. CI is the whole-suite verdict.

https://claude.ai/code/session_017hmJmH2ot6HzZcWJbwzsJb

[holder-flag-collapse](https://cheapsteak.github.io/tbd/open/?worktree=CBD29BD3-DA5F-4AFC-9B02-6268E792F821)

## Summary by CodeRabbit

* **Changes**
  * Holder sessions now follow the same hibernation, resume, reconciliation, and cleanup rules as other supported transports.
  * Garbage collection and orphan-child cleanup use shared controls rather than separate holder-specific switches.
  * Holder-session reconciliation runs consistently to repair stale or parked session state.

* **Removed**
  * Removed the experimental holder auto-hibernation setting from the app and CLI.
  * Removed dedicated holder GC, reaping, reconciliation, and hibernation configuration options, RPC controls, and CLI switches.


cheapsteak:
Maintainer sign-off on the flag-graduation MEDIUM.

The finding is correct that this departs from CLAUDE.md's flip-then-delete ladder, and correct that the justification quoted inside the PR description cannot clear a finding about that same description. Recording the decision here instead, as the review asked.

The repository owner reviewed the consequence and approved the departure: deleting the five holder flags outright, accepting that an install which had explicitly written `0` to one of them loses that as a distinct setting on upgrade.

The scope that made it acceptable, both halves verified against the tree rather than asserted:

- `Config.ptyHolderDefault` is `false` (`Sources/TBDShared/Models.swift`). A holder row cannot exist unless someone deliberately opted into the transport, so the only installs reachable at all are those that opted in AND separately dialled one of these five reclaimers off.
- Every one of the five legs was exercised on a live machine before this landed — park and wake on a real forked holder, both GC arms, the reaper leg and the reconcile arm.

For contrast, and because it is the distinction that does the work here: the same reasoning was applied to `gcHangStacksEnabled` on #760 and was **wrong** there, so that flag was restored and #760 merged with it intact. Its parent is `gcEnabled`, which resolves `gc_enabled ?? true` — it defaults ON, so removing the sub-flag would have shipped a destructive sweep to 100% of installs. `ptyHolderEnabled` defaulting OFF is precisely what makes the holder case different, and it is why this one is a narrow, opt-in-only change rather than a fleet-wide one.

The rationale is committed as `docs/specs/2026-09-07-holder-flag-consolidation-design.md`, including the rejected alternative of graduating the five one at a time.

PR #835, opened 2026-09-07
fix(tests): reclaim the fixture holders nothing was sweeping

cheapsteak:
## What's broken

A `TBDHolder` process started by a live test — plus the shell job it was
supervising — kept running on a developer machine for **24 hours** after the run
that created it had finished. It was parented to launchd, its worktree had been
archived and deleted, and it had no `terminal` row anywhere. Nothing was going to
end it; it was found by looking at the process table, and killed by hand.

Anyone who runs `scripts/test.sh` can leave one behind. The residue is
developer-machine only — it lives in `/tmp` under a test scratch root, so no user
of the product ever sees it — but it accumulates silently and each one holds a
pty and a looping shell open indefinitely.

## Why it happens

A holder is *built* to outlive whatever started it. `Sources/TBDHolder/Holder.swift`
calls `setsid()` and sets `SIGHUP`/`SIGPIPE` to `SIG_IGN` before it binds
anything, so a holder survives the daemon's death and keeps its session alive.
That is correct behaviour, and it applies just as well when the process that
spawned it is a test: the holder detaches, re-parents to launchd, and the job
keeps looping. Re-parenting is not the bug.

The bug is that **no reclaimer covered it**. Each of the obvious candidates misses
for its own reason:

- **`OrphanGC`.** Its `RowlessHolderCollector` enumerates `<base>/*.sock` for the
  *real* holders directory. A fixture's socket is under its own `/tmp` scratch
  root, so it is not in that listing at all. Separately, the orphan-process phase
  is behind `guard config.gcOrphanProcessesEnabled || dryRun`, whose default is
  `false` — so a `tbd gc sweep --dry-run` REAP line proves nothing here, because
  dry-run bypasses the gate that the real sweep stops at.
- **`AgentReaper`.** Its holder leg keys on `terminal` rows. The fixture builds
  `TBDDatabase(inMemory: true)`, so there has never been a row.
- **`scripts/test.sh`'s EXIT trap.** It *did* run — the scratch root was gone —
  and the holder still outlived it, because the trap only did `rm -rf`. Deleting
  a rendezvous socket does not kill the process listening on it; it just makes it
  unreachable.
- **The fixture's own teardown.** `HibernationFixture.tearDown` kills what its
  rows still name, which is sound and is a deliberate defence against pid reuse.
  But it reads those rows through a bounded gate, and on expiry
  `blockingTerminals` returns `[]` — its own comment concedes that a timeout
  "means the sweep above has nothing to kill by pid". A leak is the designed
  outcome of that branch, even though the fixture knew the pids all along.

## What this PR does

Three layers, one per way a fixture holder can be left behind. Each covers
exactly one, and the split is deliberate rather than defensive duplication.

**1. The fixture remembers what it spawned, identity-checked.**
`HibernationFixture` records each holder and job pid together with
`ProcessStartTime.startTime(pid:)` at spawn, and `tearDown` sweeps that list
after the by-rows pass — signalling a remembered pid only while the kernel still
reports the same start instant. That is the same answer `AgentReaper` gives to
pid reuse, and it is what makes remembering a pid safe on a box where the kernel
reissues them to other people's work. It follows `HolderProcessFixture.tearDown`,
which already killed by remembered handle; this fixture was the outlier. Covers:
**a teardown that ran and missed.**

**2. `sweep_holders` in `scripts/test.sh`'s `cleanup()`.** Before the `rm -rf`,
every `*/holders/*.sock` under the run root is resolved to its owning pid with
`lsof -t -a -U <that exact path>`; that pid's children are enumerated with
`pgrep -P` *first* (a holder's job re-parents to launchd the moment the holder
dies), then all of them are killed. Covers: **a run that died without reaching a
teardown**, where the socket is still on disk.

Two things about it are load-bearing and easy to get wrong:

- **By resource, never by pattern.** This machine runs live production holders
  and other agents' sessions; `pkill -f TBDHolder` would end them. The socket
  path is unique to the run and lives in a directory the run created, so "who has
  this exact file open" is the only safe discriminator.
- **`-a` is not optional.** lsof ORs its selection options, so `lsof -t -U <path>`
  means "every unix-socket holder on the machine, OR this path" — measured here
  at 300-odd pids, most of the process table. `-a` ANDs them. Getting this wrong
  does not fail loudly.

**3. `reclaim_abandoned_run_roots`, a start-of-run reconciler.** Before minting
its own root, a run gives every sibling `/tmp/tbd-test-home.*` that is both older
than 24 h **and** has no live owner the same two sweeps (tmux servers, then
holders), then removes it. Covers: **a SIGKILLed wrapper**, where no in-process
teardown can run at all. Because every run is a sweep, the machine heals as soon
as anybody tests again — no timer, no daemon, no scheduler.

The discriminator is two-part, and the second half is what licenses killing
anything. Age says nobody is *coming back*: no run lasts a day, and the nightly
stress harness mints one root per iteration rather than holding one. But age is a
proxy, and it reads one case exactly backwards — a run that **wedges** stops
writing, so its root ages just like an abandoned one while its holder, its tmux
servers and its `TBD_HOME` are all still in use. CI job timeouts close that for
every automated caller; a developer who starts a run by hand and walks away from
one that hangs is not covered by any of them, and `Tests/CLAUDE.md` documents
real hang classes.

So each run **claims** its root on the way in, writing its own pid and that pid's
kernel start time to `<root>/.run-owner`, and the reconciler skips any root whose
claim still checks out. The start time is what makes a day-old pid safe to read
at all — a bare number is one the kernel is free to have reissued, which is the
same reason `AgentReaper` and layer 1 above check it. Every arm fails keep-biased
except one: a root with **no** claim is reclaimed on age, because those predate
the mechanism and treating them as immortal would leak precisely what this
collects.

Alongside those, the fixture now writes the holder pid, the job pid and the
scratch root to stderr at spawn. The holder that prompted this could not be
diagnosed — its worktree was gone, so a crashed test process and a teardown that
ran and missed were indistinguishable. That line makes the next occurrence
answerable from the run log.

### Who reclaims an orphaned fixture holder

The doctrine's question
([`docs/specs/2026-08-15-named-reconciler-doctrine-design.md`](docs/specs/2026-08-15-named-reconciler-doctrine-design.md)),
answered per death mode:

- **Clean exit** — the fixture's own `tearDown`: the by-rows sweep, then
  `sweepRememberedProcesses` for anything the rows no longer name.
- **TERMed wrapper** — bash runs an EXIT trap on a fatal signal as well as a
  normal exit, so `cleanup` runs and `sweep_holders` kills whoever owns each
  surviving rendezvous socket.
- **SIGKILLed wrapper** (or a panic, or lost power) — nothing in-process runs, so
  the reclaim belongs to somebody else's run: `reclaim_abandoned_run_roots` is the
  named reconciler, sweeping and removing sibling roots older than a day.

The ordering between layers 1 and 2 is worth stating, because it looks redundant
and is not: a teardown that *ran* has already removed its scratch root, socket
included, so layer 2 has nothing left to key on. Layer 2 owns the case where no
teardown ran at all, which is exactly the case where the socket survives.

### Deliberately not touched

`OrphanGC` and `gc_orphan_processes_enabled` are untouched. That is a product
soak flag, not a cleanup lever, and this is developer-machine residue in `/tmp`
that no user ever has.

**No flag, and no spec.** Both rules in CLAUDE.md are about behaviour that
*ships*: a default-off flag is for a product feature that acts autonomously or
destroys state on a user's machine, and a brainstormed spec is for a change to
our theory of the system. Nothing here reaches a user or a daemon — it is
`scripts/`, `Tests/`, and a test fixture, and its entire blast radius is
`/tmp/tbd-test-home.*` on a developer's own box. A DB column gating a shell
script would have nowhere to be read from, and there is no gate worth having in
any case: the reconciler can only act on a directory nobody could still own. This
is a leak fix — the system already intended these processes to be reclaimed, and
nothing about that intent changes here.

## Assumptions

- **A wedged run is still a live process.** The liveness claim is what saves a
  hung run's root, and it identifies that run by pid plus kernel start time. A
  run whose wrapper process is gone while its holders are not — nothing produces
  that today, since the wrapper outlives everything it starts — would be
  reclaimed on age, which is the correct outcome anyway.
- **No run of `scripts/test.sh` lasts 24 hours** *and* leaves no claim. Either
  condition alone is now survivable; it takes both to be reclaimed while live.
- **`lsof` is on `PATH`.** Without it the holder sweep returns without doing
  anything rather than guessing; the leak reverts to today's behaviour rather
  than becoming a wrong kill.
- **A rendezvous socket under a run root belongs to that run.** The root name
  comes from `mktemp`, so no process outside the run can hold one open. This is
  what licenses killing the owner without any further check.

## Evidence & verification

**The repro.** The leak was found in the wild — pid 92079, a `TBDHolder` alive
for 24 hours, parented to launchd, its child still running
`HibernationFixture.termAwareJob`, with the worktree archived and zero terminal
rows. It was reproduced deliberately during this work: running the hibernation
suite against a copy of the fixture with `sweepRememberedProcesses()` removed left
a live holder and a live job behind after the run finished and its scratch root
had been deleted.

**Discriminating evidence.**

- `tearDownKillsWhatItSpawnedWhenNoRowNamesItAnyMore` (tier 3) spawns a real
  holder, deletes the row so the by-rows sweep has nothing to work from — the
  state an expired gate leaves teardown in, without sitting through a two-minute
  gate — tears down, and polls for both pids to disappear. **Verified failing
  with the change reverted**: 4 issues, both pids alive when the polls gave up.
- `scripts/test.test.sh` gains nineteen cases covering both new `test.sh`
  behaviours, seven of them mutation-checked against a deliberately weakened copy
  of the script, per the harness's own "edit a guard, edit its case" rule:
  a socket with a live owner has its process **and** its `pgrep -P` child killed;
  an orphaned socket with no owner is handled without failing the run; a sibling
  root minted today is left alone; one older than a day is reclaimed; an
  abandoned root has its holder killed *before* the root is removed; and the
  `lsof` invocation is pinned as ANDing its selectors, since the only way to
  drive that one is to perform the catastrophe.

  The liveness claim has five of its own: an aged root whose claimed pid is alive
  survives (and its owner is untouched); the same root is reclaimed once that pid
  dies; a **live pid claimed with somebody else's start time** — the recycled-pid
  shape no real process table can be asked for — does not save the root; an aged
  root with no claim is reclaimed; and a real run is shown claiming its own root
  with its own live pid, read from inside the run because the wrapper has exited
  by the time the case looks and neither half is derivable from outside. Dropping
  the check reclaims a live run's root on age alone; dropping the claim leaves the
  run with none.

  Two of those mutations were themselves worth the trouble. `-mmin "+0"` is not a
  widening of the age filter — BSD `find` rounds the age up to whole minutes and
  `+n` is a strict "more than", so a directory seconds old matches neither bound;
  the mutation drops the predicate instead. And the stand-in holder binds its
  socket with `python3` rather than `nc -lU`, because nc reads its stdin and exits
  the moment it sees EOF, so a binder started with its stdio detached died before
  anything could sweep it and the case passed for the wrong reason. Both cases now
  assert the stand-in is actually up before asserting it went down.

**Two silent failures caught while building this**, both worth knowing:

- `find /tmp -maxdepth 1 …` enumerates **nothing**. On darwin `/tmp` is a symlink
  to `/private/tmp`, and find will not descend into a symlinked starting point
  unless told to — it exits 0, reclaims nothing, and says nothing. `-H` is what
  fixes it, and it follows only the paths named on the command line, so a symlink
  planted *inside* the directory is still never followed.
- `lsof -t -U <path>` returns most of the process table, as described above.

**Runs.** `scripts/test.test.sh`: 457 assertions, all green, no build, no real
store touched (~50 s on a loaded developer box against a ~44 s baseline for the
same file before this change). `scripts/test.sh --no-parallel --filter
'HolderHibernationLiveTests|HolderFixtureScratchRootTests'`: 9 tests in 2 suites,
green. `swiftlint --strict`: 0 violations. Full-suite verification is CI's.

[fixture holder sweep](https://cheapsteak.github.io/tbd/open/?worktree=36B92341-FCE7-4BB4-B9B6-493942E4A067)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved test-run cleanup for abandoned processes, servers, sockets, and temporary files.
  - Added safeguards to limit cleanup to resources belonging to the relevant test run.
  - Test runs now reclaim abandoned temporary directories at startup after the configured age threshold.
  - Improved cleanup when terminal test records are removed before teardown completes.

- **Documentation**
  - Clarified test-resource lifecycle, teardown, cleanup, and abandoned-run recovery.
  - Updated regression-test coverage for resource sweeping and abandoned-directory reconciliation.
  - Added guidance for running the test suite with macOS Bash 3.2.

### DIFF
--- a/Tests/CLAUDE.md
+++ b/Tests/CLAUDE.md
@@ -99,12 +99,22 @@ one can be left behind, and each covers exactly one:
   socket with it, which is precisely why the layer above exists.
 - **A wrapper that is SIGKILLed**, where not even the EXIT trap runs, is
   reclaimed by the *next* run: `reclaim_abandoned_run_roots` gives every sibling
-  `/tmp/tbd-test-home.*` older than a day the same two sweeps and then removes
-  it. No run lasts a day — the nightly stress harness mints one root per
-  iteration — so age alone is a safe discriminator, and every run being a sweep
-  means the machine heals as soon as anybody tests again. It is the named
-  reconciler for fixture holders in the sense
+  `/tmp/tbd-test-home.*` that is both older than a day and has no live owner the
+  same two sweeps, then removes it. Every run being a sweep is what makes the
+  machine heal as soon as anybody tests again, with no timer and no daemon. It
+  is the named reconciler for fixture holders in the sense
   `docs/specs/2026-08-15-named-reconciler-doctrine-design.md` means it.
+
+  **Age is not the whole discriminator, and the missing half is the dangerous
+  one.** No run lasts a day, so age says nobody is coming back — but a run that
+  WEDGES stops writing, so its root ages exactly like an abandoned one while its
+  holder, its tmux servers and its `TBD_HOME` are all still in use. So every run
+  claims its own root on the way in, writing its pid and that pid's kernel start
+  time into `<root>/.run-owner`, and the reconciler skips any root whose claim
+  still checks out. Same identity check as everywhere else here: a pid alone
+  would be a number the kernel is free to reissue. A root with no claim in it
+  predates the mechanism and is reclaimed on age — the one arm that is
+  deliberately not keep-biased, because immortal roots are the leak.
 
 The tmux leg is the one with no teardown remedy: **tmux never unlinks its
 socket file when a server exits.** It unlinks a stale socket lazily instead, at
--- a/Tests/CLAUDE.md
+++ b/Tests/CLAUDE.md
@@ -72,6 +72,40 @@ orphan profile dirs, ~2.9k fake worktrees and ~7,100 dead tmux sockets
 accumulated that way before anyone noticed. Read `swift test …` below
 as `scripts/test.sh …`.
 
+**Two of the things a run leaves behind are processes, and no amount of path
+fencing touches either.** A tmux server outlives the socket that named it, and a
+`TBDHolder` outlives the test process that spawned it *by design* — it calls
+`setsid()` and ignores `SIGHUP` so it can survive the daemon's death
+(`Sources/TBDHolder/Holder.swift`), which means it survives a dead test process
+just as well, re-parents to launchd, and keeps its job running. Nothing in the
+product reclaims a fixture's holder: `OrphanGC`'s rowless-holder collector
+enumerates the real `~/tbd/holders`, and `AgentReaper`'s holder leg works from
+`terminal` rows an in-memory fixture database never had. One measured instance
+ran for 24 hours with its job still looping. Three layers cover the three ways
+one can be left behind, and each covers exactly one:
+
+- **A teardown that runs** kills what its rows name, then sweeps the pids it
+  remembers — refusing any whose kernel start time no longer matches the one
+  recorded at spawn. That identity check is what makes remembering a pid safe:
+  a pid is free the instant its corpse is collected, and on this box the next
+  process to take it is somebody else's. It is the same answer `AgentReaper`
+  gives to the same question.
+- **A run that dies without reaching a teardown** leaves its rendezvous sockets
+  behind, and `cleanup` kills whoever `lsof` says owns each one before the
+  `rm -rf`. **By resource, never by pattern**: the socket path is unique to the
+  run, while a `pkill` on the holder's name would end live production holders
+  and other agents' sessions on the same machine. Note the ordering this
+  implies — a teardown that ran and removed its scratch root has taken the
+  socket with it, which is precisely why the layer above exists.
+- **A wrapper that is SIGKILLed**, where not even the EXIT trap runs, is
+  reclaimed by the *next* run: `reclaim_abandoned_run_roots` gives every sibling
+  `/tmp/tbd-test-home.*` older than a day the same two sweeps and then removes
+  it. No run lasts a day — the nightly stress harness mints one root per
+  iteration — so age alone is a safe discriminator, and every run being a sweep
+  means the machine heals as soon as anybody tests again. It is the named
+  reconciler for fixture holders in the sense
+  `docs/specs/2026-08-15-named-reconciler-doctrine-design.md` means it.
+
 The tmux leg is the one with no teardown remedy: **tmux never unlinks its
 socket file when a server exits.** It unlinks a stale socket lazily instead, at
 bind time, when a new server claims that exact path — and every test mints a
@@ -156,7 +190,8 @@ The wrapper's own guards are regression-tested by `scripts/test.test.sh`, which
 runs in the `lint` CI job: it drives the symlink and ownership refusals on the
 fake home, the post-run mode-000 recheck, the fingerprint's seven arms (five
 roots, two of which are read twice), the tmux socket fence (its `sun_path`
-budget and its kill-server sweep) and the
+budget and its kill-server sweep), the holder sweep and the abandoned-run-root
+reconciler, and the
 shared-lock pin against fixture directories with a stub `swift`, so it takes
 ~11 s, builds nothing, and touches no real store. **Every case there is
 mutation-checked** — the assertion is shown going red against a deliberately
--- a/Tests/TBDDaemonLiveTests/Holder/HolderHibernationLiveTests.swift
+++ b/Tests/TBDDaemonLiveTests/Holder/HolderHibernationLiveTests.swift
@@ -430,6 +430,52 @@ struct HolderHibernationLiveTests {
         #expect(!FileManager.default.fileExists(atPath: fixture.launchArgvPath),
                 "the wake spawned a second holder instead of adopting the live one")
     }
+
+    /// Teardown's second pass, against the branch where the first one has
+    /// nothing to read.
+    ///
+    /// `tearDown` kills what the ROWS still name, which is the right default —
+    /// a park clears the pids off its row precisely because those processes are
+    /// gone, so a teardown working from a remembered list alone would signal
+    /// numbers the kernel has since reissued. But the read can come back empty:
+    /// its gate is bounded, and an expiry returns `[]`. On that branch the
+    /// by-rows sweep kills nothing, and what it failed to kill is a process
+    /// built to outlive this one — `TBDHolder` `setsid`s away and ignores
+    /// `SIGHUP` so it survives the daemon's death, which means it survives a
+    /// dead test process too, re-parents to launchd, and keeps its job running.
+    /// One did, for 24 hours.
+    ///
+    /// **The row is deleted rather than the gate starved**, because what that
+    /// branch means to `tearDown` is "there is no row here to read", and an
+    /// empty terminals table is exactly that — immediately, instead of after a
+    /// two-minute gate this test would otherwise have to sit through.
+    ///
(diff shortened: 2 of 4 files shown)


cheapsteak:
Addressed in 44cda821b — the finding was right, and mtime alone was the wrong
discriminator for exactly the reason named: a **wedged** run stops writing, so
its root ages like an abandoned one while everything inside it is still in use.
Age can say nobody is *coming back*; it cannot say nobody is *there*.

Rather than sign off on the tradeoff, the tightening suggested is in:

- Each run now **claims** its root on the way in — its pid, and that pid's
  kernel start time, written to `<root>/.run-owner` immediately after the
  `mktemp` and before anything slow (the build, the suite, the admission queue),
  so there is no window in which a live run looks unclaimed.
- `reclaim_abandoned_run_roots` skips any candidate whose claim still checks out:
  the pid is alive **and** `ps -o lstart=` still matches what was recorded. The
  start time is not decoration — a bare pid read a day later is a number the
  kernel is free to have reissued, which is the same check `AgentReaper` and the
  fixture's own remembered-pid sweep apply.
- Every arm fails keep-biased except one, and that one is deliberate: a root with
  **no** claim is reclaimed on age, because those were written by a wrapper
  predating this file and treating them as immortal would leak precisely what the
  reconciler exists to collect. It is pinned by its own case rather than left to
  be inferred.

Five new cases in `scripts/test.test.sh`, two of them mutation-checked, and they
close the coverage gap the finding names ("no test exercises an old-by-mtime-but-
still-live root"):

- an aged root whose claimed pid is **alive** survives, and its owner is
  untouched — plus the mutation, where dropping `run_root_is_live` reclaims that
  live run's root on age alone;
- the same root is reclaimed once that pid is dead and reaped;
- a **live pid claimed with somebody else's start time** — the recycled-pid shape
  no real process table can be asked to produce — does not save the root;
- an aged root with no claim is reclaimed;
- a real run is shown claiming its own root with its own live pid, observed from
  inside the run because the wrapper has exited by the time the case looks and
  neither half is derivable from outside — plus the mutation, where dropping the
  claim leaves the run with none.

The Assumptions section has been rewritten accordingly: it now takes both
conditions to reclaim a live root, not either one.

cheapsteak:
Note on the `test` job: it needed one re-run, for a flake that is live on `main`
and untouched by this diff. Both signatures seen here reproduce on `main`
independently — the clock-starvation pair `Tests/CLAUDE.md` names by name
("Archived search debounce" / "AppState appearance debounce", `EventDrivenTestClock:
no task suspended on the clock within 45.0 seconds`) failed on `main` at
[run 34154969661](https://github.com/cheapsteak/tbd/actions/runs/34154969661), and the
pty-exhaustion one (`TerminalTeardownReapTests`, `ENXIO` with 31 `/dev/ttys*`
entries) on `main` at
[run 34088233028](https://github.com/cheapsteak/tbd/actions/runs/34088233028). All of
them live in the fast parallel pass, which does not run the live target this PR
touches, and none of them reach `scripts/`. The branch has since been updated onto
current `main` and every check is green.

PR #840, opened 2026-09-08
Fix: scrolling up in a terminal replaced the draft with a past message

zionts:
## What's broken

Scroll up inside a TBD terminal running Claude Code and the transcript moves as
expected, but whatever you had typed in the composer is gone, replaced by a
prompt you sent earlier. The draft is not recoverable. It happens on ordinary
trackpad and wheel scrolling, and it happens often, because the conditions that
trigger it are common: a slow flick, a pointer resting near the bottom edge of
the terminal, or an agent that has not turned mouse reporting on yet.

## Why it happens

TBD does not let SwiftTerm report mouse events. It sets `allowMouseReporting`
to false on the attached view and installs its own event monitor that turns
wheel motion into tmux mouse buttons 4 and 5. That monitor, though, handed the
event back to AppKit in several cases: when the line delta was zero, when the
pointer sat in the slack strip below the last row or the sliver right of the
last column (the grid lookup returns nothing there, even though the pointer is
inside the view), and when tmux had not enabled mouse reporting.

A handed-back event reaches SwiftTerm's own `scrollWheel`. Because TBD turned
mouse reporting off, SwiftTerm falls back to its alternate-screen scrolling
behaviour and synthesizes Up and Down cursor keys. tmux always drives the outer
terminal on the alternate screen, so that fallback fires every time. The keys go
to tmux, tmux forwards them to the pane, and Claude Code's composer reads Up as
prompt-history recall. The scroll is delivered as typing.

Genuine wheel reports do not do this. When the monitor forwards buttons 4 and 5,
Claude Code scrolls its transcript and leaves the draft alone. The only bad path
is a wheel event reaching SwiftTerm's cursor-key fallback.

## What this PR does

The monitor now enforces one invariant: a wheel event whose pointer is inside a
visible, tmux-attached terminal, with no overlay on top, is never returned to
AppKit. It is forwarded to tmux as a mouse button, or it is swallowed. Only
events that belong to somebody else pass through.

The decision moves into a small pure helper, `TerminalWheelRouting`, so each
branch is testable without a view, a window, or a live terminal. It classifies
an event as pass-through, swallow, or forward with a button and a repeat count.
Motion is now read from both the line delta and the point-granularity delta,
because SwiftTerm's own override reads the latter, so an event with no line
delta would still have become cursor keys. Forwarded events get a grid position
clamped into the grid rather than a nil-checked one, which is what covers the
slack strip. `gridPosition(atWindowLocation:)` keeps its existing nil contract
for callers such as Cmd+Click that genuinely want "not on a cell"; the clamping
lives in a new sibling method.

Overriding `scrollWheel` in `TBDTerminalView` was the obvious alternative and is
not available: SwiftTerm declares it `public override`, not `open`, and the
fork is pinned. Fixing it in the fork would put the behaviour a version bump
away from anyone else's terminal expectations, so the routing rule stays in
TBD's own monitor, with a comment at the monitor explaining why no decline path
may be reintroduced.

Bug fix, no spec needed, per CLAUDE.md.

## Assumptions

Assumes tmux keeps running the outer terminal on the alternate screen. If that
stopped being true, SwiftTerm's fallback would scroll its own buffer instead of
sending cursor keys, and the bug would disappear on its own; the invariant here
stays correct either way.

## Evidence & verification

The repro that identified the mechanism ran in a fenced tmux against a
fake-model Claude 2.1.263. Sending three Up keys with a draft in the composer
produced a "History 1/1" banner and replaced the draft, exactly matching the
reported symptom. Sending genuine wheel mouse reports instead scrolled the
transcript and left the draft intact, which isolates the cursor-key fallback as
the only path that loses typing.

`Tests/TBDAppTests/TerminalWheelRoutingTests.swift` covers every branch: up
forwards button 4, down forwards button 5, the repeat count follows the delta
magnitude and never drops below one, a precise-scroll event with no line delta
still forwards, no motion at all passes through, an open overlay passes through,
a pointer outside passes through, and mouse reporting off swallows rather than
passing through. That last one is the regression test, and it asserts the
disposition is not pass-through rather than only that it is swallow. A further
test sweeps a grid of delta combinations and asserts that an event over the
terminal with any motion never yields pass-through. Grid clamping is covered for
an interior point against the unclamped math, the bottom slack strip, the right
sliver, a negative point, and a degenerate zero-size grid.

Test results on this branch:

```
✔ Test run with 14 tests in 1 suite passed        (--filter TerminalWheelRouting)
✔ Test run with 28 tests in 1 suite passed        (--filter TerminalPanelViewTests)
━ Test run with 3486 tests in 364 suites passed with 11 known issues   (TBDAppTests)
```

`swiftlint --strict` reports 0 violations in 1046 files. `scripts/swift-safe
build` completes clean.

Not verified locally: the change was not exercised in a running app on this
machine, and the daemon was left untouched. The reasoning above rests on the
fenced-tmux repro and the unit tests.

This PR was written by a Claude Code agent on Adam Zionts's behalf. The
maintainer is away, so Adam will carry any review feedback.

https://claude.ai/code/session_01YGCZ8b4bWY9Yqdpna639SB
[20260810-varying-wildfowl](https://cheapsteak.github.io/tbd/open/?worktree=9C81D24D-12C8-4945-88A7-431D2116B526)

### DIFF
--- a/Sources/TBDApp/Terminal/TBDTerminalView.swift
+++ b/Sources/TBDApp/Terminal/TBDTerminalView.swift
@@ -509,6 +509,26 @@ class TBDTerminalView: TerminalView {
         return (col, row)
     }
 
+    /// Converts a window-coordinate point to terminal grid (col, row), clamping
+    /// a point outside the grid to the nearest cell instead of failing.
+    ///
+    /// The view's bounds are a little larger than the grid: there is a slack
+    /// strip below the last row and a sliver right of the last column. A caller
+    /// that must act on every event landing inside the view — the scroll
+    /// monitor, which may not decline an event — needs a cell for those points
+    /// too. `gridPosition(atWindowLocation:)` keeps its `nil` for callers that
+    /// genuinely want "not on a cell".
+    func gridPositionClamped(atWindowLocation windowPoint: CGPoint) -> (col: Int, row: Int) {
+        let dims = terminalDimensions
+        return TerminalWheelRouting.clampedGrid(
+            localPoint: convert(windowPoint, from: nil),
+            boundsHeight: bounds.height,
+            cell: cellDimensions(),
+            cols: dims.cols,
+            rows: dims.rows
+        )
+    }
+
     // MARK: - Mouse click pass-through
     // Track mouseDown position to distinguish clicks from drags.
     // Single clicks are forwarded to tmux for pane switching;
--- a/Sources/TBDApp/Terminal/TerminalPanelView.swift
+++ b/Sources/TBDApp/Terminal/TerminalPanelView.swift
@@ -1553,6 +1553,20 @@ struct TerminalPanelRepresentable: NSViewRepresentable {
             // in TBDTerminalView. Instead, a local event monitor intercepts
             // scroll events and forwards them to tmux as mouse button presses.
             //
+            // The monitor may never hand an event back once it has decided the
+            // event is ours (pointer inside a visible terminal, no overlay).
+            // We turn SwiftTerm's own mouse reporting off (`allowMouseReporting
+            // = false` at attach), and its `scrollWheel` reacts to that by
+            // falling back to alternate-screen scrolling: it synthesizes Up /
+            // Down cursor keys. tmux drives the outer terminal on the alternate
+            // screen, so that fallback fires every time, and the keys land in
+            // the pane as if typed — Claude Code's composer reads Up as prompt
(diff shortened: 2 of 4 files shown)


cheapsteak (approved):

zionts:
Superseded by #899, a minimal fix for the mouse-reporting-off gap this PR reported. Closing on Adam's instruction by a Claude agent.

PR #845, opened 2026-09-08
Fix: the retire test tells a squatter from a proxy that never closed

cheapsteak:
## What's broken

One test in the model proxy's suite, `retire answers before the drain and a
successor binds the port`, has been reddening `main` for everybody. It failed on
four unrelated PRs on 2026-09-08 — runs 34258054051, 34271037853, 34275440153
and 34284111508 — always with the same line, and always passed on a rerun:

```
Expectation failed: (boundPort → nil) == (harness.port → 49260)
↳ the successor did not take the port in 4.018594166 seconds:
  bind(descriptor:ptr:bytes:): Address already in use) (errno: 48)
```

The cost is not the test: a red required check on somebody else's PR buys a
rerun, twenty minutes, and a small loss of trust in the suite.

## Why it happens

Not the handshake. The proxy's retire does exactly what the spec promises, and
the ordering is enforced by the shape of the code rather than by timing:
`ControlEndpoints.retire()` **awaits** `closeListener()` before it *returns* the
200 (`ControlEndpoints.swift:225-226`), that closure awaits the *listening*
channel's own close future (`ProxyServer.swift:110-113`, `:210-213`, against the
channel `start()` stored at `:169`), and the response is only written once
`await control.handle(…)` has returned (`ProxyServer.swift:392-406`). There is
no path on which a write outruns the close. `ProxyServer.start()` also sets
`SO_REUSEADDR` on the listener, so the connections carrying the retiring proxy's
in-flight streams do not block a successor. Both obvious suspects were checked
and neither holds.

What holds is that **a retire frees its port; it cannot reserve it.** macOS hands
TCP ephemeral ports out sequentially from a single global counter
(`net.inet.tcp.randomize_ports` is `0`) across the 16,384 numbers in
49152–65535, so the number a retire just freed comes back around after one wrap
of that range. Measured on an idle machine: 16,220 allocations, 0.15 s. In a
4,800-test parallel pass where the proxy suites, the holder suites and spawned
proxy binaries are all opening sockets beside each other, the freed number is
one an unrelated `connect()` or listener can be handed immediately — and
`SO_REUSEADDR` buys nothing against that squatter, because a socket that does
*not* set the option refuses a `SO_REUSEADDR` bind with `EADDRINUSE` for as long
as it lives. The test then spent its whole four-second allowance being refused by
a port that had been free the entire time, and blamed the handshake for it.

The test's own comment already suspected a squatter and answered it with a
bigger wall clock. Four seconds was simply less than a peer test's listener
lives.

## What this PR does

Splits the two claims the test was making through one racy observation, and
makes the surviving observation say *who* it saw.

- **The handshake claim is asked as a question with three answers, not one.** At
  the instant the 200 lands, the port is probed: refused, or `GET /tbd/status`
  through a session of its own (never the harness's, whose pooled keep-alive
  connection was accepted *before* the listener closed). A refusal is proof the
  listener is gone. A status naming this process, this test's home and this port
  can only come from the proxy that was supposed to have closed, and fails the
  test with that diagnosis — this is the assertion that still catches a retire
  which answers before it closes. A holder that is **positively identified** as
  somebody else — a proxy status naming a different pid, port or home, or a
  holder that had already let go by the time the status leg reached it, which
  the proxy under test cannot have been since it stays alive until teardown —
  is nobody's bug, and is recorded through `withKnownIssue` (`isIntermittent`)
  so the rate stays visible without reddening a check. A bare `connectRefused`
  could not tell those apart, and they have opposite verdicts.
- **The excused verdict is reached by recognising a stranger, never by failing
  to recognise anybody.** A holder that is listening and will not say who —
  a non-200, an undecodable body, a transport error on a port that is still
  listening — is a fourth verdict that **fails** the test, because a proxy that
  answered before it closed is still one of its explanations and an excused
  catch-all would be exactly the route by which that regression passed. The
  message carries what was seen, so a genuine stranger landing there can be
  reclassified on evidence rather than by default.
- **The successor's bind is only the demonstration that the port is takeable**,
  and it is skipped when the probe said the port is not free — no test can bind
  a port a stranger holds, and a proxy that never closed has already failed
  above. When the port *was* free the bind waits through the repo's one bounded
  poll, `pollUntilTrue`, for thirty seconds; a timeout now names the current
  holder with the same probe rather than reporting only "something is
  listening". `successor.stop()` between the two takes a named
  `withPhaseDeadline` of its own, like every other blocking step out here:
  outside "request"'s 60 seconds there is no outer bound left to catch a wedge
  in a listener close or an event-loop-group shutdown.
- **Skipping that wait also stops a cascade.** A thirty-second bind wait outlives
  the harness session's own thirty-second `timeoutIntervalForResource`, so the
  still-open stream read afterwards died of a `URLSession` timeout and turned one
  diagnosed failure into three. On the skip path the stream is read while it is
  still healthy.
- **`withProxy` gains `afterRequest`**, so the bind wait and the drain run under
  budgets of their own instead of sharing `body`'s 60-second "request" phase
  three ways. `withPhaseDeadline` races a plain timer that starts when "request"
  begins, so nested phases share that window rather than getting one each.
- **`waitUntil` is now `pollUntilTrue` underneath.** It used to sleep through
  `try? await Task.sleep`, which cannot tell expiry from cancellation and throws
  instantly once cancelled — a 20 ms poll becoming a busy spin over its whole
  remaining budget on a cooperative thread every other test in the pass is queued
  behind. Only its diagnostic stays local.
- **One five-second wait in `ProxyForwardingTests` is rebalanced rather than
  raised.** "a client that hangs up mid-stream still ends the relay and the tee"
  waits for the relay to release, and that wait is a *discriminating threshold*:
  it must expire before the script does, or a relay that only ended because the
  upstream ran out of events would pass. Five seconds is inside the scheduling
  latency fast pass 2 routinely shows, and it went red on run 34296850539 with
  the count still at 1. The threshold is fixed by moving the script out — 40
  events at 1 s instead of 250 ms — and the waits to 20 s, keeping the same
  ratio with four times the headroom. Nothing waits for the script to finish:
  the client cuts after the first event and the fake upstream stops rescheduling
  once its channel goes inactive.
- The post-bind `streamsInFlight == 1` assertion is gone, because a patient wait
  may legitimately outlive the six-second script. The "no stream was cut" claim
  is carried where it always really was: the in-flight count read at the moment
  the retire answered, and the six events that still arrive at the end.
- `TBDModelProxyTests` gains a dependency on `TestSupport` for `pollUntilTrue`.
  The alternative was a seventh hand-rolled poll loop, which is the thing that
  helper exists to end. Two comments that described the old dependency set are
  corrected.
- The spec gains two sentences of the same honesty: a retire frees the port
  rather than reserving it, and a port change's blast radius is reachable
  through that gap as well as through a window in which TBD was stopped.

## Assumptions

- Assumes the ephemeral range and the sequential allocator measured above are
  what CI runs on. If a future macOS randomises TCP ephemeral ports, collisions
  get rarer, not commoner, and nothing here breaks.
- Assumes the probe's fifteen seconds — the same budget `withProxy` gives its own
  requests — is enough for a live proxy on loopback to answer. It is sized
  generously on purpose, because "a stranger" is the verdict that *excuses* a
  failure and a probe that gave up early would report a proxy which never closed
  as somebody else's problem.
- Assumes a squatter releases the port inside thirty seconds when the port was
  free at probe time. If it is ever exceeded the failure message now names the
  holder.
- **Not fixed here, and worth its own decision:** the same squatter can cost a
  real installation its port. `ModelProxySupervisor.recover(from: .bindFailed)`
  mints a fresh port on the *first* refusal when nothing adoptable answers, so a
  transient holder in the gap a retire opens changes the port permanently, and
  every session already spawned against the old one loses its proxy for the rest
  of its life (Claude reads `ANTHROPIC_BASE_URL` once). A bounded retry of the
  same port before minting would close that, and it is a change to documented
  failure semantics rather than a test fix.

## Evidence & verification

- **The repro.** The four runs above, all on the same assertion, all green on
  rerun. Attempt 1 of run 34275440153 shows the test failing after 6.045 s with
  the bind refused for 4.13 s of it, while the surrounding log shows holder, PTY
  and app suites running beside it.
- **The failure that forced the discriminating probe.** Run 34296850539 attempt
  2: the connect at the instant of the 200 was *answered*, and the port then
  refused a `SO_REUSEADDR` bind for the whole thirty seconds with "something is
  listening on it". That is consistent with both explanations, and the test as
  written could not say which — which is the gap this PR closes. The code
  reading above rules out the proxy; the remaining in-process TCP listeners in
  fast pass 2 are the proxy suites' own (`FakeUpstream.swift:90`,
  `ProxyForwardingTests.swift:1103`), which the recursive `.serialized` on
  `ModelProxySuites` keeps from running beside this one, and the real proxy
  binaries `ProxyBinaryTests` spawns, which are separate processes and outlive
  their tests. Every one of those now arrives named: a different pid, or a
  different home, or no proxy status at all.
- **Discriminating evidence for the diagnosis**, measured locally on Darwin
  25.1.0 with plain sockets:
  - `net.inet.tcp.randomize_ports: 0`, `portrange.first/last = 49152/65535`; eight
    consecutive `bind(0)` calls returned 60314…60321 — the allocator is a
    sequential counter, and 49260 in the failure is near the bottom of the range
    it wraps to.
  - A freed ephemeral port was re-handed to a new socket after 16,220
    allocations, 0.15 s.
  - A listener that closes while an accepted child of it is still ESTABLISHED
    **does** let a `SO_REUSEADDR` successor bind the same port — so the retiring
    proxy's in-flight streams are not the blocker, and that option is refuted
    rather than assumed.
  - A plain socket with no `SO_REUSEADDR` holding the port **does** refuse a
    `SO_REUSEADDR` bind: `Address already in use (errno 48)`. That is the
    squatter, and it is the exact error CI reported.
- **Independent evidence that `closeListener` really closes.** Two tests in
  `ProxyBinaryTests` retire a *separately spawned* proxy binary and poll
  `connectRefused` on its port until it holds (`:539`, `:637`). Both passed in
  the same run whose in-process connect was answered.
- **The test still fails on a proxy that answers before it closes.** The probe is
  evaluated on the answer, not on the bind, and a retire that wrote its 200 and
  closed afterwards is still accepting there — it now answers its own
  `/tbd/status` and is named as the culprit rather than excused as a squatter.
- `swiftlint --strict` over the whole tree: 0 violations. Tests verified on CI
  rather than locally, per this machine's build rules.

https://claude.ai/code/session_01TggE2eESL3BcvSgFcLZkpk

[proxy retire successor bind](https://cheapsteak.github.io/tbd/open/?worktree=C968D8D1-3719-4306-8F84-B9462E443595)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved proxy transition testing to verify that retired listeners stop accepting connections immediately.
  - Increased resilience when starting successor proxies, with cancellable polling and clearer timeout diagnostics.
  - Improved reliability of long-running stream, relay-release, and tee-completion tests.

- **Documentation**
  - Clarified port availability during proxy retirement, including potential conflicts when another socket claims the freed port.
  - Documented control-endpoint behavior when successor binding fails during the transition.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -407,6 +407,11 @@ let package = Package(
             dependencies: [
                 "TBDModelProxy",
                 "TBDShared",
+                // `pollUntilTrue`, the repo's one bounded poll. It brings
+                // `TBDDaemonLib` in behind it, which nothing else here needs;
+                // the alternative is a seventh hand-rolled poll loop, which is
+                // the thing that helper exists to end.
+                "TestSupport",
                 .product(name: "NIO", package: "swift-nio"),
                 .product(name: "NIOPosix", package: "swift-nio"),
                 .product(name: "NIOHTTP1", package: "swift-nio"),
--- a/Tests/TBDModelProxyTests/ProxyControlTests.swift
+++ b/Tests/TBDModelProxyTests/ProxyControlTests.swift
@@ -1,6 +1,7 @@
 import Foundation
 import NIOHTTP1
 import NIOCore
+import TestSupport
 import Testing
 
 @testable import TBDModelProxy
@@ -114,9 +115,10 @@ extension ModelProxySuites {
             // of whatever turn is still running.
             let retired = ProxyFlagBox()
             // Six events a second apart: the stream has to still be running
-            // when the successor binds, and the successor's bind is allowed to
-            // wait out a squatter (see below), so the script's length is the
-            // budget both of those come out of.
+            // when the retire answers, which is what the in-flight count and
+            // the untouched event stream below read. The successor's wait for
+            // the port is deliberately not paid for out of this script's
+            // length; see the wait itself.
             let ticks = (1...6).map { index in
                 (delayMs: 1000, bytes: Array("event: tick\ndata: {\"n\":\(index)}\n\n".utf8))
             }
@@ -150,9 +152,23 @@ extension ModelProxySuites {
                 #expect(harness.server.streamsInFlight == 1, "the stream was cut by the retire")
                 #expect(!retired.value, "the process was handed over before the drain finished")
 
+                // The claim the handshake makes, asserted where no race can
+                // turn it: at the instant the answer lands, a connect to the
+                // port must be refused. Only a listening socket answers one, so
+                // a refusal is proof the listener is gone — and a proxy that
+                // wrote its 200 first and closed afterwards is still accepting
+                // here, however its successor's bind goes. The bind below
+                // cannot carry this claim on its own, because a retire frees
+                // the port and cannot reserve it.
+                #expect(
+                    connectRefused(port: harness.port),
+                    "the port still accepted a connection when the retire answered")
+
                 // The successor takes the port while the first proxy is still
-                // delivering. This is what `SO_REUSEADDR` on the listener buys: the
-                // connections carrying the in-flight streams still hold the port.
+                // delivering. This is what `SO_REUSEADDR` on the listener buys:
+                // the connections carrying the in-flight streams still hold the
+                // port, and a bind those refused would fail on the first
+                // attempt and on every one after it.
                 let successor = ProxyServer(
                     port: harness.port, routes: harness.routes, tee: nil,
                     status: {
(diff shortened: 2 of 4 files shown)


cheapsteak:
Status: `test` is green on this head; `claude-review` has failed 13 consecutive times on two heads with the same infrastructure error (`Claude Code native binary not found at /home/runner/.local/bin/claude` right after the action prints "successfully installed"). This is the upstream race in `anthropics/claude-code-action` v1.0.218's launcher step for Claude Code 2.1.265 (anthropics/claude-code-action#1817); a run on the same runner image passed at 22:12Z and nothing in this PR touches the workflow, so no verdict is possible until upstream ships a fix or the workflow pins the action back to v1.0.217 (`9c5ddab2e6d17b83ea679153b31f1d5f023cf636`). Pinning is a change to the merge gate under `pull_request_target`, so it is left to a maintainer. The change here is test-only.

PR #856, opened 2026-09-13
Fix: Trackpad scrolling in a fullscreen Claude tab stops firing Up/Down arrow keys into the session

cheapsteak:
## What's broken

Scrolling with a trackpad over a Claude Code tab that runs in fullscreen mode behaves like pressing the arrow keys. Scrolling up cycles through previously typed prompts instead of showing older transcript, and scrolling down opens the agents view. The transcript does move as well, so the two behaviors interleave and the tab is close to unusable. Anyone on the default tmux-subprocess transport with a trackpad or Magic Mouse hits it on every scroll; a classic wheel mouse does not, because it never produces the event shape described below.

## Why it happens

On the tmux-subprocess transport, a wheel event over a terminal is meant to be claimed by the panel's app-wide scroll monitor, which turns it into mouse wheel reports for tmux and stops it from reaching the terminal view. The monitor decided whether to bother based on the event's `deltaY`, and declined anything that read as zero.

On a trackpad, AppKit delivers many events where `deltaY` is exactly 0.0 while `scrollingDeltaY` still carries a few pixels of motion. Those slipped past the monitor and reached SwiftTerm's own `scrollWheel`. That handler accumulates `scrollingDeltaY` and, because the tmux client sits on the alternate screen and the panel disables mouse reporting on the view, it converts every cell height of accumulated pixels into an Up or Down arrow key. The session received real wheel reports from the monitor plus stray arrow keys from the emulator.

This was measured in the running app with temporary logging: for every gesture, events logged as `deltaY=0.0 scrollingDeltaY=5.0` passed through, and after about sixteen accumulated pixels the app sent `ESC O A` or `ESC O B` with a call stack ending in SwiftTerm's `sendKeyUp` / `sendKeyDown`. Claude Code itself and tmux were both ruled out first: the real CLI fed raw wheel bytes directly, and through a real attached tmux client, scrolls its transcript correctly and only opens history on a genuine Up arrow.

## What this PR does

On the tmux-subprocess transport, the monitor now claims every wheel event over a mouse-reporting terminal. A non-zero delta forwards exactly the number of wheel reports it did before, so scroll feel is unchanged. A zero delta is claimed and dropped, so it can no longer fall through to the emulator's alternate-screen fallback. An in-bounds point that lands in the sub-cell remainder strip at the view's bottom or right edge, where the grid lookup has no cell, is claimed and dropped too. The decision lives in a small pure helper, `TerminalPanelRepresentable.Coordinator.wheelReports`, following the `clickPassthroughActive` precedent, so it can be unit tested without AppKit events.

Dropping the sub-line remnants rather than accumulating them keeps the change minimal; the previous behavior already ignored them from the monitor's side, and the bug was only that the emulator was accumulating them instead.

**Scope.** The control-mode and pty-holder transports, both behind default-off flags, install no scroll monitor at all and keep mouse reporting off on the same view, so every wheel event on those paths still reaches the emulator's fallback. This PR documents that boundary in the code and leaves the fix to #857, since those paths send input straight to a pty rather than through tmux and need their own design.

## Evidence & verification

- Repro before the fix: instrumented build, one trackpad scroll over a fullscreen Claude tab on the tmux transport, log shows the pass-through events and the arrow-key sends with SwiftTerm call stacks.
- Unit tests in `Tests/TBDAppTests/ScrollWheelReportsTests.swift` cover each branch of the helper; the zero-delta case fails against the old logic, which returned the event unclaimed.
- Live verification: relaunched the app from the fix build and scrolled the same tab; the transcript scrolls and no history or agents view opens.
- Swift Testing suites could not run on the authoring machine (Command Line Tools only, no `Testing` module), so CI is the test run for this PR.

https://claude.ai/code/session_011EScFwA3EN9NER22bU5eVb


cheapsteak:
Addressed the review in the second commit: the monitor comment and helper doc now scope the guarantee to the tmux-subprocess transport and name the control-mode and holder paths as still uncovered (tracked in #857), and an in-bounds point with no grid cell is now claimed and dropped instead of passing through. PR description updated to match.

PR #859, opened 2026-09-16
fix: preserve replacement window on tmux id reuse

zionts:
## What's broken

Waking a hibernated terminal can fail after its tmux server disappears. When the recreated server reuses the terminal's previously persisted numeric window ID, wake stages a replacement window, deletes that same window during stale-window cleanup, and then fails to respawn the agent into it.

## Why it happens

The persisted tmux window ID is a coordinate, not a durable identity. The recreation path correctly creates a replacement before deleting stale state, but it did not account for tmux assigning the replacement the same textual ID as the dead pre-restart window. Bootstrap cleanup had the same hazard when its reported ID matched the replacement.

## What this PR does

- Skips stale-window cleanup when the old ID equals the freshly created replacement ID.
- Skips bootstrap-window cleanup when the bootstrap ID equals the replacement ID.
- Adds a regression for ID reuse that requires wake success, persisted replacement coordinates, a replacement respawn, and no kill of that coordinate.
- Retains the existing distinct-ID cleanup coverage.

## Evidence & verification

- Reproduced by waking a parked terminal whose tmux server no longer existed; tmux reused the persisted window coordinate for the staged replacement, cleanup killed it, and `respawn-window` failed.
- Installed the patched build and repeated that wake: it returned `woken=true`, cleared the parked markers, preserved a server-qualified live replacement (`dead=0`), and resumed the agent in the replacement pane.
- `PATH="$HOME/.swiftly/bin:$PATH" scripts/swift-safe build` passes.
- `git diff --check` passes.
- The focused regression is present, but this host cannot execute package tests: Swift 6.2.1 lacks `Testing`; Swiftly 6.3.2 reaches the test targets but the existing app test suite cannot import `XCTest` because only Command Line Tools are installed.
- `swiftlint --strict` is also blocked on this host because SourceKitten cannot load `sourcekitdInProc.framework` without a full Xcode toolchain.


cheapsteak (approved):

zionts:
Rebased onto upstream/main and pushed. Addressed both review Mediums:
- Same kill-by-id shape guarded in `WorktreeLifecycle+PreSession` and `WorktreeLifecycle+Create` (defense in depth; within one live tmux incarnation ids are unique, so this mirrors the shape rather than a confirmed reachable hazard there too).
- Added a `dryRunEnsureServerWindowID` test seam and a new test covering the bootstrap-window guard branch, which the previous test structurally couldn't exercise.

No local build was run this pass (machine load); relying on CI for this push. Will loop with the review bot from here.

PR #864, opened 2026-09-16
Fix: Opening the transcript beside a pty-holder terminal no longer replaces it with a "can't display yet" placard

cheapsteak:
## What's broken

With a session on the pty-holder transport (`pty_holder_enabled` on), opening the transcript pane next to the terminal replaces the terminal with the notice "This session runs on the pty-holder transport, which TBD can't display yet. The session is unaffected and keeps running." Closing the transcript brings the terminal back. The same happens on any layout change that turns the tab from a single pane into a split or back. The session really is unaffected, but the copy is wrong twice over: TBD does render holder sessions, and this one was rendering a moment earlier.

## Why it happens

Adding a pane changes the tab's layout node from a single pane to a split. `SplitLayoutView` renders those as different view types, so SwiftUI tears down the terminal's panel and builds a new one for the same terminal in one update.

The old panel hands its session back asynchronously: it stops its reader, serializes the screen, then sends `pane.detach`, and only that RPC releases the daemon's viewer claim. The new panel attaches immediately. The daemon's `HolderRegistry.beginAttach` correctly refuses an attach while a viewer claim stands (two readers on one pty would steal bytes), and the refusal fell into the same code path as "unsupported transport", which paints the placard with no retry. A moment later the old detach lands and the daemon resumes draining the session, which is why closing the transcript, a second remount, attaches fine.

The tmux path never showed this because tmux tolerates overlapping clients on a window and the view-session reclaim handles the overlap. A holder session has exactly one owner.

Evidence from the app's bridge debug log: every layout toggle logs a panel `cleanup`, and only every second toggle is followed by the `holder resize` line a successful attach prints.

## What this PR does

App only; the daemon is untouched and its refusal stays as it is.

- **Successor waits for predecessor, deterministically.** A new `HolderHandbackLedger` (one instance on `AppState`) records each in-flight handback task by terminal ID. `detachHolderSession` registers its task; `startHolderClient` awaits the ledger for its terminal before calling attach, then re-checks teardown. Once the detach RPC has returned, `HolderRegistry.takeBackFromViewer` has already resumed the daemon's reader and cleared the claim, so the attach that follows cannot race it. No timing budget and no matching on error strings, which is why this was chosen over a bounded retry on `attachedToViewer`.
- **Truthful copy for a real attach failure.** A refused attach, a refused `attach.ready`, or a panel whose terminal is not loaded now paints "TBD couldn't attach to this session's terminal. The session is unaffected and keeps running. Close and reopen the tab to try again." through one helper that also logs `category=holderAttachFailed` with the reason. The old `handleUnsupportedTransport` helper lost all callers and is removed. `transportPreparationNotice(for:)` and `holderTransportMessage` no longer have a production caller; they are kept for now because existing tests exercise them and a comment in `startTmuxClient` cites the decision function. Sweeping them is a follow-up.

Keeping view identity stable across the pane-to-split change would avoid the remount altogether, but it touches every pane type; the wait is the smaller change and also covers any other cause of two panels overlapping on one terminal.

## Assumptions

- Assumes `pane.detach` returns only after the daemon has re-adopted the session and cleared the viewer claim (true today in `takeBackFromViewer`). If the daemon ever cleared the claim after replying, the successor's attach could be refused again.
- Assumes both panels for a terminal live in the same app process and share `AppState`. Two app instances on one daemon are outside this fix, as before.

## Evidence & verification

- Repro before the fix: with a holder-transport Claude tab, open the transcript pane. The terminal is replaced by the placard; close the transcript and it returns. Bridge log shows cleanup without a following holder resize on the failing toggles.
- `Tests/TBDAppTests/HolderHandbackLedgerTests.swift` (new, 6 tests): returns immediately with nothing registered; blocks until a gated task finishes; clears the entry; a newer registration survives an older task finishing; a registration made mid-wait is also awaited; terminals are independent.
- `Tests/TBDAppTests/TerminalHolderTransportGateTests.swift`: new test drives the coordinator with a held handback on the ledger and asserts zero attach calls while it is held and exactly one after release, with no placard; a second test asserts no wait when nothing is in flight. The two existing refusal tests now expect the attach-failed copy and assert the transport copy is not fed, so they fail on the old code.
- Swift Testing cannot run on this machine (Command Line Tools only, no `Testing` module), so the CI `test` job is the verdict for the tests. `scripts/swift-safe build --product TBDApp` succeeds. swiftlint is not installed locally; the `lint` job covers it.
- Live verification: pending, the branch build is being launched to toggle the transcript on a holder tab; result will be noted in a comment.


cheapsteak:
Live verification on the reporting machine, branch build of this PR running as the app:

- The remount path now attaches. After a layout change on a holder tab, the bridge log shows the panel cleanup followed by the successor's holder resize (13:18:03 local), with no placard.
- Right after the app relaunch (13:17:30), the same tabs did land on the new attach-failed placard, and the logged reason was "a viewer owns the pty" for the previous app instance's claim. That is a separate daemon defect: the daemon never records the app's process identity in day-first locales, so it never reclaims a dead app's sessions. Fixed in #865. The two are independent; this PR's wait covers the same-process remount only, as its description says.

PR #870, opened 2026-09-16
test: drive GitManagerTimeoutTests on EventDrivenTestClock so the deadline arms without a polled handshake

cheapsteak:
## What's broken

`GitManagerTimeoutTests.subprocessTimeoutThrowsGitTimeoutError` (and, by the same mechanism, `timeoutThrowsPromptlyWhenGrandchildHoldsPipeOpen`) hangs to the suite's 60-second limit under CPU load. The failure signature is a time-limit trip at the `advanceWhenSuspended` line plus `an error was expected but none was thrown and "" was returned` at the `#expect(throws:)` line. The nightly flake-stress job, which runs this suite's target ten times under one spinner per core, has reproduced it on most nights: 2/10, 1/10, 1/10, 2/10 and 1/10 failures across the last five targeted nightly runs posted to #503.

## Why it happens

The suite virtualises the subprocess deadline through the injected clock, and the deadline armer in `runBoundedProcess` is an unstructured `Task` that has to be given a thread before it reaches `clock.sleep`. The tests waited for that arming with `TestClock`'s `advanceWhenSuspended`, which can only observe a registered sleeper by polling `checkSuspension()`, and every probe opens with a `megaYield` of twenty serially awaited background-QoS tasks. Under CPU saturation macOS starves background QoS, so the probes stall and the guard never sees the arming. Meanwhile the child is a real `/bin/sleep 30`: once it finishes on its own, the completion path reports success (the authoritative wall-clock check compares against a 600 s timeout), the call returns `""`, and the test that was waiting to advance virtual time has nothing left to time out. That is exactly the "no error thrown, empty string returned" half of the signature; the time-limit half is the polled wait still wedged when the limit fires.

`docs/specs/2026-08-11-event-driven-test-clock-design.md` names this mechanism, and #503 proposed the megaYield-free clock as the cheapest experiment.

## What this PR does

Migrates the suite from `TestClock` to `EventDrivenTestClock`. The clock signals a parked waiter from inside the same critical section that registers the sleeper, so there is no probe to starve: the wait ends the instant the deadline task reaches its sleep, however long the scheduler took to get it there.

- The two timeout-path tests use the strict `requireAdvanceWhenArmed`. A missed arming throws the named `NoSleeperArmed` diagnostic before virtual time moves, instead of advancing an empty ledger and then stalling on the real child (30 s, or 120 s and a bare time limit for the grandchild test). A `defer { call.cancel() }` is inert once the call has resolved and, on that thrown path, kills the child through the cancellation relay rather than leaving it to run out.
- The four happy-path tests keep a never-advanced clock, unchanged in shape.
- The arming guard stays at its 45 s default rather than `TestDeadlines.saturatedPass` (90 s): this is tier 3, run `--no-parallel` on an idle machine, and a guard past the 60 s limit could never deliver its diagnostic. The suite doc records why.
- The explicit `.timeLimit(.minutes(1))` stays, for the reasons the suite doc gives (it is the regression detector for the two drain-regression proofs). No assertion is weakened and no drain proof is touched.
- `Tests/CLAUDE.md` adds the suite to the list of migrated suites and rewords the tier-3 residual note so it no longer names a helper the suite no longer uses.

Test-only; no production code changes, so no spec.

## Evidence & verification

- **Baseline (before):** the nightly `GitManagerTimeout` target on `main`, 10 iterations under 3 spinners on a 3-core runner, as posted to #503: 2/10, 1/10, 1/10, 2/10, 1/10 failures on the last five nights it ran.
- **After:** 0/20 failed. A throwaway dispatch of the same harness target (`scripts/nightly-flake-stress.sh --target GitManagerTimeout --iterations 20`, 3 cores, 3 spinners, load1m 3.5–13 at iteration start) on this branch: https://github.com/cheapsteak/tbd/actions/runs/35162510219. Twenty iterations at zero failures against a per-iteration baseline rate of roughly 10–20 % is strong but not proof, which is why the nightly target stays in place (see the graduation plan).
- The migrated `TBDDaemonLiveTests` target compiles locally. This machine has only the CommandLineTools toolchain, which ships no `XCTest`, so the test bundle itself cannot be built or run here; the pass/fail verdict comes from this PR's CI and the stress numbers above from a throwaway dispatch of the same harness target on a scratch branch (deleted afterwards).
- Soundness note on the handshake: the arming the test waits for is `clock.sleep(for:)` inside the deadline task, which `EventDrivenTestClock` implements through `sleep(until:)`, so the strict wait observes exactly the sleeper the virtual advance fires. The remaining exposure is one-sided and bounded: if the scheduler takes more than 45 s to give the deadline task its first thread, the test fails with a named diagnostic rather than hanging.

## Graduation plan

Refs #503 rather than closing it. The nightly flake-stress job in `.github/workflows/nightly.yml` already targets `GitManagerTimeout` and posts to #503; after several consecutive green targeted nightly runs, #503 closes and the `GitManagerTimeout` target is removed from the nightly job. This PR deliberately leaves that target in place.


cheapsteak:
Admin-merging over a red fast-pass job. Three attempts of the Test workflow failed on three different suites this PR does not touch (archived-search debounce, then the model-proxy SIGTERM test, then the model-proxy retire test), while the tier-3 live pass containing the migrated suite passed on every attempt. The same two proxy suites fail on main's own head run for a8965b15: https://github.com/cheapsteak/tbd/actions/runs/35132473524. Those flakes are being triaged separately.

Issue #871, opened 2026-09-17
Flake: "SIGTERM exits zero and reclaims the pid file" — the proxy reclaims its pid file but the test's waitpid never sees it exit (every nightly since 2026-09-09)


cheapsteak:
## Signature

`TBDModelProxyTests.ModelProxySuites/ProxyProcessTests`, test **"SIGTERM exits zero and reclaims the pid file"** (`Tests/TBDModelProxyTests/ProxyBinaryTests.swift`, `sigtermExitsZero`), always the same two issues and never a third:

```
✘ Test "SIGTERM exits zero and reclaims the pid file" recorded an issue at ProxyForwardingTests.swift:1031:21: Caught error: ProxyWaitTimeout(what: "the proxy to exit", observed: "false", seconds: 30.0)
✘ Test "SIGTERM exits zero and reclaims the pid file" recorded an issue at ProxyBinaryTests.swift:476:13: Expectation failed: (status → nil) == 0
```

In the suite-level lines the nightly report keeps, it is `✘ Suite "Proxy process" failed … with 2 issues` together with `✘ Suite "Model proxy" failed … with 2 issues` (the serialized parent).

## Rate and first seen

- Nightly `FastPassWhole` (whole fast pass under 3 spinners, issue #503's reports): present in **every iteration of every night from 2026-09-09 through 2026-09-16** (3/3 per night), and in none before — the suite itself landed on 2026-09-08 in #838, so this is the suite's rate since birth, not a regression on an older suite.
- Ordinary PR runs (fast-pass-2, no spinners): one observed — PR #870's Test run 35162559624, attempt 2 (2026-09-17T02:28Z). The main-head runs that failed on 2026-09-16 failed in the *retire* test instead (a different mechanism; see the linked PR).

## What the two issues, and the missing third one, establish

- The test spawns the real `TBDModelProxy` with `posix_spawn`, waits for its pid file, sends `SIGTERM`, then polls `waitpid(pid, WNOHANG)` for 30 s through `pollUntilTrue`, which takes a **fresh read after the deadline**. So `observed: "false"` means `waitpid` had still not returned the pid 30 s after the signal.
[...]

PR #875, opened 2026-09-17
Fix: ChildReaper reaps on its own threads, so a full dispatch worker pool cannot stop a PTY child from being reaped

cheapsteak:
## What's broken

The nightly flake-stress job runs the whole fast pass under three CPU spinners on a three-core runner, and two suites are the most frequent entries in the #503 ledger: `Suite "ChildReaper"` (28 of the nights since 2026-08-17) and `Suite "Terminal teardown reaps its PTY child"` (24 nights). They fail together, and with a fixed shape: `ChildReaper` "with 2 issues" and the teardown suite "with 3–4 issues", in suites that Swift Testing times at 150–340 s. Those counts are exactly the tests that await `ChildReaper.drainPendingReaps` — two in one suite, four in the other — and the durations rule out the 240 s suite limit. Neither suite ever reproduces alone, and neither has ever reached the ledger with a per-test line: the harness keeps twelve sorted signature lines per iteration, and the suite lines sort ahead of every test line.

In production the same mechanism is a latent hazard rather than a flake: `ChildReaper` is the guaranteed `waitpid` for every torn-down terminal's PTY child (#611), and it ran on a private concurrent dispatch queue whose blocks park for the child's lifetime — which its own doc comment says can be forever for a child that ignores `SIGHUP`.

## Why it happens

libdispatch caps a process's non-overcommit workers at `kern.wq_max_constrained_threads` (64 on a stock macOS install, 70 on the box this was measured on). Every `DispatchQueue.global()` block and every private *concurrent* queue draws from that one pool; private serial queues target the overcommit roots and are exempt. Once the cap's worth of workers are parked in blocking calls, nothing dispatched to any constrained queue starts, at any QoS, until one of them returns. It is a cliff, not a slope: a probe with 60 blocked global workers ran a reaper-shaped barrier in 0.41 s (the child's lifetime), and with 80 the barrier and a `global(qos: .utility)` no-op both failed to run inside 15 s while a private serial queue and a dedicated `Thread` ran in under a millisecond.

The fast pass parks constrained workers by the dozen. `TestSupport`'s subprocess helper drains both pipes of every child with blocking reads on `DispatchQueue.global()`, Foundation's `Process` machinery holds more, and a sample of the live test process on a 14-core box at load ~160 showed 21 root-queue workers held for the whole pass. On the three-core runner every subprocess lives longer and more of them overlap, and the ledger's fingerprint is what the cliff produces: with the pool full, `ChildReaper.reap`'s block never starts, the drain barrier behind it never fires, the 30 s hang guard in `drainPendingReaps(within:)` fires instead for every drain-awaiting test at once — two in `ChildReaper`, four in the teardown suite — and the child is found still an unreaped zombie. The regression test below reproduces that verdict deterministically against the old code, down to "Suite ChildReaper failed with 2 issues".

The alternative that fits the same fingerprint — utility-QoS work simply starved of CPU by the spinners — was checked and does not: at load 200 on 14 cores, utility-QoS blocks still ran within 0.1 s. Whether CI's process actually crosses the cap on a given night is inferred from the fingerprint rather than observed, since no per-test line has ever reached the ledger; that is the one link not measured directly.

## What this PR does

`ChildReaper` runs each reap on a `Thread` of its own instead of a block on a concurrent dispatch queue. A `Thread` is a plain pthread outside the constrained pool: a reap that parks for a child's lifetime no longer spends one of the app's ~64 shared workers on waiting, and a pool that other work has filled can no longer stop a reap from starting. The reaps stay concurrent with each other, as before, and stay at utility QoS.

The test seam `drainPendingReaps` kept its contract — run the callback once every reap started before the call has finished — and lost its dependence on a queue barrier: the reaper keeps a ledger of in-flight reaps, and the thread that finishes the last reap a drain covers calls back. Nothing on the drain path touches a dispatch worker.

Test-side: `ChildReaperTests` gains a regression test that fills the constrained pool (reading `kern.wq_max_constrained_threads` from the kernel and waiting until that many workers are actually parked, since reaping before the fill completes proved nothing) and requires a reap to run and drain while it is full; `reapOffPool` moves to a dedicated thread for the same reason; the drain helper's and both suites' doc comments describe the ledger rather than the barrier.

Considered and rejected: raising the 30 s guard (the stall is indefinite, not slow); giving the barrier a higher QoS (a full pool starts nothing at any QoS); a private serial queue (serializes reaps behind a stuck child, the thing the concurrent queue was chosen to avoid). No spec: this restores the type's stated theory — the guaranteed `waitpid` — under a condition its implementation could not survive.

## Assumptions

Assumes teardowns are rare enough that a pthread per reap is negligible. The cost is one stack per reap still parked, and the count is bounded by the number of terminals torn down whose child has not yet exited — in the field that has meant a handful, and the pathological case on record was 67 children over a whole app lifetime, all of them already exited and each reaped in microseconds.

Nothing caps that thread count, and an explicit cap was rejected rather than overlooked: a cap is a queue, and a queue in front of a blocking `waitpid` reintroduces exactly the failure this PR removes — the reap that cannot start because something ahead of it will not finish. The difference is only which pool is exhausted and who else it takes down with it, and a private one would be worse than libdispatch's in one respect, since a cap sized for this subsystem is reached far sooner than 64. What makes unbounded safe here is the bound the work already has: reaps are started only by terminal teardown, one per PTY child, and a thread ends the moment its child does. If some future path could start reaps in a loop, the answer is to bound the `waitpid` itself — escalate to `SIGKILL` on a deadline, which the type's doc comment already names as the separate change it needs (an injected clock, and a decision about how the child is asked to die) — not to put a queue in front of it.

Assumes the constrained cap stays a process-wide limit that dedicated threads sit outside of; if libdispatch ever counted pthreads against it, the regression test would go red rather than the behavior degrading quietly.

## Evidence & verification

- Discriminating test: `reapsWhileTheConstrainedWorkerPoolIsExhausted` against the unchanged reaper fails after the 30 s guard with `ChildReaper's drain did not fire … pid was observed an unreaped zombie (exited, still waitable)`, and takes the concurrently running `backgroundReapSkipsObservedChild` down with it — `Suite "ChildReaper" failed … with 2 issues`, the ledger's exact shape. With the fix, all 13 tests across both suites pass, the exhaustion test in 23 ms.
- Isolated probe (a reaper-shaped `qos: .utility` concurrent queue plus one parked `waitpid`, then a barrier): 0 and 60 blocked global workers, barrier fires in ~0.41 s; 80 blocked, barrier and a global-utility no-op both absent at 15 s, private serial queue and dedicated `Thread` both under 1 ms. A `sample` of the wedged probe showed exactly 70 workers — this box's `kern.wq_max_constrained_threads` — parked in the blocking read and no thread at all for the reaper queue.
- Stress before the fix, whole fast pass through `scripts/test.sh --parallel --skip '^TBDDaemonLiveTests\.'` (10,992 tests) on a 14-core box shared with other agents: 3 spinners, 3 iterations, load 15–35, neither suite red; 14 spinners, 1 iteration, load 160–230, neither suite red. The natural rate is too low to measure outside the three-core runner, which is why the discriminating evidence is the regression test rather than the loop.
- Stress after the fix, same shape (3 spinners, 3 whole-pass iterations, load 42–164): every pre-existing test in both suites green in all three; ten iterations of the two suites alone under 3 spinners: 10/10 green, 13 tests each.
- The whole-pass runs also measured the pool directly. An earlier revision of the regression test waited up to 90 s for its own blockers to occupy every constrained slot and threw if they did not; in all three whole-pass iterations it reported "parked only 64/66/67 of the 70 constrained workers this kernel allows within 90 s" — the remaining slots were held by other suites for the entire window. That revision was dropped because holding the pool for 90 s starves every sibling's global-queue work; the shipped test waits 2 s and treats expiry as the pool being full by somebody else, which is what it observed.
- Local toolchain note: this box has CommandLineTools only, so the runs above passed `-Xswiftc -F …/Library/Developer/Frameworks` and two `-rpath`s to find `Testing.framework`, and set aside the one XCTest-only file (`ArchiveTombstoneTests.swift`) for the local runs; neither is part of the change.

[main](https://cheapsteak.github.io/tbd/open/?worktree=51605E2C-AE9C-45D5-BCF0-AF8F180A12A7)


cheapsteak:
Admin-merging over a red fast-pass job. The only failure is the pre-existing #871 proxy SIGTERM flake, which predates this branch and is unrelated to it. The two suites this PR targets both passed in that same run: "ChildReaper" in 6.3s and "Terminal teardown reaps its PTY child" in 24.4s, including the new interleaving regression test. claude-review passed after the added coverage.

